PR #1, opened 2023-10-26
Monorepo

mikevanhemert:
UDS Core is following a monorepo. This is the initial folder structure using templating defined in [uds-template-bundle](https://github.com/defenseunicorns/uds-template-bundle) and [uds-template-capability](https://github.com/defenseunicorns/uds-template-capability).


zachariahmiller (approved):

mjnagel (changes requested):
Couple things:
- codeowners per capability seems weird - it might work? but I think doing a top level codeowners with folder settings feels better to control them in one spot
- contributing and license should be identical across all capabilities?
- renovate.json only works at the root of the repo I think?
- .gitignore per folder I think is more likely to be the exception rather than norm? (aka we should be able to use the root gitignore 99% of the time)

Otherwise looks pretty good I think.

mjnagel (changes requested):
Other things:
- kyverno-policies may need a separate capability folder. BB deploys it as a separate helm chart and I don't think upstream kyverno provides any policies out of the box, or ability to make them
- kyverno-reporter could probably be smushed with kyverno if we want, but is separate in BB/DUBBD today
- prometheus is probably more accurately monitoring or prometheus-stack...it includes alertmanager and a few other things - https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack

capabilities/kyverno-report/zarf.yaml
```diff
@@ -0,0 +1,17 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/defenseunicorns/zarf/main/zarf.schema.json
```
mjnagel: I did a typo - its `kyverno-reporter` 🙃

mjnagel (approved):

PR #2, opened 2023-10-27
feat: add prometheus-stack (monitoring) capability

MxNxPx:
## Description

add prometheus-stack capability

## Related Issue

N/A

## Type of change

## Checklist before merging


jeff-mccoy (commented):
can we leave the upstream images as the default (uncommented) until we have the `only.flavor` feature released next week in Zarf?

mikevanhemert (commented):
inb4 @mjnagel does it make sense to call this capability `monitoring` instead of `prometheus` since it comes with alert manager and some other pieces in addition to Prometheus? A different alternative could be `prometheus-stack`

If it is changed, recommend updating the repository [`README.md`](https://github.com/defenseunicorns/uds-core#readme) to reflect the change

mjnagel:
Yeah - I don't love calling it Prometheus since there is more included. I'm good with prometheus-stack or monitoring or other suggestions.

MxNxPx:
> Yeah - I don't love calling it Prometheus since there is more included. I'm good with prometheus-stack or monitoring or other suggestions.

Fair point - my thought process here was that `monitoring` seemed too generic.  I'm more partial to `prometheus-stack` but am good with whatever @jeff-mccoy @mikevanhemert @mjnagel decide is the best name.

mikevanhemert:
> > Yeah - I don't love calling it Prometheus since there is more included. I'm good with prometheus-stack or monitoring or other suggestions.
> 
> Fair point - my thought process here was that `monitoring` seemed too generic. I'm more partial to `prometheus-stack` but am good with whatever @jeff-mccoy @mikevanhemert @mjnagel decide is the best name.

If you like `prometheus-stack` then I think we have a winner @MxNxPx  🥳

MxNxPx:
> can we leave the upstream images as the default (uncommented) until we have the `only.flavor` feature released next week in Zarf?

@jeff-mccoy - swapped it over now to use upstream.

jeff-mccoy (approved):

PR #3, opened 2023-10-27
feat: Add istio and preliminary ci

zachariahmiller:
## Description
Add istio and ci
...

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


zarf-runner/rusty-rogers/zarf.yaml
```diff
@@ -0,0 +1,40 @@
+kind: ZarfPackageConfig
```
mikevanhemert: Two thoughts on the Zarf runner
1. I believe we're planning to settle on a path forward for task runner, suspect this will be changed
2. Whatever the new path is will apply to more repos than UDS-Core, can we externalize it and reference it from here?

Maybe move the Zarf runner pieces to another PR so it is isolated from the Istio Capability?

mikevanhemert (dismissed):

capabilities/istio/.github/zarf-runner/zarf.yaml
```diff
@@ -0,0 +1,25 @@
+kind: ZarfPackageConfig
```
jeff-mccoy, line 1: is this duplicate code from the other zarf runner folder in the root directory?

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,134 @@
...
+          registry: ghcr.io
+          username: dummy
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Use Node.js latest
+        uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3.8.1
+        with:
+          node-version: latest
+
+
+      - name: Setup docker (missing on MacOS)
```
jeff-mccoy: I think we are mixing concerns / value by adding the macos tests. Don't think we should do those yet, but should talk about it later on.

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,134 @@
...
+      - name: Build Capability
+        shell: bash
+        run: zarf package create --confirm --no-progress
+        working-directory: "${{ needs.setup.outputs.capability-path }}/.github/zarf-runner/bob-the-builder"
+        if: ${{ inputs.capability }} != ''
+
+      - name: Deploy Capability
+        shell: bash
+        run: zarf package create --confirm --no-progress
+        working-directory: "${{ needs.setup.outputs.capability-path }}/.github/zarf-runner/deploy"
```
jeff-mccoy, line 104: confused by the working directory call here, seems pretty messy. wouldnt it make more sense (if this is necessary to pass into `zarf package create dir`?  Though not a huge fan of this format and want to talk through it more.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,28 @@
...
+  run:
+    # We need -e -o pipefail for consistency with GitHub Actions' default behavior
+    shell: bash -e -o pipefail {0}
+
+
+jobs:
+  publish-package:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest] #windows-latest?
```
jeff-mccoy: same as above, don't think macos makes sense right now.

capabilities/istio/.github/zarf-runner/bob-the-builder/zarf-config.yaml
```diff
@@ -0,0 +1,4 @@
+package:
+  create:
+    set:
+      uds_cli_version: "v0.0.6-alpha"
```
jeff-mccoy: should this at least be v7 since that has full compatibility with zarf 30?

capabilities/istio/.github/zarf-runner/bob-the-builder/zarf.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+components:
+  - name: build
+    required: true
+    actions:
+      onCreate:
+        before:
+          - description: Load UDS CLI
+            cmd: |
+              [ -f ../build/uds ] && [ "$(../build/uds version)" == "###ZARF_PKG_TMPL_UDS_CLI_VERSION###" ] && exit 0
```
jeff-mccoy, line 16: we should use brew for this instead now.

capabilities/istio/.github/zarf-runner/bob-the-builder/zarf.yaml
```diff
@@ -0,0 +1,30 @@
...
+              curl -fL https://github.com/defenseunicorns/uds-cli/releases/download/###ZARF_PKG_TMPL_UDS_CLI_VERSION###/uds-cli_###ZARF_PKG_TMPL_UDS_CLI_VERSION###_$(uname -s)_${ARCH} -o ../build/uds
+              chmod +x ../build/uds
+          - description: Build the pepr module
+            cmd: |
+              npm ci
+              npx pepr build
+            dir: ../../../pepr
+          - description: Build the zarf packages
+            cmd: |
+              ./zarf p c ../../../ --confirm --no-progress -o ../build
```
jeff-mccoy, line 31: --no-progress flats aren't required in ci for zarf (it's automatic)

capabilities/istio/.github/zarf-runner/deploy/zarf.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - name: deploy
+    required: true
+    actions:
+      onCreate:
+        before:
+          - description: Deploy the cluster bundle
+            cmd: |
+                ARCH=$(uname -m)
+                [ "$ARCH" = "x86_64" ] && ARCH="amd64"
+                ./uds deploy oci://ghcr.io/zachariahmiller/k3d-common:0.1.0-$ARCH --no-progress --confirm
```
jeff-mccoy, line 18: we shouldn't be pointing to personal repos

capabilities/istio/.github/zarf-runner/deploy/zarf.yaml
```diff
@@ -0,0 +1,30 @@
...
+            dir: ../build
+          - description: Deploy the istio
+            cmd: |
+                ARCH=$(uname -m)
+                [ "$ARCH" = "x86_64" ] && ARCH="amd64"
+                ./zarf p d ../build/zarf-package-uds-capability-istio-$ARCH-1.19.3.tar.zst  --no-progress --confirm #${COMPONENTS}          
+          - description: Inject tls certificate
+            cmd: |
+              npm install
```
jeff-mccoy: multiline suggestions are being weird, here's what it should be:

```shell
npm ci
npx --yes ts-node secret.ts
```

capabilities/istio/.github/zarf-runner/release/zarf-config.yaml
```diff
@@ -0,0 +1,4 @@
+package:
+  create:
+    set:
+      uds_cli_version: "v0.0.6-alpha"
```
jeff-mccoy: needs to be 0.0.7

capabilities/istio/gw-tls/package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "name": "burning.boats",
+    "dependencies": {
+        "kubernetes-fluent-client": "1.7.0"
```
jeff-mccoy: need to update to `1.8.2` and then run `npm update` to update the lock file

.github/workflows/build-and-test.yaml
zachariahmiller: They arent running in there right now its hardcoded to linux, but that is left over from when i was running the mac test and was required.

zachariahmiller: i can delete or not as it stands this is doing nothing unless someone added macos as a runner target.

.github/workflows/build-and-test.yaml
zachariahmiller: If its confusing whats happening and why maybe we should have a conversation synchronously

.github/workflows/publish.yaml
zachariahmiller: 100% its removed in the steps that are presently running

capabilities/istio/.github/zarf-runner/bob-the-builder/zarf-config.yaml
zachariahmiller: yes

capabilities/istio/.github/zarf-runner/bob-the-builder/zarf.yaml
zachariahmiller: but this can also run outside of CI and i want -no-progress for local dev.

capabilities/istio/.github/zarf-runner/deploy/zarf.yaml
zachariahmiller: 100%. the code that deploys this oci://ghcr.io/zachariahmiller/k3d-common:0.1.0-$ARCH needs to be moved into the org at which time this will not reference my local repo.

zarf-runner/rusty-rogers/zarf.yaml
zachariahmiller: the zarf runner stuff is not generic right now. There are multiple considerations for how an individual capability might need to be deployed and tested

MxNxPx (approved):
considering the state of tooling we have today, this gets us something we can use now for istio and CI that we could really use atm.  understanding that some of the things are just placeholders (like release & publish) which will be updated in follow up PRs.  thanks @zachariahmiller!!

capabilities/istio/.github/zarf-runner/bob-the-builder/zarf.yaml
zachariahmiller: isnt it going to take a lot longer to `brew tap defenseunicorns/tap && brew update && brew install uds`

zachariahmiller: otherwise i agree...

capabilities/istio/.github/zarf-runner/zarf.yaml
zachariahmiller: the zarf runner steps are not generic so in the root it would be for the bundle. If we had one file ala task file or like what jordan had put together (wren) where i could call a specific task this wouldnt be a problem

.github/workflows/build-and-test.yaml
zachariahmiller: Which of these is more readable:
```

      - name: Build Capability
        shell: bash
        run: zarf package create --confirm --no-progress
        working-directory: "${{ needs.setup.outputs.capability-path }}/.github/zarf-runner/bob-the-builder"
        if: ${{ inputs.capability }} != ''

        - name: Build Capability
        shell: bash
        run: |
          zarf package create \
          "${{ needs.setup.outputs.capability-path }}/.github/zarf-runner/bob-the-builder" \
          --confirm --no-progress
        if: ${{ inputs.capability }} != '' 
```
        
  I think the first one. 
        
        
   but again, the path after the github var is an artifact of how we need to do things now while we are in a holding pattern about task-runner things

.github/workflows/publish.yaml
zachariahmiller: removed

.github/workflows/build-and-test.yaml
```diff
@@ -0,0 +1,134 @@
...
+      - name: Setup docker (missing on MacOS)
+        if: runner.os == 'macos'
+        run: |
+              brew install docker
+              colima start --cpu 2 --memory 12
+              sudo ln -sf $HOME/.colima/default/docker.sock /var/run/docker.sock          
+
+      - name: Install k3d
+        run: |
+          curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | bash
```
mjnagel:
```suggestion
          curl -s https://raw.githubusercontent.com/k3d-io/k3d/main/install.sh | TAG=v5.6.0 bash
```
Pin to a version maybe + renovate tracking on this?

capabilities/istio/.github/zarf-runner/bob-the-builder/zarf.yaml
zachariahmiller: brew also isnt installed on the runners `      brew: not found` so i would have to install brew and then do the above. Worth noting we dont use brew to install zarf anywhere. Jordans wren poc had a solution for this which didnt involve brew, also the zarf action could be copied and reused to make a uds action, but that doesnt solve parity between local dev and ci.

capabilities/istio/.github/zarf-runner/deploy/zarf.yaml
zachariahmiller: I can do the supported work for this ASAP, but will need to do some refactoring to make it deploy both package/ bundle architectures.

capabilities/istio/.github/zarf-runner/deploy/zarf.yaml
```diff
@@ -0,0 +1,29 @@
...
+            cmd: |
+                ARCH=$(uname -m)
+                [ "$ARCH" = "x86_64" ] && ARCH="amd64"
+                ./uds deploy oci://ghcr.io/zachariahmiller/k3d-common:0.1.0-$ARCH --no-progress --confirm
+            dir: ../build
+          - description: Deploy the istio
+            cmd: |
+                ARCH=$(uname -m)
+                [ "$ARCH" = "x86_64" ] && ARCH="amd64"
+                ./zarf p d ../build/zarf-package-uds-capability-istio-$ARCH-1.19.3.tar.zst  --no-progress --confirm #${COMPONENTS}
```
mjnagel, line 24: Should this be `../build/zarf-package-uds-capability-istio-$ARCH-*.tar.zst` to ensure we don't have to keep it up to date with the built version?

zarf-runner/rusty-rogers/zarf.yaml
zachariahmiller: Hi Mike. I agree its temporary and its super frustrating. As for your comments per [minimumcd](https://minimumcd.org/translations/texan/#continuous-integration)

```
The bare minimums for CI are:

    [Trunk-based development](https://minimumcd.org/translations/texan/#trunk-based-development)
    Work integrates into the trunk at least daily
    Y’all automated the testin’ before mergin'
    Work is tested with other work automagically on merge
    All feature work stops when the build halts and catches fire
    New work does not mess up delivered work
```
    
 I provided this translated version for Jeff. 
 
 Anyways, we shouldnt be merging to main without working ci.

zachariahmiller: We state all over the place we follow this so 🤷

capabilities/istio/.github/zarf-runner/bob-the-builder/zarf.yaml
zachariahmiller: I switched it back for now because its working. More than happy to quickly iterate on this after this initial pr is used so other can start integrating ci into their capabilities

.github/workflows/build-and-test.yaml
zachariahmiller: We can address this once we have mvp of the task runner Jordan is back to working on.

mjnagel (approved):

zachariahmiller:
I'm merging this with the two approvals.

For someone to add their capability to CI they will need to create the .github/zarf-runner things under their capabilities folder (temp while waiting for the taskrunner kind) and then go to .github/filters.yaml and add their capability like so:

`
istio:
  - 'capabilities/istio/**/*'
neuvector:
  - 'capabilities/neuvector/**/*'
example:
  - 'capabilities/example/**/*'
  `

MxNxPx (approved):
:ship: :it:

capabilities/istio/.github/zarf-runner/deploy/zarf.yaml
zachariahmiller: If that will actually work I think you are correct. Fwiw I had tried that and thought it wasn't working but maybe I was just being dumb.

PR #6, opened 2023-10-28
Istio updates

jeff-mccoy:
## Description

Adds Istio

## Type of change

## Checklist before merging


capabilities/istio/.github/zarf-runner/bob-the-builder/zarf.yaml
```diff
@@ -25,7 +25,9 @@ components:
             cmd: |
               npm ci
               npx pepr build
-            dir: ../../../pepr
+            dir: ../../../../../
+            env:
+              - "CAPABILITY=istio"
           - description: Build the zarf packages
             cmd: |
-              ./zarf p c ../../../ --confirm --no-progress -o ../build
```
zachariahmiller, line 31: I know you dont like the no progress because its automatic in ci, but if we are running the same in local dev and ci i dont think anyone actually wants the full output from local. 

Alternatively to having the --no-progress all over, maybe we can export CI=true as a first steps in the runner package? With the task runner stuff jordan is working on this would be even easier it could just be a global var at the top of the file.

capabilities/istio/.github/zarf-runner/deploy/zarf.yaml
```diff
@@ -22,15 +22,10 @@ components:
                 ARCH=$(uname -m)
                 [ "$ARCH" = "x86_64" ] && ARCH="amd64"
                 ./zarf p d ../build/zarf-package-uds-capability-istio-$ARCH-1.19.3.tar.zst  --no-progress --confirm #${COMPONENTS}          
-          - description: Inject tls certificate
-            cmd: |
-              npm ci
```
zachariahmiller, line 27: I understand why we are removing this here, but just to clarify, do we still plan on using this in our actual deployments or just kubectl?

pepr.ts
```diff
@@ -0,0 +1,35 @@
+import { Capability, Log, PeprModule } from "pepr";
```
zachariahmiller, line 1: ❤️ will try to digest the pepr changes once this is all out of draft/after the weekend.

capabilities/istio/.github/zarf-runner/bob-the-builder/zarf.yaml
jeff-mccoy: The output is mostly the same with no-progress or not on a regular terminal, it just turns off things that repost to the same line. It was created for streaming terminal recorders to not show the progress or spinner over and over. If you want to reduce logging (which I see as an individual decision), that would be done by changing the log level.

capabilities/istio/.github/zarf-runner/deploy/zarf.yaml
jeff-mccoy: Move to the chart in a different area, it's still ugly now until helm variable things exist.

pepr.ts
jeff-mccoy: Yeah big idea is leverage monorepo for pepr -- but have a way to tell pepr to bootstrap only one "uds capability" at a time for testing.

zachariahmiller (approved):
Someone else will need to approve since i have the most recent commits on here, but the issues were resolved.

mikevanhemert (approved):
Reviewed on my phone ™️ 
Changes look good!

PR #12, opened 2023-11-02
feat: add pepr capability for istio + jobs

### LINKED ISSUE #15: Istio injection breaks k8s jobs

Unless there is something added to jobs, istio sidecars will continue running after a job has completed. Typically this means you have 2 options:
- disable injection on the job pods: sometimes possible, but not always, and breaks mTLS
- add logic to your application to terminate the sidecar: when leveraging upstream apps/charts this isn't always possible

Pepr is a great candidate to solve this issue. Either via a mutation or via a watch pepr should be able to terminate the sidecar at the appropriate time, when all other containers have finished.

The above options would work, but aren't ideal for the listed reasons.

mjnagel:
## Description

Adds a pepr capability that terminates istio sidecars after job completion. This is a well known issue which _should_ be resolved in [k8s 1.28 via a new feature flag](https://istio.io/latest/blog/2023/native-sidecars/):
- https://github.com/kubernetes/kubernetes/issues/25908
- https://github.com/istio/istio/issues/6324

This pepr capability adds a fix for today's deployments by watching all pods that are controlled by a job and sidecar injected. Once all other containers have successfully terminated pepr will exec into the sidecar and kill it using the `pilot-agent` (the only util guaranteed to always be available).

## Related Issue

Resolves https://github.com/defenseunicorns/uds-core/issues/15

## Type of change

## Checklist before merging


mjnagel:
Example job to test with:
```yaml
# Throw this into a namespace with istio injection
apiVersion: batch/v1
kind: Job
metadata:
  name: hello-world-job
spec:
  template:
    metadata:
      name: hello-world-job
    spec:
      containers:
      - name: hello-world
        image: centos:7
        command:
         - "bin/bash"
         - "-c"
         - "echo hello world"
      restartPolicy: Never
```
Pre-this capability the job will never complete. With the capability you should see it complete successfully.

jeff-mccoy:
updated, note if we were not using the watch action and instead mutate or validate, we'd need to use PeprStore instead for this.

capabilities/istio/pepr/istio-job-termination.ts
```diff
@@ -0,0 +1,69 @@
...
+    // Ensure termination isn't already in progress
+    if (inProgress[name]) {
+      return;
+    }
+
+    // Only terminate if the pod is running
+    if (status.phase == "Running") {
+      // Check if the pod has a non-istio container that has terminated
+      const canTerminate = !!status?.containerStatuses?.find(
+        ({ name, state }) => name != "istio-proxy" && state?.terminated,
```
mjnagel: I don't think finding a single terminated container is sufficient here - a job could have 2, 3, ... other containers and we wouldn't want to kill the sidecar until all are terminated. This check would kill it based on a single terminated non-istio container unless I'm mistaken?

I also think we may need to check the containerstatus exit codes? If a container is terminated, but had a non-zero exit code, it could restart depending on the job restartpolicy. We could end up in a scenario where we kill the sidecar too early as a result and mess up networking for the retries.

jeff-mccoy: On multiple containers, yes I think you're right, I was distracted by a [type bug](https://github.com/defenseunicorns/pepr/pull/356) in pepr with ramda and didn't realize this was an all vs any match when I was debugging with your example (since it was only one container). 

I do think we need to make 100% sure we understand the behavior of the job controller when a multi-container pod has one or more failed condition states. Does it restart / recreate the pod? I dont use jobs enough to say confidently how that controller handles those different scenarios.

capabilities/istio/pepr/istio-job-termination.ts
```diff
@@ -9,39 +9,60 @@ export const IstioJobTermination = new Capability({
...
+      );
+
+      if (canTerminate) {
+        // Mark the pod as seen
+        inProgress[name] = true;
```
rjferguson21: `inProgress` probably needs to be keyed off of `namespace/name` to ensure it is unique across the cluster

capabilities/istio/pepr/istio-job-termination.ts
mjnagel: https://kubernetes.io/docs/concepts/workloads/controllers/job/#handling-pod-and-container-failures - restart policy does not replace the pod on failure, just the individual failed container. I think that matches what I've seen with jobs in the past, only individual containers are restarted based on failures, not the whole pod.

cmwylie19 (approved):
lgtm

zachariahmiller (approved):

PR #14, opened 2023-11-02
test: introduce new uds task file & ci structure

jeff-mccoy:
## Description

Utilize UDS CLI task runner for all local/remote jobs. This allows us to use the same commands locally or remotely. Also introducing per-capability and complete UDS Core testing in parallel, with fail-fast + custom runners for speed. 

## Type of change

## Checklist before merging


zachariahmiller:
Preliminary thought on this is that what you have in this pr right now is pretty much 100% tracking towards what i envisioned, only caveat being that there isn't a root level run task that can do all the things if you are doing local dev. 

That would be trivial to add in the future and would basically emulate the flow you have orchestrated in github workflows on local.

zachariahmiller:
To your last commit message. You can do a fun bash one liner or use this https://github.com/philips-labs/list-folder-action and inline the filter instead of using the yaml file 🫠 

I thought the yaml file was the clearest option, albeit verbose

zachariahmiller (approved):
I don't really have much to say since I've been commenting on the progress along this journey... 🤣 LGTM! 🚀🦄

jeff-mccoy:
lol yeah yaml wasn't bad really, just started feel like one more area for us to keep up with, but probably not that big of a deal

PR #25, opened 2023-11-08
feat: release-please integration

jeff-mccoy:
## Description

The PR introduces the release-please workflow for generating releases. Additional changes:

- example-bundles -> bundles
- all schema refs for zarf/uds moved to vscode settings instead (can add other IDEs as-needed)
- updated UDS CLI to 0.1.0
- added multi-arch builds
- added basic user README info for using the packages/bundles

## Type of change

## Checklist before merging


tasks/publish.yaml
```diff
@@ -0,0 +1,50 @@
...
+    default: oci://ghcr.io/defenseunicorns/packages/uds
+
+  - name: VERSION
+    description: "The version of the packages to build"
+    # x-release-please-start-version
+    default: "0.1.0"
+    # x-release-please-end
+
+tasks:
+  - name: packages
```
mikevanhemert, line 12: I would prefer if we use the same terminology for inputs to UDS Bundles. Mattermost and UDS Core are Zarf packages that we bundle. Different terms for them adds some confusion

tasks/publish.yaml
```diff
@@ -0,0 +1,50 @@
...
+          ZARF_ARCHITECTURE=arm64 uds run -f tasks/create.yaml k3d-standard-bundle --no-progress
+
+      - description: Create the UDS Core Istio Only Bundle
+        cmd: |
+          ZARF_ARCHITECTURE=amd64 uds run -f tasks/create.yaml k3d-istio-bundle --no-progress
+          ZARF_ARCHITECTURE=arm64 uds run -f tasks/create.yaml k3d-istio-bundle --no-progress
+
+      - description: "Publish the bundles"
+        cmd: |
+          uds publish bundles/k3d-standard/uds-bundle-k3d-core-amd64-${VERSION}.tar.zst ${TARGET_REPO}/bundles --no-progress
```
mikevanhemert, line 47: Will this list grow or do you think it is relatively stable? If it might grow, we may consider asking for looping over lists of strings being added to the UDS CLI runner stuff

mikevanhemert (approved):

mikevanhemert:
I like the general approach in this PR. Couple of comments for discussion, but I don't see anything egregious and don't have any requested changes

tasks/publish.yaml
jeff-mccoy: Yeah I didn't know for sure but agree looping or task runner matrix would be better here.

zachariahmiller: Since i've brought it up over and over again already i'll make a ticket for uds cli for looping/matrix

tasks/deploy.yaml
```diff
@@ -25,12 +18,11 @@ tasks:
...
 
   - name: single-capability
     actions:
       - description: "Deploy the requested capability Zarf Package (must set CAPABILITY environment variable)"
         cmd: |
-          export ARCH=$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/')
-          zarf package deploy build/zarf-package-uds-capability-${CAPABILITY}-${ARCH}.tar.zst --confirm
-          zarf package deploy build/zarf-package-pepr-uds-core-${ARCH}-*.tar.zst --confirm || true
+          zarf package deploy build/zarf-package-uds-capability-${CAPABILITY}-${UDS_ARCH}.tar.zst --confirm
+          zarf package deploy build/zarf-package-pepr-uds-core-${UDS_ARCH}-*.tar.zst --confirm || true
```
zachariahmiller, line 28: Is UDS_ARCH coming as built in from UDS as the current system ARCH or am i missing something about where its defined?

zachariahmiller: if the former is the case i have no issues with this PR, if the latter is the case i'd like to understand, but otherwise i think this all looks good. 😄

zachariahmiller (approved):

zachariahmiller:
Similar to Mike i just had the one question, but no changes requested and as such i approved. I'll go ahead and open a ticket in uds for the loops/matrix task functionality, which would have been my only requested change if the functionality already existed in uds. 🚀

tasks/publish.yaml
zachariahmiller: for reference or additional comments https://github.com/defenseunicorns/uds-cli/issues/187

tasks/publish.yaml
jeff-mccoy: Agree we need to sort this, and as discussed, will address in another PR.

tasks/publish.yaml
jeff-mccoy: awesome thanks @zachariahmiller !

tasks/deploy.yaml
jeff-mccoy: UDS_ARCH was added a couple versions ago to provide the zarf-compatible binary arch (amd64 or arm64) for this exact usecase.

PR #32, opened 2023-11-11
feat: embed tls certs in istio package

jeff-mccoy:
## Description

Embed the uds.dev certs into the package by default in preparation for future helm override features. In the interim, the Node version still works, and the README was updated to explain this.

## Type of change

## Checklist before merging


zachariahmiller (approved):

capabilities/istio/chart/values.yaml
```diff
@@ -15,8 +15,8 @@ tls:
   # The secret name for the TLS certificate when not using passthrough mode
   secretName: gw-cert
 
-  # For TLS SIMPLE mode, the TLS certificate data for the gateway
+  # For TLS SIMPLE mode, the TLS certificate data for the gateway (not base64 encoded)
```
mjnagel, line 18: I think I'm too late to the review, but I think this update to the comment is confusing since the value passed in here has to be base64 encoded. Tracking that the secret does not do the base64 encoding anymore, but the value comment should be written from a user perspective (something like `For TLS SIMPLE mode, the base64 encoded TLS certificate data for the gateway`).

If you all agree I can PR the small change.

Issue #17, opened 2023-11-02
Handle Prometheus metrics + Istio mTLS


mjnagel:
### Is your feature request related to a problem? Please describe.

When using Istio + mTLS STRICT (which UDS Core enforces by default on all namespaces) prometheus metrics scraping can require extra configuration. Two good docs to explain this:
- https://istio.io/latest/docs/ops/integrations/prometheus/#configuration
- https://docs-bigbang.dso.mil/latest/packages/monitoring/docs/istio-mtls-metrics/

There are two common approaches:
- Allow istio to handle metrics merging for you: this relies on standardized prometheus.io annotations, and istio will merge the application metrics with the sidecar metrics. These metrics are then exposed via plain text to prometheus. 
- Customize your scraping to have prometheus scrape on TLS using the istio certs.

### Describe the solution you'd like

To avoid two downsides of metrics merging (plaintext metrics and potential for conflicting metrics between istio and app) I think that Prometheus scraping on TLS is ideal. This would require a few things to happen:
- [Prometheus configured to mount istio certificates](https://docs-bigbang.dso.mil/latest/packages/monitoring/docs/istio-mtls-metrics/#Prometheus-Config)
- [Servicemonitors configured to be https with a tlsconfig for istio certs](https://docs-bigbang.dso.mil/latest/packages/monitoring/docs/istio-mtls-metrics/#ServiceMonitor-Config)

While this could be done via values in charts in some cases pepr may be a more ideal solution to automatically inject these configs at runtime.

### Describe alternatives you've considered

Add mTLS exceptions for the metrics port: this is not ideal since some apps may expose the app and metrics on the same port.

Metrics merging: this is a good option, but does mean metrics are exposed plaintext and could conflict between app/sidecar metrics.

### Additional context

Big Bang opted for a similar approach with Prometheus mounting istio certs and using them for scraping.

mjnagel:
Tackling this with 3 options for deployment:
- Existing servicemonitor: deployed by the upstream chart, pepr will mutate to add the proper config
- Service with labels: pepr will generate a servicemonitor with the proper config based on 1+ labels to identify the port to monitor
- Configmap with labels: pepr will generate a servicemonitor with the proper config based on the service with the same name

[...]

PR #33, opened 2023-11-13
feat: add monitoring and logging

zachariahmiller:
Add promtail, fix netpols, add prometheus to standard-package, add grafana

## Type of change

## Checklist before merging


capabilities/loki/chart/templates/loki-dashboards.yaml
```diff
@@ -22,7 +22,7 @@ items:
   kind: ConfigMap
   metadata:
     name: {{ printf "%s-%s" (include "loki.name" $) $dashboardName | trunc 63 | trimSuffix "-" }}
-    namespace: monitoring
+    namespace: grafana
     labels:
       grafana_dashboard: "1"
   data:
```
mjnagel: The second half of this file isn't being used and could be removed (everything after the yaml separator on line 13).

capabilities/loki/chart/templates/loki-dashboards.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: v1
 kind: ConfigMap
```
mjnagel, line 2: Can't comment on exact line (2) but you can remove `{{- $files := .Files.Glob "src/dashboards/*.json" }}` - that's not used.

capabilities/neuvector/zarf.yaml
```diff
@@ -37,9 +41,6 @@ components:
...
 
-  - name: uds-neuvector-config
-    charts:
-      - name: uds-neuvector-config
-        namespace: neuvector
-        version: 0.1.0
-        localPath: chart
+
+
+
```
mjnagel:
```suggestion
```

capabilities/promtail/values/values.yaml
```diff
@@ -0,0 +1,166 @@
...
+    rules: []
+  relabelings: []
+  scheme: https
+  scrapeTimeout: null
+  targetLabels: []
+  tlsConfig:
+    caFile: /etc/prom-certs/root-cert.pem
+    certFile: /etc/prom-certs/cert-chain.pem
+    insecureSkipVerify: true
+    keyFile: /etc/prom-certs/key.pem
```
mjnagel, line 157: Just dropping as a note, the specific tlsConfig here should definitely be a part of a broader pepr capability - https://github.com/defenseunicorns/uds-core/issues/17

capabilities/promtail/values/values.yaml
```diff
@@ -0,0 +1,166 @@
...
+#   clients:
+#     {{- if eq .Values.loki.strategy "monolith" }}
+#     - url: http://logging-loki.logging.svc.cluster.local:3100/loki/api/v1/push
+#     {{- else }}
+#     - url: http://logging-loki-write.logging.svc.cluster.local:3100/loki/api/v1/push
+#     {{- end }}
+
+  snippets:
+    scrapeConfigs: >
+      - job_name: kubernetes-pods
```
mjnagel, line 7: I think we can nuke most of this except for the node log stuff (journal/systemd), and instead add those pieces via `snippets.extraScrapeConfigs`.

zachariahmiller: I created an issue for this

capabilities/grafana/chart/templates/datasources.yaml
```diff
@@ -0,0 +1,61 @@
...
+          tags:
+          - job
+          - instance
+          - pod
+          - namespace
+      name: Tempo
+      orgId: 1
+      type: tempo
+      uid: tempo
+      url: http://tempo.tempo.svc:3100
```
mjnagel: Probably should leave this the tempo datasource off until we decide on adding tempo or not?

capabilities/grafana/zarf.yaml
```diff
@@ -15,13 +15,14 @@ components:
...
       - name: uds-grafana-config
         namespace: grafana
         version: 0.1.0
         localPath: chart
+    images:
+      - docker.io/grafana/grafana:10.1.5
+      - docker.io/curlimages/curl:7.85.0
+      - docker.io/grafana/grafana-image-renderer:latest
+      - docker.io/library/busybox:1.31.1
+      - docker.io/bats/bats:v1.4.1
```
mjnagel:
```suggestion
      - docker.io/library/busybox:1.31.1
```
I don't think we use either of these two - image renderer is disabled and the bats image is for their test framework.

capabilities/loki/chart/values.yaml
```diff
@@ -1,2 +1,2 @@
 grafana:
-  enabled: false
+  enabled: true
```
mjnagel: nit: we could probably remove this value (along with the conditional wrapper around https://github.com/defenseunicorns/uds-core/blob/add_monitoring_and_logging/capabilities/loki/chart/templates/loki-dashboards.yaml). I don't think it provides a ton of value since the configmap won't break anything if deployed without grafana, and we don't want there to be any cases where grafana isn't deployed in our bundle.

capabilities/prometheus-stack/values/values.yaml
```diff
@@ -120,19 +30,8 @@ kube-state-metrics:
     #registry: registry1.dso.mil
     #repository: ironbank/opensource/kubernetes/kube-state-metrics
     #tag: v2.10.0
-  podAnnotations:
-    #PEPR?
-    bigbang.dev/istioVersion: 1.19.0-enterprise
   prometheus:
     monitor:
```
mjnagel: nit: should delete this so that it doesn't null the defaults out.
```suggestion
```

capabilities/prometheus-stack/values/values.yaml
```diff
@@ -157,12 +56,6 @@ kubeStateMetrics:
...
       enabled: false
-#NEEDED? (begin)
-#loki:
-#  enabled: true
-#minioOperator:
-#  enabled: false
-#NEEDED? (end)
 nameOverride: kube-prometheus-stack
 networkPolicies:
   controlPlaneCidr: 0.0.0.0/0
```
mjnagel: nit: this is a carry over from BB, along with the 3 lines after.

capabilities/promtail/chart/templates/istio/peer-authentication.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: security.istio.io/v1beta1
```
mjnagel: nit: this is redundant since the istio-system one applies globally.

capabilities/promtail/values/values.yaml
```diff
@@ -0,0 +1,166 @@
...
+  - mountPath: /var/log
+    name: varlog
+    readOnly: true
+  - mountPath: /etc/machine-id
+    name: machine-id
+    readOnly: true
+    subPath: machine-id
+
+hostname: null
+
```
mjnagel:
```suggestion
```
nit: BB carryover

capabilities/promtail/zarf.yaml
```diff
@@ -1,11 +1,28 @@
...
+        gitPath: charts/promtail
+        valuesFiles:
+          - values/values.yaml
+      - name: uds-promtail-config
+        namespace: promtail
+        version: 0.1.0
+        localPath: chart
+    images:
+      - docker.io/jimmidyson/configmap-reload:v0.8.0
+      - docker.io/busybox:1.33
```
mjnagel:
```suggestion
```
nit: don't think busybox is used?

mjnagel (commented):
Mostly some nits on extra things I noticed when scrolling through. 2 other broad comments probably to track as follow on issues:
- Some of these charts have new versions available upstream vs what is here
- Image stuff...some of the values we have for repo/image aren't really needed but pending flavor discussions might be?

capabilities/grafana/chart/templates/datasources.yaml
zachariahmiller: done

capabilities/grafana/validate.yaml
```diff
@@ -3,5 +3,8 @@ tasks:
     actions:
       - description: Validate...
         cmd: "echo Replace Me"
+
         # wait:
-        #   cluster:
+        #   network:
```
jeff-mccoy: for now we've just been adding a healthcheck to some k8s resource, we can do the network tests in another PR

capabilities/neuvector/zarf.yaml
```diff
@@ -20,26 +20,22 @@ components:
...
-          - values/neuvector-minimal-values.yaml
+          - values/values.yaml
       - name: monitor
         url: https://neuvector.github.io/neuvector-helm/
         version: 2.6.4
         namespace: neuvector
         gitPath: charts/monitor
         valuesFiles:
           - values/neuvector-monitor-values.yaml
+      - name: uds-neuvector-config
```
jeff-mccoy, line 31: not sure how I feel about combining the app portion with config things in one component. Realize it saves an extra import line, but don't know that this is necessarily the right move.

zachariahmiller: I dont have a problem changing it back in each place, but at least in the current state it didnt seem like we are gaining anything from the config charts being separate except more import statements

capabilities/promtail/values/values.yaml
```diff
@@ -0,0 +1,164 @@
...
+    name: varlog
+    readOnly: true
+  - mountPath: /etc/machine-id
+    name: machine-id
+    readOnly: true
+    subPath: machine-id
+
+resources:
+  limits:
+    cpu: 500m
```
jeff-mccoy, line 130: these limits seem a bit .... low? Are we sure this will be fine long-term?

capabilities/grafana/validate.yaml
zachariahmiller: Do you want me to add the resource checks in the validate for each of the modified apps now? I can do that.

capabilities/promtail/zarf.yaml
```diff
@@ -1,11 +1,27 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/defenseunicorns/zarf/main/zarf.schema.json
```
jeff-mccoy: we don't put these tags in zarf.yaml anymore as they are loaded via the project root config in vscode

capabilities/promtail/zarf.yaml
```diff
@@ -1,11 +1,27 @@
...
+    description: "Deploy Promtail"
+    charts:
+      - name: promtail
+        url: https://grafana.github.io/helm-charts/
+        version: 6.15.3
+        namespace: promtail
+        gitPath: charts/promtail
+        valuesFiles:
+          - values/values.yaml
+      - name: uds-promtail-config
```
jeff-mccoy, line 19: same comment regarding app vs config as their own components. maybe it's fine--but we need to decided and try to stick to a common pattern if we can I think

release-please-config.json
```diff
@@ -16,8 +16,13 @@
         "packages/standard/zarf.yaml",
         "bundles/k3d-istio/uds-bundle.yaml",
         "bundles/k3d-standard/uds-bundle.yaml",
+<<<<<<< HEAD
```
jeff-mccoy: merge conflicts

jeff-mccoy (dismissed):
I didnt take a super hard look at the netpols since I'm working on that separately right now.

capabilities/promtail/values/values.yaml
zachariahmiller: definitely not sure. That is what is set in the dubbd k3d deployment. I personally really like the idea of trying to find a better way to dynamically handle resource allocation with vpa or whatever else when it becomes an issue

capabilities/promtail/zarf.yaml
zachariahmiller: no problem i already had them there i can remove

capabilities/promtail/values/values.yaml
zachariahmiller: definitely not sure. That is what is set in the dubbd k3d deployment. I personally really like the idea of trying to find a better way to dynamically handle resource allocation with vpa or whatever else when it becomes an issue

capabilities/neuvector/zarf.yaml
mjnagel: In general I like combining the config chart + app into a single component for simplicity of imports, but happy to hear compelling reasons on why that might not be a good option.

capabilities/promtail/values/values.yaml
jeff-mccoy: I think to be safe we should just set higher limits to avoid a needless pod termination doom-loop.

capabilities/grafana/validate.yaml
jeff-mccoy: yes plz

capabilities/neuvector/zarf.yaml
jeff-mccoy: yeah it might be fine, idk--worth just deciding one way and moving on. For Istio for now, we have to keep things separate because of an outstanding helm bug. But that's not a reason to force everything else that way. Originally, components were designed to logically group apps/behavior--so you could argue that one component is reasonable in that regard. It certainly would help the imports here.

zachariahmiller: per several discussions, leaving the config-charts as part of the main components

capabilities/grafana/validate.yaml
zachariahmiller: threw in some checks on the pods. maybe should have used the deployments, stateful sets, etc instead. Assuming it makes sense can flesh out more/better checks in the future.

capabilities/promtail/values/values.yaml
zachariahmiller: i increased the limits for now, without data to back it though it feels pretty arbitrary.

zachariahmiller:
assuming the ci passes this should be good to go

mjnagel (dismissed):
LGTM

jeff-mccoy:
Added the debug stuff because something looked like it was hanging and this time it passed, probably something we need to wait and see if we can reproduce with debug logs next time. 

Screenshot 2023-11-16 at 12 16 12 AM
Screenshot 2023-11-16 at 12 15 40 AM

jeff-mccoy (approved):

Noxsios (approved):
lgtm

PR #47, opened 2023-11-17
refactor: utilize uds task includes

jeff-mccoy:
## Description

Use task includes where possible (env vars cant be passed into includes inline so publish.yaml couldn't use it right now). Also provides a top-level task file to do clean commands such as:

```bash
uds run test-uds-core
```

## Type of change

## Checklist before merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ The core applications are:
 
 ### Quickstart
 
-A common need is bootstrapping a new UDS Core environment for development or testing. The commands below will deploy the latest version of UDS Core. See the remaining sections for more details if the different bundles & packages available.
+A common need is bootstrapping a new UDS Core environment for development or testing. The commands below will deploy the latest version of UDS Core. See the remaining sections for more details on the different bundles and packages available.
 
 ```bash
 # ARM version
--- a/README.md
+++ b/README.md
@@ -37,10 +37,14 @@ The core applications are:
 
 ### Quickstart
 
-A common need is bootrapping a new UDS Core environment for development or testing. The command below will deploy a local K3d cluster with UDS Core on a Mac M1. See the remaining sections for more details if the different bundles & packages available.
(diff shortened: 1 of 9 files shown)


jeff-mccoy:
confirmed the publish bundle latest tagging action works (to match the README update):

Screenshot 2023-11-17 at 3 08 05 AM

README.md
```diff
@@ -37,10 +37,14 @@ The core applications are:
 
 ### Quickstart
 
-A common need is bootrapping a new UDS Core environment for development or testing. The command below will deploy a local K3d cluster with UDS Core on a Mac M1. See the remaining sections for more details if the different bundles & packages available.
+A common need is bootstrapping a new UDS Core environment for development or testing. The commands below will deploy the latest version of UDS Core. See the remaining sections for more details if the different bundles & packages available.
```
mjnagel:
```suggestion
A common need is bootstrapping a new UDS Core environment for development or testing. The commands below will deploy the latest version of UDS Core. See the remaining sections for more details on the different bundles and packages available.
```

tasks.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - task: create:single-package
+
+  - name: create-standard-package
+    actions:
+      - task: create:standard-package
+
+  - name: deploy-single-package
+    actions:
+      - task: deploy:single-package
+
```
mjnagel, line 23: Would it be valuable to add create/deploy:k3d-standard-bundle here? Might not be used in CI but I definitely run those tasks locally for dev and having them in this top level would be nice.

mjnagel (commented):
LGTM overall.

tasks.yaml
jeff-mccoy: Yeah mainly was thinking of my own flow, but you're probably right. I just run the test all if I've made a lot of changes otherwise I'm mainly doing single source package testing.

mjnagel: I actually totally missed the addition/change to `test-uds-core` - that matches what I would want to run locally too so I'm okay with leaving as is.

mjnagel (approved):

zachariahmiller (approved):
🚀

PR #50, opened 2023-11-22
feat: introduce Pepr common policies

jeff-mccoy:
## Description

Introduce Pepr Validations to replace Kyverno.

## Type of change

## Checklist before merging


src/policies/security.ts
```diff
@@ -0,0 +1,213 @@
...
+When(a.Pod)
+  .IsCreatedOrUpdated()
+  .Validate(request => {
+    if (exemptDropAllCapabilities(request)) {
+      return request.Approve();
+    }
+
+    const hasInvalidCapabilities = containers(request)
+      .flatMap(c => c.securityContext?.capabilities?.drop || [])
+      .find(c => c !== "ALL");
```
mjnagel: The Big Bang policy actually enforced that drop `ALL` was always present. This would only enforce `ALL` being there if `capabilities.drop` is present. This lines up with what the k8s docs reference - https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted . Edge cases here:
- `drop` is not present, pod would be allowed
- `ALL` is in the drop list, but something else is too, pod would be blocked (and shouldn't be)

It's also probably something we could/should mutate in there 🤔?

src/policies/README.md
```diff
@@ -0,0 +1,34 @@
...
+|   [ ]    | [Reqire Non-root Group](https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/require-non-root-group.yaml)                                        | Pod Security Standards (Restricted) | high     | Pod                 | Following the least privilege principle, access to the root group ID should be forbidden in containers. This policy ensures containers are running with groups > 0.                                                                                                                                                                                                                                                                                |
```
zachariahmiller: nit: In multiple places, its "Require"

src/policies/security.ts
```diff
@@ -0,0 +1,213 @@
...
+ * defined at both the Pod level and individual container levels (including initContainers and ephemeralContainers).
+ *
+ * @related https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/require-non-root-user.yaml
+ */
+When(a.Pod)
+  .IsCreatedOrUpdated()
+  .Validate(request => {
+    // Check pod securityContext
+    const podCtx = request.Raw.spec?.securityContext || {};
+    if (podCtx.runAsNonRoot || podCtx.runAsUser) {
```
mjnagel: This conditional doesn't look right. It should be checking for `!podCtx.runAsNonRoot` or `podCtx.runAsUser == 0` I think. There's also the "edge" case of not setting a runAsUser, which could also be running as root (depending on cluster config and how the image is built).

I think really we need to validate that:
- runAsUser is set || runAsNonRoot = true
- runAsUser (if set) is non-0

mjnagel (changes requested):
I think there's definitely more to review here on the actual policy contents but wanted to comment here with an overarching issue on policies not being enforced.

I crafted malicious pods + services that should have been blocked by every single policy. What I found was that only the final policy for pods in `networking` (hostport) actually blocked the pod. The same was true for service, only the final policy for service in `networking` (nodeport) blocked the service. It appears as if only the final approve/deny for the pod actually affects the admission. Other policies that the pod violates showed as logs in pepr, but the pod/svc were still admitted. I did some testing to confirm and reordering policies in `networking.ts` changes what is actually enforced (always the last policy). I also modified the `index.ts` to change the import order and confirmed that the order of the imports has an effect on what is enforced.

TLDR: Last policy, in the last import in `index.ts` is the only one that gets properly enforced.

src/policies/README.md
jeff-mccoy: lol thanks gpt

src/policies/security.ts
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * @related https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/restrict-seccomp.yaml
+ */
+When(a.Pod)
+  .IsCreatedOrUpdated()
+  .Validate(request => {
+    const authorizedTyes = ["RuntimeDefault", "Localhost"];
+
+    // Check Pod level security context
+    const podProfileType = request.Raw.spec?.securityContext?.seccompProfile?.type || "";
```
mjnagel: I think the `|| ""` is problematic here since the check on the next line requires it to be in the allow list, which `""` is not. Not sure the best fix, adding it to the list or changing this to a default of `RuntimeDefault`?

src/policies/security.ts
```diff
@@ -0,0 +1,233 @@
...
+ * The SecComp profile should not be explicitly set to Unconfined. This policy ensures
+ * that the `seccompProfile.Type` is undefined or restricted to `RuntimeDefault` or
+ * `Localhost`. Applies to Pods and all types of containers within them.
+ *
+ * @related https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/restrict-seccomp.yaml
+ */
+When(a.Pod)
+  .IsCreatedOrUpdated()
+  .Validate(request => {
+    const authorizedTyes = ["RuntimeDefault", "Localhost"];
```
mjnagel: nit: missing a `p` in `authorizedTyes`

src/policies/security.ts
```diff
@@ -0,0 +1,233 @@
...
+    if (!authorizedTyes.includes(podProfileType)) {
+      return request.Deny(`The seccomp profile at Pod level is not in the allowed list.`);
+    }
+
+    const hasOtherSecProfileType = containers(request)
+      .flatMap(c => c.securityContext?.seccompProfile)
+      .find(profile => !authorizedTyes.includes(profile?.type || ""));
+
+    if (hasOtherSecProfileType) {
+      return request.Deny(`Unauthorized seccomp profile type.`);
```
mjnagel: nit: Should have a more informative deny along the lines of `One or more containers in the Pod have ...`

src/policies/security.ts
```diff
@@ -0,0 +1,233 @@
...
+  .IsCreatedOrUpdated()
+  .Validate(request => {
+    if (exemptSELinuxTypes(request)) {
+      return request.Approve();
+    }
+
+    const allowedSeLinuxTypes = ["container_t", "container_init_t", "container_kvm_t"];
+
+    // Check Pod level security context
+    const podSeLinuxType = request.Raw.spec?.securityContext?.seLinuxOptions?.type || "";
```
mjnagel: Same issue with the default here, since it is not in the allow list pods get blocked if they don't explicitly set this.

src/policies/security.ts
```diff
@@ -0,0 +1,233 @@
...
+    if (!allowedSeLinuxTypes.includes(podSeLinuxType)) {
+      return request.Deny(`Setting SELinux type at Pod level is restricted to the allowed list.`);
+    }
+
+    const hasOtherSELinuxType = containers(request)
+      .flatMap(c => c.securityContext?.seLinuxOptions?.type)
+      .find(t => !allowedSeLinuxTypes.includes(t || ""));
+
+    if (hasOtherSELinuxType) {
+      return request.Deny(`Unauthorized SELinux type.`);
```
mjnagel: nit: same comment on listing that containers are the problem to be more verbose.

src/policies/storage.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    // Check all volumes in the pod spec, if any
+    for (const volume of volumes(request)) {
+      // Get the volume type, which will be the only key in the volume object other than "name"
+      const volumeType = Object.keys(volume).find(key => key !== "name") || "unknown";
+
+      // If the volume type is not in the allowed list, deny the request
+      if (!allowedVolumeTypes.includes(volumeType)) {
+        return request.Deny(`Volume ${volume.name} has an invalid volume type of '${volumeType}.`);
```
mjnagel:
```suggestion
        return request.Deny(`Volume ${volume.name} has a disallowed volume type of '${volumeType}'.`);
```
nit: also maybe should log allowed types?

src/policies/security.ts
mjnagel: Might also be valuable to list the specific container(s) that violate, that was always a painpoint I had with kyverno, etc (just seeing a generic "your pod is bad". Not sure if that is too much to add in the logic though.

src/policies/networking.ts
```diff
@@ -0,0 +1,97 @@
...
+ * do not use the NodePort type for enhanced security.
+ * 
+ * @related https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/disallow-nodeport-services.yaml
+ */
+When(a.Service)
+  .IsCreatedOrUpdated()
+  .Validate(request => {
+    // If the service is of type NodePort, deny the request.
+    if (request.Raw.spec?.type === "NodePort") {
+      return request.Deny("Services of type NodePort are not allowed.");
```
mjnagel: nit: the two deny messages for these service validates (externalname and nodeport) are formatted differently. would be good to align them by changing this to `NodePort services are not allowed.` or the reverse to the other.

src/policies/networking.ts
```diff
@@ -0,0 +1,97 @@
...
+  .Validate(request => {
+    // Check all containers in the pod spec, and find the first one that has a host port, if any
+    const hasHostPort = containers(request)
+      .flatMap(c => c.ports || [])
+      .find(p => p.hostPort);
+
+    // If the container has a host port, deny the request
+    if (hasHostPort) {
+      return request.Deny(
+        `Host ports are not allowed. Only exempted resources are allowed to use host ports.`,
```
mjnagel: nit: we don't mention exempted resources in all the deny messages, would be good to standardize on whether we add that phrase or not (technically there are no exemptions for this one currently either).

src/policies/exemptions/storage.ts
```diff
@@ -0,0 +1,42 @@
...
+   * `/proc`: monitoring of processes for malicious activity
+   * `/sys/fs/cgroup`: important files the controller wants to monitor for malicious content
+   * https://github.com/neuvector/neuvector-helm/blob/master/charts/core/templates/enforcer-daemonset.yaml#L108
+   */
+  neuvector.controller,
+  neuvector.enforcer,
+
+  /**
+   * NEEDS FURTHER JUSTIFICATION
+   * Promtail requires HostPath volume types
```
mjnagel:
```suggestion
   * Promtail mounts the following hostPaths:
   * - `/var/log/pods`: to tail pod logs
   * - `/var/lib/docker/containers`: to tail container logs
   * - `/run/promtail`: for Promtail's buffering and persistent state
   * Since logs can have sensitive information, it is better to exclude
   * Promtail from the policy than add the paths as allowable mounts
```
Copy-pasta exemption justification from above, should be the same reason for both.

src/policies/exemptions/security.ts
```diff
@@ -0,0 +1,12 @@
+import { registerExemptions } from ".";
+import { neuvector, promtail } from "./matchers";
+
+export const exemptDropAllCapabilities = registerExemptions([
+  neuvector.enforcer,
+  neuvector.prometheus,
+]);
```
mjnagel, line 14: Curious on these exemptions - should be able to drop all for these ones too, I don't think there is anything by default in the upstream or BB chart that would make this not-possible. We may have to `add` some capabilities but drop itself should be fine.

src/policies/security.ts
```diff
@@ -0,0 +1,233 @@
...
+When(a.Pod)
+  .IsCreatedOrUpdated()
+  .Validate(request => {
+    const hasPrivilegedContainer = containers(request)
+      .flatMap(c => c.securityContext || {})
+      .find(ctx => ctx.allowPrivilegeEscalation || ctx.privileged);
+
+    if (hasPrivilegedContainer) {
+      return request.Deny(
+        "Privilege escalation is disallowed. The allowPrivilegeEscalation and privileged fields must be undefined or set to `false`.",
```
mjnagel: UX nit: I think there's a mix of `must not be set to true` (example `hostNetwork` policy) and `must be undefined or set to false` would probably be good to standardize on one (my preference is the second).

src/policies/security.ts
```diff
@@ -0,0 +1,233 @@
...
+      return request.Approve();
+    }
+
+    const hasInvalidCapabilities = containers(request)
+      .flatMap(c => c.securityContext?.capabilities?.drop || [])
+      .find(c => c !== "ALL");
+
+    if (hasInvalidCapabilities) {
+      return request.Deny(
+        `Containers must drop all Linux capabilities by setting spec.containers[*].securityContext.capabilities.drop to 'ALL'.`,
```
mjnagel: nit: this could also be init/ephemeral containers which the deny message doesn't capture.

src/policies/security.ts
```diff
@@ -0,0 +1,233 @@
...
+    if (exemptSELinuxTypes(request)) {
+      return request.Approve();
+    }
+
+    const allowedSeLinuxTypes = ["container_t", "container_init_t", "container_kvm_t"];
+
+    // Check Pod level security context
+    const podSeLinuxType = request.Raw.spec?.securityContext?.seLinuxOptions?.type || "";
+    if (!allowedSeLinuxTypes.includes(podSeLinuxType)) {
+      return request.Deny(`Setting SELinux type at Pod level is restricted to the allowed list.`);
```
mjnagel: nit: log the allow list with the denial?

mjnagel (commented):
Number of nits and a few logic issues. Went through each policy individually to validate they worked as expected.

package.json
```diff
@@ -17,17 +17,30 @@
     "uuid": "uds-core",
     "onError": "audit",
     "alwaysIgnore": {
-      "namespaces": [],
+      "namespaces": [
+        "local-path-storage",
```
mjnagel, line 21:
```suggestion
```
I think can be removed since we moved local path to the uds-dev-stack?

src/policies/README.md
```diff
@@ -0,0 +1,34 @@
...
+|   [ ]    | [Disallow SELinux Options](https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/disallow-selinux-options.yaml)                                   | Pod Security Standards (Baseline)   | high     | Pod                 | SELinux options can be used to escalate privileges. This policy ensures that the `seLinuxOptions` specified are not used.                                                                                                                                                                                                                                                                                                                          |
```
mjnagel:
```suggestion
|    ✅    | [Disallow SELinux Options](https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/disallow-selinux-options.yaml)                                   | Pod Security Standards (Baseline)   | high     | Pod                 | SELinux options can be used to escalate privileges. This policy ensures that the `seLinuxOptions` specified are not used.                                                                                                                                                                                                                                                                                                                          |
```
Looks to be implemented?

src/policies/README.md
```diff
@@ -0,0 +1,34 @@
...
+|   [ ]    | [Restrict hostPath Volume Mountable Paths](https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/restrict-host-path-mount.yaml)                   | Best Practices (Security)           | medium   | Pod                 | hostPath volumes consume the underlying node's file system. If hostPath volumes are not universally disabled, they should be restricted to specific host paths to prevent access to sensitive information. This policy ensures that hostPath volume paths are in the allowed list.                                                                                                                                                                 |
```
mjnagel, line 26: I think this one is implemented?

```suggestion
|    ✅    | [Restrict hostPath Volume Mountable Paths](https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/restrict-host-path-mount.yaml)                   | Best Practices (Security)           | medium   | Pod                 | hostPath volumes consume the underlying node's file system. If hostPath volumes are not universally disabled, they should be restricted to specific host paths to prevent access to sensitive information. This policy ensures that hostPath volume paths are in the allowed list.                                                                                                                                                                 |
```

src/policies/networking.ts
mjnagel:
```suggestion
        `Host ports are not allowed.`,
```
Don't know if we want to add any "debug" to this one, i.e. `Found: {"name":"my-container","ports":[{hostPort: 80}]}` but at least can drop the reference to exemptions here since there are no hostport exemptions.

src/policies/networking.ts
mjnagel:
```suggestion
      return request.Deny("NodePort services are not allowed.");
```
Standardizing to align with the external name one:
```
Error from server: error when creating "test.yaml": admission webhook "pepr-uds-core.pepr.dev" denied the request: ExternalName services are not allowed.
Error from server: error when creating "test.yaml": admission webhook "pepr-uds-core.pepr.dev" denied the request: Services of type NodePort are not allowed.
```

mjnagel (commented):
Overall LGTM, love the tests. Few small comments with suggestions for your review.

package.json
mjnagel: My mistake - haven't updated to uds-k3d 0.2.0 yet so this should stay.

src/policies/README.md
jeff-mccoy: technically no, we only do writeable paths, but his is one of the examples I'd like to validate (near duplicate of another mountable path one)

mjnagel: Does [this policy](https://github.com/defenseunicorns/uds-core/blob/common-pepr/src/policies/storage.ts#L16-L50) not block hostPath type volumes completely? Might just not be thinking about this right but the source kyverno policy blocks any hostpath volumes in a similar way. The difference is just that the BB policy [allows them if they are on an allowed path](https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/restrict-host-path-mount.yaml#L42) (defaulting to nothing allowed).

mjnagel: I think I'm following maybe - this PR has a validation for volume type, but technically not the same as the BB validation of hostPaths being on a given path. I'm good with leaving this one for now and we can review in the future whether we need something like it.

mjnagel (approved):
:shipit:

PR #63, opened 2023-12-11
feat: add registry1 zarf flavors to uds core

### LINKED ISSUE #60: Add registry1 zarf flavors to uds core
Now that we have uds-core deploying with upstream images, we need a variant for IronBank only images. The repo needs to have all the packages updated with zarf component flavor functionality to support this

i can set UDS_FLAVOR=registry1 before running my package create to get a zarf package variant with only iron bank images

zachariahmiller:
## Description
Adds zarf flavors for upstream images and IB. Includes refactoring of zarf packages using composability to cleanup the config and keep things dry and refactoring of the values files to image flavors specific + common. 

Both flavors deploy successfully on their respective architectures. On arm64 (or at least darwin arm64) when using the registry1(x86) flavor it fails at metrics server regarding the istio init container, which seems to be an issue [emulating x86 on mac](https://github.com/istio/istio/issues/36762#issuecomment-1034070488). 

## Type of change

## Checklist before merging


src/promtail/values/registry1-values.yaml
```diff
@@ -0,0 +1,165 @@
...
+            separator: /
+            source_labels:
+            - namespace
+            - app
+            target_label: job
+          - action: replace
+            source_labels:
+            - __meta_kubernetes_pod_name
+            target_label: pod
+          - action: replace
```
zachariahmiller: @rjferguson21 pointed out i missed this. Deleting the redundant values when i update the merge conflicts

src/neuvector/values/registry1-values.yaml
```diff
@@ -0,0 +1,108 @@
...
+    type: ClusterIP
+  secret:
+    enabled: true
+    data:
+      oidcinitcfg.yaml:
+        always_reload: true
+        client_id: null
+        client_secret: null
+        default_role: null
+        enable: true
```
zachariahmiller: missed cleanup here too. resolving alongside the merge conflict.

.github/workflows/test.yaml
```diff
@@ -39,13 +42,20 @@ jobs:
...
+        uses: docker/login-action@v3
+        with:
+          registry: registry1.dso.mil
+          username: ${{ secrets.IRON_BANK_ROBOT_USERNAME }}
+          password: ${{ secrets.IRON_BANK_ROBOT_PASSWORD }}
+
       - name: Test a single source package
-        if: ${{ inputs.package != 'all' }}
-        run: uds run test-single-package
+        if: ${{ inputs.package != 'all' && matrix.flavor == 'upstream'}}
```
mjnagel: Don't have a _strong_ opinion here but I wonder if it's more valuable to switch this to use registry1 for the single package tests (instead of upstream). My thinking here is that people are more likely to be doing upstream tests locally (and those are more likely to "just work" since its all upstream), so those pieces are generally going to be tested before being pushed. With that in mind...short circuiting CI on registry1 might make more sense?

.yamllint
```diff
@@ -0,0 +1,38 @@
+yaml-files:
+  - '**/*.y*ml'
+  - '.yamllint'
+
+# Ignore files from upstream
+ignore:
+  - 'k3d/local/manifests/metallb/metallb-native.yaml'
+  - '**/.terraform/**'
+  - '**/chart/templates**'
```
mjnagel, line 9:
```suggestion
  - '**/chart/templates**'
  - 'node_modules/**'
```
I had a bunch of bogus linting issues locally on the `node_modules` folder.

src/istio/zarf.yaml
```diff
@@ -1,7 +1,7 @@
 kind: ZarfPackageConfig
```
mjnagel, line 1: Let me know what you think about these changes - https://github.com/defenseunicorns/uds-core/pull/108 - I think this feels closer to the common pattern with other src packages?

tasks/create.yaml
```diff
@@ -5,7 +5,7 @@ tasks:
       - task: pepr-build
 
       - description: "Create the UDS Core Standard Zarf Package"
-        cmd: "zarf package create packages/standard --confirm"
+        cmd: "zarf package create packages/standard --confirm  --flavor ${UDS_FLAVOR}"
```
mjnagel: Wondering if we should put a default flavor to simplify the dev experience (i.e. set default `UDS_FLAVOR` to `upstream` so that local dev can remain the same as previous experience.

.github/workflows/tag-and-release.yml
```diff
@@ -44,8 +47,15 @@ jobs:
...
+      - name: Login to Registry1
+        uses: docker/login-action@v3
+        with:
+          registry: registry1.dso.mil
+          username: ${{ secrets.IRON_BANK_ROBOT_USERNAME }}
+          password: ${{ secrets.IRON_BANK_ROBOT_PASSWORD }}
+
       - name: Publish UDS Zarf Package
-        run: uds run -f tasks/publish.yaml packages
+        run: uds run -f tasks/publish.yaml packages --set UDS_FLAVOR=${{ matrix.flavor }}
```
mjnagel: I think for the bundle publishing we will need something special for the naming, but might be wrong. I don't think uds-cli appends the flavor to tags yet?

.github/workflows/test.yaml
zachariahmiller: I do agree we are more likely to catch an issue on the registry1 tests. What is in here now is what we had agreed upon previously. 
I'm inclined to possibly just do both since the single capability tests are on public (free) runners and will still provide the short circuit when needed, with slightly more feedback, slightly less CI logic and no expense.

src/istio/zarf.yaml
zachariahmiller: merged. I agree with that pattern i had started going down that road and then got to the gateways which didnt fit and figure i'd try to get feedback first.

tasks/create.yaml
zachariahmiller: I would like a default. I'll test locally to see if setting UDS_FLAVOR as a top level variable works as expected to facilitate this and update accordingly if it does, otherwise we can get that sorted on the uds cli side and then add it.

.github/workflows/tag-and-release.yml
zachariahmiller: yeah... I wasnt sure on the state of that or planned state for the bundle publishing. Easy enough to solve for either way.

zachariahmiller:
I'll work on resolving the merge conflicts from the operator PR, testing bundle publish flavor behavior and setting a default flavor in the tasks tonight and tomorrow morning.

Intermittent neuvector deploy is still an issue as they dont actually expose security context in the monitor chart and it sometimes happens to work by continuing past it anyways. Not sure what the immediate fix for this is, but @mjnagel put in a pr to the upstream chart to fix on their side that we can use once it gets added. https://github.com/neuvector/neuvector-helm/pull/355

.github/workflows/tag-and-release.yml
zachariahmiller: I verified uds wasnt doing anything with the flavor. Since there isnt a way to override the tag on publish, I updated to make the registry path .../bundles/${FLAVOR}/

mjnagel: While I don't love how long that makes the artifact name/path (ex: `ghcr.io/defenseunicorns/packages/uds/bundles/registry1/k3d-core-istio:0.8.0`) I think this is a simple and reasonable solve.

One way I was considering that might be a bit more in line with how zarf handles flavors would be to dynamically tack on the flavor to the tag/version during the publish step (`yq -i '.metadata.version = "0.8.0-registry1"' bundles/k3d-standard/uds-bundle.yaml` before build/publish). While I like the outcome of that...it feels pretty hacky 🤷. We could do the same thing "statically" by duplicating the bundle files (committing with the flavor in the version), but that's not very DRY.

All that to say...I'm good with the way you handled it. I've probably already thought too much about something that isn't meant to be consumed outside of dev 😅

zachariahmiller: Yeah so I'm not a huge fan of the way i did it being different from how zarf tags it autotically, but figured it was between that, waiting for a way for UDS to handle flavors automatically or what you suggested above. I think I slightly prefer the outcome of your suggestion because it matches how zarf gets published but as you said the yq approach is hacky and the static approach feels excessive. If anyone else has strong opinions I'd welcome them to weigh in :D

zachariahmiller: @UncleGedd relevant discussion on uds and flavors.

zachariahmiller:
Before being able to merge, I think we'll need to change the repo settings for expected tests from this:
image

to these:
image

.github/workflows/test.yaml
zachariahmiller: based on how this ci testing has been going im leaning towards doing both sets of single capability tests...

.github/workflows/tag-and-release.yml
UncleGedd: Appreciate the discussion @zachariahmiller @mjnagel ! Would one of you mind opening an issue in UDS CLI to continue the conversation and discuss proposed options?

zachariahmiller:
At this point the only outstanding thing should be that loki is failing on upgrade because the gateway wont update. Deleting the old one successfully continues the upgrade to completion.
```
message: '0/1 nodes are available: 1 node(s) didn''t match pod anti-affinity rules.  
preemption: 0/1 nodes are available: 1 No preemption victims found for incoming pod
```

this might just be a single node issue. Diving into the chart to see if we can change the restart policy or something for single node.

.github/workflows/tag-and-release.yml
mjnagel: Opened https://github.com/defenseunicorns/uds-cli/issues/332 to have this discussion. Going to resolve this thread for now since I think we have a path forward and as mentioned before these bundles are pretty explicitly for dev use and changing the name down the road should not be a big deal.

README.md
```diff
@@ -28,7 +28,7 @@ The core applications are:
 | Dependency                                                     | Minimum Version |
 | -------------------------------------------------------------- | --------------- |
 | [Zarf](https://github.com/defenseunicorns/zarf/releases)       | 0.32.1          |
-| [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) | 0.5.3           |
+| [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) | 0.6.1           |
 | [NodeJS](https://nodejs.org/en/download/)                      | LTS or Current  |
 
 <!-- endtable -->
```
mjnagel, line 34: I can't comment on the actual lines but we'll want to update the quickstart `uds deploy` commands to reference the new bundle names with flavors.

src/neuvector/common/zarf.yaml
```diff
@@ -0,0 +1,35 @@
...
+        valuesFiles:
+          - ../values/values.yaml
+      # Temporarily disabled until we can get the monitor chart working via https://github.com/neuvector/neuvector-helm/pull/355
+      # - name: monitor
+      #   url: https://neuvector.github.io/neuvector-helm/
+      #   version: 2.6.4
+      #   namespace: neuvector
+      #   gitPath: charts/monitor
+      #   valuesFiles:
+      #     - ../values/monitor-values.yaml
```
mjnagel: I'm *okay* with this but in my head we could just put something like this in the registry1 values file:

```yaml
exporter:
  enabled: false
```

Ideally less churn/changes when we can re-enable it?

zachariahmiller: thats fine. I did it this way because it seemed like the most obvious/apparent way to see that it was done and why. Removing the comments and updating the values with the comment on there as a note

README.md
zachariahmiller: done

mjnagel (dismissed):
LGTM! Great work on making this more flavorful.

.github/workflows/test.yaml
zachariahmiller: we're doing both after the workflow refactor.

zachariahmiller:
@jeff-mccoy This is good to go, the only thing is the required check no longer exists and needs to be updated in the settings. 
image

I'm able to make the changes and merge, but you had said you wanted to review. PTAL or let me know if you are good with the extensive review of this with @mjnagel that has already happened so i can move forward with merging this.

.yamllint
```diff
@@ -0,0 +1,39 @@
+yaml-files:
```
jeff-mccoy, line 1: not familiar with this config file. can you esplain plz?

package.json
```diff
@@ -36,6 +36,7 @@
   },
   "devDependencies": {
     "@jest/globals": "29.7.0",
+    "@types/jest": "^29.5.11",
```
jeff-mccoy: not a pinned dep, also why is this needed here? I don't see any other TS changes anywhere.

tasks/lint.yaml
```diff
@@ -0,0 +1,8 @@
+tasks:
+  - name: run
+    description: "Lint and fix issues in the repo"
+    actions:
+      - description: "Pepr Format"
+        cmd: npx pepr format
+      - description: yaml lint
```
jeff-mccoy, line 18: still not familiar with this tool or where it is being installed. Also will this have issue with prettier?

tasks/deploy.yaml
```diff
@@ -26,8 +26,12 @@ tasks:
 
   - name: latest-package-release
     actions:
+      - description: "Get latest tag version from OCI"
```
jeff-mccoy, line 29: can you explain this change, seems unrelated to flavors?

tasks/publish.yaml
```diff
@@ -14,50 +17,62 @@ tasks:
...
 
       - description: "Tag the latest bundles"
         cmd: |
           set -e
-          pkgPath="ghcr.io/defenseunicorns/packages/uds/bundles/k3d-core"
-          zarf tools registry copy ${pkgPath}:${VERSION}-arm64 ${pkgPath}:arm64
+          pkgPath="ghcr.io/defenseunicorns/packages/uds/bundles/${FLAVOR}/k3d-core"
+          # dont tag arm64 for registry1 since IB images are only amd64
+          if [ "${FLAVOR}" != "registry1" ]; then 
+            zarf tools registry copy ${pkgPath}:${VERSION}-arm64 ${pkgPath}:arm64
```
jeff-mccoy, line 75: I'm not totally sure we really need "latest" shortcuts for the other flavors besides upstream since these are really just meant for quick testing/demo

jeff-mccoy (changes requested):
Left a few comments, but overall looks pretty good. Couple questions:

1. How will this work with Renovate with the mess of version tracking between IB and upstream?

2. Not sure we should be testing both flavors for single-app tests are those are really just circuit-breaker tests to speed up failure feedback.

jeff-mccoy:
these skipped tests also look pretty ugly, think we need to address that as well. 

Screenshot 2024-01-19 at 1 11 25 AM

.yamllint
zachariahmiller: I added a [lint task](https://github.com/defenseunicorns/uds-core/blob/b7a41e8acbcd9a5b717ba7d6304ec66cebd54dc5/tasks/lint.yaml#L8) that runs as yaml lint check and npx pepr format. This is the config file for the yaml lint. It is the same yam,l lint we use in [dubbd](https://github.com/defenseunicorns/uds-package-dubbd/blob/642e948ebdc67952d00ed6dfeb8088f326d9e863/.yamllint)

tasks/deploy.yaml
zachariahmiller: sure. The upgrade process got merged into main prior to me being able to get this merged. latest-package-release is a part of that change, not flavors.

tasks/publish.yaml
zachariahmiller: I dont really think we need them either way. I was just updating the publish step to have parity with the what was already happening.

IMO doing upstream one way and the other a different way is going to confuse someone trying to consume.

zachariahmiller:
> these skipped tests also look pretty ugly, think we need to address that as well.
> Screenshot 2024-01-19 at 1 11 25 AM

I dont disagree. Again, related to the upgrade. Not sure how to fix this without a more significant refactor of the workflows. IMO that could be addressed in a follow on PR, but whatever the team wants to do.

tasks/deploy.yaml
mjnagel: Full context on this one - https://github.com/defenseunicorns/uds-core/pull/113 - the previous logic was problematic in release please PRs.

zachariahmiller:
> Left a few comments, but overall looks pretty good. Couple questions:
> 
>     1. How will this work with Renovate with the mess of version tracking between IB and upstream?
> 
>     2. Not sure we should be testing both flavors for single-app tests are those are really just circuit-breaker tests to speed up failure feedback.

1. The goal with the way I broke out the repo/image/tag pieces for each flavors values file was supposed to deal with that. When i was testing the renovate config i was testing with a WIP version of flavors. It was mostly working, one of the issues we had was that when image versions werent referenced at all (some of upstream) the renovate pr wouldnt always be consistent, so the current pattern is to make the registry1 anbd upstream image specific pieces match completely so we shouldnt have any weirdness with renovate. I think we should be in a pretty good place as far as that goes and we can always tweak the renovate config as needed. Most of them should update correctly grouped by app including both flavors. I'm sure there will be some exceptions (wouldnt be surprised if something takes an issue with this "registry.k8s.io/ingress-nginx/kube-webhook-certgen:v20221220-controller-v1.5.1-58-g787ea74b6" for example :D ) 

2. Maybe we dont, but those tests are free. There was a discussion earlier in the review on that https://github.com/defenseunicorns/uds-core/pull/63#discussion_r1454036240 i think given it is no cost to us with some potential minor gain (aside from whether we chose to test registry1 or upstream) there isnt really a compelling reason to only test one. As it stands though if we did that, which is what i originally implemented, we'd end up with the same skipped test thing you se now for the upgrades, which would require refactoring the workflow files more.

tasks/lint.yaml
zachariahmiller: I dont think it would have an issue with prettier afaik prettier isnt doing anything with the yaml. That being said im not really familiar with prettier aside from having it baked in to some js projects i've worked on in the past. 

Right now that is a helper task and the answer is pip. This is the linter dubbd uses in their pipeline for yaml https://github.com/defenseunicorns/uds-package-dubbd/blob/642e948ebdc67952d00ed6dfeb8088f326d9e863/.github/workflows/ci-on-pull-request-dubbd-aws.yaml#L28

I had added these because i was getting annoyed i kept having to npx pepr format when pulling in new code even though i wasnt touching pepr code in my changes and wanted to make sure the ton of yaml was all being formatted in my dev.

I think these (or similar) would be added in a separate ticket as part of the ci process.

package.json
zachariahmiller: I can delete this but i dont know where its coming from... To your point no TS changes.

.yamllint
zachariahmiller: fwiw i would have preferred this to function more like pepr format and actually do the formatting but it seemed like there were limited options for that for yaml specifically that wouldnt potentially cause more issues.

jeff-mccoy: sorry I meant, what is the `yamllint` binary? I don't have it on my system or see where we're installing it

tasks/deploy.yaml
jeff-mccoy: cool thx for the clarification

tasks/publish.yaml
jeff-mccoy: k we can save for another time to think about that

tasks/lint.yaml
jeff-mccoy: yeah glad we have a step for it for sure

.yamllint
zachariahmiller: right. I dont install it because we dont install the tools in our run tasks per your request earlier on in this project (like uds and zarf install steps we originally had). yamllint can be installed via `pip install yamllint`. See dubbd's ci call: https://github.com/defenseunicorns/uds-package-dubbd/blob/642e948ebdc67952d00ed6dfeb8088f326d9e863/.github/workflows/ci-on-pull-request-k3d.yaml#L29

If you'd like i can add this as a CI step. I was initially adding so i could check formatting before pushing up changes as i said previously and wasnt trying to make this PR even larger by adding it to CI.

zachariahmiller: added this as a ci step for it does a pepr format --check-validate and the yaml lint check in ci. Locally the task is still there to fix the formatting with pepr.

In the future i'd be happy to evaluate a different tool than yamllint especially if it can auto format in a non-breaking way, which so far the tools i've looked at either wont fix or cause issues with the zarf package schema once formatted.

mjnagel (approved):

zachariahmiller:
@jeff-mccoy here is an example of what we expect the renovate updates to look like more or less. This was in my fork before making the registry1 and upstream values files both have all the image refs rather than relying on it just working in upstream. With the values file changes it will update the upstream-values, registry1-values, and the zarf.yaml images and chart versions in one PR.
image

zachariahmiller:
For clarity/transparency we were trying to determine why the fast failure started only working on a click ops cancel and not a job failure in the updated ci. 

It looks like we have two reasonable options to get it back to that functionality.

Option 1: 
Use something like this composite action but built on bash/sh not powershell https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/cancel-workflow/action.yml#L27 This should work but it looks like from @mjnagel research there are some downsides: 

> This is the simplest way I've seen to cancel other jobs. But the downside of that is that it results in everything showing cancelled so it's a bit harder to find the failed job...

Option 2: Go back to the previous workflow setup (all calling just the one test workflow) while expanding the matrix. I tested/mocked this in an outdated fork as it was the easiest way to do so and it seems to work, downside is that the actual display of the job names and lack of grouping is a little less readable. The mocked up version of that can be seen here:
https://github.com/zachariahmiller/uds-core/pull/42
https://github.com/zachariahmiller/uds-core/actions/runs/7595497465/job/20688138631?pr=42

zachariahmiller:
@mjnagel I dug into the matrix functionality a little more. It looks like with a combination of excludes and includes in the matrix definition, relying on order they are processed per the [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude) seems to work. For example:

```
    strategy:
      matrix:
        package: ${{ fromJSON(needs.check-paths.outputs.packages) }}
        flavor: [upstream, registry1]
        test_type: [install, upgrade]
        exclude:
          - test_type: upgrade
        include:
          - test_type: upgrade
            flavor: registry1
            package: all
          - test_type: upgrade
            flavor: upstream
            package: all
```

Excludes the single upgrade tests and still does both for all while maintaining the fail fast functionality. 

As you can see in the screenshot, the single upgrades were excluded, install and upgrade test ran for "all" and the failed job short circuited the all the sibling workflows spawned from the matrix config while still failing the pipeline.
image
image

mjnagel:
@zachariahmiller I think that makes sense. This slight modification (if it works) would make it a bit simpler I think?

```
    strategy:
      matrix:
        package: ${{ fromJSON(needs.check-paths.outputs.packages) }}
        flavor: [upstream, registry1]
        test_type: [install]
        include:
          - test_type: upgrade
            flavor: registry1
            package: all
          - test_type: upgrade
            flavor: upstream
            package: all
```

If I read the docs right I think that'll work? May even be able to drop the flavor from the include but feel less confident on that.

zachariahmiller:
If that works, agreed. As far as dropping flavor, I don't know if it will... With the combo of include and exclude the flavor was required. IIRC exclude worked on partial match, include didn't.

jeff-mccoy (approved):
🚀

jeff-mccoy:
Branch protection updated, g2g.

Screenshot 2024-01-21 at 3 05 06 AM

PR #66, opened 2023-12-15
feat: add UDS Operator and consolidate UDS Policies

jeff-mccoy:
## Description

This PR consolidates all Pepr capabilities into the folder `src/pepr`. This PR also introduces the UDS Operator, which manages a new CRD `packages` under the group `uds.dev/v1alpha1`.

The UDS Operator manages the lifecycle of UDS Package CRs and their corresponding resources (e.g. NetworkPolicies, Istio VirtualServices, etc.). The operator uses [Pepr](https://pepr.dev) to bind the watch operations to the enque and reconciler. The operator is responsible for:

- enabling Istio sidecar injection in namespaces where the CR is deployed
- establishing default-deny ingress/egress network policies 
- creating a layered allow-list based approach on top of the default deny network policies including some basic defaults such as Istio requirements and DNS egress
- providing targeted remote endpoints network policies such as `KubeAPI` and `CloudMetadata` to make policies more DRY and provide dynamic bindings where a static definition is not possible
- creating Istio Virtual Services & related ingress gateway network policies

### Key Files and Folders

```bash
.
├── controllers          # Core business logic called by the reconciler
│   ├── istio            # Manages Istio VirtualServices and sidecar injection for UDS Packages/Namespace
│   └── network          # Manages default and generated NetworkPolicies for UDS Packages/Namespace
├── crd
│   ├── generated        # Type files generated by `uds run -f src/pepr/tasks.yaml gen-crds`
│   ├── sources          # CRD source files
│   ├── register.ts      # Registers the UDS Package CRD with the Kubernetes API
│   └── validator.ts     # Validates UDS Package CRs with Pepr
├── enque.ts             # Serializes UDS Package CRs for processing by the reconciler
├── index.ts             # Entrypoint for the UDS Operator
└── reconciler.ts        # Reconciles UDS Package CRs via the controllers
```


bburky:
What is the plan for the lifecycle of the CR? I see you set ownerReferences, which should cleanup created resources in the namespace on deletion. But what about updates? `K8s.apply()` is SSA, so I think it will update whole resources on changes, right? (Will it update/replace every field on a second apply, or does it merge with the old changes?)

The scenario I'm most worried about is this doesn't handle deletion on CR change. If you change the network config (but don't delete the CR), some of the old NetworkPolicies will need to be deleted.

You could consider an approach like Flux's kustomize-controller: they store a list of all the created resources in `.status`. On each change compare the new list of resources to the old ones and delete any that no longer should exist (perhaps have things like `networkPolicies()` and `virtualService()` return the list of resources they created or updated as an array of strings). Listing created resources in `.status` is also really nice for debuggability by users.

src/pepr/operator/network/builder.ts
```diff
@@ -0,0 +1,131 @@
+import { V1APIGroup, V1NetworkPolicyPeer, V1NetworkPolicyPort } from "@kubernetes/client-node";
+import { K8s, Log, kind } from "pepr";
+
+import { Allow, UDSPackage } from "../crd";
+import { RemoteGenerated } from "../crd/generated/package-v1alpha1";
+
+let apiServerPeers: V1NetworkPolicyPeer[];
+
+export async function builder(
```
Racer159: Umm... Awkshually "Builder" is the name of a specific design pattern in OOP and may confuse people who see this function named that.

bundles/k3d-istio/uds-bundle.yaml
```diff
@@ -27,11 +27,11 @@ zarf-packages:
             - name: policies
               description: "Minio policies"
               path: policies
-              
+
   - name: init
     repository: ghcr.io/defenseunicorns/packages/init
-    # renovate: datasource=github-tags depName=defenseunicorns/zarf versioning=semver
-    ref: v0.31.1
+    # renovate: datasource=github-tags depName=defenseunicorns/init versioning=semver
```
mjnagel: This change would break the renovate tracking of versioning here. Renovate is expecting a github repo (https://github.com/defenseunicorns/init) that won't exist. We had switched this to reference the zarf repo since that should always be 1:1, but we could switch this to a docker datasource to track the OCI init package directly (would require a change or two to the renovate config as well - happy to help if we want to go that route).

src/pepr/istio/index.ts
```diff
@@ -15,6 +20,8 @@ When(a.Pod)
   .WithLabel("batch.kubernetes.io/job-name")
   .WithLabel("service.istio.io/canonical-name")
   .Watch(async pod => {
+    Log.info(pod, `Processing Pod ${pod.metadata?.namespace}/${pod.metadata?.name}`);
```
mjnagel:
```suggestion
    Log.info(pod, `Processing Pod ${pod.metadata?.namespace}/${pod.metadata?.name} for istio job termination`);
```
Maybe? For clarity around where/for what this is getting processed.

src/pepr/operator/controllers/namespace/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    throw new Error(`Invalid Package definition, missing namespace or name`);
+  }
+
+  const pkgLabel = `uds/${pkg.metadata.name}`;
+  const sourceNS = await K8s(kind.Namespace).Get(pkg.metadata.namespace);
+  const labels = sourceNS.metadata?.labels || {};
+
+  // Ensure Istio injection is enabled and the package label is present
+  if (labels["istio-injection"] !== "enabled" || !labels[pkgLabel]) {
+    labels["istio-injection"] = "enabled";
```
mjnagel: Previously we were only checking if the istio-injection label was missing to allow for explicitly setting disabled on a namespace. 

Was this logic intentionally changed (we now look for "enabled" explicitly and force injection on all UDS Package namespaces)?

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -0,0 +1,145 @@
...
+async function generateKubeAPI(): Promise<V1NetworkPolicyPeer[]> {
+  // Return the cached value if it exists
+  // @todo: evaluate if this ever changes with node autoscaling
+  if (apiServerPeers) {
+    return apiServerPeers;
+  }
+
+  try {
+    // Read the API server endpoints from the cluster
+    const { serverAddressByClientCIDRs } = await K8s(V1APIGroup).Raw("/api");
```
mjnagel: This will always return a single IP (for one of the server nodes), even in multi-server clusters. This is problematic if that single node has any issues, it negates the benefits of load balancing of the kube api (pods will get connection refused on all the other endpoints for the kube api).

We could grab the full list of server addresses from the `kubernetes` endpoint (`kubectl get endpoints kubernetes`) and add `/32` to each, covering all server nodes. I think that would make this more robust, thoughts?

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -0,0 +1,145 @@
...
+  namespace: string,
+  pkg: UDSPackage,
+  policy: Allow,
+  idx: number,
+): Promise<kind.NetworkPolicy> {
+  const pkgName = pkg.metadata!.name!;
+  const target = Object.values(policy.podLabels || ["all-pods"]).join("-");
+
+  // Create a unique name for the NetworkPolicy based on the package name, index, direction, pod labels, and port
+  const name = `allow-${policy.direction}-${target}-${pkgName}-${idx}`.toLowerCase();
```
mjnagel: Bit of a nit, but I don't like the index here. Tracking that some sort of index is necessary to keep these names unique but it would be nice if the index was based on the count of `allow-${policy.direction}-${target}-${pkgName}` policies rather than all policies. I think this looks confusing if anyone is looking at the names.

Quick example - `allow-egress-neuvector-updater-pod-neuvector-3` is one of the policies in a deployed cluster, but `allow-egress-neuvector-updater-pod-neuvector-1` and `allow-egress-neuvector-updater-pod-neuvector-2` don't exist. I would typically expect them to exist based on typical number conventions.

Tracking this might be more annoying to implement though since it would mean we couldn't just get the index from the list in the spec, not sure if there is a cleaner way to ensure unique names.

src/pepr/operator/controllers/network/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      remoteNamespaceLabels: {
+        "kubernetes.io/metadata.name": `istio-${gateway}-gateway`,
+      },
+      port,
+    };
+
+    // Generate the policy with a base index of 1000
+    const generatedPolicy = await generate(namespace, pkg, policy, 1000 + idx);
+    policies.push(generatedPolicy);
+  }
```
mjnagel: Should this be conditional on `UDSConfig.istioInstalled`? There's no strict dependency (i.e. CRDs) so maybe not a big deal.

src/metrics-server/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: metrics-server
+  namespace: {{ .Release.Namespace }}
+spec:
+  network:
+    allow:
+      - direction: Egress
+        remoteGenerated: KubeAPI
+        podLabels:
+          app.kubernetes.io/name: metrics-server
```
mjnagel: Metrics server actually needs egress to all server and agent nodes, not just the kube api. When deployed with a multi-node cluster I was seeing issues with this policy (until manually editing to add all node IPs). Might be a case for another `remoteGenerated` to get all nodes?

src/metrics-server/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,18 @@
...
+      - direction: Ingress
+        podLabels:
+          app.kubernetes.io/name: metrics-server
+        remoteNamespaceLabels:
+          kubernetes.io/metadata.name: "istio-system"
+        port: 443
```
mjnagel: Did you see an issue without this ingress policy? I removed it and everything seemed the same (passed validation, `kubectl top` worked, etc) - not following the need for this one 🤔

bundles/k3d-istio/uds-bundle.yaml
jeff-mccoy: no that was just a merge conflict fail

src/pepr/operator/controllers/namespace/index.ts
jeff-mccoy: Yeah my thinking here was if you're defining a UDS Package that is assuming Istio, if it's outside of that, you're not defining a UDS Package

src/pepr/operator/controllers/network/generate.ts
jeff-mccoy: hot dog I forgot about `get endpoints`

jeff-mccoy: updated, still needs further testing I think

src/pepr/operator/controllers/network/generate.ts
jeff-mccoy: I also added "uds" to this name after this screenshot, but thik this might make more sense? 

Screenshot 2024-01-04 at 1 40 04 PM

src/metrics-server/chart/templates/uds-package.yaml
jeff-mccoy: could have been a bad/copy pasta--tbh dont remember now why, but happy to delete

src/pepr/operator/controllers/network/generate.ts
mjnagel: Yeah I like that much better, makes the numbers make a bit more sense and will order them nicely in outputs.

src/neuvector/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: uds.dev/v1alpha1
+kind: Package
```
UncleGedd, line 2: This is neat! Wanted to call out the tradeoff between:
1. Having a YAML schema that the K8s community understands (ie. NetworkPolicy), but being way more verbose
2. Having a custom YAML schema generated by the fluent client that's more concise at the cost of readability (if you're not sure how the fluent client works)

src/pepr/operator/bully.ts
```diff
@@ -0,0 +1,53 @@
+import { K8s, kind } from "pepr";
+import { UDSPackage } from "./crd";
+import { Direction, Gateway } from "./crd/generated/package-v1alpha1";
+
+const intervalMS = 50;
+let inc = 0;
+
+setInterval(async () => {
```
UncleGedd: After poking around for a couple of minutes, I'm not sure what this file/fn is meant for, maybe add comments?

src/pepr/operator/controllers/network/allow-ingress-sidecar-monitoring.ts
```diff
@@ -0,0 +1,40 @@
+import { kind } from "pepr";
+
+export function allowIngressSidecarMonitoring(namespace: string): kind.NetworkPolicy {
```
UncleGedd: Probably not for this PR, but it would be awesome if we could converge on a standard way of creating NetPols. Here we use the JSON payload, but in the platform components (ie. Neuvector, etc,) we use that custom `uds-package.yaml`

UncleGedd:
Coming into this PR fairly blind, made some general comments about tradeoffs and structure, but it's tough to know how easy this will be to extend without actually extending it myself 😅

src/pepr/operator/bully.ts
jeff-mccoy: yeah thanks, I talked about it the now archived slack thread 😆 I'll add something, it's just a temporary stress test I was running

src/pepr/operator/controllers/network/allow-ingress-sidecar-monitoring.ts
jeff-mccoy: Really interesting point, I wrote these netpols when we didnt have the generator yet, moving to the generator for these as well makes a lot of sense and clearly demonstrates the difference between how UDS Operator handles these rules and how K8s does. 

Note we can't "directly use the UDS Package CRD", because these are actually processes running by the operator in response to a CR.

Screenshot 2024-01-04 at 10 37 30 PM
Screenshot 2024-01-04 at 10 37 22 PM

src/neuvector/chart/templates/uds-package.yaml
jeff-mccoy: I personally find the K8s NetPol to be one of the worst specs in core with how easy it is to misuse the matchers. Since we'll have many UDS Package CRs, I think the cost benefit is clear vs manually maintaining all these NetPols as Big Bang does today. This includes things like wiring API Server endpoints properly & auto-wiring Ingress policies for VirtualServices.  The current way (which requires maintaining a number of copy/paste policies in addition to per-app ones) is also almost 4 times as much YAML for the same config, only spread through helm chart templates and flux / helm post rendering.

Screenshot 2024-01-04 at 11 04 36 PM

src/pepr/operator/crd/validator.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (invalidNamespaces.includes(ns)) {
+    return req.Deny("invalid namespace");
+  }
+
+  // Ensure the name of each network policy is unique
+  const networkPolicy = req.Raw.spec?.network?.allow ?? [];
+
+  for (const policy of networkPolicy) {
+    // remoteGenerated cannot be combined with remoteNamespaceLabels or remotePodLabels
+    if (policy.remoteGenerated && (policy.remotePodLabels || policy.remotePodLabels)) {
```
rjferguson21: Comment seems right but the code was wrong?

```suggestion
    // remoteGenerated cannot be combined with remoteNamespace or remotePodLabels
    if (policy.remoteGenerated && (policy.remoteNamespace || policy.remotePodLabels)) {
```

jeff-mccoy: yeah thanks, that was a bad refactor

src/pepr/operator/enque.ts
```diff
@@ -0,0 +1,68 @@
+import { Log } from "pepr";
```
zachariahmiller: nit: shouldnt this file be enqueue.ts? enque is ink https://en.wiktionary.org/wiki/enque

zachariahmiller (dismissed):

src/pepr/config.ts
Noxsios, line 1: don't see any URL / domain validation for `UDSConfig.domain`, don't know how granular you wanna get but might be good to check / throw in a debug for fast verification if everything dies.

src/pepr/istio/README.md
```diff
@@ -0,0 +1,3 @@
+## Istio Pepr Capability
+
+This single capability manages the lifecycle of jobs have Istio sidecars injected into them.
```
Noxsios:
```suggestion
This capability manages the lifecycle of jobs that have Istio sidecars injected into them.
```

src/pepr/istio/index.ts
```diff
@@ -15,6 +20,11 @@ When(a.Pod)
   .WithLabel("batch.kubernetes.io/job-name")
   .WithLabel("service.istio.io/canonical-name")
```
Noxsios, line 21: what's UDS's strat for knowing that an expected CRD's label is still good / in the cluster?

src/pepr/operator/crd/register.ts
```diff
@@ -0,0 +1,38 @@
...
+            singular: "package",
+            kind: "Package",
+            shortNames: ["pkg"],
+          },
+        },
+      },
+      { force: true },
+    )
+    .then(() => {
+      Log.info("CRD registered");
```
Noxsios, line 30: I would log the CRD's queryable name (ie `k get crd <x>`) here so you can double check w/ `kubectl`

src/pepr/operator/README.md
```diff
@@ -0,0 +1,53 @@
...
+.
+├── controllers          # Core business logic called by the reconciler
+│   ├── istio            # Manages Istio VirtualServices and sidecar injection for UDS Packages/Namespace
+│   └── network          # Manages default and generated NetworkPolicies for UDS Packages/Namespace
+├── crd
+│   ├── generated        # Type files generated by `uds run -f src/pepr/tasks.yaml gen-crds`
+│   ├── sources          # CRD source files
+│   ├── register.ts      # Registers the UDS Package CRD with the Kubernetes API
+│   └── validator.ts     # Validates UDS Package CRs with Pepr
+├── enqueue.ts             # Serializes UDS Package CRs for processing by the reconciler
```
Racer159:
```suggestion
├── enqueue.ts           # Serializes UDS Package CRs for processing by the reconciler
```

src/pepr/operator/controllers/istio/virtual-service.ts
```diff
@@ -0,0 +1,91 @@
...
+import { UDSConfig } from "../../../config";
+import { Gateway, Istio, UDSPackage, getOwnerRef } from "../../crd";
+
+/**
+ * Creates a VirtualService for each exposed service in the package
+ *
+ * @param pkg
+ * @param namespace
+ */
+export async function virtualService(pkg: UDSPackage, namespace: string) {
```
Noxsios, line 13: dunno what the stance is on TS for return types in function signatures, whether to be Go-like and explicit, or be more terse and implicit / let the compiler ensure. I lean a little more towards explicit because it makes review a little easier to read (no VS Code hover feature _yet_)

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    // Add the remoteNamespace to the namespaceSelector if it exists and is not "*", otherwise match all namespaces
+    if (policy.remoteNamespace !== "" && policy.remoteNamespace !== "*") {
+      namespaceSelector.matchLabels = {
+        "kubernetes.io/metadata.name": policy.remoteNamespace,
+      };
+    }
+
+    // Add the remoteNamespace to the peers
+    peers.push({ namespaceSelector });
```
Noxsios, line 53: This logic is a little weird, the boolean checks in line 39 agree w/ line 35 and there is no other `if` logic, so this can be combined in to a single `a && b && c`

src/pepr/operator/controllers/network/generators/cloudMetadata.ts
```diff
@@ -0,0 +1,23 @@
+import { V1NetworkPolicyPeer } from "@kubernetes/client-node";
+
+export const META_IP = "169.254.169.254/32";
```
Noxsios, line 3: interesting "magic" link local address everyone chose

src/pepr/operator/controllers/network/generators/kubeAPI.ts
```diff
@@ -0,0 +1,46 @@
+import { V1NetworkPolicyPeer } from "@kubernetes/client-node";
+import { K8s, Log, kind } from "pepr";
+
+import { anywhere } from "./anywhere";
+
+let apiServerPeers: V1NetworkPolicyPeer[];
+
+void getAPIServerCIDR();
+
```
Noxsios: document why this function is invoked here, seems strange

src/pepr/operator/controllers/network/generators/kubeAPI.ts
```diff
@@ -0,0 +1,46 @@
...
+      }));
+
+      return apiServerPeers;
+    }
+  } catch (err) {
+    Log.debug(err);
+  }
+
+  // Log a warning and default to 0.0.0.0/0 if the IP is not found
+  Log.warn("Unable to get API server CIRD, defaulting to 0.0.0.0/0");
```
Noxsios:
```suggestion
  Log.warn("Unable to get API server CIDR, defaulting to 0.0.0.0/0");
```

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  // Get the current generation of the package
+  const generation = (pkg.metadata?.generation ?? 0).toString();
+
+  const policies = [
+    // All traffic must be explicitly allowed
+    defaultDenyAll(namespace),
+
+    // Allow DNS lookups
+    allowEgressDNS(namespace),
```
Noxsios, line 23: guess I know which port to tunnel out from

src/pepr/operator/enqueue.ts
```diff
@@ -0,0 +1,68 @@
...
+      this.#pendingPromise = false;
+      item.resolve();
+    } catch (e) {
+      // Reset the pending promise flag and reject the promise on error
+      this.#pendingPromise = false;
+      item.reject(e);
+    } finally {
+      // After the package is reconciled, dequeue the next package
+      await this.#dequeue();
+    }
```
Noxsios: couldn't `this.#pendingPromise = false;` happen only once, in the `finally` block, as it needs to happen no matter what

src/pepr/operator/reconciler.ts
```diff
@@ -0,0 +1,79 @@
...
+    return;
+  }
+
+  const { namespace, name } = pkg.metadata;
+
+  Log.debug(pkg, `Processing Package ${namespace}/${name}`);
+
+  // Configure the namespace and namespace-wide network policies
+  try {
+    void updateStatus(pkg, { phase: Phase.Pending });
```
Noxsios: why not `await` this?

src/pepr/policies/security.ts
```diff
@@ -67,21 +67,21 @@ When(a.Pod)
     pod.securityContext = pod.securityContext || {};
 
     // Set the runAsUser field if it is defined in a label
-    const runAsUser = metadata.labels?.["uds/user"]
+    const runAsUser = metadata.labels?.["uds/user"];
     if (runAsUser) {
-      pod.securityContext.runAsUser = parseInt(runAsUser)
+      pod.securityContext.runAsUser = parseInt(runAsUser);
```
Noxsios: i'm personally in the `"semi": false` gang for prettier. The semicolon is syntactic sugar at this point w/ all TS/JS code going usually through a build tool like esbuild/vite, so its purpose is mainly pointless.

tasks/setup.yaml
```diff
@@ -7,4 +7,4 @@ tasks:
 
       - description: "Initialize the cluster with Zarf"
         # renovate: datasource=github-tags depName=defenseunicorns/init versioning=semver
-        cmd: "zarf package deploy oci://defenseunicorns/init:v0.31.1-${UDS_ARCH} --confirm"
+        cmd: "zarf package deploy oci://defenseunicorns/init:v0.31.4-${UDS_ARCH} --confirm"
```
Noxsios: Zarf v0.32+ does not use arch in OCI ref anymore, this will break soon

src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -0,0 +1,38 @@
...
+        metadata: {
+          name: pkg.metadata.namespace,
+          labels,
+        },
+      },
+      { force: true },
+    );
+
+    // @todo: Add a finalizer to remove the label when the package is deleted
+    // @todo: Check for pods without sidecars and address them
```
Noxsios: todo

Noxsios (dismissed):

src/pepr/operator/controllers/network/defaults/default-deny-all.ts
```diff
@@ -0,0 +1,18 @@
+import { kind } from "pepr";
+
+export function defaultDenyAll(namespace: string): kind.NetworkPolicy {
+  return {
+    apiVersion: "networking.k8s.io/v1",
+    kind: "NetworkPolicy",
+    metadata: {
+      name: "uds-default-deny-all",
```
rjferguson21: This name won't be unique in `namespace`.  Is the assumption at the moment that there will be only one Package per namespace?

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -0,0 +1,134 @@
...
+
+import { Allow, RemoteGenerated } from "../../crd";
+import { anywhere } from "./generators/anywhere";
+import { cloudMetadata } from "./generators/cloudMetadata";
+import { intraNamespace } from "./generators/intraNamespace";
+import { generateKubeAPI } from "./generators/kubeAPI";
+
+export function generate(namespace: string, policy: Allow): kind.NetworkPolicy {
+  // Create a unique name for the NetworkPolicy based on the package name, index, direction, pod labels, and port
+  const name = generateName(policy);
```
rjferguson21, line 12: Similar question about uniqueness here, it does not look like `generateName` is scoping the name per package.

src/pepr/operator/controllers/network/generators/kubeAPI.ts
Racer159: Agree that it is strange - it effectively caches the apiServerPeers in memory as a singleton which IMO is not a great option (as a todo mentions) - in most scenarios node scaling doesn't change it but rolling the control plane could rotate it (and it could be intentional in a scenario where you are moving around IP spaces in a network).   Its a more advanced use case but if pepr was on a worker and kept running you would have to rekick the operator before it was able to properly regenerate the network policies.

src/pepr/operator/controllers/network/defaults/allow-egress-dns.ts
```diff
@@ -0,0 +1,16 @@
+import { Direction } from "../../../crd";
+import { generate } from "../generate";
+
+export const allowEgressDNS = (namespace: string) => {
+  const netPol = generate(namespace, {
```
Racer159, line 5: Just a note but DNS exfil is a commonish way to pull data out of a system - we might want to provide a config option to lock this down a bit more in the future

src/pepr/operator/controllers/network/generate.ts
Racer159: This _technically_ isn't the same but I do agree it is a little confusing as implemented - basically this adds an empty `namespaceSelector` which narrows the list of ingress / egress destinations to all pods in all namespaces rather than literally any and all destinations (This PR fix was helpful https://github.com/cilium/cilium/pull/5250)

src/pepr/operator/controllers/network/defaults/default-deny-all.ts
mjnagel: Might be wrong but this one should be OK since the netpol is always the same. It would get re-applied per package but otherwise have no bad effects.

src/pepr/operator/controllers/network/generators/kubeAPI.ts
Racer159: No I think he meant Canine Infectious Respiratory Disease :smile:

src/pepr/policies/security.ts
Racer159: I like them - brings me back to my roots

src/pepr/operator/controllers/network/defaults/default-deny-all.ts
rjferguson21: If I create a second package with:
```
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: foo
  namespace: grafana
```

and then delete it, the generic netpol originally created by the grafana package will be deleted as well.

src/pepr/operator/controllers/network/defaults/allow-egress-dns.ts
jeff-mccoy: yeah i originally had it as an opt-out, but dropped that since BB wasn't doing it to simplify the CR....will think about that one though, it's a valid point.

src/pepr/operator/controllers/network/defaults/default-deny-all.ts
jeff-mccoy: yeah the choices are create a second ownerref (which would require first doing a `get`) or just using a unique name, def. more performant to just use a unique name instead.

src/pepr/operator/controllers/network/generate.ts
jeff-mccoy: that is correct, that is handled in the policies.ts file

src/pepr/config.ts
jeff-mccoy: I think we can make that a separate issue if it is needed, this is populated from a zarf var fwiw

src/pepr/istio/index.ts
jeff-mccoy: not sure I understand the q

src/pepr/operator/README.md
cmwylie19, line 1: This is a really good approach for namespaced objects. It is straight forward and it makes sense and keeps the "created" objects up to date. The queue is  great data structure to handle new instances of operands and operate on them until they are reconciled.

Question -- would you ever need to control cluster scoped objects? (Not now but in the future? Just wonder if they are in scope (nodes,storageClasses,ingressClasses,persistentVolumes, etc)

This question was kicked off by noticing the ownerRefs don't allow cross namespace reference unless the owner is cluster-scoped.

We were spitballing how you would handle this and came up with two ideas should that become necessary in some future state.

**Approach 1:** Make the `UDSPackage` cluster scoped that it controls namespaced and non-namespaced items so that owner references can control namespaced and cluster scoped items...

**Approach 2**: use the delete event in Admission

```typescript
When(UDSPackage)
  .IsCreatedOrUpdated()
  // Advanced CR validation
  .Validate(validator)
  // Enqueue the package for processing
  .Watch(pkg => queue.enqueue(pkg));
```


Create a new field in the UDSOperator CR for namespaces that it controls and keep track of the objects by labels


```typescript
When(UDSPackage)
  .IsDeleted()
    // filter by package labels and namespaces to find what to delete
  .Watch(pkg => {

// get all objects with a particular label
for (const ns of pkg.metadata.ns!) {
  const virtualServices = await K8s(Istio.VirtualService)
    .InNamespace(ns)
    .WithLabel("uds/package", pkgName)
    .Get();
    
    // same for Netpol, and other namespace scoped and delete them
}

// Now take care of the deletion of your cluster scoped objects
const [nodes,storageClasses,ingressClasses,persistentVolumes, etc] = await K8s()....

}

File: src/pepr/operator/controllers/istio/virtual-service.ts (hunk 19 above)
Comment (jeff-mccoy) [reply src/pepr/operator/controllers/istio/virtual-service.ts:13]: I prefer implicit in TS and leveraging the language features (same reason we dont do [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation) or typed-jsdoc these days in TS-land. Not sure I'm tracking on lack of VSCode hover feature...

Screenshot 2024-01-09 at 1 28 11 PM
Screenshot 2024-01-09 at 1 28 04 PM

File: src/pepr/operator/enqueue.ts (hunk 25 above)
Comment (jeff-mccoy) [reply src/pepr/operator/enqueue.ts]: yeah I'm trying to remember why (if there was a good reason) I wanted that to occur prior to calling resolve/reject

File: tasks/setup.yaml (hunk 28 above)
Comment (jeff-mccoy) [reply tasks/setup.yaml]: yeah we were waiting on UDS CLI, which wasn't compatible with Zarf 0.32

File: src/pepr/operator/controllers/istio/virtual-service.ts (hunk 19 above)
Comment (Noxsios) [reply src/pepr/operator/controllers/istio/virtual-service.ts:13]: As shown in your screenshot, when you hover in VS Code you get to see the type annotations, in a GitHub review you don't get such niceties, so it can be _slightly_ harder for someone to review.

File: src/pepr/istio/index.ts (hunk 16 above)
Comment (Noxsios) [reply src/pepr/istio/index.ts:21]: If `istio` changes its `service.istio.io/canonical-name` label, is there a clean way to handle that?

File: src/pepr/config.ts
Comment (Noxsios) [reply src/pepr/config.ts:1]: sgtm

File: src/pepr/istio/index.ts (hunk 16 above)
Comment (mjnagel) [reply src/pepr/istio/index.ts:21]: In terms of this specific code the actual value of the label doesn't matter for us since it's just a shortcut to filter for job pods that are mutated/injected by istio. Istio changing the value of the label wouldn't matter for this, and if they changed the key we would just have to update here accordingly. Deploy order-wise everything should roll out fine, istio would update first, then pepr. Since this code is exclusively targeting jobs there's (in general) no issues with previous existing pods, etc (but could probably make backwards compatible to handle running jobs during the upgrade).

Could also change this to use the job label from k8s and do the "is this injected" check inside the watch, which would bypass the need for the label entirely, at the expense of watching more things.

File: src/pepr/operator/controllers/network/defaults/allow-egress-dns.ts (hunk 32 above)
Comment (mjnagel) [reply src/pepr/operator/controllers/network/defaults/allow-egress-dns.ts:5]: Might be worth checking some common distros to see if `k8s-app: kube-dns` is consistent across distros as a pod label. From a quick scan.... k3d, rke2, eks, and kind all use that label and place coredns in `kube-system`. Based on k8s docs pretty sure kubeadm does the same.

Potentially could use that as a default and if we want to have a safety net validate that selector returns a pod with a get or just expose a way to override? I don't recall offhand if there was a specific cluster that BB had in mind that used non-standard labels/ns or if it was always to err on the side of caution.

File: src/pepr/operator/controllers/istio/virtual-service.ts (hunk 19 above)
Comment (jeff-mccoy) [reply src/pepr/operator/controllers/istio/virtual-service.ts:13]: "makes review easier" -- ah you were saying GH Review lacks the VSCode feature... 😆

File: src/pepr/operator/README.md
Comment (jeff-mccoy) [reply src/pepr/operator/README.md:1]: No need for cluster-scoped resources yet, but good point for sure. We could watch deletions or play with some strategies with a finalizer + webhook or watch to ensure things were truly cleaned up.

File: src/pepr/operator/controllers/istio/injection.ts [hunk 33]
```diff
@@ -0,0 +1,38 @@
...
+      {
+        metadata: {
+          name: pkg.metadata.namespace,
+          labels,
+        },
+      },
+      { force: true },
+    );
+
+    // @todo: Add a finalizer to remove the label when the package is deleted
```
mjnagel: Think we had discussed maybe just ignoring this one since istio would be removed and the label is innocuous without istio? If we did want to remove it...might get messy since they could have added the label themselves (i.e. the operator didn't add it) and we would want to leave it in that case?

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -0,0 +1,141 @@
...
+      .join("-");
+
+  return (
+    `${policy.direction}-${name}`
+      // The name must be lowercase
+      .toLowerCase()
+      // Replace sequences of non-alphanumeric characters with a single '-'
+      .replace(/[^a-z0-9]+/g, "-")
+      // Truncate the name 245 characters
+      .slice(0, 245)
```
mjnagel: Where did 245 come from? I was looking and it seemed like [253](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) was the general rule for k8s resources (albeit lower limits for labels and pod names it seemed like)?

I think regardless this won't work as is... in `src/pepr/operator/controllers/network/policies.ts` we are prefixing this with `allow-${pkgName}-` which (assuming 253 is the correct limit) could be an additional 260 characters at most
 (`allow-<253 chars>-`). Is that true?

src/pepr/operator/controllers/istio/virtual-service.ts
```diff
@@ -0,0 +1,91 @@
...
+  // Get the list of exposed services
+  const exposeList = pkg.spec?.network?.expose ?? [];
+
+  const payloads: Istio.VirtualService[] = [];
+
+  // Iterate over each exposed service
+  for (const expose of exposeList) {
+    const { gateway = Gateway.Tenant, host, port, service, mode } = expose;
+
+    const name = `${pkgName}-${gateway}-${host}`.toLowerCase();
```
mjnagel: Do we need a safeguard here for name character limit (truncate to certain lenght)? Theoretically `pkgName` could already be at the max characters for a k8s resource I think.

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -0,0 +1,141 @@
...
+      // Remove leading and trailing '-'
+      .replace(/^-|-$/g, "")
+  );
+}
```
mjnagel: I'm not in love with the more verbose name format, I think it becomes unreadable outside of simpler/autogenerated policies...
- `allow-neuvector-ingress-neuvector-manager-pod-istio-admin-gateway-admin-ingressgateway` and the istio ones in general are pretty bad
- `allow-grafana-egress-grafana-tempo-tempo` and similar ones for promtail feel repetitive and not obvious that I'm reading ns-podselector combos
- `allow-loki-egress-istiod-communication` and others that have a description don't align well with the autogenerated names here (intentional to override with a description of course, but every other policy starts with `allow-loki-egress-loki` so it's a confusing contrast)

I'm not really sure what to suggest as an alternative though. If we kept it as `allow-<pkg>-<direction>-<remotens>` that would be relatively simple and I think capture readability well. Remote namespaces should generally help understand what "app" is being allowed, and the `remoteGenerated` would still have their descriptions to handle those cases. With that scheme the above two would look like:
- `allow-neuvector-ingress-istio-admin-gateway`
- `allow-grafana-egress-tempo`
- `allow-loki-egress-istiod-communication` (unchanged, but feels more aligned to the others now?)

That being said we'd have to deconflict policy names again using index/hash/something which I'm not sure how I feel about. I _think_ if only conflicting policies are indexed it could work well?

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -0,0 +1,100 @@
...
+    policy.metadata.labels = policy.metadata?.labels ?? {};
+    policy.metadata.labels["uds/package"] = pkgName;
+    policy.metadata.labels["uds/generation"] = generation;
+
+    // Add the package name to the name of the policy to ensure uniqueness
+    if (idx < 1) {
+      policy.metadata.name = `deny-${pkgName}-${policy.metadata.name}`;
+    } else {
+      policy.metadata.name = `allow-${pkgName}-${policy.metadata.name}`;
+    }
```
mjnagel, line 72: This doesn't fully guarantee uniqueness I don't believe - if you had two custom policies with identical direction, podLabels, remote..., but different ports their names would clash. I believe you could also make two completely different `network.allow`s but use the same description and cause a conflict as long as it was in the same package and direction?

2 policies with matching ports could be done with a single `network.allow` using the `ports` list, but I still think we need to handle that scenario. 

Not sure the cleanest way to do that without adding a lot of logic, could sort through the array for uniqueness and append an index if necessary after this name generation/before apply?

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -0,0 +1,100 @@
...
+    .Get();
+
+  // Find any orphaned polices (not matching the current generation)
+  const orphanedNetPol = policyList.items.filter(
+    netPol => netPol.metadata?.labels?.["uds/generation"] !== generation,
+  );
+
+  // Delete any orphaned policies
+  for (const netPol of orphanedNetPol) {
+    Log.debug(netPol, `Deleting orphaned VirtualService ${netPol.metadata!.name}`);
```
mjnagel:
```suggestion
    Log.debug(netPol, `Deleting orphaned NetworkPolicy ${netPol.metadata!.name}`);
```

test.yaml
```diff
@@ -0,0 +1,27 @@
+
```
mjnagel: Can be deleted?

src/pepr/operator/controllers/istio/injection.ts
jeff-mccoy: yeah I was thinking we'd prob want to add an annotation to state the original label state

src/pepr/operator/controllers/network/generate.ts
jeff-mccoy: yep good catch--I added the pkgName from another PR comment, but forgot to move this truncate function. 245 was to accommodate the prior prefix string

src/pepr/operator/controllers/network/generate.ts
jeff-mccoy: yeah I tend to agree, I think it was fairly readable without the pkgname added, but added that back has made it pretty repetitive--not really sure what the best thing to do here is now.

test.yaml
jeff-mccoy: that the default repo prettier formatting, it would come back again if someone formatted a file 🤷

src/pepr/operator/controllers/network/generate.ts
jeff-mccoy: that being said, I think this might be okayish if we simply use the `description` field regularly as it replaces the name

src/pepr/policies/security.ts
jeff-mccoy: the formatting is set by `npx pepr format` running prettier / eslint

test.yaml
mjnagel: I just meant this file in general, might've missed where/how this is getting used? If it is used I suspect there's a better place for it than at the root of the repo.

src/pepr/operator/controllers/network/policies.ts
jeff-mccoy: just make sure you do it on UDP

src/pepr/operator/controllers/istio/virtual-service.ts
jeff-mccoy: teach someone to fish, something, something....

Screenshot 2024-01-09 at 7 32 38 PM

test.yaml
jeff-mccoy: ohhhh sorry I thought this was a diff file--just saw its a random junk file, will delete thx (showed up weird like it was just a line break change in GH)

src/pepr/istio/index.ts
jeff-mccoy: yeah I think by the time they change that this capability should be null/void since this is a temp thing until sidecars are a real boy

src/pepr/operator/controllers/istio/virtual-service.ts
jeff-mccoy: yep, updated.

src/pepr/operator/controllers/istio/virtual-service.ts
Noxsios: brb, completely revamping my review flow

README.md
```diff
@@ -27,8 +27,8 @@ The core applications are:
 
 | Dependency                                                     | Minimum Version |
 | -------------------------------------------------------------- | --------------- |
```
zachariahmiller: line 15 should probably be updated to point to the pepr readme.

src/pepr/operator/controllers/istio/injection.ts
jeff-mccoy: updated with a cleanup watch on pkg delete

src/pepr/operator/controllers/network/policies.ts
jeff-mccoy: this is now fixed via a new validator step

Screenshot 2024-01-12 at 4 06 31 AM

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -0,0 +1,102 @@
...
+    // Create the NetworkPolicy for the VirtualService
+    const policy: Allow = {
+      direction: Direction.Ingress,
+      podLabels,
+      remoteNamespace: `istio-${gateway}-gateway`,
+      remotePodLabels: {
+        app: `${gateway}-ingressgateway`,
+      },
+      // Use the same port as the VirtualService if targetPort is not set
+      port: targetPort ?? port,
```
mjnagel, line 50:
```suggestion
      port: targetPort ?? port,
      description: `${Object.values(podLabels)} Istio ${gateway} gateway`,
```
I think this makes these istio ones a bit more clean without causing any collisions? Examples:
- `allow-neuvector-ingress-neuvector-manager-pod-istio-admin-gateway`
- `allow-grafana-ingress-grafana-istio-admin-gateway`

Only real change is it drops the namespace for the gateway (only ever one gateway per namespace, so no conflicts caused by this). I think that's the best we can do for this one without causing name conflicts, and wouldn't want to risk that here since these are generated (no option to add description via the package CR).

src/pepr/operator/controllers/network/generators/kubeAPI.ts
```diff
@@ -0,0 +1,46 @@
...
+export function generateKubeAPI(): V1NetworkPolicyPeer[] {
+  return apiServerPeers;
+}
+
+async function getAPIServerCIDR(): Promise<V1NetworkPolicyPeer[]> {
+  try {
+    // @todo: evaluate if this ever changes with node autoscaling
+
+    // Read the API server endpoints from the cluster
+    const { endpoints } = await K8s(kind.EndpointSlice).InNamespace("default").Get("kubernetes");
```
bburky: This will deny access to new EndpointSlices (new IPs) for the Kubernetes API, which I believe will happen if control plane nodes are replaced. Most things in-cluster will connect through the `kubernetes.default.cluster.local` Service which has a ClusterIP. These EndpointSlices are the external IPs of the k8s control plane nodes.

Not sure how to best handle this. Re-running the operator would fix it, but the operator will reconcile infrequently. I suppose you could have a watcher on EndpointSlice and update NetworkPolcies automatically, but that would also have a slight lag too (maybe that's acceptable, I think it's rare that things are connecting to the external IPs of the k8s API anyway?).

bburky: There is a todo here to check this.

If you want a quick demo:

```
k3d cluster create
kubectl get endpointslice kubernetes -o yaml

k3d node  create --role server node-foo
kubectl get endpointslice kubernetes -o yaml
```

src/pepr/operator/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  name: "uds-core-operator",
+  description: "The UDS Operator is responsible for managing the lifecycle of UDS resources",
+});
+
+export const { Store, When } = operator;
+
+const queue = new Queue();
+
+// Watch for changes to the UDSPackage CRD to remove the finalizer
+When(UDSPackage).IsDeleted().Watch(namespaceFinalizer);
```
bburky: This is not actually a real k8s finalizer (the `namespaceFinalizer` name is confusing), but it is possible to implement one in Pepr without too much effort. If the watcher somehow misses the deletion event (e.g. operator is shut down long enough to miss the event in the history), this `namespaceFinalizer` won't be invoked.

You could make a real finalizer with `.metadata.finalizers`. You would instead watch without a specific create/update/delete phase (no `IsDeleted()`) and instead check for the `.metadata.deletionTimestamp` and the presense of your finializer name in  `.metadata.finalizers`. This way if a watcher restarts and does the initial list the finalizer cleanup will be correctly invoked. The deletion of a UDSPackage would be blocked until the finalizer completes.

src/pepr/operator/controllers/network/generators/kubeAPI.ts
mjnagel: @bburky is there a different way you've seen kube api egress done before (different place to grab the proper IP(s) for a network policy). Even though pods are connecting to `kubernetes.default.cluster.local` from what I've seen they need to have an egress policy to the endpointslices in order for that connection to work consistently.

Agreed on the issue with new nodes though (captured in the todo on line 21).

bburky: Hmm, you're right. Perhaps iptables in the container transforms the ClusterIP into the external one or something?

I was just testing locally and confirmed that `kubernetes.default.cluster.local` has a different IP and does not contain the individual EndpointSlices IPs (which is expected for non-headless Services). I tested creating a NetworkPolicy without the exact EndpointSlices and indeed the kubernetes ClusterIP becomes inaccessible.

src/metrics-server/chart/templates/uds-package.yaml
mjnagel: Resolving - there is a TODO for reevaluating this which I think can be addressed in a follow on.

src/pepr/operator/controllers/network/generators/kubeAPI.ts
bburky: We could perhaps generate CNI specific policies, generate a CNI-specific allow rule, but I'm not sure if they get merged with native k8s NetworkPolicies. And doing CNI-specific things would be complicated.

https://docs.tigera.io/calico/latest/network-policy/policy-rules/service-policy#allow-access-to-the-kubernetes-api-for-a-specific-namespace

`kube-apiserver` https://docs.cilium.io/en/stable/security/policy/language/#entities-based

For a generic solution Watch() on endpoints and updating policies is all I can think of

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -0,0 +1,103 @@
+import { K8s, Log, kind } from "pepr";
```
bburky, line 1: NetworkPolcies are created after a UDSPackage is created, which is in parallel to the rest of the app's installation into its namespace (if I understand correctly). This can be racy. Nothing blocks the execution of the app's Pods, they may start to execute before the NetworkPolcies are finished being created.

Not sure if/how we want to handle this, but it's a general risk of doing policy enforcement via an operator CR that's installed parallel to an app. You _could_ have the admission controller deny Pods until the UDSPackage is status ready. But you'd also have to do things like make sure Zarf installs the UDSPackage before other resources in the namespace.

(or maybe I've missed something. I'm not fully tracking how the UDSPackage CR gets installed, is it via Helm just as another resource beside the app?)

mjnagel: It's not exactly parallel, but it's also not entirely guaranteed that everything is processed before the app is deployed. Using [grafana](https://github.com/defenseunicorns/uds-core/blob/uds-operator/src/grafana/zarf.yaml#L12-L15) as an example, the grafana-config chart contains the [UDSPackage CR](https://github.com/defenseunicorns/uds-core/blob/uds-operator/src/grafana/chart/templates/uds-package.yaml) and other configuration (in this case a configmap for grafana datasources), but will not have any pods. After install is complete then the upstream grafana helm chart is installed.

The missing bit is waiting on the UDSPackage CR to be "rolled out" (i.e. all network policies, etc applied). That could be done with a zarf wait action on `status.phase` but that would require the two charts to be split into separate components, which would make the imports in the [uds-core package](https://github.com/defenseunicorns/uds-core/blob/uds-operator/packages/standard/zarf.yaml) double. I think in the broadest sense "it works" as is but that's not guaranteed strictly.

bburky: ~It actually looks like `uds-grafana-config` is ordered after `grafana` though?~
https://github.com/defenseunicorns/uds-core/blob/10772e2c64f1e4b965b6b644b0008c81025029e9/src/grafana/zarf.yaml#L18

Edit: oops, looks like it changed between branches. But this does seem somewhat accident prone and difficult to use safely, seeing how it changed.

But yes, zarf's ordering can be a solution to install config before apps. I see how it's tricky to do use zarf's wait action though, it's tied to components which is more coarse grained than desired.

src/pepr/operator/index.ts
jeff-mccoy: yeah just a bad name, I had done it as a finalizer and then changed it as we didnt really need a finalizer here

src/pepr/operator/controllers/network/generators/kubeAPI.ts
jeff-mccoy: You can't use clusterip, netpols dont care about that, they only care about the pod ip, so we have to use endpoint slices, sadly.

src/pepr/operator/controllers/network/policies.ts
jeff-mccoy: We discussed this and have decided that moving the policies to before the istio injection ensures the fastest policy generation and the risk is narrow enough to not address further at this time (would only affect egress).

src/pepr/operator/controllers/network/defaults/allow-egress-dns.ts
jeff-mccoy: If chatgpt doesnt lie-- kube-dns selector looks safe across most distros. Here's it working with K3d. Notice the default lookup works, but trying to use an alternate dns server fails.

Screenshot 2024-01-12 at 7 19 36 PM

jeff-mccoy: same pod when you delete the CR (netpols are removed)

Screenshot 2024-01-12 at 7 22 30 PM

src/pepr/operator/controllers/network/generators/kubeAPI.ts
jeff-mccoy: okay based on our discussion, I have implemented a watch on the endpointslices and now trigger updates to netpols on changes (if required)

mjnagel (dismissed):
LGTM at this point. I did modify the loki policies and add a `Package` CR for prometheus-stack so might be good to get @mikevanhemert / @zachariahmiller to confirm those pieces at least as a validation I didn't mess things up too much.

mjnagel (approved):

PR #67, opened 2023-12-15
chore: initial renovate config

zachariahmiller:
## Description

Adds initial renovate config. This should handle the new Ironbank flavor as well but currently does not have any "intelligence" around included/supported image versions in a given helm chart version.

## Related Issue

N/A

## Type of change

## Checklist before merging


mikevanhemert (dismissed):

src/prometheus-stack/zarf.yaml
```diff
@@ -9,11 +9,11 @@ components:
     description: "Install kube-prometheus-stack operator crds using the helm chart https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-operator-crds"
     charts:
       # https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-operator-crds
+      #version: 5.1.0 (old version - for testing upgrade)
```
mjnagel:
```suggestion
```

src/prometheus-stack/zarf.yaml
```diff
@@ -22,10 +22,10 @@ components:
     description: "Install kube-prometheus-stack using the helm chart https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack"
     charts:
       - name: kube-prometheus-stack
-        namespace: monitoring
         url: https://prometheus-community.github.io/helm-charts
-        #version: 50.3.1 (old version - for testing upgrade)
         version: 54.0.1
+        #version: 50.3.1 (old version - for testing upgrade)
```
mjnagel:
```suggestion
```
Think we can clean up these comments now.

renovate.json
```diff
@@ -1,6 +1,198 @@
...
+      "matchStrings": [
+        "# renovate: datasource=github-tags depName=(?<depName>[^ ]+) versioning=(?<versioning>[^\n]+)\n.*?uds@(?<currentValue>.*)",
+        "\/\/ renovate: datasource=github-tags depName=(?<depName>[^ ]+) versioning=(?<versioning>.*?)( extractVersion=(?<extractVersion>.*?))?\n.*?(https:\/\/raw.githubusercontent.com\/defenseunicorns\/uds-cli\/v)(?<currentValue>[^\\s]+)\/"
+      ],
+      "datasourceTemplate": "github-tags",
+      "extractVersionTemplate": "^v(?<version>.*)$"
+    },
+    {
+      "fileMatch": ["\\.*\\.ya?ml$"],
+      "matchStrings": ["- (?<depName>.*):(?<currentValue>.*)"],
```
mjnagel:
```suggestion
      "matchStrings": [
        "- [\"]?(?<depName>[^\"\n]*):(?<currentValue>[^\"\n]*)[\"]?"
      ],
```
Tested a few ways but this seemed to be the best way to handle quoted images in a `zarf.yaml`. Optionally matches a quote and ensures that the capture groups don't include a quote.

mjnagel (commented):
Still testing I think there might still be some images missed in values files but they might just be innocent renovate warnings.

renovate.json
mjnagel: ~Still looking at this one, might have inadvertently broke something else with the helm matcher 🤔~

Latest suggestion is working, it was globbing in newlines previously so this one will ensure single line matches only.

src/neuvector/zarf.yaml
```diff
@@ -33,9 +33,9 @@ components:
         version: 0.1.0
         localPath: chart
     images:
-      - docker.io/neuvector/controller:5.2.2-s1
-      - docker.io/neuvector/manager:5.2.2-s1
+      - docker.io/neuvector/controller:5.2.3
+      - docker.io/neuvector/manager:5.2.3
```
mjnagel, line 37: Just noting for anyone reviewing - updated the images and chart version here because the `-s1` prefix was not being handled properly by renovate (and is not present on future chart/image versions).

mjnagel (approved):
LGTM - with these changes I was seeing all the PRs 😄 - https://github.com/mjnagel/test/pulls . Great work on this @zachariahmiller 

We will definitely need a follow on to figure out a more automated way to handle helm version vs image version lookups but as discussed that's a larger effort probably involving either custom datasources or self hosted renovate with additional upgrade tasks.

src/neuvector/zarf.yaml
zachariahmiller: TBH I'm surprised this didn't work when the -distroless and the 1.19.3-tetrate-fips-v0 tags were working

mjnagel: Yeah I was confused as well but it didn't seem to pick up the newer version until I dropped the `-s1` so I just went with it. It looks like all future (and previous?) chart versions don't have that suffix so I think it makes sense to just drop it.

jeff-mccoy (approved):

PR #97, opened 2024-01-11
ci: add upgrade test for uds core

### LINKED ISSUE #92: Test upgrades in CI

Upgrades are not being tested in CI currently, giving low confidence in a painless upgrade path for end users.

On PRs I would like both a clean install and upgrade test (from latest release or from main?) to be run both at the individual src package level and the all package/bundle level.

### Additional context

We should also ensure that concurrent runs (install/upgrade) are cancelled if one or the other fails, similar to how we cancel the all run if an individual src package fails.

mjnagel:
## Description

Updates CI to add an upgrade test. Test order:
- deploys k3d + zarf init from the PR branch
- deploys uds-core zarf package from the latest release
- creates + deploys uds-core from the PR branch
- runs validations from the PR branch

Does not include any upgrade testing at the individual source package level since these tests are just meant to short-circuit for dev cycles (this is a similar approach to our intent with flavor testing). It also does not address upgrades of the bundle (k3d/zarf init) which are out of the scope of uds-core's non-dev artifacts.

## Related Issue

Resolves https://github.com/defenseunicorns/uds-core/issues/92

(at least as a first pass for testing the full deploy/upgrade)

## Type of change

- [x] Other (CI)

## Checklist before merging


.github/workflows/test.yaml
```diff
@@ -50,3 +50,23 @@ jobs:
       - name: Save logs
         if: always()
         uses: ./.github/actions/save-logs
+
+  test-upgrade:
+    if: ${{ inputs.package == 'all' }}
+    runs-on: "uds-ubuntu-big-boy-8-core"
+    timeout-minutes: 30
```
justin-o12, line 57: Honest question since I'm new to Core and more familiar with DUBBD:  is 30 minutes enough to install and then upgrade?

mjnagel: Good question - typically the install on its own takes around 10-11 minutes. Given that the upgrade doesn't have to redo the k3d/zarf init pieces it should be relatively safe to stick with 30 minutes I hope - we haven't really done much testing of upgrades but locally I was seeing things generally take <15 minutes with a few packages cycling their pods.

The timeout is meant to keep us from a big runner eating up minutes if something is hanging - I'd prefer to start small and increase if it seems necessary with real upgrades. Open to other opinions though - not sure if others have tested upgrades and seen different results.

justin-o12: All makes sense to me, thanks for the explanation!

TristanHoladay (dismissed):

chance-coleman (dismissed):

zachariahmiller (approved):
This looks good and in line with our discussions on approach to the upgrade testing process.

jeff-mccoy (approved):
lgtm, downside being this will now double the cost of our CI runs for the full test 😆

PR #110, opened 2024-01-17
ci: add debug logging to ci output

mjnagel:
## Description

Adds basic debug output (`kubectl get all -A` + `kubectl get package -A`) to the pipeline. This is currently run with a `!cancelled()` conditional but could be switched to `failure()` if we want to keep the output clean on success.

Also modifies/fixes the PR conditional for check paths so that it properly executes only on PRs to main.

## Related Issue

N/A

## Type of change

## Checklist before merging


zachariahmiller (approved):
LGTM

PR #115, opened 2024-01-19
ci: refactor to avoid skipped job ugliness

mjnagel:
Internal PR between branches, separates upgrade job to different workflow file to avoid skipped jobs on PR display.


zachariahmiller (approved):
❤️ 🚀

PR #127, opened 2024-01-23
ci: change ci workflows for concurrency/cancellation

mjnagel:
## Description

This refactor simplifies what we have to maintain on the github workflows by adding a test_type to the matrix, for `package: all` only. This also ensures all jobs will cancel on failure as expected.

After evaluation this seems like the best way to ensure mass cancellation on an individual failure across both install and upgrade.

## Related Issue

N/A

## Type of change

## Checklist before merging


mjnagel:
This should be ready for an initial review. The required workflows on the repo will have to be adjusted again to match the new workflow names. I looked into whether we could adjust the names of the jobs to not include the matrix (`(all, upstream, install)`, etc) so that we had more flexibility in naming but it doesn't appear that GitHub offers any way to do that.

zachariahmiller (approved):
I dont love the way the job names look after this but neither of us were able to find a good workaround. 🤷 . Otherwise LGTM! I'll update the required checks for the new jobs.

PR #129, opened 2024-01-24
feat: add Istio VirtualService Requestmatch to UDS Operator

### LINKED ISSUE #114: Expanded support for VirtualService generation
During our attempt to leverage the UDS Operator Package custom resource we ran into issue trying to support GitLab's VirtualService.  It is currently impossible to specify a [HttpMatch](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPMatchRequest) via the `Package`  `network.expose` which would be necessary to emulate GitLab's current VS.

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: gitlab
  namespace: gitlab
spec:
  gateways:
  - istio-tenant-gateway/tenant-gateway
  hosts:
  - gitlab.uds.dev
  http:
  - fault:
      abort:
        httpStatus: 403
        percentage:
          value: 100
    match:
    - uri:
        prefix: /-/metrics
    route:
    - destination:
        host: gitlab-webservice-default.gitlab.svc.cluster.local
        port:
          number: 8181
  - match:
    - uri:
        prefix: /
    route:
    - destination:
        host: gitlab-webservice-default.gitlab.svc.cluster.local
        port:
          number: 8181
  - match:
    - uri:
        prefix: /admin/sidekiq
    route:
    - destination:
        host: gitlab-webservice-default.gitlab.svc.cluster.local
        port:
          number: 8080
```      

- **Given** a VirtualService that contains multi-port/prefix routes
- **When** a Package gets created to emulate the VirtualService is deployed
- **Then** A VirtualService is created that has multiple match statements routing traffic to multiple ports.

I've considered that this complexity might not be appropriate to be supported by the `Package` custom resource and the advice might be to just not rely on the UDS operator to create a VirtualService.

### Additional context
The current implementation supports routing to a single service with a single port.  As a terrible example we could do something like this: 
```
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: gitlab
spec:
  network:
    expose:
      - service: gitlab-webservice-default
        podLabels:
          app: webservice
        gateway: tenant
        host: gitlab
        ports: 
          - number: 8181
            prefix: /
          - number: 8080
            prefix: /admin/sidekiq
```

jeff-mccoy:
## Description

This PR introduces Istio VirtualService [HttpMatchRequest](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPMatchRequest) rules for the UDS Operator `expose` field. While a bit more limited than their Istio counterparts, the following fields are carried over into UDS:

- `ignoreUriCase`
- `method`
- `name` * UDS requires this field (optional in Istio)
- `queryParams`
- `uri`

Additional this PR mirrors naming validation/generation behavior UDS Operator uses for `allow` entries, including adding an optional `description` field to name the generated resource. 

## Related Issue

Fixes #114

## Type of change


rjferguson21:
Thanks for doing this so fast @jeff-mccoy.  

It might have been a conscious choice but I don't think this implementation will satisfy the GitLab example in [the issue](https://github.com/defenseunicorns/uds-core/issues/114) where a single service is exposing multiple ports.  

For other's context this is what the example VS with `matches` generates from this PR (notice the single port):
```
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: httpbin
  namespace: test-admin-app
spec:
  network:
    expose:
      - service: httpbin
        podLabels:
          app: httpbin
        gateway: admin
        host: demo
        port: 8000
        targetPort: 80
        match:
          - name: test-get-and-prefix
            method:
              # Only allow GET requests
              regex: GET
            uri:
              # Only allow routing to /status/2*, everything else should 404
              prefix: /status/2
          - name: test-exact
            uri:
              # Only allow routing to /status/410
              exact: /status/410
```
```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: httpbin-admin-demo
  namespace: test-admin-app
spec:
  gateways:
  - istio-admin-gateway/admin-gateway
  hosts:
  - demo.admin.uds.dev
  http:
  - match:
    - method:
        regex: GET
      name: test-get-and-prefix
      uri:
        prefix: /status/2
    - name: test-exact
      uri:
        exact: /status/410
    route:
    - destination:
        host: httpbin.test-admin-app.svc.cluster.local
        port:
          number: 8000
```

jeff-mccoy:
Yeah I looked at the example, this should work, but slightly differently (as in 2 expose entries) and dropping the fault entry. I looked at why they were doing that and there are other ways in istio to do that--including using regex in this match to never include the value.

rjferguson21:
That suggestion implies we will create two separate VS?  Definitely understand the trade off to keep the `Package` spec as terse as possible but wanted to clarify that is what you're suggesting as the outcome. 

The way the VS [are named today ](https://github.com/defenseunicorns/uds-core/blob/operator-vs-enhancement/src/pepr/operator/controllers/istio/virtual-service.ts#L28) it won't let you create multiple VS for a particular pkg/gateway/host combo and overwrites the previous one. 

We could potentially merge multiple `expose[]` into into the same VS...

jeff-mccoy:
Updated & tested with a few various styles to see what it would look like.

Screenshot 2024-01-24 at 7 54 31 PM
Screenshot 2024-01-24 at 7 42 37 PM
Screenshot 2024-01-24 at 7 42 15 PM
Screenshot 2024-01-24 at 7 42 10 PM

mjnagel (approved):

zachariahmiller (approved):

PR #147, opened 2024-02-02
feat: add keycloak

jeff-mccoy:
## Description

Adds Keycloak from P1, with some key changes, including UDS Core breaking change for Istio configs:

- No default admin user is created for any realm. In order to administer this instance, you must use `zarf connect keycloak` to establish the global admin account.
- TLS termination is performed by Istio via the [OPTIONAL_MUTUAL TLS config](https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings-TLSmode) to enable proper Istio traffic management and multi-point mTLS validation.
- The Keycloak instance is configured to use the [UDS Identity Config](https://github.com/defenseunicorns/uds-identity-config) image, which is a custom image that contains all the customizations needed for UDS including the core custom Keycloak plugin, custom themes, realm initialization, and PKI trust store.
- The [UDS Operator](https://github.com/defenseunicorns/uds-core/tree/main/src/pepr/operator/README.md) is used to declaratively manage Keycloak clients without any global admin-level credentials stored in the cluster.
- The [P1 Quarkus java plugin](https://repo1.dso.mil/big-bang/product/plugins/keycloak-p1-auth-plugin/-/tree/main/quarkus-ext-routing?ref_type=heads) has been replaced with Istio logic.
- The [P1 Java plugin](https://repo1.dso.mil/big-bang/product/plugins/keycloak-p1-auth-plugin) group restrictions and yaml config have been removed and replaced with Istio logic.
- The [unofficial helm chart](https://github.com/codecentric/helm-charts/tree/master/charts/keycloak) from German company Codecentric has been rewritten, simplified in this repo by Defense Unicorns.
- By default, this package deploys Keycloak in "dev mode", which disables HA and uses an H2 database persisted to a PVC. This is not suitable for production use, but is useful for development and testing. Dev mode also makes development and testing easier by increasing debug logs and disabling caching. Setting the helm value `devMode` to `false` will enable HA and use a PostgreSQL database, you must include credentials for a PostgreSQL database in the `keycloak` chart values.

BREAKING CHANGES: 

- [Path normalization for Istio configs](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-path_normalization)
- Istio GW chart configs changed

## Type of change

## Checklist before merging


src/istio/values/values.yaml
```diff
@@ -2,3 +2,5 @@ meshConfig:
   accessLogFile: /dev/stdout
   defaultConfig:
```
bburky, line 5: Enable Istio path normalization.
https://istio.io/latest/docs/reference/config/security/normalization/

I tested that this fixes some bypasses like `https://sso.uds.dev//admin` but have not yet tested this comprehensively.

```suggestion
  pathNormalization:
    normalization: DECODE_AND_MERGE_SLASHES
  defaultConfig:
```

At first I thought this only applied to AuthorizationPolicy path rules, but these docs mention that in addition to AuthorizationPolicies, the normalized path is used for routing and "the URL path proxied to the upstream service", so **this is perhaps as a breaking change across the entire cluster**.
https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-path_normalization

Istio best practices docs discussing normalization:
https://istio.io/latest/docs/ops/best-practices/security/#understand-path-normalization-in-authorization-policy

src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -0,0 +1,29 @@
...
+    matchLabels:
+      {{- include "keycloak.selectorLabels" . | nindent 6 }}
+  action: DENY
+  rules:
+  - to:
+    - operation:
+        paths: 
+          - "/admin*"
+          - "/realms/master*"
+          - "/metrics*"
```
bburky: Can we do an allowlist instead of a denylist for things going to the tenant gateway?

This also would meet Istio best practices to "Use ALLOW-with-positive-matching and DENY-with-negative-match patterns".
https://istio.io/latest/docs/ops/best-practices/security/#use-allow-with-positive-matching-and-deny-with-negative-match-patterns
In this case: do an ALLOWlist with `paths:` of the things that should be reachable from the tenant gateway (instead of a deny list of things that should be unreachable)


If you switch to an allowlist with an ALLOW rule, you should also probably create a default deny (technically "ALLOW nothing") rule by making a default ALLOW rule that has nothing in it.
See: https://istio.io/latest/docs/tasks/security/authorization/authz-http/#configure-access-control-for-workloads-using-http-traffic
```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: allow-nothing
  namespace: default
spec:
  {}
```

src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -0,0 +1,29 @@
...
+      notValues: ["istio-admin-gateway"]
+  - to:
+    - operation:
+        paths: 
+          # Never allow anonymous client registration except from the pepr-system namespace
+          # This is another fallback protection, as the KC policy already blocks it
+          - "/realms/{{ .Values.realm }}/clients-registrations/*"
+    when:
+    - key: source.namespace
+      notValues: ["pepr-system"]
```
bburky: This one's probably tricky to turn into an allowlist sadly.

I think this is equivalent to the current policy, I tested it locally:
```suggestion
    from:
    - source:
      notNamespaces: ["pepr-system"]
```

src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -0,0 +1,29 @@
...
+  rules:
+  - to:
+    - operation:
+        paths: 
+          - "/admin*"
+          - "/realms/master*"
+          - "/metrics*"
+    when:
+    - key: source.namespace
+      notValues: ["istio-admin-gateway"]
```
bburky: I think this is equivalent, I tested it locally:
```suggestion
    from:
    - source:
      notNamespaces: ["istio-admin-gateway"]
```

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,84 @@
...
+              uri:
+                prefix: /metrics
+          rewrite:
+            uri: "/realms/{{ .Values.realm }}/account"
+          headers:
+            request:
+              remove:
+                - istio-mtls-client-certificate
+              add:
+                istio-mtls-client-certificate: "%DOWNSTREAM_PEER_CERT%"
```
bburky, line 62: I think this can just be `set:` instead of `remove:` and `add:`

`set:` is documented as:
> Overwrite the headers specified by key with the given values

```suggestion
              set:
                istio-mtls-client-certificate: "%DOWNSTREAM_PEER_CERT%"
```

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,84 @@
...
+        gateway: admin
+        host: keycloak
+        port: 8080
+        advancedHTTP:
+          headers:
+            request:
+              remove:
+                - istio-mtls-client-certificate
+              add:
+                istio-mtls-client-certificate: "%DOWNSTREAM_PEER_CERT%"
```
bburky, line 91:
```suggestion
              set:
                istio-mtls-client-certificate: "%DOWNSTREAM_PEER_CERT%"
```

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,84 @@
...
+        host: sso
+        port: 8080
+        advancedHTTP:
+          headers:
+            request:
+              remove:
+                - istio-mtls-client-certificate
+              add:
+                istio-mtls-client-certificate: "%DOWNSTREAM_PEER_CERT%"
+
```
bburky, line 77:
```suggestion
              set:
                istio-mtls-client-certificate: "%DOWNSTREAM_PEER_CERT%"
```

src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: keycloak-block-admin-access-from-public-gateway
+  namespace: {{ .Release.Namespace }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "keycloak.selectorLabels" . | nindent 6 }}
+  action: DENY
```
bburky, line 11:
```
$ istioctl analyze -n keycloak
Warning [IST0133] (AuthorizationPolicy keycloak/keycloak-block-admin-access-from-public-gateway) Schema validation warning: configured AuthorizationPolicy will deny all traffic to TCP ports under its scope due to the use of only HTTP attributes in a DENY rule; it is recommended to explicitly specify the port
```

Just FYI. I don't think you need to address this?

src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: keycloak-block-admin-access-from-public-gateway
```
bburky, line 5: Out of caution, I'd strongly recommend a STRICT PeerAuthentication policy in the `keycloak` namespace.

This policy uses negation on namespaces... but requests _don't have_ a namespace if they come over a permissive connection. In practice UDS currently is STRICT by default including keycloak. I think it would be good defensive configuration to have a namespace-local PeerAuthentication (and really strong warnings that you shall not turn this off).

If the mTLS policy is PERMISSIVE, then cluster-internal HTTP requests to http://keycloak-http.keycloak.svc.cluster.local:8080 will basically bypass this DENY AuthorizationPolicy because these requests don't have a namespace set.

See https://istio.io/latest/news/security/istio-security-2021-004/ for further context.

bburky (changes requested):
Some AuthorizationPolicy hardening suggestions

Warning: potentially breaking changes from istio meshConfig updates, see comments for further detail.

src/keycloak/chart/templates/uds-package.yaml
jeff-mccoy: I ran into a problem with set and switched it--think when you didn't have a cert it was allowing any header you passed.

src/istio/values/values.yaml
jeff-mccoy: This is a pretty unusual / non-ideal behavior I think we can accept. I'll merge now, and we'll see if others disagree during the PR review

src/istio/chart/templates/gateway.yaml
```diff
@@ -4,36 +4,36 @@ kind: Gateway
 metadata:
   name: {{ .Values.name }}-gateway
   namespace: {{ .Release.Namespace }}
-  labels:
-    "uds/istio-domain": {{ $domain }}
+  annotations:
+
```
bburky: Delete empty annotations list
```suggestion
```

src/istio/common/zarf.yaml
```diff
@@ -24,3 +24,4 @@ components:
         files:
           - "manifests/envoy-filters.yaml"
           - "manifests/peer-authentication.yaml"
+          - "manifests/hack-pepr.yaml"
```
bburky: Was this just left over from testing (I'm guessing from the "hack" name) or an intended a permanent change? Why does the pepr webhook need to be permissive now?

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -0,0 +1,187 @@
...
+  try {
+    // Not including the CR data in the ref because Keycloak client IDs must be unique already
+    const name = getClientName(clientReq);
+    const token = Store.getItem(name);
+
+    let client: Client;
+
+    // If the redirectUris are not set, default to *
+    if (!clientReq.redirectUris) {
+      clientReq.redirectUris = ["*"];
```
bburky: This is a pretty insecure default. Can we just make the field required?

src/grafana/values/values.yaml
```diff
@@ -5,3 +5,30 @@ sidecar:
...
+    # https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/keycloak/
+    enabled: true
+    client_id: $__file{/etc/secrets/auth_generic_oauth/clientId}
+    client_secret: $__file{/etc/secrets/auth_generic_oauth/secret}
+    scopes: openid profile
+    email_attribute_path: email
+    login_attribute_path: preferred_username
+    name_attribute_path: name
+    name: UDS Identity Service
+    auth_url: https://keycloak.admin.###ZARF_VAR_DOMAIN###/realms/uds/protocol/openid-connect/auth
```
mjnagel: Curious on the why here for using the `keycloak.admin` VS instead of the `sso` VS? Tracking that grafana is on the admin gateway so it sorta makes sense, just not sure if that makes the user experience weird when the admin VS is typically used for admin/master realm things?

src/grafana/values/values.yaml
```diff
@@ -5,3 +5,30 @@ sidecar:
...
+    token_url: http://keycloak-http.keycloak.svc.cluster.local:8080/realms/uds/protocol/openid-connect/token
```
rjferguson21, line 33: Is this to avoid going through the gateway? Or just because internal DNS does not work?

src/istio/common/zarf.yaml
mjnagel: We've had to do this with webhooks/api services because the origin of the traffic is from the nodes/kube-proxy (not istio injected things). As to why it's added now - pepr is getting injected in this PR, previously was not.

src/istio/chart/values.yaml
```diff
@@ -4,19 +4,25 @@ name: change-me
 # Domain name for the gateway
 domain: "###ZARF_VAR_DOMAIN###"
 
-# Hosts to add to this gateway
-hosts:
-  - "*"
+# tls:
+#   # Name of the gateway port to use for TLS
+#   - name: https
```
mjnagel: Changing this to a list of TLS makes a ton of sense but it does make the overrides for bundles a bit gross I think (targeting specific list items, merging things, etc). I think we can easily switch this to a map where the key is the name, as long as we adjust the templates accordingly.
```yaml
tls:
  https:
    # The TLS mode for the gateway. One of `SIMPLE`, `MUTUAL`, 'OPTIONAL_MUTUAL', `PASSTHROUGH`
    mode: SIMPLE

    # Hosts to add to this gateway
    hosts:
      - "*"

    # The secret name for the TLS certificate when not using passthrough mode
    secretName: gw-cert

    # For TLS SIMPLE mode, the TLS certificate data for the gateway (base64 encoded)
    cert: ""

    # For TLS SIMPLE mode, the TLS key data for the gateway (base64 encoded)
    key: ""

    # For mTLS mode, the CA certificate data for the gateway (base64 encoded)
    cacert: ""
```
For your example one it would look like this, additional ones for keycloak, etc would just be `tls.keycloak`. On the template side `range` will work the same, we'll just need `{{ range $name,$tls := .Values.tls }}` to capture the key (`$name`), and then update any references to `$tls.name` to `$name`.

Should be almost entirely equivalent, but would make it a bit easier on the bundle side to target `tls.keycloak.cacert` for example.

src/istio/values/config-tenant.yaml
```diff
@@ -1,5 +1,15 @@
...
+  - name: keycloak
+    mode: OPTIONAL_MUTUAL
+    hosts:
+      - "sso"
```
mjnagel: semi-related to above comment on bundle overrides for these...would be great to not have to set the cert/key in two spots. We could make cert/key abstracted to zarf vars which I think would address it, just would look pretty gnarly in the defaults for the `zarf.yaml` to handle dev. 

Not sure on any other way to do this - but I don't think the UX of having to override 2x per gateway is great.

src/keycloak/chart/templates/istio-admin.yaml
mjnagel: Think this makes sense to do @bburky, and we could just force STRICT with no config for it (of course someone could delete but then they more likely know they did something bad).

Because I confused myself, tacking onto this thread that any access from the tenant gateway won't actually hit this DENY policy I don't think. The VS created by `src/keycloak/chart/templates/uds-package.yaml` will already redirect these requests to the `uds` realm (or other configured realm), so this DENY would primarily affect other in-cluster traffic (not originating from tenant gateway).

src/keycloak/chart/templates/vs-emulate-gitlab.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if .Capabilities.APIVersions.Has "security.istio.io/v1beta1" }}
+apiVersion: networking.istio.io/v1beta1
+kind: VirtualService
+metadata:
+  name: emulate-gitlab
+  namespace: {{ .Release.Namespace }}
```
mjnagel: I'm sure I'm just missing the connecting dots, but why is this VS created via the Istio CR instead of the UDS Package? Match and rewrite should be supported in that spec and I think you dodge name conflicts using `description`?

Also not clear to me - does this need the same `istio-mtls-client-certificate` remove/add logic?

src/istio/common/zarf.yaml
bburky: Ah, ok. Can we rename "hack-pepr" for clarity? Maybe just match the "permissive-pepr-webhook" name of the PeerAuthentiucation resource.

src/keycloak/chart/templates/istio-admin.yaml
bburky: Yes, if the VirtualService redirect is working correctly, then you'll never hit the DENY AuthorizationPolicy. However, we want to use AuthorizationPolicies for security enforcement, don't trust VirtualService configs to actually work. In practice, I don't currently know a way to bypass the VirtualService redirect. I tested this DENY AuthorizationPolicy by temporarily deleting the VirtualService redirect.

Also, this security model is important for Authservice too. There are a few ways to trick Authservice and bypass it. But its ALLOW AuthorizationPolicy does its job and denies any requests that don't have an authorized JWT.

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,84 @@
...
+        
+      # Temp workaround for any cluster pod
+      # @todo: remove this once cluster pods is a remote generated target
+      - description: "Keycloak backchannel access"
+        direction: Ingress
+        selector:
+          app.kubernetes.io/name: keycloak
+        remoteGenerate: Anywhere
+        port: 8080
+
```
mjnagel, line 26: Missing a netpol here for "egress anywhere" to cover:
- todo for egress to postgres until we can lock that down further
- egress to OCSP once that is added

I can do this in the truststore PR, just thought it might be relevant here with postgres already?

src/keycloak/chart/templates/vs-emulate-gitlab.yaml
zachariahmiller: This stuck out to me when setting up the flavors since I had to add the istio template conditional for the single package test and the reasoning wasn't apparent.

zachariahmiller (commented):
Aside from also wanting clarification on a few of the comments from @mjnagel and @rjferguson21 where I replied or reacted I didn't really have any other feedback or questions. 

I think it would maybe be helpful in the pr description to elaborate a little more on the breaking changes. Understood it's still low impact given current usage, but clarity on the implications for anyone deploying this already, even just for testing UDS packages, would be a positive thing 🦄

Happy to see how this has been coming along!

src/keycloak/chart/templates/vs-emulate-gitlab.yaml
jeff-mccoy: I actually think it might have justs been a holdover--I added the advanced http stuff in this pr originally and then we spun it out

src/grafana/values/values.yaml
jeff-mccoy: I was trying to keep all back-channel comms internal to the cluster and avoid going through the gw

src/grafana/values/values.yaml
jeff-mccoy: I guess maybe it doesnt matter? My thinking was if grafana is on the admin GW (which could be an isolated network), it might be a problem if egress on that admin network was limited

src/pepr/operator/controllers/keycloak/client-sync.ts
jeff-mccoy: yep no prob

src/grafana/values/values.yaml
mjnagel: Yeah that's a good callout - good with this and we can see if it is problematic down the road.

src/keycloak/chart/templates/istio-admin.yaml
jeff-mccoy: going to leave this for a future PR to evaluate.

mjnagel (dismissed):

zachariahmiller (dismissed):

rjferguson21 (dismissed):

mjnagel (approved):

zachariahmiller (approved):

bburky (approved):

PR #148, opened 2024-02-05
fix: network policy to allow metrics-server ingress

### LINKED ISSUE #139: Metrics server not functional on multi-node clusters
### Environment
Device and OS: MBP, macOS
App version: 0.10.0 / `main`
Kubernetes distro being used: k3d/rke2

### Steps to reproduce
1. Deploy a multinode k3d cluster (i.e. `zarf p deploy oci://defenseunicorns/uds-k3d:0.3.1 --set k3d_extra_args="--servers=3" --confirm`)
2. Zarf init
3. Deploy uds-core from the latest release or main
4. `kubectl top nodes` a few times in a row

### Expected result

Call to metrics via `kubectl top` would succeed.

### Actual Result

On a 3 server k3d cluster the call succeeds consistently every 3rd time. On an RKE2 cluster with 3 servers and 1 agent the api is consistently down (does not respond).

```console
❯ k top nodes
NAME               CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%   
k3d-uds-server-0   322m         1%     4396Mi          6%        
k3d-uds-server-1   343m         1%     4133Mi          6%        
k3d-uds-server-2   303m         1%     3710Mi          5%        
❯ k top nodes
error: Metrics API not available
❯ k top nodes
error: Metrics API not available
❯ k top nodes
NAME               CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%   
k3d-uds-server-0   322m         1%     4396Mi          6%        
k3d-uds-server-1   343m         1%     4133Mi          6%        
k3d-uds-server-2   303m         1%     3710Mi          5%        
❯ k top nodes
error: Metrics API not available
❯ k top nodes
error: Metrics API not available
```

### Severity/Priority

Medium, HPAs are not functional as a result - meaning istio gateway scaling is not functional.

### Additional Context

After a bit of exploration an ingress policy allowing ingress from anywhere -> metrics server resolves the connection but this seems overly permissive based on the requirements listed upstream - https://github.com/kubernetes-sigs/metrics-server?tab=readme-ov-file#requirements

mjnagel:
## Description

Allows ingress from anywhere to the metrics-server endpoint. After investigation, calls to `kubectl top` are source-NAT-ed when they hit metric-server so the source IP ends up being an IP in the pod CIDR range, effectively the "gateway IP" for the node (on k3d typically `10.42.<server pod cidr, 0/1/2/...>.<0/1>`). This doesn't appear to be something we can consistently grab/calculate. One way we could lock this down would be by adding the CIDRs from all nodes `podCidr`s, but no sure of the value of that vs anywhere besides optics.

Effectively we need a "KubeAPI Ingress" generated rule that would handle the above scenario. I think this is also necessary for any webhooks, in some testing the prometheus webhook is failing as well, but has an Ignore policy that hides the issue.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/139

## Type of change

## Checklist before merging

### DIFF
--- a/src/metrics-server/chart/templates/uds-package.yaml
+++ b/src/metrics-server/chart/templates/uds-package.yaml
@@ -7,7 +7,18 @@ spec:
   network:
     allow:
       - direction: Egress
-        # todo: evaluate an "all nodes" generated rule
         podLabels:
           app.kubernetes.io/name: metrics-server
+        # todo: evaluate an "all nodes" generated rule
         remoteGenerated: Anywhere
+        port: 10250
+      - direction: Egress
+        podLabels:
+          app.kubernetes.io/name: metrics-server
+        remoteGenerated: KubeAPI
+      - direction: Ingress
+        podLabels:
+          app.kubernetes.io/name: metrics-server
+        # todo: evaluate a "KubeAPI" _ingress_ generated rule
+        remoteGenerated: Anywhere
+        port: 10250


zachariahmiller (approved):
Based on our chat... assuming we're okay with this for the short term as a path forward/ until we find a better solution (if it exists) LGTM.

cmwylie19:
It does seem like ipBlocks and cidr ranges are the way to go for Kubelets

jeff-mccoy (approved):
Agree with this now until we establish a solid generated solution.

PR #153, opened 2024-02-07
feat: add authservice to uds-core

mjnagel:
## Description

Adds authservice as a package to UDS Core. Out of the box this doesn't mean much without IDAM but it gets the package in place with proper UDS-isms.

## Related Issue

None

## Type of change

## Checklist before merging


src/pepr/config.ts
```diff
@@ -1,4 +1,4 @@
-const isZarfEnv = process.env.UDS_DOMAIN !== "###ZARF_VAR_DOMAIN###";
+const isZarfEnv = process.env.UDS_DOMAIN === "###ZARF_VAR_DOMAIN###";
```
jeff-mccoy: so i do think this might be a problem, if you are using `npx pepr dev` or `npx pepr deploy`, you will not be using Zarf and the template string will still exist. If you are using Zarf it will be empty or the value of the Zarf variable.

mjnagel: I'll have to poke at this some more - maybe we just need a different check. I tossed a debug log in there while doing a UDS deploy to dump `process.env.UDS_DOMAIN` and it was still the zarf template string, not the actual value. I think with pepr dev it was undefined if I remember right, not sure on pepr deploy. Will test those scenarios out.

mjnagel: Scenarios and what `process.env.UDS_DOMAIN` is set to:
- `npx pepr dev`: undefined
- `npx pepr deploy`: `###ZARF_VAR_DOMAIN###`
- `uds run test-uds-core`: `uds.dev`
- `UDS_PKG=authservice uds run test-single-package`: `###ZARF_VAR_DOMAIN###`

So the issue is single capability overlap with `npx pepr deploy`. With what's on `main` it will assume istio on single capability, with what's here on my branch it will assume NO istio on `npx pepr deploy` (both of which are the opposite of what we want). I checked `process.env.UDS_WITH_ISTIO` and we'd hit the same overlap with those as well

Not sure if there's fundamentally any difference between `npx pepr build` + `zarf p deploy` (what is used by `uds run test-single-package`) and `npx pepr deploy`? If not we are probably looking at a runtime check against the cluster or (feels janky) we toss a `--set` on https://github.com/defenseunicorns/uds-core/blob/authservice-add/tasks/deploy.yaml#L27 for something like `SINGLE_CAPABILITY=true` that we can check on the pepr config side. Going to mess with this and see if I can figure out a happy path that doesn't seem clunky.

mjnagel: This is the check we ended up with - https://github.com/defenseunicorns/uds-core/pull/153/commits/70054945600b966d5593129f7aeb9fb84aa60132

If handles the undefined case correctly + sets domain on the single capability deploy test. Not sure its the cleanest solve, seems a bit hidden, but without checking against the cluster for istio CRDs/pods/ns I'm not sure we have anything else to go off of.

src/authservice/chart/values.yaml
```diff
@@ -0,0 +1,162 @@
...
...XUwmcSyH8LP92NVIuFlD3HYTDGosVbfA_u_H25Z4XCGKW_vLDTNrl8PcA3HqIoD-vNavysdxAq_KNw7iLLc0KLsjFYSdJL_54H7QubsGR0AyIrLLurJbqAtvttGJK38k5XYWKIwYGtu6iiJwjSb7UtonVdPh8Vw\",\"e\":\"AQAB\",\"x5c\":[\"MIICoTCCAYkCBgFyLIEqUjANBgkqhkiG9w0BAQsFADAUMRIwEAYDVQQDDAliYWJ5LXlvZGEwHhcNMjAwNTE5MTAzNDIyWhcNMzAwNTE5MTAzNjAyWjAUMRIwEAYDVQQDDAliYWJ5LXlvZGEwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQCGIwvWSPD6zDbkGBpkjUDJ+BwJEE8ko8T7DC22prs03Gm2v/lEIEa4leoPKjenW+NxtvwqN0qdcjuRZ6vxvY/P9E/Wb9Bw+so7FKW6TsZkXwBGxzoU8ZvNiCLyjxwFVKxqaXodnOk3dmcfgMVnbuJ8z5SX8/IzFnZrB6iEhqLNen6uAXtGqlq/k1dTCZxLIfws/3Y1Ui4WUPcdhMMaixVt8D+78fblnhcIYpb+8sNM2uXw9wDceoigP681q/Kx3ECr8o3DuIstzQouyMVhJ0kv/ngftC5uwZHQDIissu6sluoC2+20YkrfyTldhYojBga27qKInCNJvtS2idV0+HxXAgMBAAEwDQYJKoZIhvcNAQELBQADggEBAIVkoDYkM6ryBcuchdAL5OmyKbmmY4WDrMlatfa3uniK5jvFXrmVaJ3rcu0apdY/NhBeLSOLFVlC5w1QroGUhWm0EjAA4zyuU63Pk0sro0vyHrxztBrGPQrGXI3kjXEssaehZZvYP4b9VtYpus6oGP6bTmaDw94Zu+WrDsWdFs+27VEYwBuU0D6E+ENDGlfR+9ADEW53t6H2M3H0VsOtbArEutYgb4gmQcOIBygC7L1tGJ4IqbnhTYLh9DMKNklU+tq8TMHacps9FxELpeAib3O0J0E5zYXdraQobCCe+ao1Y7sA/wqcGQBCVuoFgty7Y37nNL7LMvygcafgqVDqw5U=\"],\"x5t\":\"mxFIwx7EdgxyC3Y6ODLx8yr8Bx8\",\"x5t#S256\":\"SdT7ScKVOnBW6qs_MuYdTGVtMGwYK_-nmQF9a_8lXco\"}]}"
```
jeff-mccoy: not sure I love leaving these here, even if it does show an example of how to use it

mjnagel: I'll go back through and clean up some of these example values. Definitely a bit messy and unnecessary.

jeff-mccoy: sounds good, def a tomorrow us problem

mjnagel: Removed some of the example values where they didn't make sense - especially with this one I think we'd likely be using `jwks_uri` so less concerned about an example for `jwks` - might make sense to drop down that down the road even.

I also dropped a few extra knobs that are for wiring authservice to non-keycloak endpoints (Google SSO, etc) which I don't think we need? I can restore if we want those for something but think dropping them fits with our opinionation on uds-core?

src/authservice/chart/templates/authn.yaml
```diff
@@ -0,0 +1,20 @@
...
+apiVersion: security.istio.io/v1beta1
+kind: RequestAuthentication
+metadata:
+  name: jwt-authn
+  namespace: istio-system
+spec:
+  selector:
+    matchLabels:
+      {{ .Values.selector.key }}: {{ .Values.selector.value | quote }}
+  jwtRules:
```
bburky, line 12: I always wanted to include `audiences:` in this, but it would need to be done per namespace and not a global istio-system policy. This Helm chart was originally written to kind of do everything and made some compromises.

Currently this RequestAuthentication  policy allows any JWT from the correct issuer and does not check the audience `aud` claim (the Keycloak client ID). In theory with something like an XSS attack, you could steal one app's JWT and use it to authenticate do a different app.

In practice, many apps don't send the JWT down to the client... but some do. Also Authservice now strips the client-provided `Authentication:` header... but there have been ways to bypass Authservice entirely in the past. I don't consider these perfect mitigations and would prefer per client per namespace enforcement done in Istio.

Have we considered something like requesting Authservice protection in the UDS CR? Then Pepr could generate the RequestAuthentication per-namespace with a specific Client ID.

mjnagel: Definitely would trust your judgement and knowledge here way more than mine. Assuming a follow on PR handles client creation + wiring to authservice + creates a per client `RequestAuthentication`...seem wise to remove this global one from the chart now?

bburky: Whichever, remove now, leave it and opt out, or remove later. We can leave this if you want. Apps can opt out of the global policy by not using the `protect: keycloak` label or whatever it's configured to. If they don't match they label they don't get the AuthorizationPolicies either, but I think if you're doing per-namespace RequestAuthentication, then per-namespace AuthorizationPolicies may make sense too anyway?

mjnagel: Leaning towards leaving until we know better what the follow on / replacement looks like. Not sure it makes a big difference either way since keycloak isn't here yet but I think this is a good callout for the next PR for sure. Going to resolve this thread and keep it in mind for IDAM PR to follow :+1:

jeff-mccoy (approved):
Lgtm, as discussed we can tweak some of these configs in the other pr 🦄

PR #154, opened 2024-02-07
feat: introduce advancedHTTP for expose field & change podLabels to selector

jeff-mccoy:
This PR deprecates `podLabels`, 'remotePodsLabels', and `match` and changes this  to `selector`, `remoteSelector` and `advancedHTTP.match`, respectively. This PR also introduces `advancedHTTP` which brings support for the following `VirtualService` fields:

- `corsPolicy`
- `directResponse`
- `headers`
- `retries`
- `rewrite`
- `timeout`
- `weight`


chance-coleman (dismissed):

src/pepr/config.ts
```diff
@@ -1,8 +1,14 @@
-const isZarfEnv = process.env.UDS_DOMAIN !== "###ZARF_VAR_DOMAIN###";
+import { Log } from "pepr";
+
+// We need to handle `npx pepr <>` commands that will not template the env vars
+const domain = process.env.UDS_DOMAIN;
+const isZarfEnv = domain ? domain !== "###ZARF_VAR_DOMAIN###" : false;
```
mjnagel, line 5: If I'm reading this right I think it handles the undefined scenario I was tracking as a problem but not sure it addresses single package issue completely? See https://github.com/defenseunicorns/uds-core/pull/153/commits/70054945600b966d5593129f7aeb9fb84aa60132 where we set domain explicitly in the single-package deploy. Happy to address and resolve that out of band since it doesn't actually seem like an issue here...maybe something else going on with my authservice stuff 🙃

src/pepr/operator/controllers/istio/virtual-service.ts
```diff
@@ -105,10 +117,10 @@ export async function virtualService(pkg: UDSPackage, namespace: string) {
 }
 
 export function generateVSName(pkg: UDSPackage, expose: Expose) {
-  const { gateway = Gateway.Tenant, host, port, service, match, description } = expose;
+  const { gateway = Gateway.Tenant, host, port, service, description, advancedHTTP } = expose;
 
   // Ensure the resource name is valid
-  const matchHash = match?.flatMap(m => m.name).join("-") || "";
+  const matchHash = advancedHTTP?.match?.flatMap(m => m.name).join("-") || "";
```
mjnagel, line 118: Small nit: this section doesn't handle the deprecated `expose.match` I don't think?

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -23,7 +23,7 @@ export function generate(namespace: string, policy: Allow): kind.NetworkPolicy {
     spec: {
       policyTypes: [policy.direction],
       podSelector: {
-        matchLabels: policy.podLabels,
+        matchLabels: policy.selector,
```
mjnagel, line 26: Might be missing somewhere else this is happening but do we need/want to still support the deprecated `podLabels` + `remotePodLabels` here? Looks like they were dropped entirely from the logic.

Alternatively I'm okay with dropping them but we'd want to remove from the CRD as well.

src/pepr/operator/crd/sources/v1alpha1.ts
```diff
@@ -1,6 +1,162 @@
...
+        type: "array",
+        items: {
+          minimum: 1,
+          maximum: 65535,
+          type: "number",
+        },
+      },
+      // Deprecated fields
+      podLabels: {
+        description: "Deprectated: use selector",
```
mjnagel:
```suggestion
        description: "Deprecated: use selector",
```

src/pepr/operator/crd/sources/v1alpha1.ts
```diff
@@ -1,6 +1,162 @@
...
+        type: "object",
+        additionalProperties: {
+          type: "string",
+        },
+      },
+      remotePodLabels: {
+        description: "Deprectated: use remoteSelector",
```
mjnagel:
```suggestion
        description: "Deprecated: use remoteSelector",
```

src/pepr/operator/crd/sources/v1alpha1.ts
```diff
@@ -1,6 +1,162 @@
...
+        maximum: 65535,
+        type: "number",
+      },
+      // Deprecated field
+      match: {
+        description: "Deprecated: use advancedHTTP.match",
+        ...advancedHTTP.properties?.match,
+      },
+      podLabels: {
+        description: "Deprectated: use selector",
```
mjnagel:
```suggestion
        description: "Deprecated: use selector",
```

mjnagel (commented):
Couple minor comments, overall good. Probably a good pattern we can set here for deprecating things and removal later on.

mjnagel (dismissed):
I like the approach with the migrate fn. Might need to regen the CRD file(s) after those spelling changes to update the comments? LGTM overall.

mjnagel (approved):

PR #156, opened 2024-02-08
feat: switch loki to simple scalable

### LINKED ISSUE #146: Switch loki to scalable mode (r/w/b pods)

Currently loki is deployed in a single binary mode which is not production ready.

UDS Core should deploy scalable mode out of the box, with it wired up to dev-minio for testing. We may need to expose zarf vars to make this a pleasant experience for connecting to S3 compatible storage.

Leave this to the end user, which ends up being a lot to figure out on their own /cc @brianrexrode

rjferguson21:
## Description
Switches Loki from monolith to simple scalable

Updated CI to use a bundle for some of the tasks to be able to configure minio with overrides.  Open to suggestions about other ways to do this.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/146
## Type of change

## Checklist before merging


bundles/k3d-standard/uds-bundle.yaml
```diff
@@ -39,6 +50,30 @@ packages:
     ref: 0.12.0
     # x-release-please-end
     overrides:
+      loki:
```
mjnagel, line 42: Not sure where this should live, but I think it would be good to document these "required" overrides in the repo somewhere.

src/loki/chart/values.yaml
```diff
@@ -1 +1 @@
-scalable: false
+scalable: true
```
mjnagel: Think we can drop this value with the change to the peerauth?

bundles/k3d-test/uds-bundle.yaml
```diff
@@ -0,0 +1,30 @@
...
+            - path: "buckets"
+              value:
+                - name: "uds-loki-chunks"
+                - name: "uds-loki-rules"
+                - name: "uds-loki-admin"
+            - path: "users"
+              value:
+                - accessKey: logging
+                  secretKey: "logging-secret"
+                  policy: readwrite
```
mjnagel: I'm wondering if it makes sense to just create a default user in uds-k3d. We already have a default bucket `test` and I think Loki can use the same bucket for all 3 storage pieces. If we made a default user/pass we could skip the need for this extra test bundle I think?

src/loki/values/values.yaml
```diff
@@ -50,6 +60,12 @@ test:
   enabled: false
 # Configuration for the write pod(s)
 write:
+  # Remove default anti-affinity since this is a 1-replica setup
+  affinity: ""
```
mjnagel: If affinity is an issue we might just want to do multi-node k3d in our demo bundle...I think between this and the metrics-server issue there's validation that its worth testing that way, assuming it doesn't have a massive impact on resource usage or startup time.

src/loki/values/values.yaml
```diff
@@ -50,6 +60,12 @@ test:
...
+
+  # replicas to 1 to save resources
+  replicas: 1
+
```
mjnagel: I think I'd rather have this HA by default and drop the resources blocks from our values. Resources are more likely to need tweaking per environment, but >1 replica is likely to be a need everywhere so I think its good to ensure we are testing multi-replica setups. Dropping default resource values does makes HPA a bigger effort to setup, but I think its the right move given how environment specific that tends to be.

src/loki/values/values.yaml
zachariahmiller: I would be surprised if that doesnt result in potentially having to go to the next size runner. If that is the case maybe a better middle ground solution would be to run this (multinode) test on nightly and/or tag/release pipeline before creating release.

bundles/k3d-test/uds-bundle.yaml
zachariahmiller: you can just use the built in admin user generated by the chart. If you wanted to test using a role to connect using the overrides here would be a good option. I think for this use case if the test is going to be with  a single bucket writing to folders you can probably just use the generated admin creds for minio and the test bucket.

Totally a viable option to add some defaults to the uds-k3d stack though and it should be minimal overhead if they arent being used.

src/loki/values/values.yaml
zachariahmiller: For gitlab i left the "real" setup as default and used variable overrides + a prebuilt uds-config.yaml for the "dev/ci/lite" set of values. I think that's a good pattern to follow for all the things. Thoughts?

bundles/k3d-test/uds-bundle.yaml
rjferguson21: Added an explicit user and bucket this [uds-k3d PR](https://github.com/defenseunicorns/uds-k3d/pull/41).  Not sure how I feel about adding hard-coded minio credentials over there but it would certainly make testing a lot easier by alleviating the need to configure the credentials/buckets with bundle overrides.  It seems like the least of all of the evils at this point.

@zachariahmiller I considered using the admin credentials but I was hoping to avoid creating tasks to 'setup' minio, or lookup credentials, plus I would not have anyway to set those credentials in the uds-core package unless I added ZARF_VARS.  Maybe there is an easy way to do that that I am not aware of but it seems like my choices are:

* pre-baked credentials for a minio user that are known ahead of time
* bundle overrides (I believe our ideal outcome is to avoid wrapping the single package tasks in a bundle)

src/loki/values/values.yaml
rjferguson21: @mjnagel This comment plus the one about HA leaves me thinking you'd like to see a default values.yaml for uds-core loki that would include:

* the default affinity which prevents multiple replicas from being scheduled on the same node
* replicas > 1  per component (read/write/backend) (default is 3)

I believe this would mean the default configuration would no longer work on a single node uds-k3d cluster.  Are we ok with that? I think it would mean folks would have to apply a dev config via a uds bundle config if they wanted to continue testing single node.  This would also affect us in CI for the tasks that previously were not wrapped in bundles.

zachariahmiller: I agree the default config should be the real/ha config. The overrides being only at bundle level right now vs zarf is a little frustrating, but imo overriding the default for CI/local dev with the overrides I think makes more sense. Understood on the impact to bundle-less tasks 😢

zachariahmiller: Maybe bundle level component selection could help work around that though  (for single package tests) 😉

bundles/k3d-test/uds-bundle.yaml
zachariahmiller: Quite the conundrum

mjnagel: > pre-baked credentials for a minio user that are known ahead of time

I think this is the right answer since it simplifies everything to a point where we can _hopefully_ do package tests without a bundle. It should also help with other charts that need s3 similarly, to some extent, without having to use zarf actions/tasks in the middle of a bunch/deploy to grab existing users somehow.

zachariahmiller: It's funny I was originally going to add that to the minio chart values and didn't because it was a package specific implementation of what was supposed to be a generic tool/stack. I think this is fine at the moment, hopefully we can get to a better way in the future.

src/loki/values/values.yaml
mjnagel: Yeah in agreement with the statements re:HA/prod config being the default. I think we definitely need to be testing the prod-ready setup with the full stack (not just single capability), so we have a few considerations here:
- single capability: this probably "has to" be the prod ready config if we want to avoid bundles for single capability, which would mean in this case it likely needs to be multi node cluster 🤔. We want this test pretty much exclusively for fast feedback, so if multi-node slows this down we'll have to rethink that.
- full stack lightweight dev: we could override with bundle overrides to make this a lightweight setup rather since we have bundle context here
- full stack prod ready: I kind of like the nightly idea here for testing, not sure if that's still in k3d vs rke2...if we want to really test prod setup k3d makes less sense. We also should balance testing in CI vs testing in our own long-lived staging, both of which are good.

This is a problem we're likely to have with other pieces of the puzzle - it's a balancing act of having uds-core in a "smart" setup out of the box but also providing the simplified dev experience. We prioritized dev experience first which I think is the right move but we need to figure out how to provide and test the prod side.

src/loki/values/values.yaml
mjnagel: Yeah agreed here on that - think the tricky part here is single capability which doesn't exist in gitlab + figuring out how/where we want to validate the prod setup.

bundles/k3d-test/uds-bundle.yaml
mjnagel: I think as long as we keep the user/bucket generic (currently `uds` in @rjferguson21 's PR there) it should be okay and generic enough for usage. Doesn't solve GitLab and other places where multiple buckets are needed, or handle fine-grained auth, but I think that's okay.

bundles/k3d-standard/uds-bundle.yaml
rjferguson21: Updated the README.md for that bundle - https://github.com/defenseunicorns/uds-core/pull/156/files#diff-0bb4d161ebc7c1143b9b02a825ca72a61e16aec663414b221d1e0f1c3f7a0b04

src/loki/values/values.yaml
zachariahmiller: I think I agree with everything you said here. What wasn't clicking before is that we could look at running the multi node setup just on the loki single package test.

src/loki/values/values.yaml
zachariahmiller: Technically gitlab is a single package test, but yeah the deploy test running in a bundle vs in a zarf package prevents getting around the issue with overrides. 🫠

zachariahmiller: I'm also a big fan of nightly "full" prod config test. RKE2 for now, potentially down the road rke2, eks, etc.

That will allow us to keep the frequent CI tests lightweight and quick still getting the feedback and confidence in the full deployment without large compute cost implications.

rjferguson21:
@mjnagel @zachariahmiller 

Updated to:
* test loki single-package with a 3 node cluster (other singe-package tests run single node)
* run upgrade test with 3 nodes (because it includes loki)
* standard bundle includes configuration to continue to support single node clusters for demos

Ended up modifying the cluster creation task to use `inputs` after I realized the upgrade test was also going to need the three node cluster and could not rely solely on checking `UDS_PKG`.

bundles/k3d-standard/README.md
```diff
@@ -8,26 +8,76 @@ This bundle is used for demonstration, development, and testing of UDS Core. In
 - [MetalLB](https://metallb.universe.tf/) - Provides type: LoadBalancer for cluster resources and Istio Gateways
 - [HAProxy](https://www.haproxy.org/) - Utilizes k3d host port mapping to bind ports 80 and 443, facilitating local FQDN-based routing through ACLs to MetalLB load balancer backends for Istio Gateways serving *.uds.dev, keycloak.uds.dev, and *.admin.uds.dev.
 
-## Configuration
+## Available Overrides
+## Package: uds-k3d
```
mjnagel: Feels like these should be different heading levels? Same as `## Package: core` below.

src/loki/chart/values.yaml
```diff
@@ -1 +0,0 @@
-scalable: false
```
mjnagel, line 1: Should we still have an empty default values file? Thought helm would blow up somewhere without it, but apparently it kinda works 🤔

src/loki/values/values.yaml
```diff
@@ -5,11 +5,28 @@ fullnameOverride: loki
...
 loki:
   storage:
-    type: "filesystem"
+    bucketNames:
+      chunks: uds
+      ruler: uds
+      admin: uds
+    type: s3
+    s3:
+      endpoint: http://minio.uds-dev-stack.svc.cluster.local:9000
```
mjnagel, line 21: So...I can't think of a good way to handle this...but I believe for IRSA you have to leave endpoint empty? Which means someone using IRSA will have to wipe out this value to get it to work. Can't think of a way to leave this empty by default but make CI happy...so maybe we just document the IRSA setup? Doesn't have to be done in this PR unless there is an obvious fix.

mjnagel (commented):
Tracking that CI is still being weird with promtail/loki but wanted to post some other comments from my review.

src/loki/values/values.yaml
rjferguson21: Yea, I think you're right.  I believe `s3forcepathstyle` would have to be unset (or false) as well.  Happy to document in a follow on PR or maybe as an addition to the EKS nightly testing.

bundles/k3d-standard/README.md
rjferguson21: :+1:  Should be fixed now.

src/loki/values/values.yaml
mjnagel: Yeah that works - we'll probably have a better grasp with the EKS testing and can point to an example.

mjnagel (approved):

PR #162, opened 2024-02-08
ci: use uds task instead of GH action for registry login

MxNxPx:
## Description

Due to intermittent login issues that fail CI workflow runs, we would benefit from being able to retry logins with some delay between retries.  This would move from using the GH Actions docker/login-action to a uds run task for registry logins. 

## Type of change

## Checklist before merging


tasks/setup.yaml
```diff
@@ -1,3 +1,7 @@
+variables:
```
zachariahmiller: @MxNxPx Any reason we wouldnt add this to the common tasks repo and then include it here?

MxNxPx: @zachariahmiller great question.

to do so we need to have [this uds-cli feature](https://github.com/defenseunicorns/uds-cli/issues/414) in order to import the remote task while supporting use of inputs via the `--with` flag.

.github/workflows/test.yaml
```diff
@@ -49,10 +49,12 @@ jobs:
...
-        with:
-          gh_token: ${{ secrets.GITHUB_TOKEN }}
-          ib_user: ${{ secrets.IRON_BANK_ROBOT_USERNAME }}
-          ib_password: ${{ secrets.IRON_BANK_ROBOT_PASSWORD }}
+
+      - name: Login to GHCR
+        run: uds run -f tasks/setup.yaml registry-login --with registry=ghcr.io --with registry-username=dummy --with registry-password=${{ secrets.GITHUB_TOKEN }} --with registry-retry-interval=5
+
+      - name: Login to registry1
+        run: uds run -f tasks/setup.yaml registry-login --with registry=registry1.dso.mil --with registry-username=${{ secrets.IRON_BANK_ROBOT_USERNAME }} --with registry-password=${{ secrets.IRON_BANK_ROBOT_PASSWORD }} --with registry-retry-interval=5
```
mjnagel: nit: since the default `registry-retry-interval` is 5 we can simplify the commands here and drop it I think?

tasks/setup.yaml
mjnagel: The only alternative I see here is not using the new `with`/`inputs` syntax and just using ENVs...could probably make common tasks work if we did that but not sure which makes more sense?

tasks/setup.yaml
```diff
@@ -12,3 +16,26 @@ tasks:
...
         cmd: "uds zarf package deploy oci://defenseunicorns/init:v0.32.1 --confirm"
+
+  - name: registry-login
+    inputs:
+      registry:
+        default: registry1.dso.mil
+      registry-username:
+      registry-password:
+      registry-retry-interval:
+        default: 5
```
mjnagel: Per the schema these are all invalid even though they functionally work:
- `description` is required
- `registry-username:` and the other empty ones are invalid (adding description should resolve)
- Apparently `default` has to be a string...

tasks/setup.yaml
MxNxPx: considering that the `--with` flag may be going away, the `--set` flag seems to work, and a common task exists... swapping to import that rather than bake into the uds-core repo

.github/workflows/test.yaml
MxNxPx: agreed, removed it so it takes the default

tasks/setup.yaml
MxNxPx: this is resolved in the common tasks repo now

tasks.yaml
```diff
@@ -35,6 +36,15 @@ tasks:
...
       - task: setup:k3d-test-cluster
 
+  - name: registry-login
+    actions:
+      - task: common-setup:registry-login
+        with:
+          registry: ${REGISTRY}
+          registryUsername: ${REGISTRY_USERNAME}
+          registryPassword: ${REGISTRY_PASSWORD}
+          registryRetryInterval: ${REGISTRY_RETRY_INTERVAL}
```
mjnagel, line 46: I think I recall this being problematic if you don't pass in a `REGISTRY_RETRY_INTERVAL` var? It ends up nulling out the default, so `sleep` errors out with no seconds set.

mjnagel:
```suggestion
```

tasks/setup.yaml
```diff
@@ -1,3 +1,7 @@
+variables:
+  - name: REGISTRY1_TEST_IMAGE
+    default: registry1.dso.mil/ironbank/opensource/defenseunicorns/zarf/zarf-agent
+
```
mjnagel:
```suggestion
```
Think we can drop this.

mjnagel (dismissed):

tasks.yaml
MxNxPx: good catch!
rather than remove, i am thinking we may want to tweak this number in the future if/when we encounter failures.  so i am adding back the `--set` in the CI job.

tasks.yaml
```diff
@@ -10,6 +10,7 @@ includes:
   - deploy: ./tasks/deploy.yaml
   - test: ./tasks/test.yaml
   - lint: ./tasks/lint.yaml
+  - common-setup: https://raw.githubusercontent.com/defenseunicorns/uds-common-tasks/v0.2.0/tasks/setup.yaml
```
mjnagel: Sorry for another comment - might be good to update to latest tag and repo name here just so we don't forget down the line. Also tracking this with renovate?

MxNxPx: @mjnagel good call - this has been updated along with the renovate config

mjnagel (dismissed):

mjnagel (approved):
Thanks for tackling this - hoping it gives us better insight into whether registry1 login flakiness.

Issue #130, opened 2024-01-24
Exemptions for policies in UDS Package CR


mjnagel:
### Is your feature request related to a problem? Please describe.

When I deploy my app I would like to specify a few policy exemptions because my app needs a certain privileged setup.

### Describe the solution you'd like

When I deploy my app I would like to be able to specify exemptions to policies via the UDS Package spec.

### Describe alternatives you've considered

Exemptions could be added to uds-core directly but this creates a longer feedback loop to wait on UDS Core to release the exemption.

### Additional context

While I do believe this is necessary I think it would be good for us to think through a safe way to add this support, and I think we absolutely need to provide a way to audit/display exemptions in one place so that they don't have to be discovered from across all UDS Packages + in the UDS Core code.

zachariahmiller:
Additional feedback relating to this issue from delivery:

**Persona:** Delivery, author of zarf packages / bundles that my end user will consume

**Feature Request:** Ability to ignore certain namespaces and/or certain pods from security policies (previously enforced by Kyverno, now by Pepr in UDS Core).

**Reason:**
1. Some workloads still require things like `runAsUser: 0` , even in prod.
2. When doing rapid development sometimes it is advantageous to turn off security policies to ensure that the application will work without them, with the plan to turn them back on once the application is functional ("start from known good state")

**Severity/Urgency:** High. UDS Core is likely unusable in its current state for any workload that requires these policy exceptions due to this missing capability.

**Acceptance Criteria:**
- I don't have to rebuild any docker images and can use the "source of truth" images that are hosted in Iron Bank
- I don't have to rebuild any Zarf packages and can use the "source of truth" OCI artifacts in GHCR
- If deploying a single Zarf package, I'm able to apply the policy exceptions at deploy time using simple configuration
- If the package is part of a Bundle, I'm able to apply the policy exception at bundle build time and decide for myself whether to allow my end user to modify it or not when they go to deploy the bundle

**Nice-To-Have:**
- I don't have to write any javascript/typescript to be able to do something that I used to be able to do with Kyverno in ~4 lines of YAML.

PR #165, opened 2024-02-09
feat: add policy exemptions

TristanHoladay:
## Description
This PR adds a way for users to register policy exemptions at runtime using the uds-core operator. 

## Related Issue
Relates to defenseunicorns/uds-core#130 

## Type of change

## Checklist before merging


src/pepr/operator/reconciler.ts
```diff
@@ -65,17 +65,55 @@ export async function reconciler(pkg: UDSPackage) {
...
+
 /**
  * Updates the status of the package
  *
- * @param pkg The package to update
+ * @param cr The package to update
  * @param status The new status
  */
-async function updateStatus(pkg: UDSPackage, status: Status) {
+async function updateStatus(cr: UDSPackage | UDSExemption, status: Status) {
```
jeff-mccoy: might be worth creating an `interface` for this

TristanHoladay: Were you thinking an actual interface like `interface UDSCR extends GenericKind {}` since both Package and Exemption extend GenericKind.

or a type for the intellisense `type UDSCR = UDSPackage | UDSExemption`

jeff-mccoy: eh either is fine, I'd probably choose interface

TristanHoladay:
Ran several options with setting exemptions in the store.

1. synchronously only using `setItem()` -- not too surprisingly doesn't work when writing to the same policy key multiple times (neuvector for example does this). Because of how the store works, if not given enough time only the last write makes it to the store and then most pods needing exemptions for that policy are denied.

2. asynchronously with `setItemAndWait()` -- solves the concurrent write problem but is slow and exacerbates the issue of pods starting up before exemptions have made it into the store. In neuvector's case, this means the enforcer pod tries to run and is denied one or more times before successfully deploying. 
  - trial 1: `"msg":"Time to complete exemption write: 118758.07304999232"`
  - trial 2:  `"msg":"Time to complete exemption write: 118881.12059801817"}`
  
3. synchronously with local aggregation first -- solves the concurrent write issue by first creating a local map and aggregating all matchers from the Exemption into lists attached to their relevant policy. Then a single write is made for each policy to the store without any waiting for the previous call to finish. This drastically reduces the time for the exemptions to show up in the store. There still seems to be cases of pods getting denied while waiting. However I noticed that they seemed to take less attempts and the neuvector deployment finished faster in this case.
  - trial 1: `"msg":"Time to complete exemption write: 2.490468978881836"}`
  - trial 2: `"msg":"Time to complete exemption write: 0.44616997241973877"}`
  - trial 3: `"msg":"Time to complete exemption write: 0.9177009463310242"}`
  - trial 4: `"msg":"Time to complete exemption write: 0.4052410125732422"}`
  > Note: these write times are not truly accurate for how long it took for the store to be updated with all exemptions. These numbers reflect how long it took to create the local map and then send each update to the store.

jeff-mccoy (changes requested):
Won't be able to review for a couple days due to in-person things, but wanted to flag to make sure I review this one as it's going to touch some of the same areas as the identity work.

blancharda:
Will exemptions be required to be defined in the `uds-policy-exemptions` namespace?

It seems all of the included definitions ([such as the Neuvector exemption](https://github.com/defenseunicorns/uds-core/pull/165/files#diff-87adef2fab7f3c3aebd48a2f5e8953e58d3bd39fbd548e73b22b09fcf72cdc6cR5)) are -- but (imo) it feels odd for the Neuvector chart to be creating resources in other namespaces.

blancharda:
> Will exemptions be required to be defined in the `uds-policy-exemptions` namespace?

It seems the [current answer is yes](https://github.com/defenseunicorns/uds-core/pull/165/files#diff-b5316f6ab132ae25f900cb362582c788675e57c8642e9de596960ede8eb9efecR53) and after reviewing previous discussion of the decision, it seems intentional.

I still think it feels like an antipattern for our charts to be creating resources in a non-chart namespace. Even helm chart dependencies (which can be whole other applications) are all installed in the same namespace as the parent chart. 

Perhaps an alternative would be to pull the exemption definitions out into a separate package/chart, or consolidate them under `/src/pepr/policies/exemptions`?

TristanHoladay:
@blancharda yeah based on previous conversations the single namespace iirc was decided on to make restricting who can create them easier for cluster admins. I agree it's a little weird to have a chart create a resource in a separate namespace, but i don't personally find it too off-putting. 

From a user perspective, the Exemption CR could be applied from a separate chart / package, as long as it's before the main chart so the pods can be exempted. That shouldn't matter to the operator. 

For uds-core, the desire was to take the hard-coded exemptions out of that path `src/pepr/policies/exemptions` and add them to the `uds-config` charts as CRs for those tools (Neuvector, Promtail, Prometheus). Is your thought to leave them as CRs but move them back under that path?

blancharda:
I think I'm struggling on two points.
1. I don't think the charts should be installing things to multiple namespaces.. it's not unheard of (and is certainly possible) but makes assumptions about pre-existing namespaces that helm charts don't typically make when it can be avoided
2. CRDs are cluster-wide objects and it seems like it will introduce confusion to limit how/where we respond to corresponding CRs that are created. 

If the CRD is installed in my cluster, I would expect for it to have the same effect, regardless of the namespace I put it in. I'm having trouble thinking of another CRD that is namespace limited in this way.

IMO, how I (as an admin) choose to lock down where that resource can be created should be independent of where our operator looks for the CRs and how we respond to them when they are created..

I would feel better about moving the exemptions back to the individual packages if it were part of the [uds package object](https://github.com/defenseunicorns/uds-core/blob/main/src/neuvector/chart/templates/uds-package.yaml).. but if it's just a CRD/CR, I think (by default at least) we should respect it no matter where it is created in the cluster. Perhaps we could offer `valid-exemption-namespaces` as a uds-core configuration or something to scratch the admin simplification itch.

src/pepr/operator/controllers/exemptions/exemptions.ts
```diff
@@ -0,0 +1,110 @@
...
+      return true;
+    }
+  }
+}
+
+const policyList = Object.values(Policy);
+
+// *** Use Local Map to then Update Store ***
+// Add Exemptions to Pepr store as "policy": "[{matcher}]"
+export function processExemptions(exmpt: UDSExemption) {
```
mjnagel: Won't comment on all of them, but lots of places where `exempt` is missing the second `e` like this. Should be an easy find/replace.

TristanHoladay: i purposely left it out i guess as a shorter abbreviation, but yeah that seems kind of stupid now. I'll change it.

mjnagel: `xempt` is the way to go 😛, but yeah I think full name is better so that anyone new to the code isn't guessing at the right vars to pass around/use. Especially when we have exemption and others fully written out.

mjnagel:
> If the CRD is installed in my cluster, I would expect for it to have the same effect, regardless of the namespace I put it in. I'm having trouble thinking of another CRD that is namespace limited in this way.

@blancharda its actually similar to kyverno, although [kyverno is opt-in](https://kyverno.io/docs/installation/customization/#container-flags) (`exceptionNamespace`) to this behavior of locking them down to a specific namespace. I've seen the same behavior with a number of operators (again mostly opt-in), but its not uncommon to have it as an option. By watching a single namespace you reduce load for what is watched, etc. I do see your point on confusion around how this functions, I think at minimum we should document that Exemptions only function in that namespace. We could also throw a validator on to deny all others, which might work, but if we wanted to keep the smaller watch area it would have to be done separate from the in-namespace watch.

> I don't think the charts should be installing things to multiple namespaces.. it's not unheard of (and is certainly possible) but makes assumptions about pre-existing namespaces that helm charts don't typically make when it can be avoided

I don't feel as strongly about this because (at least in UDS Packages) we would be adding this specifically to the uds-config chart and its a UDS CR. I think we are more than safe making assumptions about what exists when the operator exists and inside of our opinionated uds-config charts. I don't see it much different than assuming a CRD exists that your chart doesn't install (i.e. UDS Package CRD).

blancharda:
> but its not uncommon to have it as an option

@mjnagel That's a fair point, and I would definitely be more comfortable if it were an opt-in process. IMO that just gives system administrators more flexibility in how they want to secure and optimize their cluster, which feels like exactly what we should be promoting. 

> I don't feel as strongly about this because (at least in UDS Packages)

I mostly agree here as well... my general concern with charts that install to pre-existing namespaces is that it gets very messy if you want/need to deploy a second release of the same chart for debugging / testing / etc. The likely hood of that with these charts seems low, but it would still be annoying if/when we try to do it.

mjnagel:
> IMO that just gives system administrators more flexibility in how they want to secure and optimize their cluster, which feels like exactly what we should be promoting.

I could maybe get behind defaulting to a single namespace for a best practice, but allowing config for someone to decrease security if they want to hurt themselves - would be similar to Istio mTLS at that point. Would have to think through this to figure out whether this is something we could support easily without duplication of watches. I see this very much as a security thing - if locked down to a single namespace most app dev in a "software factory", etc, would likely already have reduced RBAC to only deploy to a single namespace.

blancharda:
> most app dev in a "software factory", etc, would likely already have reduced RBAC to only deploy to a single namespace

Shouldn't that RBAC also be type limited anyways?

The design of our CRD and operator design are independent from how any particular environment will be administered. At the end of the day, we won't always have control over it. That's not to say we shouldn't promote best practices, and educate through clear and simple defaults and documentation -- because we absolutely should! 

I just want to be careful that we aren't limiting the design of our system based on assumptions about, or even well intentioned promotion of, best practices for cluster administration.

blancharda:
To be clear, I think as long as the optionality is there, it will be flexible enough for any env. My preference would be to watch all namespaces by default for the sake of clarity, with the option (and perhaps documented recommendation of locking it down to a single one) -- but I'm not going to die on that hill..

src/pepr/policies/exemptions/index.ts
```diff
@@ -1,52 +1,39 @@
...
-   */
-  name?: string | RegExp;
-};
-
-export type ExemptList = Array<Exempt>;
+import { Policy } from "../../operator/crd";
+import { Store } from "../common";
 
 /**
  * Register a list of exemptions to be used by the validation action.
```
mjnagel:
```suggestion
 * Check a resource against an exemption list for use by the validation action.
```
Something like this to reword I think?

src/pepr/policies/exemptions/index.ts
```diff
@@ -1,52 +1,39 @@
...
+  const exemptList = JSON.parse(Store.getItem(policy) || "[]");
 
-      // If the exempt name is specified, check it
-      const name = request.Raw.metadata?.name || request.Raw.metadata?.generateName;
-      if (exempt.name && !name?.match(exempt.name)) {
-        continue;
-      }
+  // Loop through the exempt list
+  for (const exempt of exemptList) {
+    // If the exempt name is specified, check it
```
mjnagel:
```suggestion
    // If the exempt namespace is specified, check it
```

src/pepr/policies/networking.ts
```diff
@@ -16,7 +17,7 @@ import { exemptHostNamespaces } from "./exemptions/networking";
 When(a.Pod)
   .IsCreatedOrUpdated()
   .Validate(request => {
-    if (exemptHostNamespaces(request)) {
+    if (isExempt(Policy.DisallowPrivileged, request)) {
```
mjnagel: This looks like the wrong policy? Should be`DisallowHostNamespaces`:
```suggestion
    if (isExempt(Policy.DisallowHostNamespaces, request)) {
```

src/prometheus-stack/chart/templates/uds-exemption.yaml
```diff
@@ -0,0 +1,21 @@
...
+        namespace: monitoring
+        name: "^prometheus-kube-prometheus-stack-prometheus-.*"
+      
+    - policies:
+        - Restrict_Volume_Types
+      matcher:
+        namespace: monitoring
+        name: "^kube-prometheus-stack-prometheus-node-exporter-.*"
+
+
```
mjnagel: Noticed that yamllint is not being run on these files. It looks like we can make yamllint sort of work with helm by doing these changes:
- add `'!**/chart/templates/uds-*.yaml'` to the ignore file list, after `'**/chart/templates**'`
- change `braces` to disable

With those changes `uds run lint-check` shows a couple issues with the new exemption files as well as one issue with a package file.

mjnagel (commented):
Still need to look through the meat of the logic here but wanted to get these initial small comments published for you to take a look at them.

jeff-mccoy:
Read back through these comments--think it's worth a sync meeting to make a decision, but my thinking here is we want to insert friction/discourage policy exemptions--unless it's a security tool, you're probably doing something wrong if you are introducing exemptions. I agree we could leave in each NS and let RBAC do it's job, but it does feel like it would be easier to mess that up as NS is often a way we will grant / zone responsibility in the future as we think beyond just SWF things. 

Not sure I read helm NS relationships the same as others--I'd say if anything helm ignores the NS, which is why we can pre-create them in Zarf without breaking helm things. I personally wouldn't want a user with a given NS RBAC to control their app to have the ability to exempt the cluster security rules, and it sounds like we all agree RBAC is the way to do that. I think my point is, if you're already requiring different RBAC to perform that function you can't do that in the same chart anyway, by definition.

src/pepr/policies/networking.ts
TristanHoladay: ohp. thanks for catching that.

src/pepr/operator/controllers/exemptions/exemptions.ts
```diff
@@ -0,0 +1,87 @@
...
+      } else {
+        // check if policy has this matcher from previous version of CR
+        for (const m of matchers) {
+          if (m.name === name) {
+            Log.debug(`Removing ${name} from ${p}`);
+            exemptionMap.set(
+              p,
+              matchers.filter(m => m.name !== name),
+            );
+          }
```
mjnagel: I think this check for removal misses a case where you delete an entire matcher (rather than just a policy under a matcher). 

Probably a broader question/discussion on how we architect removals...need to have some way to get context of previous state/tie a matcher back to the `exemption` CR it came from.

AustinAbro321:
How do you guys feel about adding `kind` on the exemption matcher? I believe there is potential for unintended exemptions without `kind`

Let's say we have an exemption with the following spec:
```yaml
spec:
  exemptions:
    # This is meant for services
    - policies:
        - Restrict_ExternalName
      matcher:
        namespace: neuvector
        name: "^neuvector-enforcer.*"

    # This is meant for pods
    - policies:
        - Restrict_ExternalName
      matcher:
        namespace: neuvector
        name: "^neuvector-controller-pods.*"
```

The intent behind this exemption may be to allow pods called `neuvector-controller-pods` and service called `neuvector-enforcer`, but assuming the following uds code (which doesn't exist but potentially could) you could unintentionally allow a service called `neuvector-controller-pods`. The potential for abuse is more likely when short name patterns with wildcards are used like `^promtail-.*`

```javascript
// Let's say we make a service with policy.RestrictExternalNames 
// We make a service "nuevector-controller-pods"
When(a.Service)
.IsCreatedOrUpdated()
.Validate(request => {
  if (isExempt(Policy.RestrictExternalNames, request)) {
    return request.Approve();
  }
  if (request.Raw.spec?.type === "ExternalName") {
    return request.Deny("ExternalName services are not allowed.");
  }

  return request.Approve();
});

// Let's say we make a service with policy.RestrictExternalNames 
// We call the service "nuevector-controller"
When(a.Pod)
.IsCreatedOrUpdated()
.Validate(request => {
  if (isExempt(Policy.RestrictExternalNames, request)) {
    return request.Approve();
  }
  if (request.Raw.spec?.type === "ExternalName") {
    return request.Deny("ExternalName services are not allowed.");
  }

  return request.Approve();
}); 
```


Additionally, trying to build a [clusterPolicyReport](https://htmlpreview.github.io/?https://github.com/kubernetes-sigs/wg-policy-prototypes/blob/master/policy-report/docs/index.html) without `kind` on `matcher` is impossible as we don't know which resources are actually validated against the exemption. The matcher could be referring to a service or a pod, and they can have the same name in the same namespace. The only way to determine what is being exempted is to look through the UDS codebase which can't be automated. Adding `kind` would also make the intent of the exemptions more clear.

mjnagel:
@AustinAbro321 interesting idea. I think we would likely have any policies applying to svc vs pod as separate "policy names" and separate "exemption names" so I'm not sure we would ever encounter the exact scenario you laid out above but I see where that could be problematic. I had to refresh my memory but Kyverno actually has a similar approach where you [specify the resource kind](https://kyverno.io/docs/writing-policies/exceptions/) in addition to the namespace and name.

I think there might be a broader issue that you're alluding to though. Even if you had the exemption name with the resource kind - you would still be referencing the policy itself in code to know what the policy does and what it _should_ apply to. This is typically where other policy engines have another CRD for policies themselves, which I don't really like but there might be value in having an easy to lookup list of "applied policies" in cluster. Will have to think on that a bit more because we don't want to invent a policy language - intent would just be discoverability of what is enforced, on what types of resources.

AustinAbro321:
Yea @mjnagel that's the issue myself, @btlghrants, and @Arsinic are trying to solve over dash days. Originally we were going to load the exemptions into our pepr capability to make and maintain a [clusterPolicyReport](https://htmlpreview.github.io/?https://github.com/kubernetes-sigs/wg-policy-prototypes/blob/master/policy-report/docs/index.html) tracking every resource that is exempted in the cluster. We would basically look at when a generic resource meets the conditions set by the matcher, and if it did we would add it to the report as a resource that was skipped over because of an exemption. We figured this could then be consumed by Lula during the report, or at least compared to the final Lula report so when Lula reports that certain NIST or whatever system control is being violated we have a reason why it's being violated. 

But talking about it more, we thought it'd ultimately be better if when UDS exempts something there are labels or annotations stating what was exempted. That we way don't have the trust the UDS exemptions resource are being implemented to spec. We also wouldn't have to worry about lingering resources from a UDS exemption that was deleted. We can just compile a list of resources that have the exemption labels. 

Not saying this has to go into this PR, just some things we've been thinking about over dash days

mjnagel:
@AustinAbro321 I might've been missing some of the context on what you are looking at. Is the goal a report with just the exemptions that have happened? I was thinking of a policy report in the broader sense where it would include:
- things denied
- things that adhered to the policy
- things that were exempted

It sounds like maybe your focus is on that third one only? I was thinking beyond that about other issues you'd have knowing about the policies for general approve/denies (outside of exemptions).

> But talking about it more, we thought it'd ultimately be better if when UDS exempts something there are labels or annotations stating what was exempted. That we way don't have the trust the UDS exemptions resource are being implemented to spec. We also wouldn't have to worry about lingering resources from a UDS exemption that was deleted. We can just compile a list of resources that have the exemption labels.

There is currently an annotation `uds-core.pepr.dev/uds-core-policies: succeeded` that validates something went through the validating webhook, but it doesn't say whether it was exempted or passed. I think that's worth trying to add something to state that it was exempted.

AustinAbro321:
@mjnagel Looking at exemptions and the reporting around them was the main focus of our dash days this month.  I think for sure we'd want to report on all three though. We will really have to see how we want to format the report. Just thinking out loud there's a few possible ways: 
- A report per policy, we could have everything the policy approved, exempted, or denied. 
- A report for each resource showing which policies passed and failed on it. It looks like kyverno does this https://kyverno.io/docs/policy-reports/#viewing-policy-report-summaries or plans to. 
- A namespace or cluster wide report showing everything that passed, was exempted, or denied. 

Lula would be the consumer of this report (I think) so we'll probably end up talking to them on what they want to see. I imagine people working in the cluster will also at least care about what is getting denied, so we'll probably want to make a report on that as well. Not sure how obvious pepr denies are when working in tools like zarf, uds-cli or helm, or if people just have to look through the pepr controller logs.  

We are doing a series, so this may be our focus in March.

docs/CONFIGURE_POLICY_EXEMPTIONS.md
```diff
@@ -0,0 +1,29 @@
+## Configuring UDS-CORE Policy Exemptions
+
+By default policy exemptions ([UDSExemptions](../src/pepr/operator/crd/generated/exemption-v1alpha1.ts)) are only allowed in a single namespace -- `uds-policy-exemptions`. While this is an anti-pattern for how CRDs typically work in Kubernetes, we believe this is a best practice for these reasons:
+
```
jeff-mccoy, line 4:
```suggestion
By default policy exemptions ([UDSExemptions](../src/pepr/operator/crd/generated/exemption-v1alpha1.ts)) are only allowed in a single namespace -- `uds-policy-exemptions`. We recognize this is not a conventional pattern in K8s, but believe it is ideal for UDS for the following reasons:
```

docs/CONFIGURE_POLICY_EXEMPTIONS.md
```diff
@@ -0,0 +1,29 @@
...
+- makes maintaining rbac for controlling exemptions easier
+- reduces the risk that an unintentional mis-configuration of rbac allows a cluster exemption that would otherwise be denied
+- increases organization for cluster management, monitoring, and reporting
```
jeff-mccoy:
```suggestion
- highlights the fact that an exemption can reduce the overall security posture of the cluster 
- makes maintaining RBAC for controlling exemptions more straightforward
- reduces the risk that an unintentional mis-configuration of RBAC allows a cluster exemption that would otherwise be denied
```

jeff-mccoy: multiline suggestions are weird....

src/pepr/operator/crd/sources/exemption/v1alpha1.ts
```diff
@@ -0,0 +1,99 @@
...
+            exemptions: {
+              type: "array",
+              description: "Policy exemptions",
+              items: {
+                type: "object",
+                required: ["policies", "matcher"],
+                properties: {
+                  description: {
+                    type: "string",
+                    description:
```
jeff-mccoy: I think we should probably do something a little different here than Pkg. Probably should make this `title` and include a freetext `description` prop as well so we can include the exemption data in the CR itself and it won't be lost when the YAML/JSON is processed. This will be important with UDS Engine to show why something was exempted. This would also mean all the various CR YAML comments we currently have in this PR should move to this description field. I'd also keep at least `title`, but maybe also `description` in the CR status table.

src/pepr/operator/crd/sources/exemption/v1alpha1.ts
```diff
@@ -0,0 +1,99 @@
...
+                      "A description of this exemption, this will become part of the exemption name",
+                  },
+                  policies: {
+                    description: "A list of policies to override",
+                    type: "array",
+                    items: {
+                      type: "string",
+                      enum: [
+                        "Disallow_Host_Namespaces",
```
jeff-mccoy: this will have a small maintenance tail I think, but like seeing this as an enum as the type system will give us TS validation and the API server will give us free validation on the runtime side.

src/pepr/policies/security.ts
```diff
@@ -292,7 +348,11 @@ When(a.Pod)
 When(a.Pod)
   .IsCreatedOrUpdated()
   .Mutate(request => {
-    if (exemptDropAllCapabilities(request)) {
+    if (isExempt(Policy.DropAllCapabilities, request)) {
```
jeff-mccoy: I like the idea of this annotation to track exemptions--but think we could probably generalize this into a common function such that we could do `.Mutate(markExemption)` or something like that and let generics earn their pay.

src/pepr/policies/storage.ts
```diff
@@ -1,7 +1,8 @@
 import { a } from "pepr";
 
 import { When, containers, volumes } from "./common";
```
jeff-mccoy, line 3: nit, but for some reason our rules don't pick up import organization rn so if I organize imports in vscode, this import order changes.

src/pepr/operator/crd/register.ts
```diff
@@ -1,6 +1,6 @@
 import { K8s, Log, kind } from "pepr";
-
 import { v1alpha1 } from "./sources/v1alpha1";
```
jeff-mccoy: probably worth changing this to `as pkg`

src/pepr/operator/crd/index.ts
```diff
@@ -1,6 +1,7 @@
 import { V1OwnerReference } from "@kubernetes/client-node";
+import { GenericKind } from "kubernetes-fluent-client";
 
-import { Package as UDSPackage } from "./generated/package-v1alpha1";
+export interface UDSCR extends GenericKind {}
```
jeff-mccoy: this seems odd as these two types sould already extend `GenericKind`

src/pepr/operator/crd/sources/exemption/v1alpha1.ts
```diff
@@ -0,0 +1,99 @@
...
+                        "Restrict_Host_Ports",
+                        "Restrict_Proc_Mount",
+                        "Restrict_Seccomp",
+                        "Restrict_SELinux_Type",
+                        "Restrict_Volume_Types",
+                      ],
+                    },
+                  },
+                  matcher: {
+                    description: "Name and namespace of pod to exempt. Regex allowed for name.",
```
jeff-mccoy: This is probably fine if we believe Policies are always for pods. Is there every a time when we would have a Policy for something else? If so, then I think we'd need to make `kind` a prop and maybe default to `pod`

jeff-mccoy (dismissed):
left some notes, overall great work and direction in excellent, just some smallish changes I think

src/pepr/operator/crd/sources/exemption/v1alpha1.ts
TristanHoladay: Austin Abro and his dash days team brought this up with us while trying to implement a policy reporter. There are 2 instances in the policies/networking.ts that validate services. So you're feedback further confirms we should go this route.

src/pepr/operator/crd/sources/exemption/v1alpha1.ts
TristanHoladay: Oh yeah, i'd totally forgotten about the description for the name. i like that, making the description as the reasons for the exemption. Was chatting with @mjnagel and we were wondering about the `title`. Since we don't necessarily need that for processing and storing the exemptions, would that be more for runtime engine and reporting?

src/pepr/operator/crd/sources/exemption/v1alpha1.ts
```diff
@@ -0,0 +1,99 @@
...
+                    type: "string",
+                    description:
+                      "A description of this exemption, this will become part of the exemption name",
+                  },
+                  policies: {
+                    description: "A list of policies to override",
+                    type: "array",
+                    items: {
+                      type: "string",
+                      enum: [
```
rjferguson21, line 75: Would be curious what others think, but I'd advocate we use `UPPER_SNAKE_CASE` for these values. 
```
                      enum: [
                        "DISALLOW_HOST_NAMESPACES",
                        "DISALLOW_NODEPORT_SERVICES",
                        "DISALLOW_PRIVILEGED",
                        "DISALLOW_SELINUX_OPTIONS",
                        "DROP_ALL_CAPABILITIES",
                        "REQUIRE_NON_ROOT_USER",
                        "RESTRICT_CAPABILITIES",
                        "RESTRICT_EXTERNAL_NAMES",
                        "RESTRICT_HOSTPATH_WRITE",
                        "RESTRICT_HOST_PORTS",
                        "RESTRICT_PROC_MOUNT",
                        "RESTRICT_SECCOMP",
                        "RESTRICT_SELINUX_TYPE",
                        "RESTRICT_VOLUME_TYPES",
                      ],
```

src/pepr/policies/security.ts
TristanHoladay: i like that idea. had some issues getting it to work nicely with the two cases that do actual mutating if it's not exempt, but with @rjferguson21 's help i think it's still cleaner than before.

src/pepr/operator/crd/sources/exemption/v1alpha1.ts
mjnagel: I think constant case / `UPPER_SNAKE_CASE` does somewhat make sense based on general conventions I see upstream. Would point out that our codebase already has several enums that are `PascalCase` (`remoteGenerated` is a good example), so would be good to keep them standardized regardless of what we choose here - `PascalCase` seems to be somewhat common for enums too 🤷

src/pepr/policies/storage.ts
TristanHoladay: added this to the `.vscode/settings.json`

```json
"[typescript]": {
    "editor.codeActionsOnSave": {
        "source.organizeImports": "always"
    }
  }
  ```

File: src/pepr/operator/crd/sources/exemption/v1alpha1.ts (hunk 17 above)
Comment (TristanHoladay) [reply src/pepr/operator/crd/sources/exemption/v1alpha1.ts:75]: In general i agree with you @rjferguson21. In the past i've always made my enums `UPPER_SNAKE_CASE`, and probably prefer that. Went with the `PascalCase` cause like @mjnagel said it's somewhat already in our code base. For the sake of getting this PR merged in, i say we keep it as is and change it later after some further thought etc...

File: package.json [hunk 18]
```diff
@@ -25,6 +25,8 @@
     },
     "env": {
       "UDS_DOMAIN": "###ZARF_VAR_DOMAIN###",
+      "UDS_WITH_ISTIO": "###ZARF_CONST_WITH_ISTIO###",
```
mjnagel:
```suggestion
```
Think this can be dropped - in the keycloak PR we switched to `UDS_SINGLE_TEST` so with istio shouldn't be used anywhere now?

src/pepr/operator/crd/generated/exemption-v1alpha1.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+export interface Matcher {
+  kind?: Kind;
+  name: string;
+  namespace: string;
+}
+
+export enum Kind {
+  Pod = "pod",
+  Services = "services",
```
mjnagel: Might be good for consistency to use singular for service here?

src/pepr/operator/crd/generated/exemption-v1alpha1.ts
```diff
@@ -0,0 +1,83 @@
...
+  RequireNonRootUser = "Require_Non_Root_User",
+  RestrictCapabilities = "Restrict_Capabilities",
+  RestrictExternalNames = "Restrict_External_Names",
+  RestrictHostPathWrite = "Restrict_HostPath_Write",
+  RestrictHostPorts = "Restrict_Host_Ports",
+  RestrictProcMount = "Restrict_Proc_Mount",
+  RestrictSELinuxType = "Restrict_SELinux_Type",
+  RestrictSeccomp = "Restrict_Seccomp",
+  RestrictVolumeTypes = "Restrict_Volume_Types",
+}
```
mjnagel: Definitely not the "convention" expert here, but this feels a bit strange to use `PascalCase` for the enum names but `_` for the value? Don't want to keep asking about naming conventions 🙃, but the remote generated stuff for example uses pascal for both. Tracking that this would need updates across a number of files if we change it though...

src/pepr/policies/security.ts
```diff
@@ -56,9 +54,9 @@ When(a.Pod)
 When(a.Pod)
   .IsCreatedOrUpdated()
   .Mutate(request => {
-    if (exemptPrivileged(request)) {
+    markExemption(Policy.RequireNonRootUser)(request);
+    if (request.HasAnnotation(`uds-core.pepr.dev/uds-core-policies.${Policy.RequireNonRootUser}`))
       return;
```
mjnagel: This bothers me for whatever reason - but `npx pepr format` doesn't change it. Any reason for dropping the explicit `{}` around return here (and the other spot or two where this same logic is used)?

mjnagel (commented):
Think this is in a great spot overall - couple small comments here.

src/pepr/operator/index.ts
```diff
@@ -40,3 +42,47 @@ When(UDSPackage)
...
+    Log.debug(exempt, `Processing Exemption ${namespace}/${name}`);
+
+    try {
+      await updateStatus(exempt, { phase: Phase.Pending });
+
+      processExemptions(exempt);
+      await updateStatus(exempt, {
+        phase: Phase.Ready,
+        observedGeneration: exempt.metadata.generation,
+        titles: exempt.spec?.exemptions?.map(e => e.title || ""),
```
mjnagel: Thinking about the titles here...if someone leaves them off it could give a weird impression on a status view that there are no exemptions included? Not sure if that's a reason to require titles, or try to auto-gen something based on the matcher? Information is obviously there regardless, just less clear if I'm looking at k9s for example.

src/pepr/policies/security.ts
TristanHoladay: ahh yeah, when it sits on one line i think it looks nice, but since it got pushed down looks weird and easy to misread. I'll add the `{}`

src/pepr/operator/index.ts
TristanHoladay: that's a good point. i guess we could use the matcher name if there's no title. might look a little weird if there's regex in the name but probably still better than nothing.

.yamllint
```diff
@@ -7,11 +7,12 @@ ignore:
   - 'k3d/local/manifests/metallb/metallb-native.yaml'
   - '**/.terraform/**'
   - '**/chart/templates**'
+  - '!**/chart/templates/uds-*.yaml'
```
mjnagel: Seems like this is more pain than its worth, so I think we can just drop it. Might want to evaluate helm lint smoke tests in CI to handle the charts (could also do `helm template` -> `yamllint`). Regardless - out of the scope of your PR so can probably just revert the changes here for now.

mjnagel: Opened https://github.com/defenseunicorns/uds-core/issues/231 as a follow on to this.

mjnagel (commented):
One other spot to add explicit `{}` back - https://github.com/defenseunicorns/uds-core/blob/feat-exemptions/src/pepr/policies/security.ts#L309-L310

src/pepr/operator/reconcilers/reconciler.ts
```diff
@@ -1,12 +1,13 @@
-import { K8s, Log } from "pepr";
```
mjnagel, line 1: Small nit: I like the foldering of reconcilers, but might be good to rename this one to `package-reconciler.ts` (and same with the spec file).

mjnagel (commented):
99% there, couple small things and then this should be g2g

mjnagel (approved):

PR #172, opened 2024-02-12
ci: debug events output

mjnagel:
## Description

In some CI runs we are seeing pending pods but without events output its hard to tell if there is a resourcing issue or something else going on. Conditionally running this on failure since its pretty verbose.

## Related Issue

N/A, issues seen in CI that this seeks to help debug.

## Type of change

## Checklist before merging

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -85,7 +85,7 @@ jobs:
         if: ${{ failure() }}
         run: |
           echo "::group::kubectl get events"
-          kubectl get events -A
+          kubectl get events -A --sort-by='.lastTimestamp'
           echo "::endgroup::"
 
       - name: Save logs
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -81,6 +81,13 @@ jobs:
           kubectl get package -A
           echo "::endgroup::"
 
+      - name: Debug Events Output
+        if: ${{ failure() }}
+        run: |
+          echo "::group::kubectl get events"
+          kubectl get events -A
```
MxNxPx:
```suggestion
          kubectl get events -A --sort-by='.lastTimestamp'
```

MxNxPx: might be helpful to have it sort events by timestamp, but that is more of a personal preference.

MxNxPx (dismissed):

.github/workflows/test.yaml
mjnagel: Good call, I always forget the sort. More useful debug = more betterer.

MxNxPx (approved):
:ship: :it:

Issue #149, opened 2024-02-06
Generated netpol for webhooks / apiservices


mjnagel:
### Describe what should be investigated or refactored

https://github.com/defenseunicorns/uds-core/commit/f1d434a68ef1f2a29ab3b13608bc16ce78211ed4 added an open ingress to metrics-server to allow the proper ingress from `kubectl top`, but this is overly permissive. 

While testing for that PR we discovered that webhooks appear to behave the same way/are missing ingress similarly. We will want to evaluate an "Ingress from KubeAPI" generated rule that would cover this scenario, allowing ingress from the source-NAT-ed IPs for each node.

### Links to any relevant code

https://github.com/defenseunicorns/uds-core/blob/main/src/metrics-server/chart/templates/uds-package.yaml

### Additional context

I have not found any other repos where similar network policies are employed so this feels like a bit of uncharted territory. The closest I've seen is a cilium netpol that uses the `cluster` entity - https://docs.cilium.io/en/latest/security/policy/language/#entities-based

mjnagel:
In testing using the cert-manager package/webhook I "got away with" an Ingress from KubeAPI policy on single-node k3d clusters, so reproducing/debugging the problems may require a multi-node cluster.

I also seemed to experience consistent failures with ingress from kubeapi/kubenodes IF the webhook pod was scheduled on an agent node. With multi-server clusters I saw the issues intermittently.

robgieringer:
Refs CORE-2

PR #192, opened 2024-02-20
fix: validating/mutating webhook networkpolicies and mtls

mjnagel:
## Description

Adds the proper ingress rules to the webhooks and PERMISSIVE mTLS to allow calls to operate as expected. Validated by turning the webhook failure policies to Fail and applying CRs. The fix applied is similar to the approach taken with metrics-server, and can be re-evaluated as a generated rule instead of Anywhere in the future.

## Related Issue

Related to https://github.com/defenseunicorns/uds-core/issues/149 (not a solve for it, but would also be modified by that issue)

## Type of change

## Checklist before merging


MxNxPx (approved):
:ship: :it:

PR #194, opened 2024-02-21
chore: identity flavors

zachariahmiller:
## Description

Broke out flavored values files
Created common keycloak component imported into flavored components
Added templating for security context for keycloak image to facilitate using registry1 image
Added templating to all helm resources that require istio to enable single capability tests to work
Commented out keycloak validate task that checks the exposed endpoints as single capability tests dont have istio and cannot test that functionality

Note: if using colima on a mac the local image in keycloak doesnt "just work" as crane requires the default docker sock unix:///var/run/docker.sock  or DOCKER_HOST to be set. This behavior is unique to this scenario and colima does not symlink anymore so docker desktop can be colocated on the same machine. This can be resolved by running `export DOCKER_HOST=unix:///Users/<username>/.colima/<colima-profile>/docker.sock` prior to attempting to build the package containing keycloak.

Tested on registry1 flavor on ubuntu amd64 and upstream with darwin arm64.

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


zachariahmiller:
FYSA: This is failing ci because of a zarf bug https://github.com/defenseunicorns/zarf/issues/2320, which is impacting main due to a change (change metrics server from `required: true` to `default: true` will likely be reverted until the issue is fixed in zarf

zachariahmiller:
Okay this should be good again. @jeff-mccoy not sure what you want to do with the keycloak specific interface/endpoint checks. Definitely nice to have on the full runs, but without istio they break the single package tests. I left them commented out for now just so ci would pass rather than deleting in case there were specific thoughts on keeping them in some capacity.

zachariahmiller:
Of course the full install failed immediately after i commented on the pepr jest tests, but dont think that is related to anything i implemented here.

PR #198, opened 2024-02-22
chore: test artifacts before publish

### LINKED ISSUE #150: Trigger test workflows prior to release

Currently the release pipeline runs a [create + publish](https://github.com/defenseunicorns/uds-core/blob/main/tasks/publish.yaml) on the zarf packages / uds bundles, but does not test the artifacts prior to publish. 

Due to Ironbank's mutable tags, and consuming from a variety of upstream sources that could take similar approaches, the artifact published could differ from what was last tested in CI.

When a release is triggered a package should run through these steps:
- create (already happening)
- deploy/test (using the artifact from create)
- publish (using the artifact from create)

This will ensure that the exact artifact published was run through the CI tests.

Pin every image to a sha. This might be worth exploring but not sure if all charts easily support this, how much churn we might incur with renovate, etc. It also seems like even in that scenario it would be prudent to run tests before publish to perform a validation.

### Additional context

This shouldn't typically be required but I believe its critical due to the way Ironbank images are known to be mutable. An issue occurred previously on DUBBD where a PR tested fine, but by the time the release happened it pulled a new image for one of the packages and did not work as expected.

TristanHoladay:
## Description
Test artifacts before publishing them to make sure nothing has changed with upstream or ironbank images.

## Related Issue
Resolves #150 

## Type of change

## Checklist before merging


TristanHoladay:
currently some extra conditionals to not deploy the arm packages. should we start an issue for adding arm testing to our workflows?

.github/workflows/tag-and-release.yml
```diff
@@ -46,12 +48,38 @@ jobs:
       - name: Login to registry1
         run: uds run registry-login --set REGISTRY=registry1.dso.mil --set REGISTRY_USERNAME=${{ secrets.IRON_BANK_ROBOT_USERNAME }} --set REGISTRY_PASSWORD=${{ secrets.IRON_BANK_ROBOT_PASSWORD }} --set REGISTRY_RETRY_INTERVAL=90
 
+      - name: Create Standard Package and Bundle
+        if: ${{ matrix.type == 'standard-package' }}
+        run: |
+          set -e
```
mjnagel: I don't think this is necessary, github should do this by default.
```suggestion
```

tasks/publish.yaml
```diff
@@ -5,63 +5,48 @@ variables:
...
+
   - name: VERSION
     description: "The version of the packages to build"
     # x-release-please-start-version
     default: "0.13.1"
     # x-release-please-end
 
 tasks:
-  - name: packages
+  - name: standard package
```
mjnagel: I think these names all need to be hyphenated? Expect this would throw a syntax issue/something about the space when called, and the workflows are calling with a hyphen currently.
```suggestion
  - name: standard-package
```

.github/workflows/tag-and-release.yml
```diff
@@ -46,12 +48,38 @@ jobs:
...
+          ZARF_ARCHITECTURE=${{ matrix.arch }} uds run -f tasks/create.yaml istio-package --no-progress --set FLAVOR=${FLAVOR}
+          ZARF_ARCHITECTURE=${{ matrix.arch }} uds run -f tasks/create.yaml k3d-istio-bundle --no-progress
+
+      - name: Deploy Standard Package via Bundle
+        if: ${{ matrix.type == 'standard-package' && matrix.arch == 'amd64' }}
+        run: uds run -f tasks/deploy.yaml k3d-standard-bundle --no-pogress --set uds_arch=${{ matrix.arch }}
+
+      - name: Deploy Istio Packages via Bundle
+        if: ${{ matrix.type == 'istio-package' && matrix.arch == 'amd64' }}
+        run: uds run -f tasks/deploy.yaml k3d-istio-bundle --no-pogress --set uds_arch=${{ matrix.arch }}
```
mjnagel: Commenting here, but this applies to all the items here. I think if we changed the matrix `type`s to `standard` and `istio` we could DRY this up:
```
      - name: Deploy ${{ matrix.type }} Package via Bundle
        if: ${{ matrix.arch == 'amd64' }}
        run: uds run -f tasks/deploy.yaml k3d-${{ matrix.type }}-bundle --no-pogress --set uds_arch=${{ matrix.arch }}
```

Something like this _should_ handle both standard/istio and work for create, deploy, and publish I think?

mjnagel (changes requested):
I like the matrix approach in general but one of my concerns with this currently is parallelizing the architecture publishes. Maybe not critical but currently this would publish the `arm64` packages regardless of whether or not `amd64` testing fails I think? I don't think we necessarily need to test the arm build in CI, but it would be good to only publish when the test succeeds. We could just switch back to publishing both in the same task and drop arch from the matrix, but might be other approaches that are more optimal.

.github/workflows/tag-and-release.yml
TristanHoladay: mmm, nice point of refactor. i like it.

.github/workflows/tag-and-release.yml
```diff
@@ -46,12 +47,25 @@ jobs:
...
+      - name: Create Packages and Bundles
+        run: |
+          ZARF_ARCHITECTURE=amd64 uds run -f tasks/create.yaml ${{ matrix.type}}-package --no-progress --set FLAVOR=${{ matrix.flavor }}
+          ZARF_ARCHITECTURE=amd64 uds run -f tasks/create.yaml k3d-${{ matrix.type }}-bundle --no-progress
+
+          ZARF_ARCHITECTURE=arm64 uds run -f tasks/create.yaml ${{ matrix.type}}-package --no-progress --set FLAVOR=${{ matrix.flavor }}
+          ZARF_ARCHITECTURE=arm64 uds run -f tasks/create.yaml k3d-${{ matrix.type }}-bundle --no-progress
+
+      - name: Deploy amd64 Bundles
+        run: uds run -f tasks/deploy.yaml k3d-${{ matrix.type }}-bundle --no-pogress --set uds_arch=amd64
```
mjnagel:
```suggestion
      - name: Test amd64 Bundle
        run: |
          uds run -f tasks/deploy.yaml k3d-${{ matrix.type }}-bundle --no-progress --set uds_arch=amd64
          uds run -f tasks/test.yaml validate-packages --no-progress
```
pogress -> progress the main issue here, but think that `Test` might fit better than `Deploy`? I think it would also be good if we ran the `validate-packages` task here after the deploy.

tasks/publish.yaml
```diff
@@ -5,60 +5,36 @@ variables:
...
   - name: VERSION
     description: "The version of the packages to build"
     # x-release-please-start-version
     default: "0.15.0"
     # x-release-please-end
 
 tasks:
-  - name: packages
+  - name: standard-package
     description: "Build and publish the UDS packages"
```
mjnagel:
```suggestion
    description: "Publish the UDS package"
```

tasks/publish.yaml
```diff
@@ -5,60 +5,36 @@ variables:
...
       - description: "Publish the packages"
         cmd: |
-          uds zarf package publish build/zarf-package-core-amd64-${VERSION}.tar.zst ${TARGET_REPO}
-
-          # dont publish arm64 for registry1 since IB images are only amd64
+         ZARF_ARCHITECTURE=amd64 uds run -f tasks/create.yaml standard-package --no-progress --set FLAVOR=${FLAVOR}
           if [ "${FLAVOR}" != "registry1" ]; then
-            uds zarf package publish build/zarf-package-core-arm64-${VERSION}.tar.zst ${TARGET_REPO}
+            ZARF_ARCHITECTURE=arm64 uds run -f tasks/create.yaml standard-package --no-progress --set FLAVOR=${FLAVOR}
           fi
```
mjnagel: I think we dropped the wrong section here - looking at [the task](https://github.com/defenseunicorns/uds-core/blob/update-publish-ci/tasks/publish.yaml#L22-L25) (because the diffs are overwhelming) this isn't doing a publish anymore, just a create. Should drop the create and only do the publish I believe?

mjnagel (changes requested):

tasks/publish.yaml
TristanHoladay: ohhh, yep that's right. i only tested the publish bundle locally so i didn't catch that.

tasks/publish.yaml
TristanHoladay: removing the action description since task description would be the same.

mjnagel (approved):
Good with this to start - can think through optimizations around create, etc as a follow on but want to get this testing into our process now.

Issue #103, opened 2024-01-16
Add netpols to istio components


mjnagel:
# Current Scope

Add NetworkPolicy resources to secure all Istio components, including:

- istiod
- Istio ingress/egress gateways
- istio-cni
- ztunnel

Network policies should follow a default-deny approach with explicit rules for required ingress and egress traffic. Where possible, leverage upstream Istio-provided NetworkPolicies to reduce maintenance overhead and stay aligned with upstream best practices. This will improve isolation and reduce the attack surface while ensuring all Istio functionality, including ambient mode components, continues to operate correctly.

# OLD ISSUE

### Is your feature request related to a problem? Please describe.

https://github.com/defenseunicorns/uds-core/pull/66 introduced the new operator to UDS Core but does not include UDS Package CRs for istio. A basic `Package` CR would be beneficial to control network policies on Istio.

### Describe the solution you'd like

Add a `Package` CR to Istio as part of its [common chart](https://github.com/defenseunicorns/uds-core/tree/main/src/istio/common/chart/templates). Note that we may need to code some specific logic for `istio-system` to _NOT_ apply istio injection since this is a special case.

Definition of done: Istio contains a `Package` CR with network policies and monitoring as needed. All Istio functionality continues to work as expected (make sure to validate with ambient components)

### Describe alternatives you've considered

1. Network policies and any other necessary objects could be created/deployed "manually" rather than via the operator. While this would work and sidestep any dependency on Pepr it feels strange to not leverage the operator's benefits.
2. Leave Istio unprotected with no network policies.

TristanHoladay:
Spoke with Barrett from the Pepr team about the use case of adding a chart programatically to the generated pepr module zarf.yaml. Conclusion was that it's not a good fit for Pepr to handle this. We could always write a script do this but probably not worth it just to keep from declaring two pepr packages in the standard package zarf.yaml.

mjnagel:
Noting here that istio will need egress to the SSO host for jwks/authservice pieces - https://github.com/defenseunicorns/uds-core/pull/201#discussion_r1653139220

mjnagel:
Since Pepr is deployed before Istio now, this issue can be worked - rescoped to just Istio (for now). We could apply a `Package` CR for Pepr as part of an additional component (after Pepr) but would need to be careful to not cause issues locking our operator out (i.e. restrictive netpols, cycling itself for istio injection, etc).

For now let's just scope this to Istio.

mjnagel:
[...]

PR #201, opened 2024-02-22
feat: enable authservice integration

rjferguson21:
## Description
* Enable Authservice integration via 
```
enableAuthserviceSelector:
        app: httpbin
```
* Update to use authservice-go
* Creates per application AuthorizationPolicies to enable authservice
 
To test:
See https://github.com/defenseunicorns/uds-core/blob/authservice-pepr/docs/IDAM.md

## Related Issues

N/A


jeff-mccoy:
> sso.uds.dev needs to be resolvable inside the cluster for authservice, I am currently running with this command to workaround: "uds zarf package deploy oci://defenseunicorns/uds-k3d:0.3.1 --confirm --set=K3D_EXTRA_ARGS=\"--subnet 192.168.0.0/16 --host-alias 192.168.0.1:sso.uds.dev\""

I wonder if we could use the backchannel like we do with grafana? `http://keycloak-http.keycloak.svc.cluster.local:8080/realms/uds/protocol/openid-connect/`, unless it would get agree because Istio is managing TLS?

docs/IDAM.md
chance-coleman, line 1: I probably missed a conversation about this, does this replace a README for `src/authservice`? Just curious if we want both a docs folder and README's in different places or if it should be unified into one or the other.

rjferguson21: I am open to opinions on this but I created IDAM.md which then links to the package README.md, but also goes into more "how to use" explanation. 

But I did just realize there was no README.md for authservice so I added one.

chance-coleman (dismissed):

docs/IDAM.md
```diff
@@ -0,0 +1,86 @@
...
+
+For details on the AuthService Package see [README.md](../src/authservice/README.md)
+
+## Protecting a UDS Package with AuthService
+To enable authentication for applications that do not have native OIDC configuration, UDS Core can utilize AuthService as an authentication layer.
+
+Follow these steps to protect your application with AuthService:
+
+* Set the `isAuthSvcClient` field to `true` in the `sso` configuration of the Package.
+* Ensure that the pods of the application are labeled with: `protect: keycloak`.
```
mjnagel: Throwing a comment here since we discussed offline - thoughts on automatically handling this (`protect: keycloak`)? I think if we did per-workload authn/z and netpols for ingress -> authservice it would be doable (use workload specific selectors rather than the generic `protect: keycloak`)? We would have to add `selector` to the spec and force the user to specify that instead of having them label their workload, but otherwise I think this is doable? Not sure which is a better flow.

bburky: I'm unsure what the best option is.

Yes, you can do per namespace by just dropping a copy of the RequestAuthentication and AuthorizationPolicy resources into every namespace you want protected, yes.

However there are sometimes reasons to want to skip authservice protection on a few Pods for various reasons. (Something completely cluster-internal only perhaps that breaks when authservice is added.) And there's no way to  opt out of the policies if they have no `matchLabels` (and negated `matchLabels` isn't a thing in Istio).

So I think the safest option is to continue to do `protect: keycloak`. But perhaps a Pepr mutating admission webhook could auto-inject `protect: keycloak` into every Pod unless it's opted out? (unsure where opt-out goes, it is security relevant. Could be the UDS CR for most security, but Pod annotations for opt out is probably fine)

mjnagel: @bburky to clarify my thought was per-namespace, with selector on a specific workload. So in essence someone would pass a selector into the SSO config for the pod(s) that they want protected. Then we would generate the applicable authn/z pieces for that namespace + selector. This prevents us from having to do any mutation of `protect: keycloak` which I though would be pretty messy (not every chart exposes labels/annotations, so it gets tough to find an easy place to opt out).

bburky: Cool. Yes per namespace plus a selector should work great. Some reason I didn't consider providing a customizable selector in the UDS CR or something like that. We could probably make it so if you omit the selector it does everything in the namespace, which is the most common case I think.

mjnagel (commented):
Could you also add a renovate packageRule to group the authservice PRs?

See the istio one as an example: https://github.com/defenseunicorns/uds-core/blob/authservice-pepr/renovate.json#L162-L167

src/pepr/operator/crd/generated/package-v1alpha1.ts
```diff
@@ -438,13 +438,14 @@ export interface Sso {
    */
   description?: string;
   /**
-   * Whether the SSO client is enabled
+   * Labels to match pods to automatically protect with authservice. Leave empty to disable
```
bburky, line 478: I think an empty selector of `{}` like: `enableAuthserviceSelector: {}` should enable Authservice for all Pods in a namespace, correct?

I'd like to include this in the docs and recommend it as a default. I'd rather people only write narrow selectors if they need to exclude some Pods.

And this should probably say "leave null to disable" instead of "empty", to disambiguate between `{}` which is actually all.

(I may have this wrong and Istio `selector: { labelSelector: {} }` and `selector: nil` may actually be different. I want to make sure we provide a way to let users do "all Pods in namespace")

rjferguson21: `enableAuthserviceSelector: {}` should enable authservice for all pods in the namespace, but I am curious to hear why that would be your recommendation for the default? 

Does that lean into an assumption of there being a single pod/deployment in the namespace? If a user deployed an application with an internal API or database, how would we recommend they authenticate?

src/authservice/README.md
```diff
@@ -0,0 +1,13 @@
...
+including [Authentication Policy](https://istio.io/docs/tasks/security/authn-policy/) and [RBAC](https://istio.io/docs/tasks/security/rbac-groups/).
+Together, they allow developers to protect their APIs and web apps without any application code required.
+
+See [IDAM.md](../../docs/IDAM.md) for guidance on using the [UDS Package](../pepr/operator/README.md) custom resource to generate Authservice chains.
+
+### Changes from the DoD Platform One Big Bang package
+
+This package differs from the [DoD Platform One Big Bang package](https://repo1.dso.mil/big-bang/product/packages/authservice) in a few key ways:
+
+* Leverages the [authservice-go](https://github.com/tetrateio/authservice-go) project which is the successor to the original [authservice](https://github.com/istio-ecosystem/authservice) project.
```
mjnagel: nit: BB has switched to 1.0.0 already - https://repo1.dso.mil/big-bang/product/packages/authservice/-/blob/main/chart/values.yaml?ref_type=heads#L62

src/authservice/values/registry1-values.yaml
```diff
@@ -1,3 +1,3 @@
 image:
-  repository: registry1.dso.mil/ironbank/istio-ecosystem/authservice
-  tag: "0.5.3"
+  repository: registry1.dso.mil/ironbank/istio-ecosystem/authservice-go
+  tag: "1.0.2"
```
mjnagel: We should probably point to standard authservice now as the tetrate go-fork is archived - https://github.com/tetrateio/authservice-go - all future updates should end up in the main repo.

src/pepr/operator/crd/generated/package-v1alpha1.ts
bburky: Just that I'd encourage auth on all webservers that are reachable (in-cluster/SSRF and external) and don't already have auth. If you're careful to write good policies it's fine to exclude things, but it's a safe default to encourage I think.

I'd recommend docs something like:
> Use `enableAuthserviceSelector: {}` to add authservice to your app. If you need to exclude some Pods, edit it as needed and verify that the excluded Pods are not reachable externally or from other namespaces without authentication.

Doing auth on `{}` is also future proof to the app changing in the future and adding new Pods that don't match the old selector.

bburky: (there's also the actual "zero trust" answer, but it only applies to custom developed apps: Encourage people to develop apps that don't access their backend unauthenticated. Have them pass the user's JWT from the frontend Pod to the backend/db Pod as auth. In this setup, you can reuse istio JWT authn/authz checking... but you actually don't want the CUSTOM authservice authz. Not sure if this is something we want to build support for, but it actually results in good design of custom apps which create a great audit log in isto access logs)

src/authservice/chart/values.yaml
```diff
@@ -45,9 +45,9 @@ global:
 
   oidc:
     # -- OpenID Connect hostname.  Assumption of Keycloak based on URL construction
-    host: login.uds.dev
+    host: sso.uds.dev
```
mjnagel: Should use domain zarf var here.

mjnagel (commented):
Forgot to add this review at some point.

src/authservice/README.md
rjferguson21: Updated to 1.0.1, which is the latest from that repo.

src/authservice/values/registry1-values.yaml
rjferguson21: Updated to point to the non-archived repo.

src/authservice/chart/values.yaml
rjferguson21: Updated

docs/IDAM.md
```diff
@@ -0,0 +1,87 @@
+## Keycloak Package
```
mjnagel: @UnicornChance should we move this to a different section of the docs/update the format to fit the docs site?

renovate.json
```diff
@@ -93,6 +98,11 @@
       "groupName": "grafana",
       "commitMessageTopic": "grafana"
     },
+    {
+      "matchFileNames": ["src/authservice/**"],
+      "groupName": "authservice",
+      "commitMessageTopic": "authservice"
+    },
```
mjnagel: Duplicated the addition here it looks like?

src/authservice/README.md
mjnagel: Unresolving this one because I remembered why I commented this - it's technically no longer a difference from the BB package (since they adopted authservice-go as well). I'm not sure if its worth calling out the dynamic selector/client creation pieces with pepr or just drop this part of the readme?

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,127 @@
...
+  let config = JSON.parse(atob(authSvcSecret!.data!["config.json"])) as AuthserviceConfig;
+
+  // update config based on event
+  config = buildConfig(config, event);
+
+  // update the authservice secret
+  await updateAuthServiceSecret(config);
+}
+
+// might be neccesary to rebuild the entire config
```
mjnagel:
```suggestion
// might be necessary to rebuild the entire config
```

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -63,6 +63,24 @@ export async function networkPolicies(pkg: UDSPackage, namespace: string) {
...
+  const ssos = pkg.spec?.sso?.filter(sso => sso.enableAuthserviceSelector);
+
+  for (const sso of ssos || []) {
+    const policy: Allow = {
+      direction: Direction.Egress,
+      selector: sso.enableAuthserviceSelector,
+      remoteNamespace: "authservice",
+      remoteSelector: { "app.kubernetes.io/name": "authservice" },
+      port: 10003,
+      description: "Authservice Egress",
```
mjnagel: I think this has potential to conflict on naming if multiple authservice clients are used in the same namespace + package CR. If we add the selector to the description that should make it safe.

src/test/zarf.yaml
```diff
@@ -14,8 +14,12 @@ components:
       - name: app-tenant
         files:
           - "app-tenant.yaml"
+      - name: app-authservice-tenant
+        files:
+          - "app-authservice-tenant.yaml"
     images:
       - docker.io/kong/httpbin:latest
+      - arunvelsriram/utils
```
mjnagel: I don't see anywhere this is used, was it a carryover from a previous test?

tasks.yaml
```diff
@@ -51,24 +52,52 @@ tasks:
...
+  - name: registry-login
+    description: "Authenticate and Validate registry connection"
+    actions:
+      - task: common-setup:registry-login
+        with:
+          registry: ${REGISTRY}
+          registryUsername: ${REGISTRY_USERNAME}
+          registryPassword: ${REGISTRY_PASSWORD}
+          registryRetryInterval: ${REGISTRY_RETRY_INTERVAL}
+
```
mjnagel: I don't think we have this common-setup imported anymore.
```suggestion
```

src/test/tasks.yaml
```diff
@@ -57,6 +57,37 @@ tasks:
...
+          cluster:
+            kind: Deployment
+            name: authservice
+            namespace: authservice
+
+      - description: Verify the authservice tenant app is protected by checking redirect
+        maxRetries: 3
+        cmd: |
+          set -e
+          SSO_REDIRECT=$(uds zarf tools kubectl run curl-test --image=cgr.dev/chainguard/curl:latest -q --restart=Never --rm -i -- -Ls -o /dev/null -w %{url_effective} "https://protected.uds.dev")
```
mjnagel, line 78: Was this run inside k8s for a specific reason? We have a somewhat similar grafana check that uses the host curl to test Grafana - https://github.com/defenseunicorns/uds-core/blob/main/src/grafana/tasks.yaml#L11-L25 - okay with either but wanted to check on the "why" here.

src/test/app-authservice-tenant.yaml
```diff
@@ -0,0 +1,81 @@
...
+        - image: docker.io/kong/httpbin
+          imagePullPolicy: IfNotPresent
+          name: httpbin
+          resources:
+            limits:
+              cpu: 100m
+              memory: 128Mi
+            requests:
+              cpu: 100m
+              memory: 128Mi
```
mjnagel: Trying to be mindful of us pressing up against the runner resource limit - can you cut these in half to match the other tenant/admin apps? Also maybe worth asking...could we just do a single tenant app w/ authservice protection (vs two separate ones)? I know it starts to combine tests potentially so open to your opinion.

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -120,8 +136,13 @@ async function syncClient(
...
-    if (isAuthSvcClient) {
-      // Do things here
+    if (enableAuthserviceSelector) {
+      // Add additional authservice config
+      await reconcileAuthservice(
+        { client, name, action: Action.Add },
+        enableAuthserviceSelector,
+        pkg,
+      );
     }
```
mjnagel: This currently doesn't allow for modifications, at least to client-name (and unsure how others might be handled). If I deploy the test http-bin app and change the client name in the `Package` CR to `uds-core-httpbin-foo` I end up with:
- a new `uds-core-httpbin-foo` client in Keycloak
- the old `uds-core-httpbin` client deleted from Keycloak
- `RequestAuthentication` and `AuthorizationPolicy` resources for _both_ clients
- Authservice chain for _both_ clients

The net result is that I hit `protected.uds.dev` and it errors on client not found (since that chain is first in authservice). TLDR: Need to make sure we are handling modifications as expected here - I also tested changes to redirectUri and you end up with similar behavior for chains (although the Istio resources work out OK in that case).

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,127 @@
...
+  labelSelector: { [key: string]: string },
+  pkg: UDSPackage,
+) {
+  await updateConfig(event);
+  await updatePolicy(event, labelSelector, pkg);
+}
+
+// write authservice config to secret
+export async function updateConfig(event: AuthServiceEvent) {
+  const authSvcSecret = await K8s(kind.Secret).InNamespace(namespace).Get(secretName);
```
mjnagel: There's an issue here with managing this secret and assuming it exists outside the scope of pepr. If you run an "upgrade" of uds-core it will end up resetting the authservice secret to the state of the chart (namely just the local chain). It might make sense for us to entirely manage that secret with pepr, rather than trying to build upon something controlled by the helm chart?

Alternatively we could have Authservice make its own secret and then pepr makes a new one that "extends" it. There are probably other options with mutations - just comes down to how we want to manage and whether we want other people to be able to use Authservice chains outside of the `Package` flow.

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// might be neccesary to rebuild the entire config
+export function buildConfig(config: AuthserviceConfig, event: AuthServiceEvent) {
+  let chains: Chain[];
+
+  if (event.action == Action.Add) {
+    // add the new chain to the existing authservice config
+    chains = config.chains.concat(buildChain(event));
+  } else {
```
mjnagel: A side effect of this logic and the way we `Reconcile` `Package`s on pepr restart - everytime I cycle the watcher pod I end up duplicating the chain in the authservice secret. We likely need to handle some logic around modify/de-dupe for the chains here.

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -66,14 +70,26 @@ export async function purgeSSOClients(pkg: UDSPackage, refs: string[] = []) {
     if (token) {
       Store.removeItem(ref);
       await apiCall({ clientId }, "DELETE", token);
+
+      // find sso by clientId
+      const sso = pkg.spec?.sso?.find(sso => sso.clientId === clientId);
+
+      // if sso.enableAuthserviceSelector was set, remove from authservice config
+      if (sso && sso.enableAuthserviceSelector) {
```
mjnagel: This conditional currently doesn't work due to checking the "active" spec for the client ID vs the "previous state". We'll need to make sure we have a way to check the previous state to determine if something was an authservice client and got removed/modified.

renovate.json
mjnagel:
```suggestion
```

mjnagel (commented):
Base functionality seems pretty good, but I have some concerns here around upgrades/modifications and the way we are handling those.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,127 @@
...
+    spec: {
+      selector: {
+        matchLabels: labelSelector,
+      },
+      rules: [
+        {
+          from: [
+            {
+              source: {
+                requestPrincipals: ["https://sso.uds.dev/realms/uds/*"],
```
mjnagel: This should use `UDSConfig.domain`

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,127 @@
...
+    metadata: {
+      name: `${name}-jwt-authn`,
+      namespace,
+    },
+    spec: {
+      jwtRules: [
+        {
+          forwardOriginalToken: true,
+          issuer: "https://sso.uds.dev/realms/uds",
+          jwksUri: "https://sso.uds.dev/realms/uds/protocol/openid-connect/certs",
```
mjnagel: These should use `UDSConfig.domain`

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,127 @@
...
+      action: Action.Custom,
+      provider: {
+        name: "authservice",
+      },
+      rules: [
+        {
+          when: [
+            {
+              key: "request.headers[authorization]",
+              notValues: ["*"],
```
bburky, line 42: This policy allows bypassing Authservice (but not Istio's JWT verification) by including an `Authorization:` header in the incoming request.

If we have a real use case for this and we're confident in the Istio JWT verification (it needs to be improved), we _could_ keep this. Otherwise remove it.

Some possible reasons to keep this code are to allow:
* A mission app frontend may need to send a request to it's backend. Because the backend is authservice protected, this allows the frontend to send the user's JWT to the backend and pass Istio JWT valdiation
    *  I think this specific use case really makes a lot of sense./
    * Unless we expect people to remove authservice protection from mission app backend Pods?
* If any apps use something like the `grant_type=client_credentials` OAuth flow for (non-browser) API access to Authservice-protected mission apps.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,127 @@
...
+  namespace: string,
+): RequestAuthentication {
+  return {
+    kind: "RequestAuthentication",
+    metadata: {
+      name: `${name}-jwt-authn`,
+      namespace,
+    },
+    spec: {
+      jwtRules: [
```
bburky, line 96: Please check `audiences` to enforce restrictions on the `aud` claim (which matches a Keycloak client). Template the Keycloak client name into the `audiences` array here.

This prevents a (somehow leaked) JWT from a _different_ client being used to authenticate to the wrong app. 

Authservice itself does enforce this too, but we generally try to enforce everything with Istio too. ...Also the CUSTOM AuthorizationPolicy actually allows bypassing Authservice.

There are some use cases where users may want some additional clients though. Specifically if they want to use Keycloak clients as service accounts with `grant_type=client_credentials` to implement (non-browser) API access to Authservice-protected mission apps.
Probably future work: It may make sense to add a new field to the UDS package CR like `additionalNonAuthserviceClients` (name TBD). If users use something like `grant_type=client_credentials`
...I suppose you don't really need to ever make this configurable, they could just make a second RequestAuthentication in the same namespace which gets merged with this.

src/test/app-authservice-tenant.yaml
rjferguson21: Updated

src/pepr/operator/controllers/network/policies.ts
rjferguson21: Updated

src/test/tasks.yaml
rjferguson21: I did this because I did not want to have to ensure curl existed when the tasks run.  Happy to conform as it looks like its already required but my thought was that this test package + validate would be more portable.

pepr.ts
```diff
@@ -15,6 +16,7 @@ import { prometheus } from "./src/pepr/prometheus";
   await registerCRDs();
   // KFC watch for exemptions and update in-memory map
   await startExemptionWatch();
+  await setupAuthserviceSecret();
```
bburky, line 22: `pepr.ts` runs during `pepr build`, but this function will query a real cluster which is only available at deploy time. I think the other two functions here work because they check the existance of `process.env.PEPR_WATCH_MODE`, which ensures they only run at runtime?

The error I get is:
```
{"level":50,"time":1719327352861,"pid":85781,"hostname":"Blakes-MacBook-Pro.local","ok":false,"status":400,"statusText":"Unknown error"}
```

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -0,0 +1,139 @@
...
+  // create secret if it doesn't exist
+  try {
+    const secret = await K8s(kind.Secret)
+      .InNamespace(operatorConfig.namespace)
+      .Get(operatorConfig.secretName);
+    Log.info(`Authservice Secret exists, skipping creation - ${secret.metadata?.name}`);
+  } catch (e) {
+    Log.info("Secret does not exist, creating authservice secret");
+    await updateAuthServiceSecret(buildInitialSecret(), false);
+  }
```
bburky: need to handle errors in this path too

docs/configuration/uds-operator.md
```diff
@@ -154,6 +154,30 @@ spec:
...
+spec:
+  sso:
+    - name: Demo SSO httpbin
+      clientId: uds-core-httpbin
+      redirectUris:
+        - "https://httpbin.uds.dev/login"
+      enableAuthserviceSelector:
+        app: httpbin
+```
+For a complete example, see [app-authservice-tenant.yaml](https://github.com/defenseunicorns/uds-core/blob/main/src/test/app-authservice-tenant.yaml)
```
bburky: I would suggest a more conservative default of the following, which protects all Pods in the namespace by default:
```yaml
enableAuthserviceSelector: {}
```

Particularly, I want to ensure all Pods have authentication of some type. Postgres or other in-cluster databases can be excluded, but I actually would consider protecting mission app's backend HTTP services by default. This _does_ require the mission app developers to pass the JWT from requests to the frontend to the backend app though.

One attack this protects mission apps from is XSRF between the frontend and backend Pods of the mission app itself. It also protects the backend Pod from other in-cluster attacks, but NetworkPolicies _should_ block other in-cluster requests. Intra-namespace between frontend and backend Pods will always be allowed though and really is vulnerable to some things like XSRF vulnerabilities in the frontend mission app code.

As needed, sure people can configure `enableAuthserviceSelector:` to only specific labels. But then authservice doesn't actually protect the excluded Pods. I'd rather see us recommend a secure by default configuration. The cluster administer needs to do a careful review of any code that has no authservice-protection and ensure it has sufficient authentication (which may involve a deep code review of a mission app).

We should discuss this though. I think this was the default mission app deployment model in Party Bus. @ntwkninja, did you say you were requiring mission apps to pass a JWT from frontend to backend?

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,164 @@
...
+        {
+          forwardOriginalToken: true,
+          issuer: `https://sso.${UDSConfig.domain}/realms/uds`,
+          jwksUri: `https://sso.${UDSConfig.domain}/realms/uds/protocol/openid-connect/certs`,
```
bburky: It looks like we currently don't have any NetworkPolicies in `istio-system` currently? Please create an issue for that to follow up on later.

Using `jwksUri` requires `istiod` to reach out to a server and requires it's egress to be allowed. But it's not currently an issue because there isn't actually any NetworkPolicies in `istio-system`, it's something you'll need to add when you add them to `istio-system` though.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,164 @@
...
+    },
+  };
+}
+
+async function updatePolicy(
+  event: AuthServiceEvent,
+  labelSelector: { [key: string]: string },
+  pkg: UDSPackage,
+) {
+  const operation = event.action === AuthServiceAction.Add ? "Apply" : "Delete";
```
bburky: This is a typesafe and future proof equivalent (it will throw type errors if there are new items added to the Enum):
```suggestion
  const operation = {[AuthServiceAction.Add]: "Apply", [AuthServiceAction.Remove]: "Delete"}[event.action];
```

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,164 @@
...
+      p!.metadata!.labels = {
+        "uds/package": pkg.metadata!.name!,
+        "uds/generation": generation,
+      };
+      const kind = getKind(p.kind);
+      await K8s(kind)[operation](p);
+    });
+  } catch (e) {
+    Log.error(e, `Failed to update auth policy for ${event.name} in ${namespace}: ${e}`);
+  }
```
bburky: I think this is clearer. `getKind()` isn't really typesafe if there are more policy types added in the future, this avoids that by removing the loop and switching what the helper function is doing.

If you want to make long lines shorter you can add a temp variable, but I don't really have strong opinion on that part.

```suggestion

  function updateMetadata(resource: RequestAuthentication | AuthorizationPolicy) {
    resource!.metadata!.ownerReferences = getOwnerRef(pkg);
    resource!.metadata!.labels = {
      "uds/package": pkg.metadata!.name!,
      "uds/generation": generation,
    };
    return resource;
  }

  try {
    await K8s(AuthorizationPolicy)[operation](updateMetadata(authserviceAuthorizationPolicy(labelSelector, event.name, namespace)));
    await K8s(RequestAuthentication)[operation](updateMetadata(authNRequestAuthentication(labelSelector, event.name, namespace)));
    await K8s(AuthorizationPolicy)[operation](updateMetadata(jwtAuthZAuthorizationPolicy(labelSelector, event.name, namespace)));
  } catch (e) {
    Log.error(e, `Failed to update auth policy for ${event.name} in ${namespace}: ${e}`);
  }
```

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  const orphanRequestAuth = requestAuthentications.items.filter(
+    requetAuth => requetAuth.metadata?.labels?.["uds/generation"] !== generation,
+  );
+
+  for (const orphan of orphanRequestAuth) {
+    Log.debug(orphan, `Deleting orphaned RequestAuthentication ${orphan.metadata!.name}`);
+    await K8s(RequestAuthentication).Delete(orphan);
+  }
+}
```
bburky: I think this shorter version is clearer?

Also, what do you want the behavior of `resource.metadata?.labels?.["uds/generation"] !== generation` to be when the left side evaluates to `undefined` (no generation label)? Currently it will delete it. I just wanted to make sure that's expected

```suggestion
async function purgeOrphanPolicies(generation: string, namespace: string, pkgName: string) {
  for (const kind of [AuthorizationPolicy, RequestAuthentication]) {
    const resources = await K8s(kind)
      .InNamespace(namespace)
      .WithLabel("uds/package", pkgName)
      .Get();

    for (const resource of resources.items) {
      if (resource.metadata?.labels?.["uds/generation"] !== generation) {
        Log.debug(resource, `Deleting orphaned ${resource.kind!} ${resource.metadata!.name}`);
        await K8s(kind).Delete(resource);
      }
    }
  }
}
```

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export async function authservice(pkg: UDSPackage) {
+  // Get the list of clients from the package
+  const authServiceClients = R.filter(
+    sso => R.isNotNil(sso.enableAuthserviceSelector),
+    pkg.spec?.sso || [],
+  );
+
+  for (const client of authServiceClients) {
+    const name = `sso-client-${client.clientId}`;
```
bburky: This `sso-client-${clientName}` store name templating happens in a few different places (also in `client-sync.ts`). To ensure that they always format in the same way, should we pull this string templating into a function?

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+    const token = Store.getItem(name);
+    if (!token) {
+      throw new Error(`Failed to get token for client ${client.clientId}`);
+    }
+    const keycloakClient = await apiCall(client, "GET", token);
```
bburky: Because `authservice()` runs after `keycloak()` in `packageReconciler()`, it's actually guaranteed that `syncClient()` already ran on every client and the API call to create the Keycloak client has already happened and has been put into the store

We probably should refactor this into a `getSecret(sso: Sso)` function in `client-sync.ts`. There's actually a bug that `client.secret` is ignored in both this code and `syncClient()` (I think the correct behavior is no not create the Keycloak client if the secret already exists? I think it's intended usage is manually creating keycloak clients? That is kind of a problem with this code though which _has_ to configure the client. Perhaps assert check that `client.secret` is not set and throw an error if we can't handle the client being preconfigured).

```suggestion
```

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+  return authserviceClients;
+}
+
+export async function purgeAuthserviceClients(pkg: UDSPackage, authserviceClients: string[] = []) {
+  // compute set difference of pkg.status.authserviceClients and authserviceClients using Ramda
+  R.difference(pkg.status?.authserviceClients || [], authserviceClients).forEach(async clientId => {
+    Log.info(`Removing stale authservice chain for client ${clientId}`);
+    await reconcileAuthservice({ name: clientId, action: Action.Remove }, {}, pkg);
+  });
+}
```
bburky: Considering this function is called `purgeAuthserviceClients()` it kind of feels that like the expected behavior is to purge what you pass to it, so rename the argument to `newAuthserviceClients` for clarity.
```suggestion
export async function purgeAuthserviceClients(pkg: UDSPackage, newAuthserviceClients: string[] = []) {
  // compute set difference of existing pkg.status.authserviceClients and newAuthserviceClients using Ramda
  R.difference(pkg.status?.authserviceClients || [], newAuthserviceClients).forEach(async clientId => {
    Log.info(`Removing stale authservice chain for client ${clientId}`);
    await reconcileAuthservice({ name: clientId, action: Action.Remove }, {}, pkg);
  });
}
```

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+    const name = `sso-client-${client.clientId}`;
+    const token = Store.getItem(name);
+    if (!token) {
+      throw new Error(`Failed to get token for client ${client.clientId}`);
+    }
+    const keycloakClient = await apiCall(client, "GET", token);
+
+    client.secret = keycloakClient.secret;
+    await reconcileAuthservice(
+      { name: client.clientId, action: Action.Add, client: client as unknown as Client },
```
bburky: Can we avoid casting `client` into `Client` and just have it defined as `Sso` in `AuthServiceEvent`? This will unexpectedly leave lots of fields undefined after conversion. As far as I can tell, the `Sso` type would work everywhere and this bad conversion isn't needed.

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+      client.enableAuthserviceSelector as { [key: string]: string },
```
bburky: This is much clearer about what the assertion is:
```suggestion
      client.enableAuthserviceSelector!,
```
Also, if you just pass the `client` with the correct `Sso` type, you don't need to pass this argument separately anyway.

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+  let chains: Chain[];
+
+  if (event.action == Action.Add) {
+    // add the new chain to the existing authservice config
+    chains = config.chains.filter(chain => chain.name !== event.name);
+    chains = chains.concat(buildChain(event));
+  } else {
+    // search in the existing chains for the chain to remove by name
+    chains = config.chains.filter(chain => chain.name !== event.name);
+  }
```
bburky: We don't have typescript configured with `strictNullChecks`, so please do a type safe exhaustive match.
```suggestion
  if (event.action == Action.Add) {
    // add the new chain to the existing authservice config
    chains = config.chains.filter(chain => chain.name !== event.name);
    chains = chains.concat(buildChain(event));
  } else if (event.action == Action.Remove) {
    // search in the existing chains for the chain to remove by name
    chains = config.chains.filter(chain => chain.name !== event.name);
  } else {
    throw new Error(`Unhandled Action: ${event.action satisfies Never}`);
  }
```
(Alternately, you can use a switch statement with no default case if you set `strictNullChecks: true` in tsconfig, but enabling `strictNullChecks` have more side effects)

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  // add the new chains to the existing authservice config
+  return { ...config, chains } as AuthserviceConfig;
+}
+
+export function buildChain(update: AuthServiceEvent) {
+  // TODO: get this from the package
+  // parse the hostname from the first client redirect uri
+  const hostname = new URL(update.client!.redirectUris[0]).hostname;
```
bburky, line 94: I guess using the first redirectUri is ok... but we really need to document that the hostname of the first one is extracted and used this way

You do need to handle the error case `redirectUris` being empty somewhere (you could do it in the validator if you really want to)

bburky (commented):
Not done yet, but submitting this review for now

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
mjnagel: https://github.com/defenseunicorns/uds-core/issues/103 was opened for netpols on both istio and pepr previously, dropped a note in there for jwksUri.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
bburky: Recently there was a UDS request to allow end users use CLI tools to get JWTs (for sigstore/cosign/etc). This means either public clients redirecting to localhost, or what I'd prefer, a public client doing the device flow:
https://github.com/defenseunicorns/uds-core/issues/509
This doesn't really make the security of anything in UDS _more_ vulnerable, but it does make missing `aud` verification more easily exploitable.

I verified that this missing audiences claim verification is indeed vulnerable. I first manually created a public Keycloak client allowing the device flow using the Keycloak admin console GUI. Then I used the [`step oauth`](https://smallstep.com/docs/step-cli/reference/oauth/) CLI to get a JWT. I then passed this to protected.uds.dev in an Authorization Bearer header.

_This is not the only to leak JWTs though, it's just one of the easiest ways. Do not assume we can prevent the user from getting a JWT even without using the device flow._

Proof of concept:
```bash
# First, manually create a public client using the device flow. Then:
id_token=$(step oauth --provider https://sso.uds.dev/realms/uds/ --client-id=device-test --console-flow device | jq -r .id_token)
curl -H "Authorization: Bearer $id_token" https://protected.uds.dev/get
```

Please ensure we have `aud` claim checking by adding `audiences:` to RequestAuthentication here. (this is my reccomendation)

Alternately, disable bypassing the authservice CUSTOM AuthorizationPolicy here: https://github.com/defenseunicorns/uds-core/pull/201/files#r1643449596 (I think it's fine to keep this actually, if we check the JWT better)

docs/configuration/uds-operator.md
rjferguson21: Opened a follow on issue to evaluate our posture here: https://github.com/defenseunicorns/uds-core/issues/514

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+          token_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/token`,
+          callback_uri: update.client!.redirectUris[0],
+          client_id: update.client!.clientId,
+          client_secret: update.client!.secret,
+          scopes: [],
+          logout: {
+            path: "/local",
+            redirect_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/token/logout`,
+          },
+          skip_verify_peer_cert: true,
```
bburky: This is an uneeded override of the global config. The rest of the JWKS config is global, remove this here
```suggestion
```

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+}
+// this initial secret is only a placeholder until the first chain is created
+function buildInitialSecret(): AuthserviceConfig {
+  return {
+    allow_unmatched_requests: false,
+    listen_address: "0.0.0.0",
+    listen_port: "10003",
+    log_level: "trace",
```
bburky: This seems like a really low `log_level` to have as default. Raise it? Or make it configurable if needed?

Please do ensure any default log levels used don't actually write secrets (JWTs) to the logs. Really low debug log levels sometimes do

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+          id_token: {
+            header: "Authorization",
```
bburky: This is an incomplete override of the global `id_token` config. We could probably remove this? Otherwise, please set `preamble` here too.

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -0,0 +1,141 @@
...
+    default_oidc_config: {
+      skip_verify_peer_cert: false,
+      authorization_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/auth`,
+      token_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/token`,
+      jwks_fetcher: {
+        jwks_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/certs`,
+        periodic_fetch_interval_sec: 60,
+        skip_verify_peer_cert: "false",
```
bburky: ~~This seems like a bad default. This skips TLS verification on fetching the JWKS. It's _maybe_ safe if the JWKS URI is an in-cluster istio-protected host, but really is entirely unsafe if it's an external URL. If we can use an in-cluster `.svc.cluster.local` keycloak URL with HTTPS, that's maybe best. Otherwise, please leave this enabled by default.~~

If the deployment needs to use non-WebPKI for this, they should be configuring `trusted_certificate_authority`/`trusted_certificate_authority_file` instead.

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -0,0 +1,146 @@
...
+      },
+      client_id: "global_id",
+      client_secret: "global_secret",
+      id_token: {
+        preamble: "Bearer",
```
bburky, line 62: I'm unsure if we really need this. I think it is mostly around as a backwards compatibility thing with old Party Bus apps and things? Most clients should use the `Authorization: Bearer ...` header instead, never this `JWT:` header

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -0,0 +1,146 @@
...
+        header: "Authorization",
```
bburky, line 63: As part of re-enabling `skip_verify_peer_cert`, this CA needs to become configurable.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
rjferguson21: Refactor suggestion looks good.

Good call out about the generation, I think it is ok.  The only way that label would be undefined is if we did not create the policy?

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
rjferguson21: Updated

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,120 @@
...
+          authorization_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/auth`,
+          token_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/token`,
+          callback_uri: update.client!.redirectUris[0],
+          client_id: update.client!.clientId,
+          client_secret: update.client!.secret,
+          scopes: [],
+          logout: {
+            path: "/local",
+            redirect_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/token/logout`,
+          },
```
bburky, line 114: I'm unsure if we currently support HA authservice (which includes Redis). But if we ever do, we need to ensure the `cookie_name_prefix` is unique per chain. This is a requirement if using Redis.

See docs:
https://github.com/istio-ecosystem/authservice/blob/v1.0.1/config/v1/oidc/config.proto#L168-L176


```suggestion
          },
          "cookie_name_prefix": update.client!.clientId,
```

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
rjferguson21: Updated

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
rjferguson21: Thanks for this suggestion. Technically I needed the `client.secret` inside of this method, which the Sso type does not have but since I was already fetching the client I updated to use that object instead and it made everything a bit cleaner.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
rjferguson21: Updated.  Definitely looks better.

src/authservice/chart/templates/secret.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ include "authservice.fullname" . }}
+  name: {{ include "authservice.fullname" . }}-static
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "authservice.labels" . | nindent 4 }}
```
bburky: I think the Helm-based configuration's (defaults at least) should _exactly_ match what the operator code would generates?

Please see my other comments on the Authservice operator code and make sure the global config is in sync. This actually looks _more_ configurable (good, assuming the defaults match).

I assume the reason for this duplication is to allow people a way to inject some custom config into the authservice config? The code duplication seems not great. But I guess that use case makes sense somewhat

bburky (commented):
Some Authservice config comments.

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
rjferguson21: Updated with a note in the operator docs - https://github.com/defenseunicorns/uds-core/blob/authservice-pepr/docs/configuration/uds-operator.md#protecting-a-uds-package-with-authservice

src/pepr/operator/controllers/keycloak/authservice/config.ts
bburky: Sorry, I misread this, "skip_verify_peer_cert: false` is actually _good_ (I misread the double negative).

But `trusted_certificate_authority` and/or `trusted_certificate_authority_file` do need to be configurable if the deployment doesn't use WebPKI/enclave's custom CA/selfsigned certs/etc

bburky:
I don't think there's an existing thread for this, but I tried editing `groups:` with `kubectl edit` and it failed to apply the new group changes. It got stuck in `.status.phase=pending` and had an error in the Events. @ntwkninja saw similar behavior when applying an updated Bundle/Package too.
```
  Warning  ReconciliationFailed  2s    uds.dev/operator  Failed to process client request 'uds-core-httpbin' for authservice-test-app/httpbin-other. This can occur if a client already exists with the same ID that Pepr isn't tracking.. RETRY FAILED, aborting: {}
```

(Some what related, please delete the `apiCall()` mentioned here https://github.com/defenseunicorns/uds-core/pull/201#discussion_r1653261256 as part of fixing this. It's kind of related but probably not the cause of this bug.)

EDIT: actually... it _eventually_ reconciled. Weird. But it threw an error for quite a while

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -54,10 +55,12 @@ export async function packageReconciler(pkg: UDSPackage) {
 
     // Configure SSO
     const ssoClients = await keycloak(pkg);
+    const authserviceClients = await authservice(pkg);
```
bburky: There is a potential security issue where this can fail open:

1. Create a Package without groups, and wait for UDS operator to create the VirtualService
2. Update the Package to have authservice (and optionally groups).
   * If the operator has an error during processing of either of these SSO functions, these functions will throw an error and halt the reconcile... but the old VirtualService will stay around. This means that it fails open.

(similarly you could go from Package+authservice → Package+authservice+groups and get an error in group application and get similar bad fail-open)

I'm unsure what the best solution is. Should we maybe catch errors in the authservice code and tear down the virtual services? That would mean the operator fails closed? Perhaps move `istioResources()` below this code and pass it a bool `destroy` option that triggers teardown (this allows better code reuse)?

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
rjferguson21: I've updated the config but currently does not use Redis. Definitely an important follow on.  Created an issue https://github.com/defenseunicorns/uds-core/issues/518

src/pepr/operator/reconcilers/package-reconciler.ts
rjferguson21: :+1: Created an issue as a follow up here - https://github.com/defenseunicorns/uds-core/issues/519

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
rjferguson21: Talked about this out of band  but basically the client secret is currently not 'stored' by the operator code which makes this HTTP call necessary.   It is definitely possible to return the client (with secret) from the keycloak() reconciliation and avoid this but it would require some refactoring of the keycloak client status process.

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
rjferguson21: We have pending changes in [this PR](https://github.com/defenseunicorns/uds-core/pull/495) related to the `name` so I'm going to punt on this until that PR comes.

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
jeff-mccoy: I think that small refactor is worth it to reduce the dependency on yet another keycloak api call

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
mjnagel: @bburky can you confirm that the below addition of `audiences` satisfies your ask here and resolve this thread if it looks good?

src/test/tasks.yaml
mjnagel: Good with this, resolving.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
bburky: yes, the new code specifies `audiences:`

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
bburky: The https://github.com/defenseunicorns/uds-core/pull/201#discussion_r1643450893 fix should address this

src/authservice/chart/templates/secret.yaml
mjnagel: Yeah - need to do some cleanup here. This secret is not currently being used so we will likely be deleting the template.

The only things we need to follow up on here is:
- ensuring we have a way to get the CA cert to pepr (this could be as simple as another env passed to pepr for the config, we may need this anyways for some other CA related issues)
- deciding if we want to support non-pepr managed authservice chains (I think short-term answer is no)

src/pepr/operator/controllers/keycloak/authservice/config.ts
mjnagel: 👍 will be working on passing in the CA.

src/pepr/operator/controllers/keycloak/authservice/config.ts
bburky: sorry, this GitHub comment landed on the wrong line. It's about the `JWT` header

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -0,0 +1,143 @@
...
+    listen_port: "10003",
+    log_level: "info",
+    default_oidc_config: {
+      skip_verify_peer_cert: false,
+      authorization_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/auth`,
+      token_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/token`,
+      jwks_fetcher: {
+        jwks_uri: `https://sso.${UDSConfig.domain}/realms/${operatorConfig.realm}/protocol/openid-connect/certs`,
+        periodic_fetch_interval_sec: 60,
+        skip_verify_peer_cert: false,
```
bburky: https://github.com/istio-ecosystem/authservice/blob/main/config/v1/oidc/config.proto#L116
> Deprecated: Use the one from the OIDCConfig instead.
```suggestion
```

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -0,0 +1,143 @@
...
+        .InNamespace(operatorConfig.namespace)
+        .Get(operatorConfig.secretName);
+      Log.info(`Authservice Secret exists, skipping creation - ${secret.metadata?.name}`);
+    } catch (e) {
+      Log.info("Secret does not exist, creating authservice secret");
+      try {
+        await updateAuthServiceSecret(buildInitialSecret(), false);
+      } catch (err) {
+        Log.error(err, "Failed to create UDS managed authservice secret.");
+      }
```
bburky: The added log message is good, but this should still cause the entire reconcile to fail.

Maybe re-throw or wrap the exception and throw it. I'm not sure how you all are handling exceptions.

src/istio/values/values.yaml
```diff
@@ -6,3 +6,11 @@ meshConfig:
...
     gatewayTopology:
       forwardClientCertDetails: SANITIZE
+  extensionProviders:
+    - name: "authservice"
+      envoyExtAuthzGrpc:
+        service: "authservice.authservice.svc.cluster.local"
+        port: "10003"
+
+pilot:
+  jwksResolverExtraRootCA: "" # todo: this needs to have the non-base64 encoded CA cert for keycloak
```
mjnagel: To make a CA cert fully work with authservice we will need to set this or provide a different way for istiod to hit the jwks endpoint. This is necessary for jwtRules to work in the RequestAuthentication.

mjnagel: Tested locally and `PILOT_JWT_ENABLE_REMOTE_JWKS` = `hybrid` env on istiod does work and is a pretty simple way to do. The tradeoff is that the sidecar on the workload (i.e. mission app) is the one hitting the jwks endpoint, so it needs egress to the keycloak service. I don't think there's any risk here?, and this is something we could easily generate as part of the authservice netpol additions. We discussed this yesterday but for reference [this is the issue](https://www.github.com/istio/istio/issues/40553) where that is mentioned as an approach. Would love @bburky's review on whether that makes sense and is safe to do...

The alternative of setting the CA for istiod + hitting keycloak on the external side is a bit of a pain since the istiod chart expects the CA as a "plain PEM" (not base64 encoded). It's generally easier, and more aligned with how we handle the gateway certs, if we expect it to be passed in base64 encoded. To make that work we could either load the certs via volume/mount on istiod ourselves (creating a secret outside of the upstream chart) or just contribute upstream to istio for them to support the CA in a secret/base64 encoded format instead of the current configmap. Not opposed to these options, but they don't feel as seamless.

Another random note - this is already identified as a need for us to manage providing CAs to other packages - https://github.com/defenseunicorns/uds-core/issues/464 - so in the future this might be something for pepr to manage and inject in different spots. Current state that wouldn't help us with istiod since pepr deploys after istio (we could still make it work, but would have to modify + cycle istiod after deploy rather than mutating on deploy).

bburky: I think in-mesh egress from app sidecar to keycloak is fine. The AuthorizationPolicy DENY rules will still block `/admin*`, `/realms/master*`, `/metrics*` and `/realms/{{ .Values.realm }}/clients-registrations/*`, and I think that's all we care about. This doesn't really grant any additional access then.

src/pepr/operator/controllers/keycloak/authservice/config.ts
rjferguson21: Re-throwing this now as it is a critical failure.

docs/configuration/uds-operator.md
```diff
@@ -163,6 +163,33 @@ spec:
...
+      clientId: uds-core-httpbin
+      redirectUris:
+        - "https://httpbin.uds.dev/login"
+      enableAuthserviceSelector:
+        app: httpbin
+```
+> [!IMPORTANT]
+> The UDS Operator uses the first `redirectUris` to populate the `match.prefix` hostname and `callback_uri` in the authservice chain.
+>
+For a complete example, see [app-authservice-tenant.yaml](https://github.com/defenseunicorns/uds-core/blob/main/src/test/app-authservice-tenant.yaml)
```
mjnagel: There's a few formatting things here that look off:
- add newline on 166 before the header
- add newline on 189 before the important block quote
- I think we want a hugo alert instead of standard markdown (see [this example](https://github.com/defenseunicorns/uds-core/blob/e2efdf9b059f698369721412409509cc702593bc/docs/deployment/uds-deploy.md?plain=1#L28-L30))
- In the "IMPORTANT" note line 191 looks to be a trailing `>` (shouldn't matter once switched to hugo alert)

cc @UnicornChance unsure if I missed anything here.

mjnagel: Should we also call authservice out as a feature in the [package functions list](https://github.com/defenseunicorns/uds-core/blob/e2efdf9b059f698369721412409509cc702593bc/docs/configuration/uds-operator.md?plain=1#L9-L26) above? Possibly with a similar "ALPHA" note until further testing/security review has concluded?

pepr.ts
```diff
@@ -46,6 +48,6 @@ import { prometheus } from "./src/pepr/prometheus";
     });
   }
 })().catch(err => {
-  Log.error(err);
+  Log.error(err, "Critical error during startup. Exiting...");
```
mjnagel: I guess we missed this one and the above debug log in the previous logging PR, these should probably be using our new `setupLogger` child log stream?

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,167 @@
...
+    );
+    await K8s(RequestAuthentication)[operation](
+      updateMetadata(authNRequestAuthentication(labelSelector, event.name, namespace)),
+    );
+    await K8s(AuthorizationPolicy)[operation](
+      updateMetadata(jwtAuthZAuthorizationPolicy(labelSelector, event.name, namespace)),
+    );
+  } catch (e) {
+    log.error(e, `Failed to update auth policy for ${event.name} in ${namespace}: ${e}`);
+    throw new Error(`Failed to update auth policy for ${event.name} in ${namespace}: ${e}`, {
```
mjnagel: nit: could pull this common message into `const msg` to use in both spots?

```suggestion
    const msg = `Failed to update auth policy for ${event.name} in ${namespace}: ${e}`
    log.error(e, msg);
    throw new Error(msg, {
```

mjnagel (commented):
Working my way through the meat of the authservice code, a few small nits up front though.

docs/configuration/uds-operator.md
chance-coleman: I believe the header should be `###` to match the formatting of the sections before and after this section.

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,122 @@
...
+  let config = await getAuthserviceConfig();
+
+  // update config based on event
+  config = buildConfig(config, event);
+
+  // update the authservice secret
+  await updateAuthServiceSecret(config);
+}
+
+// might be necessary to rebuild the entire config
```
mjnagel: Is this comment still applicable?

pepr.ts
chance-coleman: yeah we did skip this, what would we call this component? just "pepr" maybe?

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -0,0 +1,122 @@
...
+  } else {
+    throw new Error(`Unhandled Action: ${event.action satisfies never}`);
+  }
+
+  // add the new chains to the existing authservice config
+  return { ...config, chains } as AuthserviceConfig;
+}
+
+export function buildChain(update: AuthServiceEvent) {
+  // TODO: get this from the package
```
mjnagel, line 92: Is this comment still necessary, can't remember if we wanted to change logic here to source the hostname from something else still?

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -0,0 +1,145 @@
...
+    log.error(e, `Failed to write authservice secret`);
+    throw new Error("Failed to write authservice secret", { cause: e });
+  }
+
+  log.info("Updated authservice secret successfully");
+
+  if (checksum) {
+    log.info("Adding checksum to deployment authservice secret successfully");
+    await checksumDeployment(configHash);
+  }
```
mjnagel, line 127: Worth noting that this checksum gets added/applied to the deployment once per chain change (add or delete) rather than batched per-package or per-interval. In practice this seemed okay when applying a package with multiple authservice clients and doing updates across them, I only saw the authservice pod actually bring up a new pod one time (suspect the transactions are fact enough that only one replicaset/pod ends up getting created).

I think this is fine for now, but maybe worth considering a batching approach in the future to limit authservice cycling - let me know if you think this is worthy of a follow on issue or not.

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -65,24 +64,25 @@ export async function keycloak(pkg: UDSPackage) {
...
   for (const ref of toRemove) {
-    const token = Store.getItem(ref);
-    const clientId = ref.replace("sso-client-", "");
+    const storeKey = `sso-client-${ref}`;
+    const token = Store.getItem(storeKey);
     if (token) {
-      Store.removeItem(ref);
-      await apiCall({ clientId }, "DELETE", token);
+      Store.removeItem(storeKey);
+      await apiCall({ clientId: ref }, "DELETE", token);
```
mjnagel: Not a change in this PR but maybe worth considering swapping the order of these two so that we don't lose a token if the deletion fails? Unsure if we can really recover in that case, but probably best to reorder?

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -188,7 +184,7 @@ export function handleClientGroups(clientReq: Sso) {
   delete clientReq.groups;
 }
 
-async function apiCall(sso: Partial<Sso>, method = "POST", authToken = "") {
+export async function apiCall(sso: Partial<Sso>, method = "POST", authToken = "") {
```
mjnagel: I think with the change to not need the GET we don't need this exported anymore?

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -217,7 +213,7 @@ async function apiCall(sso: Partial<Sso>, method = "POST", authToken = "") {
   }
 
   // Remove the body for DELETE requests
-  if (method === "DELETE") {
+  if (method === "DELETE" || method === "GET") {
```
mjnagel, line 216: Ditto to above comment, don't think we need this conditional since we aren't using GET anymore? Unless there was a reason to leave this around... _in case_?

src/pepr/operator/crd/index.ts
```diff
@@ -20,17 +20,19 @@ export {
...
-  Resolution as IstioResolution,
   Endpoint as IstioEndpoint,
+  Location as IstioLocation,
   Port as IstioPort,
+  Resolution as IstioResolution,
+  ServiceEntry as IstioServiceEntry,
 } from "./generated/istio/serviceentry-v1beta1";
 
+export { Action, AuthorizationPolicy } from "./generated/istio/authorizationpolicy-v1beta1";
+export { RequestAuthentication } from "./generated/istio/requestauthentication-v1";
```
mjnagel: nit: maybe `export as` with a prefix of `Istio` to align with other istio resources?

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,167 @@
...
+  pkg: UDSPackage,
+) {
+  // type safe map event to operation (either Apply or Delete)
+  const operation = operationMap[event.action];
+  const namespace = pkg.metadata!.namespace!;
+  const generation = (pkg.metadata?.generation ?? 0).toString();
+  const ownerReferences = getOwnerRef(pkg);
+
+  const updateMetadata = (resource: AuthorizationPolicy | RequestAuthentication) => {
+    resource!.metadata!.name = resource!.metadata!.name + "-" + generation;
```
mjnagel: Discussed offline but think we can delete the generation suffix here and simplify the purge logic.

src/authservice/chart/values.yaml
```diff
@@ -114,20 +29,7 @@ tolerations: []
...
 affinity: {}
 
-# -- Log level for the deployment
-config:
-  logLevel: trace
-
 # -- Label to determine what workloads (pods/deployments) should be protected by authservice.
 selector:
   key: protect
   value: keycloak
```
mjnagel: We can likely drop these values from here at this point AND remove the corresponding `remotePodLabels` from the [Package CR](https://github.com/defenseunicorns/uds-core/blob/authservice-pepr/src/authservice/chart/templates/uds-package.yaml#L21-L28).

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
rjferguson21: No, I will remove

src/pepr/operator/controllers/keycloak/authservice/config.ts
rjferguson21: As we discussed it does not seem to be creating undue churn but it's probably worth doing/fixing just in case so I would sleep better with an issue.

src/pepr/operator/controllers/keycloak/client-sync.ts
rjferguson21: Good call.  Will update

src/pepr/operator/controllers/keycloak/client-sync.ts
rjferguson21: I'm going to leave it just in case.

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
rjferguson21: Created this as a follow on - https://github.com/defenseunicorns/uds-core/issues/560

src/istio/values/values.yaml
mjnagel: Resolving this thread - validated custom CA + certs on the latest commits.

pepr.ts
rjferguson21: Updated to use a `startup` component.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
rjferguson21: Updated

src/pepr/operator/controllers/keycloak/authservice/config.ts
mjnagel: https://github.com/defenseunicorns/uds-core/issues/561

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -0,0 +1,173 @@
...
+  pkg: UDSPackage,
+) {
+  // type safe map event to operation (either Apply or Delete)
+  const operation = operationMap[event.action];
+  const namespace = pkg.metadata!.namespace!;
+  const generation = (pkg.metadata?.generation ?? 0).toString();
+  const ownerReferences = getOwnerRef(pkg);
+
+  const updateMetadata = (resource: IstioAuthorizationPolicy | IstioRequestAuthentication) => {
+    resource!.metadata!.name = resource!.metadata!.name;
```
mjnagel: This line can be deleted I think.

mjnagel (dismissed):

mjnagel (approved):

PR #206, opened 2024-02-27
chore: add truststore for keycloak

mjnagel:
## Description

Adds the truststore to keycloak.

## Related Issue

N/A

## Type of change

## Checklist before merging


src/keycloak/chart/values.yaml
```diff
@@ -8,7 +8,14 @@ image:
   pullPolicy: IfNotPresent
 
 # renovate: datasource=github-tags depName=defenseunicorns/uds-config versioning=semver
-configImage: ghcr.io/defenseunicorns/uds/identity-config:0.2.1
+# configImage: ghcr.io/defenseunicorns/uds/identity-config:0.3.0
+configImage: ghcr.io/mjnagel/uds-identity-config:0.3.0-dev
```
jeff-mccoy: 👀

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -90,11 +95,11 @@ spec:
...
             
             # X509 configuration
-            # - name: KC_SPI_TRUSTSTORE_FILE_FILE
-            #   value: /truststore/truststore.jks
-            # - name: KC_SPI_TRUSTSTORE_FILE_PASSWORD
-            #   value: changeit
-            
+            - name: KC_SPI_TRUSTSTORE_FILE_FILE
+              value: /opt/keycloak/conf/truststore.jks
+            - name: KC_SPI_TRUSTSTORE_FILE_PASSWORD
```
jeff-mccoy, line 97: since we're [using the default](https://www.keycloak.org/server/enabletls#_setting_the_truststore_password), think we can drop this

mjnagel: I added this as a helm value in case someone is loading an alternate truststore for their environment and "secured" it with a real password. This way they can override the helm value with their bundle to use whatever password is needed. Thoughts on that?

I could wrap this in a conditional and leave it off in the default case to keep things a bit cleaner though.

src/keycloak/chart/values.yaml
mjnagel: 😅 temp change while the other PR was in flight so that I could test e2e and easily send this branch to others for validation.

src/keycloak/chart/templates/statefulset.yaml
mjnagel: Apparently despite the default in docs this is required, or something else is nulling out the default:
```
keycloak ERROR: Unexpected error when starting the server in (development) mode
keycloak ERROR: Failed to start quarkus
keycloak ERROR: Attribute 'password' missing in 'truststore':'file' configuration
```

Will look at that a bit more and open upstream issue if that's a bug/bad docs.

rjferguson21 (approved):

PR #210, opened 2024-02-29
feat: add velero package

blancharda:
## Description
Adds a Velero package to support deploying a backup utility as part of uds-core.


src/velero/values/values.yaml
```diff
@@ -0,0 +1,56 @@
+kubectl:
+  # Container Level Security Context for the 'kubectl' container of the crd jobs. Optional.
+  # See: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
+  containerSecurityContext:
+    capabilities:
+      drop:
+        - ALL
+  # Annotations to set for the upgrade/cleanup job. Optional.
+  annotations:
+    sidecar.istio.io/inject: 'false'
```
mjnagel: Should be able to remove this I think? Big Bang had this in the past because of the job sidecar termination issue if I remember right. We have a pepr capability that should handle that so this should be able to be injected?

src/velero/tasks/deployment-validation.yaml
```diff
@@ -0,0 +1,79 @@
+variables:
+  - name: BACKUP_NAME
+    default: backup-test
+tasks:
```
mjnagel: Any reason to not inline these in `src/velero/tasks.yaml`? Having a new `tasks` folder isn't wrong, but seems unnecessary here since the top level tasks is just doing an include?

src/velero/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,45 @@
...
+      # Todo: wide open for now for pushing to s3
+      - direction: Egress
+        podLabels:
+          app.kubernetes.io/name: velero
+        remoteGenerated: Anywhere
+
+      - direction: Egress
+        podLabels:
+          batch.kubernetes.io/job-name: "velero-upgrade-crds"
+        remoteGenerated: Anywhere
```
mjnagel: Should this be KubeAPI rather than Anywhere? Would expect this job to hit kubeapi but not much else?

src/velero/values/values.yaml
mjnagel: Could also remove the `capabilities.drop` section too since that gets mutated in by Pepr 🤷

src/velero/values/values.yaml
```diff
@@ -0,0 +1,56 @@
...
+    schedule: "0 3 * * *"
+    useOwnerReferencesInBackup: false
+    template:
+      csiSnapshotTimeout: 0s
+      includeClusterResources: true
+      snapshotVolumes: false
+      excludedNamespaces:
+        - kube-system
+        - flux
+        - velero
```
mjnagel: Cross referencing the [config we did for DUBBD](https://github.com/defenseunicorns/uds-package-dubbd/blob/main/aws/values/aws-velero.yaml#L26-L31) - do you see any plusses/minuses to exclude vs include here? If we stick with exclude we can drop `flux`.

For what its worth I think excluded makes more sense so we backup things on top of uds-core as well?

src/velero/zarf.yaml
```diff
@@ -1,11 +1,42 @@
...
+      path: common
+    charts:
+      - name: velero
+        valuesFiles:
+          - values/upstream-values.yaml
+    images:
+      - velero/velero:v1.13.0
+      - velero/velero-plugin-for-csi:v0.7.0
+      - velero/velero-plugin-for-aws:v1.9.0
+      - velero/velero-plugin-for-microsoft-azure:v1.8.2
```
mjnagel: Commenting to get other opinions here - we currently have the chart configured to expect AWS storage (s3). Someone could certainly helm override to use a different provider, so to some extent makes sense to include the other plugins. Just wary of including extra images that aren't used - these ones are likely small so makybe less of a problem?

src/velero/zarf.yaml
```diff
@@ -1,11 +1,42 @@
...
+    import:
+      path: common
+    charts:
+      - name: velero
+        valuesFiles:
+          - values/registry1-values.yaml
+    images:
+      # The kubectl image used by default in the repo1 chart does not support the runtime command of the upstream chart (`which` has been removed in the IB image) -- suggest replacing with an alternative kubectl image
+      # - registry1.dso.mil/ironbank/opensource/kubernetes/kubectl:v1.28.6
+      - registry1.dso.mil/ironbank/opensource/kubernetes-1.21/kubectl:v1.21.14
```
mjnagel: Placeholder comment - wanted to revisit the ironbank images and see if I could find any that provide a newer/supported kubectl with `which`.

mjnagel (commented):
Really appreciate the work on this @blancharda - some preliminary comments, hopefully small things.

src/velero/tasks/deployment-validation.yaml
blancharda: The intention was to keep the test case output (health check vs backup functionality) separate while still matching the top level `validate` task name. I think if I in-line them both under the `validation` task, they would both look like part of the same test when they really do very different things. 

I'm certainly open to changing it if that's the pattern we're following elsewhere though.

src/velero/chart/templates/uds-package.yaml
blancharda: That's very fair. The networking rules were the thing I understood the least here. I'm sure I just copy/pasted this from another pkg 😆 

I could use a hand on the correct syntax though unless it's literally just replacing `Anywhere` -> `KubeAPI`.

mjnagel: Yep should be literally that! https://github.com/defenseunicorns/uds-core/blob/main/src/neuvector/chart/templates/uds-package.yaml#L25
```suggestion
      - direction: Egress
        podLabels:
          batch.kubernetes.io/job-name: "velero-upgrade-crds"
        remoteGenerated: KubeAPI
```
Should've added the suggestion on the first comment - just wasn't sure if there was something I was missing with other egress here.

src/velero/tasks/deployment-validation.yaml
mjnagel: Ah gotcha. So you should still be able to achieve the same goal, but all in one file for `tasks.yaml`. Rough example:
```yaml
variables:
  - name: BACKUP_NAME
    default: backup-test

tasks:
  - name: validate
    actions:
      - task: health-check
      - task: execute-backup
  - name: health-check
    actions:
      - description: Velero StatefulSet Health Check
        wait:
          cluster:
            kind: Deployment
            name: velero
            namespace: velero
  - name: execute-backup
    actions:
...
```

Basically just adding them as separate tasks, but in the same file still so that we don't need the include layer?

src/velero/values/values.yaml
blancharda: Depends which [dubbd config you crossreference](https://github.com/defenseunicorns/uds-package-dubbd/blob/main/rke2/values/rke2-velero-s3.yaml) 🙈 

That said.. I think exclude makes the most sense as a DEFAULT config. As you said, I think the intention would be to capture a backup of everything on top of uds-core.

In practice, I fully expect downstream users to override this config to meet environment specific requirements so I would almost treat this one as more of a simple example..

blancharda: I can drop flux though. That's definitely leftover from the dubbd config that I was referencing.

src/velero/values/values.yaml
blancharda: Done 👍

src/velero/values/values.yaml
mjnagel: :+1: seems legit, and yeah expect that others will override to customize what, where, when they backup things.

src/velero/zarf.yaml
blancharda: I (obviously) leaned towards including them because we've needed them (AFS driver and the generic CSI driver) in past efforts. I think the safe assumption is that the majority of use cases will be S3 or S3 compatible which is why the default values for the pkg make it easy to configure for that use case.. but I'm open to the argument that since we don't currently need it maybe we shouldn't include them. It's not like it would be difficult to add them back later as needed 🤷 

FWIW, these are the image sizes:
```
$ docker images
REPOSITORY                                                           TAG            IMAGE ID       CREATED         SIZE
velero/velero-plugin-for-csi                                         v0.7.0         af809012f8ca   5 weeks ago     55.6MB
velero/velero-plugin-for-aws                                         v1.9.0         1f561b2b60e1   5 weeks ago     85MB
velero/velero-plugin-for-microsoft-azure                             v1.8.2         4d08719dba26   3 months ago    54.5MB
```

src/velero/tasks/deployment-validation.yaml
blancharda: Doh!
Somehow I didn't realize you could reference task defs from the same file rather than an import.

Agreed that's cleaner -- I'll update accordingly

src/velero/zarf.yaml
mjnagel: Haven't identified any drop in replacements - I opened an IB issue [here](https://repo1.dso.mil/dsop/opensource/kubernetes/1.28/kubectl/-/issues/20) to request adding `which`. I'm tentatively okay with this as a shortterm. My fears:
- velero could use some newer kubectl args/commands that don't work with 1.21 (not super likely)
- security posture/CVEs with this image - it will never get updated so we can't adhere to any CVE SLAs/bugfixes

It would be ideal if we could just disable this job and install the CRDs separately, similar to how we handle with some of the other pieces of uds-core...but not sure that is cleaner in this case since they don't publish CRDs in a separate chart.

src/velero/zarf.yaml
mjnagel: Leaning towards drop the extras (for now) and explicitly call out the s3 compatible assumption. Still would be an option if the image was sideloaded (spearate zarf package) but would make it more clear what our opinion is with the default values, and keep things slimmer to start. Very happy to re-evaluate as this rolls out to more users if we find that csi/azure are common asks.

src/velero/zarf.yaml
blancharda: Agreed. I say we leave it for now, but track that issue and update if/when an alternative is available. In the meantime, if it becomes a CVE issue, we can fall back to adding each individual file as a manifest, [like we have done elsewhere](https://github.com/defenseunicorns/uds-package-postgres-operator/blob/main/zarf.yaml#L10).

src/velero/zarf.yaml
blancharda: Sounds good, I'll drop them and add a note in the README.

src/velero/zarf.yaml
```diff
@@ -1,11 +1,38 @@
...
+      - registry1.dso.mil/ironbank/opensource/velero/velero:v1.12.3
+      - registry1.dso.mil/ironbank/opensource/velero/velero-plugin-for-aws:v1.8.2
```
TristanHoladay: Not sure if there was a different reason for these versions compared to upstream, other than IB probably just being behind, but it looks like they have `v1.13.0` and `v1.9.0` now.

src/velero/README.md
```diff
@@ -1 +1,99 @@
...
+```
+> NOTE: requires [the velero CLI](https://velero.io/docs/v1.3.0/velero-install/)
+
+Alternatively, manually create a `backup` object with `kubectl`:
+```bash
+uds zarf tools kubectl apply -f - <<-EOF
+  apiVersion: velero.io/v1
+  kind: Backup
+  metadata:
+    name: ${BACKUP_NAME}
```
TristanHoladay: nit - probably better as a hard-coded name for simplicity when someone's testing this out.

src/velero/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,45 @@
...
+          - 8080
+        description: "Promtail Log Storage"
+
+      - direction: Egress
+        remoteNamespace: tempo
+        remoteSelector:
+          app.kubernetes.io/name: tempo
+        port: 9411
+        description: "Tempo"
+
```
mjnagel:
```suggestion

```
I feel bad for missing this earlier but I don't think either of these are actually needed. The promtail one seems like maybe a copy/paste from somewhere it was necessary? Tempo I lean towards leaving off since it isn't in core yet (tracking that this is in other packages though). If/when we add Tempo it would be autogenerated and added to all packages since its necessary everywhere.

mjnagel (dismissed):

TristanHoladay (changes requested):

src/velero/zarf.yaml
mjnagel: Updated to latest, pending CI this should be resolved.

src/velero/README.md
mjnagel: Updated to `test-backup`, good call.

blancharda: I think that would prevent running the validation more than once which could be a desirable feature outside of CI

blancharda: merp -- disregard, this is the README lol

rjferguson21 (approved):

TristanHoladay (approved):

PR #219, opened 2024-03-04
fix: kubeapi netpol generation now also includes the ip from the kubernetes service

corang:
Fixes a bug in EKS vpc-cni for gitlab where the gitlab pods couldn't talk to the api server

## Type of change

## Checklist before merging


TristanHoladay (approved):

mjnagel (approved):
I think this is a reasonable change to unblock the issue in EKS - shouldn't be any change to the security posture since we were allowing this traffic already (albeit using the endpoints vs clusterip).

PR #224, opened 2024-03-05
fix(keycloak): add missing postgres host and port secret keys

jeff-mccoy:
This PR adds the postgres host/port entries missing from the secret along with a conditional secret sha256sum annotation to force a roll on secret update. Note this behavior is different than than the [Codecentric chart](https://github.com/codecentric/helm-charts/blob/58a318a16a5ed220601da39fcd4e82127c3f911a/charts/keycloak/templates/statefulset.yaml#L98) that only puts the password in the secret and the rest direct-injected into the `Statefulset`.


brianrexrode (approved):
👀 👍 2️⃣ me

mjnagel (approved):

PR #241, opened 2024-03-11
fix(keycloak): only use PVC for devMode

jeff-mccoy:
Only use PVC for keycloak devMode to avoid issues in HA mode.


brianrexrode (approved):
👀 👍 2️⃣ me

rjferguson21:
Tested this with HA deployment and everything worked.

PR #245, opened 2024-03-12
feat(operator): add events and improve lifecycle ops

jeff-mccoy:
This PR enables stronger idempotent patterns with the automatic reprocessing of all CRs when seen for the first time on a new pod boot. It also adds better error data in the form of K8s event creation in when failures occur. Lastly some code refactoring for the reconcilers was done in addition to more unit tests added.


jeff-mccoy:
Examples of new event creation on failure

Screenshot 2024-03-12 at 4 27 05 AM
Screenshot 2024-03-12 at 4 26 43 AM

jeff-mccoy:
Note the behavior in this PR won't permanently resolve CR resync on pod restart until [Pepr #643](https://github.com/defenseunicorns/pepr/pull/643) is released. However, because the inner function contents are changing with this PR, the resync will occur on upgrade the first time.

package.json
```diff
@@ -33,7 +33,7 @@
     "k3d-setup": "k3d cluster delete pepr-dev && k3d cluster create pepr-dev --k3s-arg '--debug@server:0'"
   },
   "dependencies": {
-    "pepr": "0.27.0"
+    "pepr": "0.28.3"
```
mjnagel, line 36: Commenting here for visibility - not sure if we want to wait on the [Ironbank](https://repo1.dso.mil/dsop/opensource/defenseunicorns/pepr/controller) update here...but that image is technically still pinned at 0.27.0 - https://github.com/defenseunicorns/uds-core/blob/making-better-controllers/tasks/create.yaml#L7

jeff-mccoy: Yeah I was hoping to get this wrapped up this week, but pushed on IB side, so I guess we'll see what happens with them tomorrow.

mjnagel: 0.28.3 is published in IB now - prioritizing the rest of this review to get it merged and released today.

mjnagel (approved):
LGTM - the events on failure are a big plus.

rjferguson21 (approved):

PR #250, opened 2024-03-12
feat: add nightly testing eks

### LINKED ISSUE #213: Nightly "production like" testing

Currently uds-core has CI smoke testing on k3d, with a "dev" setup (single replicas, single node, minimal external dependencies). When I deploy to my RKE2 or EKS cluster I don't have a good example of an HA setup and low confidence that has been tested before I deploy in my environment.

Nightly CI tests against k8s distros, suggest targeting rke2 and eks, with a "production" bundle configuration used for CI testing only (not published).

Beyond just the basic deploy it would also be good to add more full e2e testing as part of this nightly deploy.

End users should have staging environments to test the deploy before they move to prod so this in theory isn't a high risk, but I think it's very valuable for us to test before cutting releases rather than relying on end users to test for us.

We could also run the testing on every PR, but that is likely not desirable since it slows down the fast feedback on PRs.

### Additional context

We know of end users interested in RKE2 and EKS so that is part of the rationale for targeting those distros.

We also don't want to reproduce any testing that should be done upstream, so we likely should stick to cross-application functionality (i.e. hitting istio virtual services, SSO logic, etc).

TristanHoladay:
## Description
Add EKS setup for deploying and testing uds-core.

## Related Issue
related to #213 

## Type of change

## Checklist before merging


TristanHoladay:
latest run of nightly testing https://github.com/defenseunicorns/uds-core/actions/runs/8471443672 after commit [9e2d57b](https://github.com/defenseunicorns/uds-core/pull/250/commits/9e2d57b3566233766229f58b23b23658e4f01426)

.github/test-infra/ci-iac-aws/loki/main.tf
```diff
@@ -0,0 +1,153 @@
+# test tf
```
zachariahmiller: Just going to comment this here but applies to all the terraform folders. These should be able to be consolidated into one with a loop on a couple of values to keep the repo DRY

.github/test-infra/ci-iac-aws/zarf.yaml
```diff
@@ -0,0 +1,186 @@
...
+  # x-release-please-start-version
+  version: "0.15.1"
+  # x-release-please-end
+  architecture: amd64
+  source: https://github.com/defenseunicorns/uds-core
+  documentation: https://github.com/defenseunicorns/uds-core
+  vendor: Defense Unicorns
+
+variables:
+  - name: CLUSTER_NAME
```
zachariahmiller: I dont think this should be a zarf package. It could be a couple of uds tasks called directly in CI. It would be much faster that way as well.

.github/test-infra/eks/zarf.yaml
```diff
@@ -0,0 +1,45 @@
...
+    default: ami-068ab6ac1cec494e0
+
+components:
+  - name: load-eksctl
+    required: true
+    actions:
+      onDeploy:
+        after:
+          # Remove existing eksctl
+          - cmd: rm -f eksctl
```
zachariahmiller: This also probably shouldnt be a zarf package (see other comment). 

Separately, i could question if it makes sense to just use the basic eksctl deployment instead of a config that is more representative of an EKS deployment that would be done in a prod env (for example like some of deliveries IaC). This would be beneficial for actually catching issues that may not come up in such a basic test. Additionally, it would bring the testing mroe in line functionally with any rke2 testing.

tasks/create.yaml
```diff
@@ -55,3 +58,10 @@ tasks:
           rm -fr dist
           npm ci
           npx pepr build $CUSTOM_PEPR_IMAGE
+
+  - name: no-pepr-package
```
zachariahmiller: This can be removed if the eks things arent zarf packages

zachariahmiller (changes requested):

.github/test-infra/ci-iac-aws/zarf.yaml
mjnagel: ^ ditto, I know this was probably path of least resistance for copy/paste from DUBBD but viewing now I think we can do this way simpler with tf/eksctl direct in a task. Would also handle multi-arch things without you having to think about it (currently terraform/eksctl for arm/mac are not in these packages).

.github/test-infra/eks/zarf.yaml
mjnagel: Generally agree with sentiment on different eks deployment, but think this is acceptable for first pass - we should still be able to catch something like "irsa no worky" with this.

.github/workflows/nightly-testing.yaml
```diff
@@ -0,0 +1,11 @@
+name: Nightly Testing
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # Runs at midnight every day
```
mjnagel: This is technically UTC time, so 6pm mountain. I think that's OK but we could probably move it a few hours later (6 UTC?) so that its outside the timeframe of our normal work/after all changes for a day.

tasks/create.yaml
```diff
@@ -55,3 +58,4 @@ tasks:
           rm -fr dist
           npm ci
           npx pepr build $CUSTOM_PEPR_IMAGE
+
```
zachariahmiller:
```suggestion
```

tasks/iac.yaml
```diff
@@ -0,0 +1,155 @@
...
+              overrideBootstrapCommand: |
+                #!/bin/bash
+                /etc/eks/bootstrap.sh ${CLUSTER_NAME} --container-runtime containerd
+          EOF
+
+      - cmd: eksctl create cluster --dry-run -f cluster-config.yaml
+      - cmd: sleep 5
+      - cmd: eksctl create cluster -f cluster-config.yaml
+      - cmd: eksctl utils write-kubeconfig -c ${CLUSTER_NAME}
+
```
zachariahmiller:
```suggestion

```

tasks/iac.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - cmd: |
+          terraform init -force-copy \
+          -backend-config="bucket=${STATE_BUCKET_NAME}" \
+          -backend-config="key=${BUCKETS_STATE_KEY}" \
+          -backend-config="region=${REGION}" \
+          -backend-config="dynamodb_table=${STATE_DYNAMODB_TABLE_NAME}"
+        dir: .github/test-infra/buckets-iac
+      - cmd: terraform apply -auto-approve
+        dir: .github/test-infra/buckets-iac
+
```
zachariahmiller:
```suggestion

```

tasks/iac.yaml
```diff
@@ -0,0 +1,155 @@
...
+          -backend-config="region=${REGION}" \
+          -backend-config="dynamodb_table=${STATE_DYNAMODB_TABLE_NAME}"
+        dir: .github/test-infra/buckets-iac
+      - cmd: terraform apply -auto-approve
+        dir: .github/test-infra/buckets-iac
+  
+  - name: terraform-outputs
+    actions:
+      - cmd: terraform output -raw loki_s3_bucket
+        setVariables:
```
zachariahmiller:
```suggestion
        setVariables:
```

tasks/iac.yaml
```diff
@@ -0,0 +1,155 @@
...
+        dir: .github/test-infra/buckets-iac
+  
+  - name: terraform-outputs
+    actions:
+      - cmd: terraform output -raw loki_s3_bucket
+        setVariables: 
+          - name: "LOKI_S3_BUCKET"
+        dir: .github/test-infra/buckets-iac
+      - cmd: terraform output -raw aws_region
+        setVariables:
```
zachariahmiller:
```suggestion
        setVariables:
```

tasks/iac.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - cmd: terraform output -raw loki_s3_bucket
+        setVariables: 
+          - name: "LOKI_S3_BUCKET"
+        dir: .github/test-infra/buckets-iac
+      - cmd: terraform output -raw aws_region
+        setVariables: 
+          - name: LOKI_S3_AWS_REGION
+        dir: .github/test-infra/buckets-iac
+      - cmd: terraform output -raw loki_irsa_role_arn
+        setVariables:
```
zachariahmiller:
```suggestion
        setVariables:
```

tasks/iac.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - cmd: terraform output -raw aws_region
+        setVariables: 
+          - name: LOKI_S3_AWS_REGION
+        dir: .github/test-infra/buckets-iac
+      - cmd: terraform output -raw loki_irsa_role_arn
+        setVariables: 
+          - name: LOKI_S3_ROLE_ARN
+        dir: .github/test-infra/buckets-iac
+      - cmd: terraform output -raw velero_s3_bucket
+        setVariables:
```
zachariahmiller:
```suggestion
        setVariables:
```

tasks/iac.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - cmd: terraform output -raw loki_irsa_role_arn
+        setVariables: 
+          - name: LOKI_S3_ROLE_ARN
+        dir: .github/test-infra/buckets-iac
+      - cmd: terraform output -raw velero_s3_bucket
+        setVariables: 
+          - name: VELERO_S3_BUCKET
+        dir: .github/test-infra/buckets-iac
+      - cmd: terraform output -raw aws_region
+        setVariables:
```
zachariahmiller:
```suggestion
        setVariables:
```

tasks/iac.yaml
```diff
@@ -0,0 +1,155 @@
...
+                eks.amazonaws.com/role-arn: "${LOKI_S3_ROLE_ARN}"
+              velero_use_secret: false
+              velero_irsa_annotation:
+                eks.amazonaws.com/role-arn: "${VELERO_S3_ROLE_ARN}"
+              velero_bucket: ${VELERO_S3_BUCKET}
+              velero_bucket_region: ${VELERO_S3_AWS_REGION}
+              velero_bucket_provider_url: ""
+              velero_bucket_credential_name: ""
+              velero_bucket_credential_key: ""
+          EOF
```
zachariahmiller:
```suggestion
          EOF

```

.gitignore
```diff
@@ -13,3 +13,5 @@ insecure*
 zarf
 tmp-tasks.yaml
 cacert.b64
+run/
+extract-terraform.sh
```
zachariahmiller:
```suggestion
extract-terraform.sh

```

.github/workflows/test-eks.yaml
```diff
@@ -0,0 +1,88 @@
...
+        run: ZARF_ARCHITECTURE=amd64 uds run -f tasks/create.yaml standard-package
+
+      - name: Create Core Bundle
+        run: uds create .github/bundles --confirm
+
+      - name: Create Cluster
+        run: uds run -f tasks/iac.yaml create-cluster
+        timeout-minutes: 60
+
+      - name: Create IAC
```
zachariahmiller:
```suggestion
      - name: Create IAC
```

.github/workflows/test-eks.yaml
```diff
@@ -0,0 +1,88 @@
...
+        uses: actions/checkout@v4
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.AWS_COMMERCIAL_ROLE_TO_ASSUME }}
+          role-session-name: ${{ github.job || github.event.client_payload.pull_request.head.sha || github.sha }}
+          aws-region: ${{ env.UDS_REGION }}
+          role-duration-seconds: 21600
+
```
zachariahmiller:
```suggestion

```

zachariahmiller (commented):
@TristanHoladay I think this looks good. I added some suggestions to fix for the lint things. I do think it makes sense to have a follow up ticket/s to use an os that has been hardened as well as probably move the terraform or an eksctl config a little closer to what the one used in the delivery tf is.

TristanHoladay:
eks run after last iac refactors: https://github.com/defenseunicorns/uds-core/actions/runs/8529515192/job/23365445948?pr=250

zachariahmiller (approved):

PR #256, opened 2024-03-13
feat: add `monitor` to operator, fix monitoring setup

### LINKED ISSUE #17: Handle Prometheus metrics + Istio mTLS

mjnagel:
## Description

This adds a new portion to the `Package` spec, `monitor`. This functionality is briefly documented [here](https://github.com/defenseunicorns/uds-core/blob/operator-monitor-magic/docs/MONITOR.md). Also included is a mutation for all service monitors to handle them in the expected istio mTLS way. This PR also fixes some missing dashboards and enables all default dashboards from the upstream kube-prometheus-stack chart (this may be excessive?).

Currently monitored:
- Prometheus stack (operator, self, alertmanager)
- Loki
- kube-system things (kubelet, coredns, apiserver)
- Promtail
- Metrics-server
- Velero
- Keycloak
- Grafana
- All istio envoy proxies (podmonitor)

Not added here:
- NeuVector: Currently has limited config options with regards to auth, keeping this disabled in anticipation of SSO on NeuVector, with a desire to contribute upstream to enable our use case of SSO-only auth.

In addition this PR switches single package testing to use/add Istio. This appears to add around 1 minute to each pipeline run, but:
- allows us to test the istio VS endpoints in single package checks (some quirks with this in current state due to SSO redirects, but will allow us to do e2e testing in those pipelines in the future)
- allows us to assume istio always and not build bespoke pepr code for the specific no-istio scenario (ex: prometheus can assume certs)

This would still allow someone to run locally without istio for some packages / un-inject, mess around with mTLS, etc - which may be useful still to identify where Istio is causing problems. This just switches our CI posture so that we assume istio always.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/17

## Type of change

## Checklist before merging


src/prometheus-stack/values/values.yaml
```diff
@@ -132,3 +94,13 @@ prometheusOperator:
...
+
+# todo: this does not work on single package testing because istio certs are missing
+alertmanager:
+  alertmanagerSpec:
+    scheme: "https"
+    tlsConfig:
+      caFile: /etc/prom-certs/root-cert.pem
+      certFile: /etc/prom-certs/cert-chain.pem
+      insecureSkipVerify: true
+      keyFile: /etc/prom-certs/key.pem
```
mjnagel: todo: fix this for single package testing...

TristanHoladay:
neuvector chart `2.7.4` was released with your changes @mjnagel . https://github.com/defenseunicorns/uds-core/pull/333 tracking the upgrade. waiting for IB to update images to `5.3.1`

src/prometheus-stack/values/values.yaml
```diff
@@ -35,6 +19,8 @@ nodeExporter:
   serviceMonitor:
     interval: ""
 prometheus:
```
mjnagel, line 21: Add istio pod monitor?

src/pepr/prometheus/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Mutate a service monitor to enable mTLS metrics
+ *
+ */
+When(Prometheus.ServiceMonitor)
+  .IsCreatedOrUpdated()
+  .Mutate(async sm => {
+    // Provide an opt-out of mutation to handle complicated scenarios
+    if (sm.Raw.metadata?.labels?.["uds/skip-sm-mutate"]) {
```
rjferguson21: Would an annotation be a better fit here?

src/pepr/operator/controllers/monitoring/service-monitor.ts
```diff
@@ -0,0 +1,86 @@
...
+        namespace,
+        labels: {
+          "uds/package": pkgName,
+          "uds/generation": generation,
+        },
+        ownerReferences: getOwnerRef(pkg),
+      },
+      spec: {
+        endpoints: endpoints,
+        selector: promSelector,
```
rjferguson21: nit: I'd probably just inline these variables.

src/pepr/operator/controllers/monitoring/service-monitor.ts
```diff
@@ -0,0 +1,86 @@
...
+  );
+
+  // Delete any orphaned VirtualServices
+  for (const sm of orphanedSM) {
+    Log.debug(sm, `Deleting orphaned ServiceMonitor ${sm.metadata!.name}`);
+    await K8s(Prometheus.ServiceMonitor).Delete(sm);
+  }
+
+  // Return the list of monitor names
+  return [...new Set(payloads.map(sm => sm.metadata!.name!).flat())];
```
rjferguson21: Could this just be the `payloads.map`?

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -59,6 +59,28 @@ export async function networkPolicies(pkg: UDSPackage, namespace: string) {
...
+  const monitorList = pkg.spec?.monitor ?? [];
+  // Iterate over each ServiceMonitor
+  for (const monitor of monitorList) {
+    const { selector, targetPort, podSelector } = monitor;
+
+    // Create the NetworkPolicy for the ServiceMonitor
+    const policy: Allow = {
+      direction: Direction.Ingress,
+      selector: podSelector ?? selector,
+      remoteNamespace: `monitoring`,
```
rjferguson21:
```suggestion
      remoteNamespace: "monitoring",
```

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -59,6 +59,28 @@ export async function networkPolicies(pkg: UDSPackage, namespace: string) {
...
+      remoteSelector: {
+        app: `prometheus`,
```
rjferguson21:
```suggestion
        app: "prometheus",
```

src/pepr/prometheus/index.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (ns === undefined) {
+        return;
+      }
+      const namespace = await K8s(kind.Namespace).Get(ns);
+      if (
+        namespace.metadata?.labels &&
+        namespace.metadata.labels["istio-injection"] === "enabled"
+      ) {
+        istioInjected = true;
+      }
```
rjferguson21: I think this could be pulled out into a helper function like `isIstioInjected`.  I believe you are after if any of the namespaces (or namespace of the SM) are istio injected.

src/pepr/operator/controllers/monitoring/service-monitor.ts
```diff
@@ -0,0 +1,86 @@
...
+        ownerReferences: getOwnerRef(pkg),
+      },
+      spec: {
+        endpoints: endpoints,
+        selector: promSelector,
+      },
+    };
+
+    // Apply the VirtualService and force overwrite any existing policy
+    await K8s(Prometheus.ServiceMonitor).Apply(payload, { force: true });
```
rjferguson21, line 27: Not going to hold up this PR but if we could do our best to avoid mixing the code for "resource building" and us using KFC to apply it will make unit testing this code a lot easier.  

It is a small thing but wherever we can segregate pure functions we should try to.

mjnagel: Lmk if this fits what you expected - https://github.com/defenseunicorns/uds-core/pull/256/commits/9cbd7db3cc8334930761c52b6c52c631fe79c86d - unit test is a bit eh, but attempted to just pull that generation piece out. Think I could've recreated/grabbed some of the vars in the function itself rather than passing them, but they're needed in both spots so it seemed 50/50.

chance-coleman (dismissed):

rjferguson21 (approved):

PR #259, opened 2024-03-14
feat: debug output

rjferguson21:
## Description
Testing CI

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


mjnagel (commented):
Is the desire here to have an easy way for local dev debugging? My leaning is that devs would mostly want to do debug in their own way, but if we want to add this would recommend we drop the github specific CI grouping headers.

PR #262, opened 2024-03-14
chore: add keycloak to dev bundle and rename

### LINKED ISSUE #251: Update core-istio-dev to optionally deploy keycloak
A common need for testing packages that depend on uds core is to be able to test the integration points with uds-core. While there are several, one of the biggest places is integration with keycloak. All SWF packages will need this for example and there is a desire to do this in a way that is able to run locally and on lower resourced runners for efficiency and cost. This feature request is to modify the internal core-istio package to have an optional keycloak component and to update the core-istio-dev bundle to reimport the package so that the istio pieces and the keycloak pieces are able to be selected via the uds-cli's --packages flag. This will enable developers to be able to select just istio or istio and keycloak as their deployment baseline for the bundles. 

This solution was discussed as a least gross, lowest maintenance way to implement this ability for developers without significant features being added to uds-cli that don't exist today. It is obvious that this could be a slippery slope to add other packages to test integration, but for now this is a a pretty important piece. As the need arises this similar pattern _could_ be used for other packages like Prometheus, but maybe its reasonable for that integration testing to happen in larger tests. That need can be evaluated on a case-by-case basis in the future and handled according to this pattern, objected to or may drive a discussion about adding a capability in uds-cli to better facilitate this process (like component deselection in a package) when that team has more capacity.

This wont be a breaking change for consumers of the core-istio package, but may require slight modifications to their use to remove keycloak if it causes resource issue on their test platform. We will mitigate this by accounting for this functionality in the uds-common tasks and prescribing their use when possible. 

- many other permutations of this basic idea, however most of those involved producing an additional release artifact which is gross.
- making uds-core have required: false and default: true and implementing a component deselection feature in uds-cli. This is not ideal due to the way we want to force users to always have the right tools deployed and also due to a lack of capacity on the uds-cli teams end while they focus on engine things. 

### Additional context

Discussed with most of the uds-core team to weigh pros and cons. Barring any implementation issues. Came to an agreement with @mjnagel on this path to move forward.

zachariahmiller:
## Description
This change replaces the core-istio bundle with a dev-slim bundle that includes keycloak in addition to istio and pepr.

## Related Issue
https://github.com/defenseunicorns/uds-core/issues/251


zachariahmiller:
Will update to fix linting errors and add a couple lines to to docs in a bit, but if there is other feedback to address lmk

MxNxPx (dismissed):

README.md
```diff
@@ -70,6 +72,16 @@ uds deploy k3d-core-istio-dev:0.15.1
 
 <!-- x-release-please-end -->
 
+Deploy just Istio and Pepr:
+
+<!-- x-release-please-start-version -->
+
+```bash
+uds deploy k3d-core-istio-dev:0.15.1 -p "uds-k3d-dev, init, core-istio"
```
mjnagel: Wondering if we should rename this since its no longer just "istio". I don't love changing things around but if we intend to extend this model in the future with other optionality we might want to change the name to `k3d-core-slim-dev` or something? Thoughts?

zachariahmiller: I have no issue renaming it I was trying to introduce as low of an impact of a change as possible and changing the name would be ceasing to publish the existing artifact.

zachariahmiller: Not sure why I didn't see this comment earlier today 😭

zachariahmiller: @mjnagel @jeff-mccoy @Racer159 I updated this PR based on the discussions that were had to move to a "slim dev" package/bundle with keycloak + istio. Is this more or less in line with the desired outcome?

README.md
```diff
@@ -60,16 +60,19 @@ uds deploy k3d-core-demo:0.16.1
 
 #### UDS Package Development
 
-In addition to the demo bundle, a [k3d-core-istio-dev bundle](./bundles/k3d-istio/README.md) also exists to work with UDS Core with only Istio & Pepr installed. Run the command below to use it:
+In addition to the demo bundle, a [k3d-core-istio-dev bundle](./bundles/k3d-istio/README.md) also exists to work with UDS Core with only Istio, Keycloak & Pepr installed. Run one of the commands below to use it.
```
mjnagel:
```suggestion
In addition to the demo bundle, a [k3d-slim-dev bundle](./bundles/k3d-slim-dev/README.md) also exists to work with UDS Core with only Istio, Keycloak & Pepr installed. Run the command below to use it.
```

tasks/deploy.yaml
```diff
@@ -15,8 +15,8 @@ tasks:
 
   - name: k3d-istio-bundle
```
mjnagel: nit: could change the name here, although I don't think this task is used anywhere?

mjnagel (changes requested):
We'll also need to update `release-please-config.json` to have the new filepaths (instead of `packages/istio/zarf.yaml` and `bundles/k3d-istio/uds-bundle.yaml`).

Otherwise this LGTM - validated the task changes for create/publish and all seems in line.

README.md
zachariahmiller: thank i knew i was going to miss something. Will update.

tasks/deploy.yaml
zachariahmiller: oof yeah okay. Funny thing is i changed the zarf package name in that task and still missed that.

zachariahmiller:
@mjnagel Updated with your suggested changes and re-requested review. Assuming CI passes, which it should, I will go ahead and update the pr description and title after this next meeting.

zachariahmiller:
@mjnagel i resolved your comments and updated the description and pr title.

mjnagel (commented):
(last one I promise) [this readme](https://github.com/defenseunicorns/uds-core/blob/optiona_keycloak_in_dev/packages/slim-dev/README.md) should be updated to the new name/description.

mjnagel (approved):

PR #276, opened 2024-03-20
feat: introduce sso secret templating

### LINKED ISSUE #263: SSO Secret Templating
**As** Ezra **I want** a way to define a secret template for Pepr to populate **so that** I can more easily create SSO secrets in ways that application Helm charts expect them

- **Given** I have an application configured to integrate with a `UDSPackage` CRD
- **When** I add a new secret template field to the `sso` section
- **And** I apply that CRD to my cluster
- **Then** Pepr templates the specified client information into the secret
- **And** saves that to the cluster ready for an application to pickup

We could create intermediary mechanisms for this but those would be likely to have race conditions, or other issues in implementation.

### Additional context
This should expose the values available in the sso secret today and should map custom secret keys in the generated secret to templated string values that represent the data that the application wants (i.e. json files or properties files).

See for an example of what GitLab expects: https://github.com/defenseunicorns/uds-package-gitlab/issues/60

jeff-mccoy:
This PR introduces SSO secret templating within the UDS Operator. This allows dynamic property bindings for the secret and custom file types such as nested JSON, YAML, etc. 

Fixes #263

Screenshot 2024-03-20 at 2 47 24 AM

### DIFF
--- a/src/pepr/operator/controllers/keycloak/client-sync.ts
+++ b/src/pepr/operator/controllers/keycloak/client-sync.ts
@@ -163,11 +163,14 @@ async function apiCall(sso: Partial<Sso>, method = "POST", authToken = "") {
 
 export function generateSecretData(client: Client, secretTemplate?: { [key: string]: string }) {
   if (secretTemplate) {
+    Log.debug(`Using secret template for client: ${client.clientId}`);
     // Iterate over the secret template entry and process each value
     return templateData(secretTemplate, client);
   }
 
   const stringMap: Record<string, string> = {};
+
+  Log.debug(`Using client data for secret: ${client.clientId}`);
 
   // iterate over the client object and convert all values to strings
   for (const [key, value] of Object.entries(client)) {
--- a/src/pepr/operator/controllers/keycloak/client-sync.ts
+++ b/src/pepr/operator/controllers/keycloak/client-sync.ts
@@ -3,6 +3,7 @@ import { K8s, Log, fetch, kind } from "pepr";
 import { UDSConfig } from "../../../config";
 import { Store } from "../../common";
 import { Sso, UDSPackage } from "../../crd";
+import { getOwnerRef } from "../utils";
 import { Client } from "./types";
 
 const apiURL =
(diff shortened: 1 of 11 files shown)


src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -86,9 +92,9 @@ async function syncClient(
       metadata: {
         namespace: pkg.metadata!.namespace,
         // Use the CR secret name if provided, otherwise use the client name
-        name: clientReq.secretName || name,
+        name: secretName || name,
       },
-      stringData: clientToStringmap(client),
+      stringData: generateSecretData(client, secretTemplate),
```
mjnagel: This currently acts as a patch, adding new keys to the secret but not deleting old ones if you edit a `secretTemplate` in your package CR. In some testing it seems like `data` (instead of `stringData`) operates more as expected, updating the secret contents and deleting old keys. Seems like we should switch to that?

mjnagel (commented):
Only other comment - would be good to have some tests on this functionality, I think @TristanHoladay has some stuff in flight to add there.

TristanHoladay:
As discussed earlier today with @mjnagel and @rjferguson21 , i wanted to test from the top entrypoint -- `keycloak()` -- all the way down, but the blockers hit with all the different kinds of mocking made it not worth it. so exported `generateSecretData()` to just catch any changes to the regex that could break expected output.

mjnagel (approved):

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -45,6 +52,7 @@ export async function purgeSSOClients(pkg: UDSPackage, refs: string[] = []) {
     const token = Store.getItem(ref);
     const clientId = ref.replace("sso-client-", "");
     if (token) {
+      Store.removeItem(ref);
       await apiCall({ clientId }, "DELETE", token);
```
bburky, line 56: Not sure which order of these is best, I think I would put `Store.removeItem(ref);` after `await apiCall()` and wrap both of these in a try/catch. It looks like there's no error handling at the moment and the error won't be caught.

Need to decide what behavior you want in case of Keycloak API errors. My suggestion would be to not drop the client from the store unless the DELETE succeeds. Unsure what's best, if it goes wrong someone may need to manually drop the item and I don't think it's really possible to hack the Pepr store directly. Deleting an item from the `.spec` manually is possible though, I don't think the client names are stored there though?

Issue #266, opened 2024-03-16
Testing of release workflows


mjnagel:
### Describe what should be investigated or refactored

Currently the release workflows are not fully tested until a release is cut. This has lead to two recent failed releases, 0.14.0 and 0.16.0. We should evaluate a way to safely test these workflows (without cutting releases/publishing "real" artifacts) so that we have more confidence in PRs that modify those.

### Links to any relevant code

https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/tag-and-release.yml

mjnagel:
@MxNxPx I see that you referenced this in your snapshot addition. Was just thinking/wondering, maybe the fact that our snapshot release process runs through the release workflow is sufficient here? Might still be helpful to have some sort of a test in PRs that modify the release workflows but it's a good first step.

mjnagel:
Going to close this out - it's not not an issue but through the snapshot workflow running nightly we have a partial test of this.

PR #278, opened 2024-03-20
ci: add daily or ad-hoc snapshot releases

MxNxPx:
## Description

add a daily (or ad-hoc) github workflow for `snapshot` releases.  this is to support the ability to publish a release of uds-core before official releases to enable folks to perform testing.

## Related Issue

Relates to https://github.com/defenseunicorns/uds-core/issues/266

## Type of change

## Checklist before merging


.github/workflows/experimental-release.yaml
```diff
@@ -0,0 +1,86 @@
...
+      #  run: |
+      #    uds run -f tasks/deploy.yaml k3d-standard-bundle --no-progress
+      #    uds run -f tasks/test.yaml validate-packages --no-progress
+
+      - name: Publish Standard Package
+        run: uds run -f tasks/publish.yaml standard-package --set FLAVOR=${{ matrix.flavor }} --set VERSION=experimental
+
+      - name: Publish Upstream Flavored Bundles
+        if: ${{ matrix.flavor != 'registry1'}}
+        run: uds run -f tasks/publish.yaml bundles --set VERSION=experimental
```
mjnagel: Would love if we could pull this part into a common workflow/action so that the "normal" publish and the experimental release run through the same code. Might've missed a discussion on this - tracking that currently the `latest` versioning would be a mess but that could be pulled out of the main publish so that it runs as a separate task, only on releases.

Thoughts on doing that? Hoping we could DRY this up so that we don't duplicate the code and I think that would also help align more with https://github.com/defenseunicorns/uds-core/issues/266 so that we are testing the exact same workflow rather than something that could diverge.

.github/workflows/snapshot-release.yaml
```diff
@@ -0,0 +1,88 @@
...
+      contents: write
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Update snapshot-latest tag
+        run: |
+          if gh release view snapshot-latest; then
+            gh release delete snapshot-latest --cleanup-tag -y
+          fi
+          gh release create snapshot-latest -t "snapshot-latest" -n "" --prerelease
```
mjnagel: Not sure I like having the release/git-tag named `snapshot-latest` but the published OCI artifact using the SHA. I'm not opposed to either one, but they aren't aligned then its not obvious to an end user what OCI artifact to consume to get the latest snapshot (unless I'm missing some way these are linked). _Yes_ the sha does show on the release page but if we don't make it very easy to just grab the snapshot I fear that few people would? This would also result in a lot of "orphaned" OCI artifacts that aren't referenced from a release.

Couple other options:
- go all in on `snapshot-latest`, tags+OCI: downside here is that the snapshot gets overwritten nightly/whenever, but not sure we care about that?
- go all in on sha, tags+OCI: downside of lots of releases, harder to clean these up + lots of release artifacts (OCI)
- hybrid...
  - oci tags: both sha and a `latest` equivalent
  - git tags/releases: only latest
  - provides history of all snapshot without cluttering releases, but does anyone care to pull an old snapshot if they don't see it in the release page?

I don't generally mind using the sha in our naming but I think we need to be careful about cluttering our releases and OCI artifacts by doing these snapshots and not cleaning them up in a responsible way.

mjnagel: Discussed briefly synchronously - path forward is along these lines:
- OCI publishing:
  - publish to different snapshot path
  - publish every snapshot (nightly/adhoc) without overwriting old ones
  - maybe some cleanup job to wipe out anything older than a week / the latest minor release (pending discussion on value/cost here)
  - tags for these would be sha+some date/timestamp
  - also tag the latest one with snapshot-latest
- Releases:
  - no releases per sha
  - single snapshot-latest release that is updated everytime we cut a new snapshot
  - basic release notes diffing what is included (diff between the snapshot and most recent real release?) - not 100% required, but nice to have

.github/workflows/experimental-release.yaml
MxNxPx: DRY is good!  updated to extract the `publish` pieces of the workflow and based on the `snapshot` input provided by the calling workflows (either typical or snapshot), then it will release accordingly.

.github/workflows/snapshot-release.yaml
MxNxPx: will plan to open a follow up issue (if this change is merged) for cleaning up snapshots.  otherwise, all these items should be addressed now in the latest changes.

.github/workflows/publish.yaml
```diff
@@ -46,6 +31,21 @@ jobs:
       - name: Login to registry1
         run: uds run registry-login --set REGISTRY=registry1.dso.mil --set REGISTRY_USERNAME=${{ secrets.IRON_BANK_ROBOT_USERNAME }} --set REGISTRY_PASSWORD=${{ secrets.IRON_BANK_ROBOT_PASSWORD }} --set REGISTRY_RETRY_INTERVAL=90
 
+      - name: Get snapshot version using git commit short sha and date
+        if: ${{ github.event.inputs.snapshot == 'true' }}
```
mjnagel: nit: `${{ inputs.snapshot }}` should work here (and other place the input is used).

https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs

> The workflow will also receive the inputs in the github.event.inputs context. The information in the inputs context and github.event.inputs context is identical except that the inputs context preserves Boolean values as Booleans instead of converting them to strings. The choice type resolves to a string and is a single selectable option.

Would want to validate but since this would preserve booleans we could do a true bool vs a string equality check.

.github/workflows/snapshot-release.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+on:
+  schedule:
+    - cron: "0 10 * * *"
+  workflow_dispatch:
+    inputs:
+      snapshot:
+        required: true
+        default: 'true'
+        type: string
```
mjnagel: I don't think we want this to be an input here anymore (it's always passed as true to the publish workflow on line 20).

.github/workflows/publish.yaml
```diff
@@ -46,6 +31,21 @@ jobs:
...
+          RELEASE_DATE=$(date +'%Y-%m-%d')
+          echo "SNAPSHOT_VERSION=${RELEASE_DATE}-${SHORT_SHA}" >> $GITHUB_ENV
+
+      - name: Set versions to snapshot
+        if: ${{ github.event.inputs.snapshot == 'true' }}
+        run: |
+          yq -ei '.metadata.version=env(SNAPSHOT_VERSION), (.packages[]|select(has("ref"))|select(.name=="k3d-core-demo")).ref=env(SNAPSHOT_VERSION)' bundles/k3d-standard/uds-bundle.yaml
+          yq -ei '.metadata.version=env(SNAPSHOT_VERSION), (.packages[]|select(has("ref"))|select(.name=="k3d-core-slim-dev")).ref=env(SNAPSHOT_VERSION)' bundles/k3d-slim-dev/uds-bundle.yaml
+          yq -ei '.metadata.version=env(SNAPSHOT_VERSION)' packages/standard/zarf.yaml
+          yq -ei '.metadata.version=env(SNAPSHOT_VERSION)' packages/slim-dev/zarf.yaml
```
mjnagel: Okay with this, but might make sense to move this to the snapshot specific file, and then pass an input into this workflow for the version?

.github/workflows/publish.yaml
```diff
@@ -71,12 +71,21 @@ jobs:
...
+        if: ${{ github.event.inputs.snapshot != 'true' && matrix.flavor != 'registry1' }}
         run: uds run -f tasks/publish.yaml bundles
 
+      - name: Publish Standard Package Snapshot
+        if: ${{ github.event.inputs.snapshot == 'true' }}
+        run: uds run -f tasks/publish.yaml standard-package --set FLAVOR=${{ matrix.flavor }} --set TARGET_REPO="ghcr.io/defenseunicorns/packages/uds/snapshots" --set VERSION="${SNAPSHOT_VERSION}"
+
+      - name: Publish Upstream Flavored Bundles Snapshot
+        if: ${{ github.event.inputs.snapshot == 'true' && matrix.flavor != 'registry1' }}
+        run: uds run -f tasks/publish.yaml bundles --set TARGET_REPO="ghcr.io/defenseunicorns/packages/uds/snapshots" --set VERSION="${SNAPSHOT_VERSION}"
```
mjnagel: In my head was hoping these would be identical (not need separate one for snapshot) but for that to work I think we'd need to pass in release-please version + default repo (and lose out on the defaults in the tasks). Not sure which route is better in this case, either works - so feel free to resolve this based on your preference.

mjnagel (commented):
Hopefully minor comments - think this is looking pretty good overall.

.github/workflows/publish.yaml
MxNxPx: @mjnagel i am leaning towards leaving as-is or perhaps we do the conditional inside the job step (setting the input to an env var) like this...

```yaml
      - name: Publish Standard Package
        run: |
          if [ "${SNAPSHOT}" = "true" ]; then
             uds run -f tasks/publish.yaml standard-package --set FLAVOR=${{ matrix.flavor }} --set TARGET_REPO="ghcr.io/mxnxpx/packages/uds/snapshots" --set VERSION="${SNAPSHOT_VERSION}"
          else
             uds run -f tasks/publish.yaml standard-package --set FLAVOR=${{ matrix.flavor }}
          fi

      - name: Publish Upstream Flavored Bundles
        if: ${{ matrix.flavor != 'registry1' }}
        run: |
          if [ "${SNAPSHOT}" = "true" ]; then
             uds run -f tasks/publish.yaml bundles --set TARGET_REPO="ghcr.io/mxnxpx/packages/uds/snapshots" --set VERSION="${SNAPSHOT_VERSION}"
          else
             uds run -f tasks/publish.yaml bundles
          fi
```

what do you think?

.github/workflows/publish.yaml
MxNxPx: @mjnagel i was struggling with this as well, but i think if we did this in the calling workflow we would need to checkout there to get the sha and then checkout again in this workflow?  i could be wrong, but that is why i landed on doing it all here.

.github/workflows/snapshot-release.yaml
MxNxPx: @mjnagel agreed - i will remove

.github/workflows/publish.yaml
MxNxPx: @mjnagel - good call, i will simplify to true boolean

.github/workflows/publish.yaml
mjnagel: Hm yeah good point. Checkout takes ~1sec in CI so it's not a big time waste, but unsure if double checkout has other consequences. Okay leaving for now and we could refactor down the road - at first just felt off to have this in our publish workflow but its all conditional-ed so it doesn't hurt.

.github/workflows/publish.yaml
mjnagel: My thought was more along the lines of:
```console
      - name: Publish Upstream Flavored Bundles
        if: ${{ matrix.flavor != 'registry1'}}
        run: uds run -f tasks/publish.yaml bundles --set TARGET_REPO="${{ inputs.repo }}" --set VERSION="${{ inputs.version }}"
```

And then passing inputs from each of the calling workflows. I think we can get the version from release-please outputs? And then hardcoding repo based on the calling workflow? Probably makes sense to remove the defaults from the tasks if we went this way.

.github/workflows/publish.yaml
MxNxPx: made the necessary changes

.github/workflows/snapshot-release.yaml
MxNxPx: removed

.github/workflows/publish.yaml
MxNxPx: went down this path, but the changes needed seemed a bit harder to follow.  maybe we leave it like this for now and revisit later?

mjnagel: Yep, as discussed offline going to press with current state and revisit for different refactoring in the future.

.github/workflows/publish.yaml
```diff
@@ -61,20 +64,29 @@ jobs:
       # Standard Package by default tests what's in the Istio Package
       - name: Test amd64 Bundle
         run: |
-          uds run deploy-standard-bundle --no-progress
+          uds run -f tasks/deploy.yaml k3d-standard-bundle --no-progress
```
mjnagel: Will want to switch this back to `uds run deploy-standard-bundle --no-progress` - otherwise the k3d image doesn't get populated from the default var in `tasks.yaml`.

mjnagel (commented):
LGTM otherwise

.github/workflows/publish.yaml
MxNxPx: @mjnagel - updated now

mjnagel (approved):

PR #284, opened 2024-03-22
fix: cacert fix and cleanup

chance-coleman:
## Description

* There was some extra fields accidentally brought into the cacert for keycloak istio gateways, removed those. 

* Task to copy cacert generated in identity-config into gateways doesn't work on non-mac users machines. Fix for that task and a change will be added to the identity side as well [in this branch](https://github.com/defenseunicorns/uds-identity-config/tree/smoke-test).

## Type of change

## Checklist before merging


src/keycloak/tasks.yaml
```diff
@@ -34,9 +34,8 @@ tasks:
     actions:
       - task: config:cacert
       - cmd: |
-          export cacert=$(cat cacert.b64)
-          yq e -i '.tls[0].cacert = env(cacert)' src/istio/values/config-tenant.yaml
-          yq e -i '.tls[0].cacert = env(cacert)' src/istio/values/config-admin.yaml
+          yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' src/istio/values/config-tenant.yaml tls_cacert.yaml
+          yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' src/istio/values/config-admin.yaml tls_cacert.yaml
```
mjnagel, line 34: Is it correct that this shouldn't be merged until an updated identity-config image/task is actually producing `tls_cacert.yaml`?

chance-coleman: @mjnagel correct. This task is not actively being used anywhere so we wouldnt see anything break. The identity change is captured in my registration integration test branch which is still wip.

PR #248, opened 2024-03-12
chore: refactor exemptions controller

TristanHoladay:
## Description
Refactor exemptions validator and controller.

## Related Issue

Resolves #249 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging

- [ ] Test, docs, adr added or updated as needed
- [ ] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md)(https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md#submitting-a-pull-request) followed

PR #290, opened 2024-03-26
fix: unwanted exemption deletions

### LINKED ISSUE #289: Exemption Matchers Deleting One Another
### Environment
Device and OS: Intel Nuc / Ubuntu Linux
App version: core-slim 0.17.0
Kubernetes distro being used: k3d

### Description
Updating an exemption CR with an exemption block that contains the same matcher as a previous block but is of a different kind (i.e. service versus pod) deletes the exemptions of the first block.

### Steps to reproduce
1. uds run test-single-package --set uds_pkg=neuvector
2. check `pepr.dev/v1/peprstores` and describe `pepr-uds-core-store`
3. update the exemption package of neuvector with an exemption containing same matcher as another block but use policy `DisallowNodePortServices` and set kind to `service`.
4. without tearing down the cluster, re-apply the neuvector package.
5. check `pepr.dev/v1/peprstores` and describe `pepr-uds-core-store`
6. the output from the store will show that the original exemptions are gone.

### Expected result
Adding an exemption block with same matcher name but different kind (from save exemption CR) should not overwrite other exemption blocks with that matcher name.

### Actual Result
Exemptions are overwritten or deleted.

### Severity/Priority
medium

TristanHoladay:
## Description
Fix the cases when matchers are deleted because they're nearly identical but not actually identical.

## Related Issue

Fixes #289 

### Additional Context
To help in bug fix and refactoring i merged #248 into this PR

## Type of change

## Checklist before merging


src/pepr/operator/controllers/exemptions/exemptions.ts
```diff
@@ -46,117 +34,127 @@ function addIfIncludesPolicy(
...
+  // Check stored matchers that have same owner ref as current UDSExemption
+  for (const pm of policyMatchers.filter(m => m.owner === ownerId)) {
+    let shouldBeRemoved = true;
+
+    // check if stored matcher exists in current list of UDSExemption matchers
+    for (const m of currExemptMatchers) {
+      if (isSame(pm, m)) {
+        shouldBeRemoved = false;
+      }
+    }
```
mjnagel, line 75: So I'm not sure this handles all the issues. Here's a minimal reproduction of the issue I'm seeing:
```yaml
apiVersion: uds.dev/v1alpha1
kind: Exemption
metadata:
  name: test
  namespace: uds-policy-exemptions
spec:
  exemptions:
    - policies:
        - DisallowHostNamespaces
      matcher:
        namespace: test
        name: "^test-pod.*"
    - policies:
        - DisallowPrivileged
      matcher:
        namespace: test
        name: "^test-pod.*"
    - policies:
        - RequireNonRootUser
      matcher:
        namespace: test
        name: "^test-pod.*"
    - policies:
        - DropAllCapabilities
      matcher:
        namespace: test
        name: "^test-pod.*"
```

If you apply this manifest the peprstore ends up only having the `DropAllCapabilities` exemption. This would be true anytime someone uses the same exact matcher in the same CR (i.e. anytime name, namespace, kind, and owner line up), but with multiple separate policy lists. I do think is a quite viable situation that we should support if we can - someone might want to separate out exemptions within the same CR so that they could put specific descriptions on each for example. Happy to help brainstorm how to handle this.

mjnagel: Extra context for why this might be useful - currently the promtail/neuvector exemptions share a single description for a number of policies that are exempted. In reality those descriptions only explain why one or two of the policies are exempted, not all of them. If we had an easy way to add a description per policy I think we'd be able to provide better info here (that is eventually pulled in UDS Engine?).

TristanHoladay: the mode for giving justifications is not the best atm for sure. I think our options right now are:

#### 1. make policies objects
```yaml
    - policies:
        - name: DropAllCapabilities
          reason: "cause i want to"
        - name: DropAllCapabilities
          reason: "cause i want to"
      matcher:
        namespace: neuvector
        name: "^neuvector-prometheus-exporter-pod.*"
      title: "neuvector-prometheus-exporter-pod"
```
pros:
* nicely attaches a justification to each policy for this matcher, so easier for reporting purposes.
* wouldn't require any controller logic changes  

cons: 
* doesn't actually fix the above issue. it does give less reason to structure an Exemption that way though.
* could make the exemption blocks messy and less readable if there are big justifications mixed into the policy list.
* have to update CRD / CR, which is not too big a deal since uds-core isn't widely used but is a consideration.

#### 2. Squish
Like you mentioned in chat, we could take all the duplicate matchers and combine them into a single exemption block at time of processing.

pros:
* no CRD / CR updates, just controller logic changes.
* handles the actual bug you found.

cons:
* might or might not solve the justification clarity issue and doesn't promote our pattern

#### 3. Make titles unique and required
If we make titles unique and required, we can do comparisons based of those and not just kind, ns, name, and owner. 

pros:
* users have to give titles, which helps with reporting
* we don't have to generate titles for them if they're missing
* would be a fix for the bug

cons:
* might just make controller logic more complicated not less.
* CRD / CR has to change a little bit
* might or might not solve the justification clarity issue and doesn't promote our pattern

#### 4. Do an additional check
Since `deleteIfPolicyRemoved()` is the culprit here, we could probably do a check for any duplicate matcher that does have the policy before deleting the matcher off the policy. 

pros:
* no CRD / CR changes needed
* handles bug

cons:
* could add a performance hit (probably negligible though)
* probably make the code more bleh
* might or might not solve the justification clarity issue and doesn't promote our pattern

I think i'm liking option 1 and / or option 2.

mjnagel: Opened an issue for this, I think we can resolve in a follow on.

mjnagel (approved):

PR #307, opened 2024-03-28
chore: support headless keycloak admin user

### LINKED ISSUE #293: Enable keycloak admin initialization for headless installs
### Overview
When installing core in restricted environments, the admin user / installer may not be able to easily access a browser from the same workstation where they have cluster admin (or zarf/uds) access.

An example of this might be a situation with a restricted workstation computer that allows remote access to a headless server (bastion host) for development actions. In that case, the user may be able to get to a terminal on the bastion, but may not nececarially be able to access a port-forwarded (or zarf-connected) connection to keycloak in order to perform the admin initialization.

### Proposed Solution
Allow a user to optionally generate a random password for keycloak, and require that it be updated on first login.

Per @jeff-mccoy:
1.  Add `insecureAdminPasswordGeneration` or something equally not-ideal as a boolean for the chart, and generate a secret only if true
1. Document the user should update as soon as possible and only use if unable to use zarf connect keycloak for some reason
1. Also document that the secret will no longer be valid/useful after the account is updated for security reasons.

### Additional context // Alternative solutions
(Also per @jeff-mccoy)

We can highly discourage it's use and strongly encourage password change on first login.  Here's the other things I found:
* With quarkus, they got rid of the ability to create an admin acct with the cli script
* You can get around this by running the server in the container on an alternate http port (works):  `KEYCLOAK_ADMIN=jeff KEYCLOAK_ADMIN_PASSWORD=pass /opt/keycloak/bin/kc.sh start-dev --http-port 30000`
* There's actually a hardcoded admin:admin cred if you enable a test property for quarkus, which is interesting
* Maybe could use lynx with zarf/connect  (untested),
get fancy with zarf + curl...but the admin page has a cookie with CSRF protection, so it get's weird (untested):
```
if [ -z "${USERNAME}" ]; then
            echo "USERNAME is not set"
            exit 1
          fi
          if [ -z "${PASSWORD}" ]; then
            echo "PASSWORD is not set"
            exit 1
          fi
          ./uds zarf connect keycloak --cli-only --local-port 45000
          curl -c cookies.txt -o welcome_page.html "http://127.0.0.1:45000/"
          stateChecker=$(grep 'name="stateChecker"' welcome_page.html | sed 's/.*value="\([^"]*\)".*/\1/')
          curl -b cookies.txt -d "username=${USERNAME}" -d "password=${PASSWORD}" -d "passwordConfirmation=$PASSWORD" -d "stateChecker=$stateChecker" "http://127.0.0.1:45000/"
          rm cookies.txt welcome_page.html
          ```

### PR DESCRIPTION
**Pull request by mjnagel**
## Description

Provides support for an initial admin user password to be generated. This is documented as the non-ideal path with instructions to change the password after initial login.

This toggle is exposed as a variable in the demo/dev bundles to support simpler CI/Dev testing workflows.

## Related Issue

Resolves https://github.com/defenseunicorns/uds-core/issues/293

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/README.md
+++ b/src/keycloak/chart/README.md
@@ -29,7 +29,7 @@ To be able to access the Keycloak administrative console you must create an init
 
 If you are unable to open a browser to create this user (i.e. a headless install), the chart provides values under `insecureAdminPasswordGeneration` which will generate the initial admin user for you and place the password in a secret (`keycloak-admin-password`) in cluster. To use this set `insecureAdminPasswordGeneration.enabled` to `true`, and (optionally) set `insecureAdminPasswordGeneration.username` to the desired username.
 
-Note that this is highly discouraged as it is often used as a shared admin account rather than a being tied to a specific user. If you do use this toggle, you must should immediately change the password after initial login so that your admin account information is not exposed in the cluster secret. This secret will only contain the initial password and will not be updated after you change the password during initial login.
+Note that this is highly discouraged as it is often used as a shared admin account rather than a being tied to a specific user. If you do use this toggle, you should immediately change the password after initial login so that your admin account information is not exposed in the cluster secret. This secret will only contain the initial password and will not be updated after you change the password during initial login.
 
 ## Why StatefulSet?
 
--- a/src/keycloak/chart/README.md
+++ b/src/keycloak/chart/README.md
@@ -12,7 +12,7 @@ For more information on Keycloak and its capabilities, see its [documentation](h
 
 When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off.
 
-#### Autoscaling
+### Autoscaling
 
 The example autoscaling configuration in the values file scales from three up to a maximum of ten Pods using CPU utilization as the metric. Scaling up is done as quickly as required but scaling down is done at a maximum rate of one Pod per five minutes.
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by mjnagel** [comment]
Need to validate some of the helm soup around secret upgrade (not changing it) and also want to think through the bundle config piece a bit more.

Edit: fixed helm soup, it shouldn't change the secret ever, which is what we want.

File: src/keycloak/chart/README.md [hunk 1]
```diff
@@ -23,6 +23,14 @@ autoscaling:
   enabled: true
 ```
 
+### Admin User Creation
+
+To be able to access the Keycloak administrative console you must create an initial admin user. You can do this using `zarf connect keycloak` which will launch a port-forward session to Keycloak. From there you will be prompted for the initial admin username and password.
+
+If you are unable to open a browser to create this user (i.e. a headless install), the chart provides values under `insecureAdminPasswordGeneration` which will generate the initial admin user for you and place the password in a secret (`keycloak-admin-password`) in cluster. To use this set `insecureAdminPasswordGeneration.enabled` to `true`, and (optionally) set `insecureAdminPasswordGeneration.username` to the desired username.
+
+Note that this is highly discouraged as it is often used as a shared admin account rather than a being tied to a specific user. If you do use this toggle, you must should immediately change the password after initial login so that your admin account information is not exposed in the cluster secret. This secret will only contain the initial password and will not be updated after you change the password during initial login.
```
jeff-mccoy:
```suggestion
Note that this is highly discouraged as it is often used as a shared admin account rather than a being tied to a specific user. If you do use this toggle, you should immediately change the password after initial login so that your admin account information is not exposed in the cluster secret. This secret will only contain the initial password and will not be updated after you change the password during initial login.
```

jeff-mccoy (dismissed):
Small typo

jeff-mccoy (approved):

PR #311, opened 2024-03-28
feat: pre core exemption handling

### LINKED ISSUE #272: Policy Exemptions For Packages Like Rook-Ceph
We have a [custom zarf init package that deploys rook-ceph and then initializes zarf](https://github.com/defenseunicorns/uds-capability-rook-ceph). If we then deploy uds-core on top of that we can no longer upgrade rook-ceph because pepr will block the recreation of the rook-ceph resources. There are no exemptions or mechanisms currently available to allow rook-ceph to deploy successfully on an upgrade.

We need to be able to exempt the policies being enforced in the screenshot below. There were talks on different ways to solve this problem. I will list some of that below in the alternative section.

Screenshot 2024-03-15 at 4 19 11 PM

- **Given** a cluster with rook-ceph deployed along with uds-core
- **When** an upgrade to rook-ceph is deployed
- **Then** pepr doesn't block the rook-ceph resources from upgrading and recycling

- An overall uds cluster config for some things as an alternative to our current env injection stuff
-- We could allow specific flags that generate exemptions for “rook mode” “eks mods” “Jeff’s an idiot mode” And those would generate the requisite exemptions
- We could throw a helm template somewhere in core to allow you to craft exemptions with helm values
- Allow for a namespace wide exemption?

### Additional context
This is blocking the ability to provide bundle upgrades that include rook-ceph upgrades.

TristanHoladay:
## Description
Introducing a UDS Exemption CRD package, which allows for applying UDSExemption CRs before uds-core has deployed. 
Also, integrating a uds-core scoped pre-core-exemption chart that allows users to pass exemptions (for things that might exist pre-core) through their bundle.

## Related Issue

resolves #272 

## Type of change

## Checklist before merging


TristanHoladay:
For testing manually:

1. Initial deploy and then cycle
   * `uds run crd-test`
   * after everything comes up delete podinfo pod and make sure it re-creates with the exemptions

2. Upgrade (podinfo exemption only)
   * leave cluster from previous test up (comment out the `setup:create-k3d-cluster` in `crd-test` in `tasks.yaml`
   * un-comment the commented exemption block for node port services in the podinfo config chart (`test-podinfo/config-chart/templates/uds-exemption.yaml`)
   * change the `service.type` in the podinfo values (`test-podinfo/chart/values.yaml`) to NodePort
   * re-run `uds run crd-test`
   * you should see the podinfo exemption update, as well as the pepr store, and the podinfo service to should recreated successfully as a NodePort service

3. Upgrade (exemption CRD)
    * you can leave the cluster up from the previous test
    * change the exemption CRD by adding a required field (e.g. make a new property in to .spec like .spec.test in `src/pepr/operator/crd/sources/exemption/v1alpha1.ts`)
    * update the podinfo exemption with new required field 
    * generate the crd via `uds run -f src/pepr/tasks/gen-crds` which will update the crd.yaml in the exemption crd zarf package
    * you'll need to clean up the fields from the generated crd in the crd.yaml. remove all fields that are unique to the applied instance (eg. uid, timestamps etc...)
    * re-run `uds run crd-test`
    * verify that the podinfo exemption is updated and that podinfo cycles

test-podinfo/zarf.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+components:
+  - name: demo-helm-charts
+    required: true
+    charts:
+      - name: uds-podinfo-config
+        namespace: podinfo
+        version: 0.1.0
+        localPath: config-chart
+      - name: podinfo-local
```
rjferguson21: Is it possible to use the OCI published version as opposed to pulling in the whole chart?

TristanHoladay:
example bundle using CRD package:

```yaml
kind: UDSBundle
metadata:
  name: example CRD

packages:
  - name: exemption-crd
    repository: ghcr.io/defenseunicorns/...
    ref: 0.20.0

  - name: init
    repository: ghcr.io/defenseunicorns/packages/init
    # renovate: datasource=github-tags depName=defenseunicorns/zarf versioning=semver
    ref: v0.33.1

  # has UDS Exemption CR
  - name: test-podinfo
    path: ../../build/
    ref: 0.1.0

  - name: core-slim-dev
    path: ../../build/
    ref: 0.20.0
 ```

example bundle using helm overrides:

```yaml
kind: UDSBundle
metadata:
  name: example helm overrides

packages:
  - name: init
    repository: ghcr.io/defenseunicorns/packages/init
    # renovate: datasource=github-tags depName=defenseunicorns/zarf versioning=semver
    ref: v0.33.1

  # UDS Exemption CR does not live here but is passed to core
  - name: test-podinfo
    path: ../../build/
    ref: 0.1.0

  - name: core-slim-dev
    path: ../../build/
    ref: 0.20.0
    # x-release-please-end
    overrides:
      pre-core-exemptions:
        pre-core-exemptions:
          values:
            - path: enabled
              value: true
            - path: exemptions
              value: |
                    exemptions:
                      - policies:
                          - DisallowPrivileged
                          - RequireNonRootUser
                          - DropAllCapabilities
                        title: "podinfo1"
                        matcher:
                          namespace: podinfo
                          name: "^podinfo.*"
                          test: 1
                      - policies:
                          - DisallowNodePortServices
                        title: "podinfo2"
                        matcher:
                          namespace: podinfo
                          name: "^.*-local.*"
                          kind: service
                          test: 2
```

packages/exemption-crd/zarf.yaml
```diff
@@ -0,0 +1,16 @@
+kind: ZarfPackageConfig
+metadata:
+  name: uds-core-exemption-crd
+  yolo: true
+  # x-release-please-start-version
+  version: 0.20.0
```
mjnagel:
```suggestion
  version: 0.21.0
```

packages/exemption-crd/zarf.yaml
```diff
@@ -0,0 +1,16 @@
+kind: ZarfPackageConfig
+metadata:
+  name: uds-core-exemption-crd
```
mjnagel: I'm tempted to just call this "uds-crds". I don't know if we want to add the `Package` CRD here right now, but in theory this is more flexibility to provide other pre-core custom resource down the line without need to rename. Maybe that's more confusing if we aren't including the `Package` CRD right now though....thoughts?

src/pre-core-exemptions/README.md
```diff
@@ -0,0 +1,53 @@
+# Pre Core Exemptions
+
+This package serves as a way for users deploying uds-core to pass exemptions to core for things that are deployed before core.
+
+For example, if a team is deploying a bundle containing a custom init package that has rook-ceph, the first time it
+deploys everything will be fine, but if it cycles for whatever reason once core is deployed then rook-ceph will be denied by Pepr policies. Thus an exemption CR is needed.
+The problem, though, is the init package can't deploy an exemption resource when the exemption CRD has not yet been deployed by core.
```
mjnagel:
```suggestion
For instance, a team is deploying a bundle that includes a custom init package with rook-ceph. During install, everything goes smoothly since Pepr's validation is not present. However, during an upgrade after the core deployment, Pepr policies might deny new rook-ceph pods. 
```

bit of a nit but this wording was hard for me. slight chatgpt on this so feel free to tweak.

src/pre-core-exemptions/README.md
```diff
@@ -0,0 +1,53 @@
+# Pre Core Exemptions
+
+This package serves as a way for users deploying uds-core to pass exemptions to core for things that are deployed before core.
```
mjnagel:
```suggestion
This package serves as a way for users deploying uds-core to create exemption resources for things that are deployed before core.
```
Maybe? Technically is a "pass through" but I think its more clear to just say this is a way to create exemptions? No strong opinion here.

.github/bundles/exemption/uds-bundle.yaml
```diff
@@ -0,0 +1,75 @@
+kind: UDSBundle
```
mjnagel: Don't have a great answer here, but wondering if we can reduce duplication and reuse our existing dev or demo bundle here. I don't think it would hurt to expose `exemptions` as a variable in either/both of those, so then it's a question of whether testing with a post-core podinfo deployment works here. I think I'm _okay_ with that solution since this exemption passthrough is not strictly "pre-core" and it would still validate our chart/override works, thoughts? Hoping to find a nice middle ground where we could test but not have to make a new CI bundle just for that...maybe this would also help some of the conditional logic?

.github/bundles/exemption/uds-bundle.yaml
```diff
@@ -0,0 +1,75 @@
...
+metadata:
+  name: k3d-core-slim-dev
+  description: A UDS bundle for deploying Istio from UDS Core on a development cluster
```
mjnagel: If we do keep this bundle, probably need some metadata updates 😄

mjnagel (changes requested):
Hopefully small comments overall - I think this is a good 2-pronged approach to provide solves for this issue.

packages/exemption-crd/zarf.yaml
TristanHoladay: i like that idea. would actually have helped with an issue i ran into yesterday testing something. we could go ahead and include the Package CRD (no reason not too that i can think of).

.github/bundles/exemption/uds-bundle.yaml
TristanHoladay: hmmm, yeah i like getting rid of duplication / an extra bundle to maintain. guessing the post-core podinfo deploy would just be in the validation task?

mjnagel: Yeah that was my thought. Could be done similar to what we do with the test package in CI - https://github.com/defenseunicorns/uds-core/tree/main/src/test

mjnagel:
Going to close this PR for now - we need to reevaluate if there's still a need here and how to prioritize if there is. Noting that there are a few known ways to accomplish exemptions for things that come before core:
- Exemptions can be added as a standalone "post-core" package, which ensures they can be re-scheduled/upgraded
- Pre-core items are typically "k8s/infra" related and could/should be deployed in `kube-system` oftentimes (which is exempted from any pepr policies)
- Exemptions can be added directly in pre-core packages, wrapped with helm capability checks which ensures they are present at an upgrade

PR #330, opened 2024-04-05
fix: drop path normalization to MERGE_SLASHES to allow apps to handle encoded slashes

### LINKED ISSUE #288: Istio DECODE_AND_MERGE_SLASHES option breaks GitLab
### Steps to reproduce
1. Make a request to gitlab like `api/v4/projects/namespace%2Fproject` through the istio tenant gateway

### Expected result

Project info is received

### Actual Result

Gitlab serves a 404 since it's expecting the url to not have been decoded

image

image

### Severity/Priority

Medium: at least gitlab web ide is broken, but any other api requests that need encoded slashes are too

### Additional Context

Racer159:
## Description

PR to address `DECODE_AND_MERGE_SLASHES` causing issues in certain applications.  Gives the ability to selectively turn this off in a namespace with the `UDSPackage` CR.

## Related Issue

Fixes #288

## Type of change

## Checklist before merging

- [X] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md)(https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md#submitting-a-pull-request) followed


Racer159:
Post changes with retain slashes set to true:
image
Pre changes (and without retain slashes):
image

(Decided to leave path normalization on except for decoding slashes since it shouldn't really hurt anything - I don't know of any apps that expect double slashes or other weird slash configs)

src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -20,9 +20,9 @@ export async function enableInjection(pkg: UDSPackage) {
   const annotations = sourceNS.metadata?.annotations || {};
   const pkgKey = `uds.dev/pkg-${pkg.metadata.name}`;
 
-  // Save the original value of the istio-injection label only if it's not already set
-  if (!annotations[injectionLabel]) {
-    annotations[injectionAnnotation] = labels[injectionLabel] || "non-existent";
+  // If the original namespace injectionLabel did not exist, mark it as non-existent in the annotations
+  if (!labels[injectionLabel]) {
```
Racer159: changed this because it wasn't connected to anything on the other end - can change it back if that was intended

jeff-mccoy (dismissed):
Holding this one hostage until we discuss further, I really don't think we should do this and would rather remove the [Istio Path Normalization](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-path_normalization) than add all this complexity.

Racer159:
> Holding this one hostage until we discuss further, I really don't think we should do this and would rather remove the [Istio Path Normalization](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-path_normalization) than add all this complexity.

Yep got no problem with that, was fun learning more about Istio anyway :sweat_smile:

bburky:
@jeff-mccoy 
> Holding this one hostage until we discuss further, I really don't think we should do this and would rather remove the [Istio Path Normalization](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig-path_normalization) than add all this complexity. 

Ok.

We'll likely be unable to use path-based AuthorizationPolices with any confidence across all of UDS without careful review of each app individually. I would have liked to have had the ability to use AuthorizationPolices to apply CVE mitigations on UDS apps, see https://github.com/defenseunicorns/uds-core/issues/288#issuecomment-2023640139 for more context. In theory we can still do this, but we'll have to review each app's URL decoding implementation before doing so.

For Keycloak specifically, `MERGE_SLASHES` _seems_ to be sufficient. I did some _light_ manual testing, and `//admin` is still blocked (redirected) with `MERGE_SLASHES`, `/%2fadmin` is now a keycloak error (good, not allowed through). A cursory review of the source code (in vertx via quarkus via keycloak), suggests only alpha, didgit, hyphen, period, underscore and tilde characters are URL decoded: 
https://github.com/eclipse-vertx/vert.x/blob/5832acf2fb22ed0301c396d4c3a7c316005e2939/src/main/java/io/vertx/core/http/impl/HttpUtils.java#L317-L329

At least `MERGE_SLASHES` level of normalization is required if we use path-based AuthorizationPolicies on Keycloak,  because Keycloak merges `//` into `/` (example: `//admin` is a valid URL). This is because Keycloak implements this non-standard (but common, as they note) normalization:
https://github.com/eclipse-vertx/vert.x/blob/5832acf2fb22ed0301c396d4c3a7c316005e2939/src/main/java/io/vertx/core/http/impl/HttpUtils.java#L390-L393

src/pepr/operator/controllers/istio/injection.ts
mjnagel: I got a big confused on the logic initially, but I think the original code seems right? Intent was to be able to revert the label back to its original value [here](https://github.com/defenseunicorns/uds-core/blob/9518634b24f2d5c285e598f8620849bbc6288ba4/src/pepr/operator/controllers/istio/injection.ts#L69-L77).

What we want is to revert the label to its original value if set OR delete it if it was non-existent. So I think we still need to set the annotation to `labels[injectionLabel]` if that has a value, or `non-existent` if it doesn't. TLDR think we can revert this change and it should work as intended?

mjnagel (approved):
LGTM! Thanks for the fix here!

PR #341, opened 2024-04-12
feat: keycloak PVC customization

blancharda:
## Overview
Add values file options and templating for persistence to allow tweaking PVC setting in keycloak.

## Type of change

### DIFF
--- a/src/keycloak/chart/README.md
+++ b/src/keycloak/chart/README.md
@@ -10,7 +10,7 @@ For more information on Keycloak and its capabilities, see its [documentation](h
 
 ### Dev Mode
 
-When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off. Devmode also levarages PVCs by default for `data` and `themes`.
+When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off. Devmode also leverages PVCs by default for `data` and `themes`.
 
 ### Autoscaling
 
--- a/src/keycloak/chart/README.md
+++ b/src/keycloak/chart/README.md
@@ -10,7 +10,7 @@ For more information on Keycloak and its capabilities, see its [documentation](h
 
 ### Dev Mode
 
-When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off.
+When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off. Devmode also levarages PVCs by default for `data` and `themes`.
 
 ### Autoscaling
 
--- a/src/keycloak/chart/templates/pvc.yaml
+++ b/src/keycloak/chart/templates/pvc.yaml
@@ -12,12 +12,8 @@ spec:
   resources:
     requests:
       storage: {{ .Values.persistence.providers.size | quote }}
-  {{ if .Values.persistence.storageClassName }}
-  {{- if (eq "-" .Values.persistence.storageClassName) }}
-  storageClassName: ""
-  {{- else }}
+  {{- if .Values.persistence.storageClassName }}
   storageClassName: "{{ .Values.persistence.storageClassName }}"
-  {{- end }}
   {{- end }}
 ---
 {{- end }}
@@ -36,12 +32,8 @@ spec:
   resources:
     requests:
       storage: {{ .Values.persistence.conf.size | quote }}
-  {{ if .Values.persistence.storageClassName }}
-  {{- if (eq "-" .Values.persistence.storageClassName) }}
-  storageClassName: ""
-  {{- else }}
+  {{- if .Values.persistence.storageClassName }}
   storageClassName: "{{ .Values.persistence.storageClassName }}"
-  {{- end }}
   {{- end }}
 ---
 {{- end }}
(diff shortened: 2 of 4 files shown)


src/keycloak/chart/templates/pvc.yaml
```diff
@@ -22,8 +82,15 @@ metadata:
...
   resources:
     requests:
-      storage: 512Mi
+      storage: {{ .Values.persistence.themes.size | quote }}
+  {{ if .Values.persistence.storageClassName }}
+  {{- if (eq "-" .Values.persistence.storageClassName) }}
+  storageClassName: ""
+  {{- else }}
+  storageClassName: "{{ .Values.persistence.storageClassName }}"
+  {{- end }}
```
mjnagel: Think I'm a bit confused on the template here - what is the intent of the `-` check? `storageClassName: ""` would be equivalent to no storage class I believe? Is no storage class an actual need here, or can we just simplify the template to:
```suggestion
  {{ if .Values.persistence.storageClassName }}
  storageClassName: "{{ .Values.persistence.storageClassName }}"
  {{- end }}
```

src/keycloak/chart/README.md
```diff
@@ -10,7 +10,7 @@ For more information on Keycloak and its capabilities, see its [documentation](h
 
 ### Dev Mode
 
-When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off.
+When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off. Devmode also levarages PVCs by default for `data` and `themes`.
```
mjnagel:
```suggestion
When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off. Devmode also leverages PVCs by default for `data` and `themes`.
```

src/keycloak/chart/values.yaml
```diff
@@ -91,6 +91,24 @@ resources:
     cpu: "1"
     memory: "1Gi"
 
+# Persistence settings for PVC management (when enabled=false, an emptyDir will be used)
+# `devMode` overrides the enabled flag for `data` and `themes` to preserve legacy behavior
```
```suggestion
# Persistence settings for PVC management (when enabled=false, an emptyDir will be used)
# `devMode` overrides the enabled flag for `data` and `themes` to allow hot-reloads of the theme/plugin in k3d
```

Also not sure if there's a better way to capture why you might want to toggle this on (either here or in the docs). At first glance I would be afraid someone would see persistence for data and think of course they should turn it on, when really they probably don't need it.

mjnagel (commented):
Overall LGTM - couple of small comments.

src/keycloak/chart/templates/pvc.yaml
blancharda: It's probably overcomplicated.. but there is technically a difference between an emptystring storageClassName and a nil/non-existent one. In theory, you might want to explicitly set `storageClassName: ""` to prevent automatic provisioning, which is what the special `"-"` case is meant to handle. Not providing `storageClassName` at all results in using the default SC.

I'm not sure I've ever found myself in that particular situation, but it's a fairly common [pattern I've seen elsewhere](https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_storage.tpl#L15).

mjnagel: Yep - was tracking the difference just wasn't sure if we needed to add that support here (we already assume storage class for other charts that I don't think support this). All good with leaving though if you think it's a value add.

src/keycloak/chart/values.yaml
blancharda: I was intentionally a little vague about it, as I'm not sure we necessarily want to encourage what I'm doing 😅 
But it may be worth adding a little more detail regarding the potential need to share/load data from other pods or workloads.

You raise another good point though.. 
When keycloak is running in HA -- do any of these PVCs need to be updated to read-write-many, or is everything hotloaded from the init container?

I'm not sure I'm familiar enough with KC operation to answer that one..

mjnagel: All should be loaded from the config image:
- data: import of realm, no additional data gets stored there after
- conf: truststore, always loaded from image
- theme: loaded from image
- provider: plugin jar, also from image

None of them would get used to store additional data so as long as the init image has the above four pieces (it would if built following the indentity-config process) then the emptydir volumes would always get loaded on startup, without the need for a PVC.

mjnagel: Your question is basically why I was hoping we could document something to express that most people don't need this 😅. Persistence comes across as something that you would want, but in this case it's not necessary (and potentially could cause probably with multi-replica if it wasn't RWX volumes?).

src/keycloak/chart/templates/pvc.yaml
blancharda: It's probably overkill in this case.. #deletedeletedelete 😉

src/keycloak/chart/values.yaml
blancharda: Yea, I think we can add a cautionary note to the README that the default config should be sufficient for MOST deployments and you should really only need to tweak persistence for custom changes/situations.

mjnagel (approved):

PR #351, opened 2024-04-17
feat: enable sso for neuvector

### LINKED ISSUE #91: Setup SSO for NeuVector, disable basic auth

Deploying UDS Core by default provides NeuVector with a default `admin` user (password `admin`).

Automate the SSO setup for NeuVector using the UDS Package CR, and (if possible) disable basic/local auth. Map existing groups/roles to NeuVector roles as applicable (admin/auditor should get different permissions in NeuVector).

Note: The neuvector prometheus exporter does require auth as of last check, so we will need to identify a way to handle this NPE account, or maintain basic auth for metrics only (read-only).

rjferguson21:
## Description
Adds Keycloak client creation and SSO configuration for Neuvector.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/91

## Type of change

## Checklist before merging


src/neuvector/values/values.yaml
```diff
@@ -13,32 +13,19 @@ controller:
 
   apisvc:
     type: ClusterIP
-  secret:
-    enabled: true
+  configmap:
+    enabled: false
```
mjnagel: This should be true for the configmap I think? To make sure we load in our telemetry/scan config?

src/neuvector/chart/templates/uds-package.yaml
```diff
@@ -4,6 +4,25 @@ metadata:
...
   namespace: {{ .Release.Namespace }}
 spec:
+  sso:
+    - name: Neuvector
+      clientId: uds-core-admin-neuvector
+      redirectUris:
+        - "https://neuvector.admin.{{ .Values.domain }}/openId_auth"
+      secretName: neuvector-secret
+      secretTemplate:
+        oidcinitcfg.yaml: |-
```
mjnagel, line 14: Did you see any weird timing issues here/potential for them (i.e. similar to grafana failing to mount, but worse in this case since the volume is optional)? 

Another thing I'm thinking is that we no longer benefit from the upstream checksum on this secret to cycle the deployment. I'm not sure of a good answer, but it would be nice to ensure the deployment rolls if this secret changes.

rjferguson21:
@mjnagel FYI I tested this trying to use the internal address for keycloak and Neuvector complains that the issuer is not consistent between the configured issuer and the issuer from the `.well-known` endpoint.

src/neuvector/chart/templates/uds-package.yaml
```diff
@@ -4,6 +4,25 @@ metadata:
   name: neuvector
   namespace: {{ .Release.Namespace }}
 spec:
+  sso:
+    - name: Neuvector
+      clientId: uds-core-admin-neuvector
+      redirectUris:
+        - "https://neuvector.admin.{{ .Values.domain }}/openId_auth"
+      secretName: neuvector-secret
```
mjnagel, line 12: Not a real comment, but I think this makes sense to use instead of `neuvector-init` because it allows us to continue using `secret.enabled` if we needed it for anything else. Kinda weird that the controller even mounts this since its not created anywhere in the chart.

rjferguson21: Oh, good catch.  This works but I don't think I consciously chose that secret name.

src/neuvector/values/values.yaml
rjferguson21: Updated.

src/neuvector/chart/templates/uds-package.yaml
rjferguson21: Added a patch to set a checksum annotation for the deployment.

mjnagel (dismissed):
LGTM - noting that I opened https://github.com/defenseunicorns/uds-core/issues/354 as a follow on/more generic way to approach the secret checksum thing.

src/neuvector/common/zarf.yaml
```diff
@@ -43,3 +43,7 @@ components:
                 name: neuvector
                 namespace: neuvector
                 condition: "'{.status.phase}'=Ready"
+          - description: Annotate Controller deployment for Neuvector SSO secret
+            cmd: |
+              SSO_CHECKSUM=$(./zarf tools kubectl get secret neuvector-secret -o json -n neuvector | sha256sum | cut -d' ' -f1)
+              ./zarf tools kubectl patch deployment neuvector-controller-pod -n neuvector -p "{\"spec\":{\"template\":{\"metadata\":{\"annotations\":{\"neuvector-uds-secret\":\"$SSO_CHECKSUM\"}}}}}"
```
mjnagel: very nit but...
```suggestion
              ./zarf tools kubectl patch deployment neuvector-controller-pod -n neuvector -p "{\"spec\":{\"template\":{\"metadata\":{\"annotations\":{\"checksum/uds-sso-secret\":\"$SSO_CHECKSUM\"}}}}}"
```
is a bit more in line with the upstream checksum annotation/standard ones I've seen.

mjnagel (approved):

PR #352, opened 2024-04-17
feat: add keycloak sso realm values

chance-coleman:
## Description
Add values for configuring google sso.

Three values:
* Enable SSO
* SSO Client ID
* SSO Client Secret

Default values create the Google IDP, however it is disabled and if enabled would require client id and client secret to be added to work properly.

Values put into secret and then create environment variable from those which are referenced in the uds-identity-config realm.json. A PR will be created in uds-identity-config as well for that side of this process.

## Related Issue
[
uds-identity-config PR](https://github.com/defenseunicorns/uds-identity-config/pull/58)

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/secret-kc-realm.yaml
+++ b/src/keycloak/chart/templates/secret-kc-realm.yaml
@@ -9,8 +9,8 @@ type: Opaque
 data:
   {{- range $key, $value := .Values.realmInitEnv }}
   {{- if eq (typeOf $value) "bool" }}
-  realm_{{ $key }}: {{ toString $value | b64enc }}
+  REALM_{{ $key }}: {{ toString $value | b64enc }}
   {{- else }}
-  realm_{{ $key }}: {{ $value | b64enc }}
+  REALM_{{ $key }}: {{ $value | b64enc }}
   {{- end }}
   {{- end }}
--- a/src/keycloak/chart/templates/secret-kc-realm.yaml
+++ b/src/keycloak/chart/templates/secret-kc-realm.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "keycloak.fullname" . }}-realm-env
+  namespace: {{ .Release.Namespace }}  
+  labels:
+    {{- include "keycloak.labels" . | nindent 4 }}
+type: Opaque
+data:
+  UDS_GOOGLE_SSO_ENABLED: {{ .Values.google.sso.enabled | toString | b64enc }}
+  UDS_GOOGLE_SSO_CLIENT_ID: {{ .Values.google.sso.client_id | b64enc }}
+  UDS_GOOGLE_SSO_CLIENT_SECRET: {{ .Values.google.sso.client_secret | b64enc }}
--- a/src/keycloak/chart/templates/secret-kc-realm.yaml
+++ b/src/keycloak/chart/templates/secret-kc-realm.yaml
@@ -7,6 +7,10 @@ metadata:
     {{- include "keycloak.labels" . | nindent 4 }}
 type: Opaque
 data:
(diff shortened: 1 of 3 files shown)


src/keycloak/chart/values.yaml
```diff
@@ -15,6 +15,13 @@ domain: "###ZARF_VAR_DOMAIN###"
 # The primary Keycloak realm
 realm: uds
 
+# Google SSO Values
```
rjferguson21: One thought I am having is would this be more flexible if we changed this to accept arbitrary keys/values?

If we did it that way we would not have to update the chart Values if we need to add other values or downstream users wanted to define their own templated env vars in their realm.json (assuming they are using a custom  uds-identity-config)?

src/keycloak/chart/values.yaml
```diff
@@ -15,6 +15,10 @@ domain: "###ZARF_VAR_DOMAIN###"
 # The primary Keycloak realm
 realm: uds
 
+# UDS Identity Config Environment Variables
+identityConfigEnv:
+  ssoEnabled: false
```
mjnagel: nit: this name feels weird. assume this is tied to the specific google idp enabled flag on the realm side? Maybe we can name more specifically?

chance-coleman: At the moment its specific to google idp, however if we were to add a different idp it would still be required..although if we had two idp's and only wanted one then we'd need separate values so maybe it's better to be specific.

src/keycloak/chart/values.yaml
```diff
@@ -15,6 +15,13 @@ domain: "###ZARF_VAR_DOMAIN###"
 # The primary Keycloak realm
 realm: uds
 
+# UDS Identity Config Environment Variables. More info here: https://github.com/defenseunicorns/uds-identity-config/blob/main/docs/CUSTOMIZE.md#override-default-realm
+realmInitEnv:
+  googleIdpEnabled: false
+  # Other UDS Identity Config fields that will be used in the realm.json initalization of keycloak
+  # googleIdpClientId: ""
+  # googleIdpClientSecret: ""
```
rjferguson21: IMO I think it would be best if these are named like environment variables.
```suggestion
# UDS Identity Config Environment Variables. More info here: https://github.com/defenseunicorns/uds-identity-config/blob/main/docs/CUSTOMIZE.md#override-default-realm
realmInitEnv:
  GOOGLE_IDP_ENABLED: false
  # Other UDS Identity Config fields that will be used in the realm.json initalization of keycloak
  # GOOGLE_IDP_CLIENTID: ""
  # GOOGLE_IDP_CLIENT_SECRET: ""
```

src/keycloak/chart/templates/secret-kc-realm.yaml
```diff
@@ -0,0 +1,16 @@
...
+metadata:
+  name: {{ include "keycloak.fullname" . }}-realm-env
+  namespace: {{ .Release.Namespace }}  
+  labels:
+    {{- include "keycloak.labels" . | nindent 4 }}
+type: Opaque
+data:
+  {{- range $key, $value := .Values.realmInitEnv }}
+  {{- if eq (typeOf $value) "bool" }}
+  realm_{{ $key }}: {{ toString $value | b64enc }}
```
rjferguson21:
```suggestion
  REALM_{{ $key }}: {{ toString $value | b64enc }}
```

src/keycloak/chart/templates/secret-kc-realm.yaml
```diff
@@ -0,0 +1,16 @@
...
+  {{- else }}
+  realm_{{ $key }}: {{ $value | b64enc }}
```
rjferguson21:
```suggestion
 REALM_{{ $key }}: {{ $value | b64enc }}
```

rjferguson21 (approved):

mjnagel (approved):

PR #356, opened 2024-04-19
feat: add `expose` service entry for internal cluster traffic

Racer159:
## Description

This adds a service entry to allow traffic to stay inside the cluster and enable things like proper network policies when clients need to access this endpoint.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging

- [X] [Contributor Guide Steps](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md)(https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md#submitting-a-pull-request) followed


chance-coleman (dismissed):

mjnagel (commented):
I think this seems okay in general. Broad strokes though, I think this could be done for every virtualservice (and done via pepr generating it based on the `spec.expose` entry in the Package CR)?

I'm not opposed to rolling forward with just this one, but if we are seeing issues due to cluster -> external LB -> cluster it may make sense to just cover all the things rather than adding one specific one for Keycloak.

Racer159:
two ServiceEntries made for `keycloak`:
image

tasks.yaml
```diff
@@ -30,6 +30,9 @@ tasks:
       - description: "Deploy the Istio source package with Zarf Dev"
         cmd: "uds zarf dev deploy src/istio --flavor ${FLAVOR}"
 
+      - description: "Deploy the Prometheus-Stack source package with Zarf Dev to only install the CRDs"
+        cmd: "uds zarf dev deploy src/prometheus-stack --flavor crds-only"
+
```
mjnagel, line 36: Not sure I understand this addition - was this something while this was WIP or am I missing a connection somewhere?

src/pepr/operator/controllers/istio/service-entry.ts
```diff
@@ -0,0 +1,114 @@
...
+} from "../../crd";
+import { getOwnerRef, sanitizeResourceName } from "../utils";
+
+/**
+ * Creates a ServiceEntry for each exposed service in the package
+ *
+ * @param pkg
+ * @param namespace
+ */
+export async function serviceEntry(pkg: UDSPackage, namespace: string) {
```
mjnagel: This has a lot of overlap with the `virtualService` loop we do over `expose`. If we could consolidate to a single loop over `expose` that would help with the maintenance of this going forward I think (i.e. if we changed host/gateway in the spec somehow we would only need to update in one spot).

No overall objections to splitting into a separate file+function for the payload construction, but I think we can draw out some commonality here.

src/pepr/operator/controllers/istio/service-entry.ts
```diff
@@ -0,0 +1,114 @@
...
+        resolution: IstioResolution.DNS,
+        ports: [serviceEntryPort],
+        endpoints: [serviceEntryEndpoint],
+      },
+    };
+
+    Log.debug(payload, `Applying ServiceEntry ${payload.metadata?.name}`);
+
+    // Apply the ServiceEntry and force overwrite any existing policy
+    await K8s(IstioServiceEntry).Apply(payload, { force: true });
```
mjnagel: I wouldn't say this is strictly required here but if we could split out the object creation from things that hit the kubeapi that would help with being able to unit test these pieces (without needing to mock KFC things).

See the [monitoring split](https://github.com/defenseunicorns/uds-core/blob/add-keycloak-service-entry/src/pepr/operator/controllers/monitoring/service-monitor.ts) for an example of this - this is basically the same as @rjferguson21 comment on that PR - https://github.com/defenseunicorns/uds-core/pull/256#discussion_r1577260406 . If you could do this and add a test I think that would help.

src/pepr/operator/controllers/istio/service-entry.ts
```diff
@@ -0,0 +1,114 @@
...
+  // Iterate over each exposed service
+  for (const expose of exposeList) {
+    const { gateway = Gateway.Tenant, host } = expose;
+
+    const name = generateSEName(pkg, expose);
+
+    // If we have already made a ServiceEntry with this name, skip (i.e. if advancedHTTP was used)
+    if (serviceEntryNames.get(name)) {
+      continue;
+    }
```
mjnagel: I think this works but want to put a placeholder comment here to remind myself to validate we're not missing any edge cases. fwiw the logic looks right to me - in general we should be able to skip duplicates since nothing could share the same fqdn but be on a different gateway...

tasks.yaml
Racer159: Nope, `dev-setup` is broken on main - Pepr will fail for anything needing a service monitor (like `keycloak`) - this abuses flavors to just put the crds in the cluster so Pepr can continue with it's mutations.

mjnagel: Ah, okay tracking. I missed the first time around that this was an addition to the dev-setup task. Honestly might be good to revisit in general and see what makes sense for dev-setup...maybe good to look at this the same way we do slim-dev (i.e. adding keycloak to it as well). Probably a separate discussion there...

I'm hesitant to add a `--flavor` that doesn't exist, at least without adding a clarifying/explanatory comment on what is happening here. This would behave the same if we just dropped `--flavor` entirely, correct? I don't mind using/abusing flags like this but I don't think it's super apparent why it's there at first glance.

Racer159: Correct - I added the flavor flag to more indicate that it is an abuse but will clarify that in a comment as well

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -0,0 +1,93 @@
+import { K8s, Log } from "pepr";
+
+import { IstioVirtualService, IstioServiceEntry, UDSPackage } from "../../crd";
+import { getOwnerRef } from "../utils";
+import { generateVirtualService } from "./virtual-service";
+import { generateServiceEntry } from "./service-entry";
+
+/**
+ * Creates a VirtualService for each exposed service in the package
```
mjnagel:
```suggestion
 * Creates a VirtualService and ServiceEntry for each exposed service in the package
```

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -0,0 +1,93 @@
...
+    // Apply the VirtualService and force overwrite any existing policy
+    await K8s(IstioVirtualService).Apply(vsPayload, { force: true });
+
+    vsPayload.spec!.hosts!.forEach(h => hosts.add(h));
+
+    // Generate a ServiceEntry for this `expose` entry
+    const sePayload = generateServiceEntry(expose, namespace, pkgName, generation, ownerRefs);
+
+    // If we have already made a ServiceEntry with this name, skip (i.e. if advancedHTTP was used)
+    // TODO: (@MJNAGEL) - want to validate that skipping this here doesn't miss any edge cases
```
mjnagel:
```suggestion
```

Went back over this and can't find any current holes so I think we can press on! My thinking:
- name is pkgname + host + gateway
- a second service entry is only needed if the host or gateway are different (everything else is relatively "hardcoded")
- since both of these are captured in the name this logic would only skip extra entries if neither changed

mjnagel (commented):
Sorry for the delayed review - LGTM overall just some comment cleanup.

rjferguson21 (approved):

mjnagel (approved):

Issue #363, opened 2024-04-23
Kubernetes service changes should be processed in a queue in the order in which they came in


cmwylie19:
### Describe what should be investigated or refactored

We are watching for changes to the Kubernetes service and Neuvector Jobs in order to update network policy. This triggers a cascade of events on each change leading to likely thrash in the kube-apiserver. If we put them into a queue then they will be processes one at a time in the order in which the event came in. Cutting down the load on the API Server.

Visual Proof
Look at the wild spikes in CPU on the watcher. Now with this ordered processing, it seems to throttle the amount used and therefore is not hitting that strange frozen state. 

From 24m to 8m
![image](https://github.com/defenseunicorns/uds-core/assets/1096507/fd841f9e-e1be-49cf-ae2c-1f8f8f88d682)

from 32m to 15m
![image](https://github.com/defenseunicorns/uds-core/assets/1096507/aadbfc10-edb4-413c-821b-5656ac07ec22)

Ive seen it drop at low at 3m of CPU

### Links to any relevant code
https://github.com/defenseunicorns/uds-core/blob/da3eb5ab4f5e6ced50f838456999995d5be601b7/src/pepr/operator/index.ts#L43

https://github.com/defenseunicorns/uds-core/blob/da3eb5ab4f5e6ced50f838456999995d5be601b7/src/pepr/istio/index.ts#L22

### Additional context

Add any other context or screenshots about the technical debt here.

PR #364, opened 2024-04-23
chore: 🐛 use reconcile to reduce thrash in pod watch and service watch

cmwylie19:
## Description

We are watching for changes to the Kubernetes service and Neuvector Jobs in order to update network policy. This triggers a cascade of events on each change leading to likely thrash in the kube-apiserver. If we put them into a queue then they will be processes one at a time in the order in which the event came in. Cutting down the load on the API Server.

**Visual Proof**

Look at the wild spikes in CPU on the watcher. Now with this ordered processing, it seems to throttle the amount used and therefore is not hitting that strange frozen state.

See the issue for additional screenshots

image

52 mins no errors

image
image

First failure after 78 mins (memory seems to be growing)
image

## Related Issue

Fixes - Hopefully 🤞 363 

Relates to #363  https://github.com/defenseunicorns/pepr/issues/745

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/istio/index.ts
+++ b/src/pepr/istio/index.ts
@@ -19,7 +19,7 @@ When(a.Pod)
   .IsCreatedOrUpdated()
   .WithLabel("batch.kubernetes.io/job-name")
   .WithLabel("service.istio.io/canonical-name")
-  .Watch(async pod => {
+  .Reconcile(async pod => {
     Log.info(
       pod,
       `Processing Pod ${pod.metadata?.namespace}/${pod.metadata?.name} for istio job termination`,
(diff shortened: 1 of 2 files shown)


mjnagel:
I opened a parallel PR to added reconcile for the service watch but not the pod one - https://github.com/defenseunicorns/uds-core/pull/362. Might let this one simmer for a bit and see if it seems to help before pressing forward - thinking about the case here for our job termination...unless we are seeing duplicate attempts to terminate the sidecar, I'm not sure sequential/queue-based processing is needed here.

cmwylie19:
no problem, we can close this. thanks! Close as you see fit

PR #391, opened 2024-05-08
feat: uds common renovate config

### LINKED ISSUE #389: Adopt uds-common renovate config

We currently maintain a length and very specific renovate config for uds-core. UDS-Common maintains a [great genericized config](https://github.com/defenseunicorns/uds-common/blob/main/config/renovate.json5) that we could adopt and help to extend if it doesn't catch all of our current dependencies. We will likely want to maintain our packagerules and maybe some other one-offs (neuvector image comes to mind), but the rest of these should be able to leverage the common config.

This should absolutely be validated to ensure we don't lose track of any dependencies.

### Additional context

Worth noting that we are currently missing a check for one place zarf init is used in CI, which is covered by the common config - https://github.com/defenseunicorns/uds-core/commit/1b5f604ec3502a6b7d22985fe12fa06b4c0e6e9c

chance-coleman:
## Description
Utilize the uds-common renovate config.
* Now includes major and minor upgrades instead of only major
* No longer uses `latest` bucket, instead uses the major minor buckets
* Removes lots of false positive matches on the keyword docker
* Properly gets versions from task files ( except for the pepr controller in task/setup.yaml )

Four repo specific regex matchers are needed for the pepr controller and for managing neuvector values files.

Also includes an update to Keycloak package for catching two versions with upstream docker image.

## Related Issue

Fixes #389 

## Type of change

## Checklist before merging


renovate.json
```diff
@@ -1,121 +1,39 @@
...
-    ":dependencyDashboard",
+    "github>defenseunicorns/uds-common//config/renovate.json5",
     ":semanticCommits",
     ":semanticCommitTypeAll(chore)"
   ],
-  "forkProcessing": "enabled",
-  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "branchConcurrentLimit": 0,
-  "prConcurrentLimit": 0,
-  "prHourlyLimit": 0,
```
mjnagel: I think we will want to keep these 3 still - by default renovate has some rate limiting and with the amount of dependencies we have on core it might limit our ability to see PRs. Common config doesn't have any overrides to the limits.

renovate.json
```diff
@@ -1,121 +1,39 @@
...
-  "rebaseStalePrs": true,
-  "rebaseWhen": "conflicted",
-  "commitBodyTable": true,
-  "separateMajorMinor": false,
```
mjnagel, line 16: Maybe we don't care/need this but it [defaults to true](https://docs.renovatebot.com/configuration-options/#separatemajorminor) and common does not override unless its somewhere in the base config?

renovate.json
```diff
@@ -1,121 +1,39 @@
...
-      "matchStrings": [
-        "# renovate: datasource=docker( versioning=(?<versioning>.*?))?\n\\s+.+:\\s+(?<depName>[^:]+):(?<currentValue>.*)"
-      ],
-      "datasourceTemplate": "docker"
+        "fileMatch": ["\\.*\\.ya?ml$"],
+        "matchStrings": [
+          "# renovate: datasource=docker( versioning=(?<versioning>.*?))?\n\\s+.+:\\s+(?<depName>[^:]+):(?<currentValue>.*)"
+        ],
+        "datasourceTemplate": "docker"
     },
```
mjnagel: Is this not covered by https://github.com/defenseunicorns/uds-common/blob/main/config/renovate.json5#L62-L71 ?

mjnagel: Assuming this is for the keycloak additions - I used the linked regex test (https://regex101.com/r/d9t0lt/1) and it seemed to match those both as expected.

chance-coleman: Originally this was for a neuvector dep, however i think it gets picked up by one of the other neuvector specific regex or maybe the one you referenced. Either way, you're right it shouldn't be there.

mjnagel (approved):

PR #407, opened 2024-05-14
fix: exemption race conditions

### LINKED ISSUE #314: Exemptions with same matcher twice cause erroneous deletes
Here's a minimal reproduction of the issue I'm seeing:

```yaml
apiVersion: uds.dev/v1alpha1
kind: Exemption
metadata:
  name: test
  namespace: uds-policy-exemptions
spec:
  exemptions:
    - policies:
        - DisallowHostNamespaces
      matcher:
        namespace: test
        name: "^test-pod.*"
    - policies:
        - DisallowPrivileged
      matcher:
        namespace: test
        name: "^test-pod.*"
    - policies:
        - RequireNonRootUser
      matcher:
        namespace: test
        name: "^test-pod.*"
    - policies:
        - DropAllCapabilities
      matcher:
        namespace: test
        name: "^test-pod.*"
```

If you apply this manifest the peprstore ends up only having the `DropAllCapabilities` exemption. This would be true anytime someone uses the same exact matcher in the same CR (i.e. anytime name, namespace, kind, and owner line up), but with multiple separate policy lists. I do think is a quite viable situation that we should support if we can - someone might want to separate out exemptions within the same CR so that they could put specific descriptions on each for example. Happy to help brainstorm how to handle this.

_Originally posted by @mjnagel in https://github.com/defenseunicorns/uds-core/pull/290#discussion_r1541820991_

### LINKED ISSUE #409: Exemption Race Conditions
When users package multiple Exemption CRs together, such that they're processed back to back without any delay, the Pepr store does not update in time before the next exemption is processed and thus overwrites the previous store updates.

### Steps to reproduce
1. group several exemption CRs together into 1 chart / package
2. deploy that chart / package on top of uds core

### Expected result
all exemptions should show up in the pepr store

### Actual Result
only the last exemption shows up in the pepr store

### Severity/Priority
medium -- there are workarounds but it's a bad user experience

### Additional Context
The issue lies in the time it takes for writes to solidify in the Pepr store. Using `SetItemAndWait()`, even if all writes for a single CR are batched into a single write, takes too long. There are ways to arbitrarily make the exemption controller wait a few seconds for synchronous writes to finish before moving onto the next exemption, which works and is faster than using `SetItemAndWait` but is not ideal.

### Solution
Given guidance from @jeff-mccoy, we're going to switch the Pepr watch for a KFC watch, which will run in the admission controllers, allowing us to use an in-memory list of exemptions when validating resources against policies.

TristanHoladay:
## Description
Fixes race conditions with exemptions that result in overwrites of previous exemptions in the Pepr store or mutating then allowing pods that were meant to be exempted from mutation.

## Issue

Fixes #409 

Fixes https://github.com/defenseunicorns/uds-core/issues/314

## Type of change

## Checklist before merging


src/pepr/policies/index.ts
```diff
@@ -1,6 +1,31 @@
...
+  if (process.env.PEPR_WATCH_MODE === "false") {
+    const watcher = K8s(UDSExemption).Watch(async (exemption, phase) => {
+      console.log(`Exemption ${exemption.metadata?.name} is ${phase}`);
+
+      processExemptions(exemption, phase, policyExemptionMap);
+    });
+
+    // This will run until the process is terminated or the watch is aborted
+    await watcher.start();
+  }
```
cmwylie19:
```suggestion
  // only run in admission controller
  if (process.env.PEPR_WATCH_MODE === "false") {
    const watcher = K8s(UDSExemption).Watch(async (exemption, phase) => {
      console.log(`Exemption ${exemption.metadata?.name} is ${phase}`);

      processExemptions(exemption, phase, policyExemptionMap);
    });

    // This will run until the process is terminated or the watch is aborted
    await watcher.start();
  }
```

src/pepr/operator/controllers/exemptions/exemptions.ts
```diff
@@ -1,9 +1,14 @@
 import { Log } from "pepr";
-import { policies } from "../../../policies/index";
-import { ExemptionElement, Matcher, Policy, UDSExemption } from "../../crd";
-
-type StoredMatcher = Matcher & { owner: string };
-type PolicyMap = Map<Policy, StoredMatcher[]>;
+import { PolicyMap, StoredMatcher } from "../../../policies";
+import { ExemptionElement, Policy, UDSExemption } from "../../crd";
+
+export enum WatchPhase {
```
TristanHoladay, line 4: Was trying to use the enum from the fluent client but got errors when importing the type. so i just remade it here.

cmwylie19 (dismissed):

src/pepr/operator/reconcilers/index.ts
```diff
@@ -12,7 +12,7 @@ const uidSeen = new Set<string>();
  * @param cr The custom resource to check
  * @returns true if the CRD is pending or the current generation has been processed
  */
-export function shouldSkip(cr: UDSExemption | UDSPackage) {
+export function shouldSkip(cr: GenericKind) {
```
mjnagel: Should this still be `UDSPackage` instead of `GenericKind`?

mjnagel (commented):
One other broad question/thing to validate - would want to make sure that this no longer being queue based doesn't lead to any weird behaviors with rapid create/edit/delete events (today those get processed in order on a queue with `Reconcile`, unsure exactly how that would behave with KFC watch here).

src/pepr/operator/reconcilers/index.ts
TristanHoladay: i went back and forth on that one. I don't think it hurts to set it to `GenericKind` but i'll change it to `UDSPackage`. it'll be more consistent with the other functions that way.

TristanHoladay:
I was thinking the same @mjnagel . so far i haven't seen any issues, but i'll see about creating a test for it.

src/pepr/operator/controllers/exemptions/exemptions.ts
```diff
@@ -1,160 +1,114 @@
...
+    const matcherToStore = {
+      ...e.matcher,
+      owner: exemption.metadata?.uid || "",
+    };
+
+    const policies = e.policies ?? [];
+    for (const p of policies) {
+      const storedMatchers = map.get(p) ?? [];
+      if (log) {
+        console.log(`Adding to ${p}: ${JSON.stringify([...storedMatchers, matcherToStore])}`);
```
mjnagel: Still working my way through the rest of the PR, but one comment upfront - anywhere we are using `console.log` I think we should really be using `Log.debug` or `Log.info` depending on the specific case. This will ensure we have consistent logging and are using the same log method as all the other parts of the operator + admission pieces.

I'm also on the fence on the conditional for logging here. I think I see how it is used (only logging adds, not modifies?), but would prefer that we err on the side of more logging rather than less.

src/pepr/operator/crd/validators/exempt-validator.ts
```diff
@@ -30,11 +30,6 @@ export async function exemptValidator(req: PeprValidateRequest<UDSExemption>) {
     }
   }
 
-  // Validate there's at least 1 exemption element
-  if (exemptions.length === 0) {
-    return req.Deny("Invalid number of exemptions: must have at least 1");
-  }
-
```
mjnagel, line 37: Is there a specific reason this validation was removed? I don't think we strictly need to block empty exemptions (although it could be confusing), just wondering if I'm missing something here that necessitated this change?

Worth testing this but I think we can actually add `minItems: 1` into the exemption source (lined up with the `type: "array"`) and that should enforce at the schema level rather than here.

src/pepr/operator/controllers/exemptions/exemptions.ts
TristanHoladay: good point about the logger. I forgot about `Log` and just defaulted to `console`. I added the boolean so that we don't log adding to the temporary map. the updates will still get logged, just not from the `addToMap`.

src/test/tasks.yaml
```diff
@@ -49,5 +49,15 @@ tasks:
...
+      - cmd: uds zarf tools kubectl get pods -n podinfo -o=jsonpath='{.items[0].metadata.name}'
+        setVariables:
+          - name: PODINFO_NAME
+      - description: Check for podinfo to come back up
+        wait:
+          cluster:
+            kind: Pod
+            name: ${PODINFO_NAME}
+            namespace: podinfo
+
```
mjnagel: This failed for me locally on the first run:
```
Error creating: admission webhook "pepr-uds-core.pepr.dev" denied the request: Host ports are not allowed.; Unauthorized container securityContext. Containers must not run as root. Authorized: [runAsNonRoot = false | runAsUser > 0] Found: {"name":"podinfo","ctx":{"capabilities":{"drop":["ALL"]},"runAsUser":0,"runAsGroup":0,"runAsNonRoot":false}}
```

I see all 5 exemption resources created, and actually see `Adding to RequireNonRootUser:` with podinfo in the logs too 🤔. I'm going to re-run a few times and look through the logic but it worries me a bit that this could fail 😬.

src/pepr/operator/controllers/exemptions/exemptions.ts
mjnagel: Ah okay, tracking. Sounds good.

src/test/tasks.yaml
```diff
@@ -49,5 +49,15 @@ tasks:
...
 
+      - cmd: uds zarf tools kubectl get pods -n podinfo -o=jsonpath='{.items[0].metadata.name}'
+        setVariables:
+          - name: PODINFO_NAME
+      - description: Check for podinfo to come back up
+        wait:
+          cluster:
+            kind: Pod
+            name: ${PODINFO_NAME}
+            namespace: podinfo
```
mjnagel:
```suggestion
      - description: Check for podinfo to come up
        wait:
          cluster:
            kind: Pod
            name: app.kubernetes.io/name=podinfo
            namespace: podinfo
```
nit: this supports labels so we should be able to skip the "get pod name" part.

src/test/tasks.yaml
```diff
@@ -49,5 +49,15 @@ tasks:
...
+
       - description: Remove the test resources
         cmd: "uds zarf package remove build/zarf-package-uds-core-test-apps-*.zst --confirm"
```
mjnagel:
```suggestion
        cmd: "uds zarf package remove build/zarf-package-uds-core-test-apps-*.zst --confirm --no-progress"
```
would be a nice addition while we're modifying this 😄

src/test/tasks.yaml
mjnagel: Tentatively going to resolve this, I haven't been able to reproduce this issue and have made a few refactors locally so am hoping this is a non-issue at this point. Something to keep an eye on in CI/local dev though.

mjnagel (dismissed):

rjferguson21 (dismissed):

rjferguson21 (approved):

MxNxPx (approved):

cmwylie19 (approved):

PR #423, opened 2024-05-22
feat: add reconciliation retries for CRs

mjnagel:
## Description

Adds re-tries to Package CR status + logic to increment and handle retries. Currently will attempt package reconcile 5x before failing.

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/operator/crd/generated/exemption-v1alpha1.ts
+++ b/src/pepr/operator/crd/generated/exemption-v1alpha1.ts
@@ -67,6 +67,7 @@ export enum Policy {
 export interface Status {
   observedGeneration?: number;
   phase?: Phase;
+  retryAttempt?: number;
   titles?: string[];
 }
 
@@ -80,4 +81,5 @@ RegisterKind(Exemption, {
   group: "uds.dev",
   version: "v1alpha1",
   kind: "Exemption",
+  plural: "exemptions",
 });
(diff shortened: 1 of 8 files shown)


MxNxPx (approved):

rjferguson21 (approved):

mjnagel:
Posting additional context on this shift retroactively since this change was rather significant and resulted in a few issues. 

Retries were introduced here to account for a specific error we ran into during pepr upgrades/pods cycling. With the introduction of service monitor generation in the operator, we have a flow where the watcher pod generates a service monitor that the admission pods then mutate. Across upgrades we encountered intermittent failures due to webhook timeouts - the watcher would fail to apply the service monitors, erroring out reconciliation of a Package on something that should be retry-able (thinking about normal helm/zarf flow, multiple apply attempts would be made).

Rather than introduce a targeted retry for just the servicemonitor behavior we decided it would potentially solve more intermittent issues (ex: intermittent networking related problems) if we just did a generic 5x retry on all Packages. This was reviewed synchronously and tested against a few scenarios where retries did resolve issues. For history sake linking bugs introduced here:
- [Packages can get stuck in Pending when a retry is occuring](https://github.com/defenseunicorns/uds-core/issues/492) - If the Package successfully reconcile once already (uidSeen) then retries will not actually take place
- [Packages can get stuck in a Failed <-> Pending loop](https://github.com/defenseunicorns/uds-core/issues/525) - If the Package has never successfully reconciled it will infinitely retry.

PR #424, opened 2024-05-23
feat: velero csi plugin

justicorn:
## Description
Adding the CSI plugin to Velero
...

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


blancharda:
[In the original PR](https://github.com/defenseunicorns/uds-core/pull/210#discussion_r1511759778) we opted to leave out additional plugins by default, but left the door open to circle back later..

I think it's later now 😆

cc @mjnagel -- I think the original concern was around clarity on supported defaults (totally valid). There was also an assumption that the aws plugin would support S3 compatible storage providers.  

Regarding the S3 compatibility assumption: I think that was bunk (and I'm the one that pushed it lol). Initial testing with rook-ceph appeared to be working, but we also may have inadvertently still had all the plugins enabled when we ran those tests. While the assumption may hold true for some providers, I don't think there's a guarantee, and we're much more likely to encounter storage providers that support the CSI API.

Regarding default support: I think the need for this is present in multiple envs, and this feels like a pretty minor change that expands default capability without negatively impact existing users or change current workflows..

If it's not clear -- I'm sold 😆
Curious your thoughts though.

blancharda:
> Regarding the S3 compatibility assumption: I think that was bunk (and I'm the one that pushed it lol).

I also think in the original discussion around it, I/we may have partially conflated S3 compatible object storage (not really related to velero) with the storage driver (very much related to velero lol) which is easy to do when they're both installed by the same tool at the same time and often named very similarly.. 

TLDR.. bad assumptions are bad 🙈

mjnagel (commented):
I think this makes sense to add to expand the plugin/provider support. To make sure I understand - it may be helpful to have an example backup config we could show/link for how to use CSI (even just in this PR for myself to reference/validate). I'm not sure if that's something we could easily add a test for here also or not?

src/velero/values/registry1-values.yaml
```diff
@@ -14,3 +14,11 @@ initContainers:
...
       - mountPath: /target
         name: plugins
+  - name: velero-plugin-for-csi
+    image: registry1.dso.mil/ironbank/opensource/velero/velero-plugin-for-csi:v0.7.0
+    imagePullPolicy: IfNotPresent
+    volumeMounts:
+      - mountPath: /target
+        name: plugins
+configuration:
+  features: EnableCSI
```
mjnagel: Could you resolve the handful of linting issues here (just newlines at the end of the files) + merge the latest main changes in?

justicorn:
> I think this makes sense to add to expand the plugin/provider support. To make sure I understand - it may be helpful to have an example backup config we could show/link for how to use CSI (even just in this PR for myself to reference/validate). I'm not sure if that's something we could easily add a test for here also or not?

I included a section in the readme identifying that additional configuration would be required to get this working, this is just enabling the CSI plugin itself. Turning on `snapshotVolume` and configuring the CSI's snapshot location, etc would be required to test it. In the case of CI testing, it appears to be using the AWS EBS CSI, and that would utilize the AWS plugin, which was already included.

mjnagel (approved):

Issue #327, opened 2024-04-03
Matrix k8s version testing [latest, n-1] on k3d


MxNxPx:
### Describe what should be investigated or refactored

Currently we test on a single k3d version in PRs. In order to expand test coverage to handle other k8s versions we should matrix this test across several minor versions. To start this should be for k3d only, and cover the latest minor version + n-1 minor version. To reduce PR CI time this should only be done nightly to start.

Definition of done: Addition of nightly testing on matrixed k3d versions, covering n-1 and latest. Ideally this is easily extensible to additional versions in the future (matrixed, task inputs, etc).

### Additional context

Considering there will be folks using UDS on older K8s versions than what is used to publish, ought to consider adding some testing of n-1 K8s version.

mjnagel:
Updating this ticket:
- we are [1.29.5](https://github.com/defenseunicorns/uds-core/blob/main/tasks.yaml#L8) for k3d/k3s
- we are on [1.27](https://github.com/defenseunicorns/uds-core/blob/main/tasks/iac.yaml#L29) for EKS
- RKE2 infra is not maintained as a part of our repo/CI currently but we have other testing environments that we should update to 1.29+

mjnagel:
Rescoping this effort to focus on testing across multiple k8s distros. To start we should do nightly test on the latest k8s minor version and n-1 minor version, at minimum on k3d - possibly on eks as well.

MxNxPx:
> Updating this ticket:
> 
> * we are [1.29.5](https://github.com/defenseunicorns/uds-core/blob/main/tasks.yaml#L8) for k3d/k3s
> * we are on [1.27](https://github.com/defenseunicorns/uds-core/blob/main/tasks/iac.yaml#L29) for EKS
> * RKE2 infra is not maintained as a part of our repo/CI currently but we have other testing environments that we should update to 1.29+

Now on 1.30 for EKS per https://github.com/defenseunicorns/uds-core/pull/605

mjnagel:
Updated ticket description to focus on k3d, nightly, and 2 versions to start.

noahpb:
I would add that our nightly ci tests across the various k8s distros (k3d, rke2, eks, aks) should all run against the same minor version of k8s. I added an issue to track this here https://github.com/defenseunicorns/uds-core/issues/784.

mjnagel:
[...]

PR #426, opened 2024-05-23
fix: use updated k3s

### LINKED ISSUE #253: CI install/upgrade timing out waiting on pending pods

Periodically in CI, on the `all` tests, the test will hang/timeout on a package. It is commonly Grafana or Prometheus stack that the timeout occurs on. Generally Pending pods can be observed in the debug output, ex: https://github.com/defenseunicorns/uds-core/actions/runs/8268437284/job/22621379639?pr=210#step:9:55

We should identify where/why this is happening to make our CI more robust and less prone to flaky failures.

### Additional context

A number of suspicions have been proposed but so far debug hasn't yielded anything concrete:
 - k8s resource issues (typically would show in events, haven't seen this)
 - pvc binding issues (typically would show in events, haven't seen this)
 - "host system" resource issues (the runner actually hitting max cpu/memory)

rjferguson21:
## Description
Testing CI with upgraded k3s

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/253

## Type of change

## Checklist before merging

### DIFF
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -5,7 +5,7 @@ variables:
   - name: PKG
 
   - name: K3D_IMAGE
-    default: "ghcr.io/defenseunicorns/oss/uds-k3d-k3s:v1.28.7-k3s1"
+    default: "rjferguson21/k3s:v1.29.5-k3s1"
 
 includes:
   - create: ./tasks/create.yaml
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -5,7 +5,7 @@ variables:
   - name: PKG
 
   - name: K3D_IMAGE
-    default: "rjferguson21/k3s:v1.29.5-k3s1"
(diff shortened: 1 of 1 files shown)


tasks.yaml
```diff
@@ -5,7 +5,7 @@ variables:
   - name: PKG
 
   - name: K3D_IMAGE
-    default: "ghcr.io/defenseunicorns/oss/uds-k3d-k3s:v1.28.7-k3s1"
+    default: "rjferguson21/k3s:v1.29.5-k3s1"
```
mjnagel, line 8: sus

mjnagel (approved):

mjnagel:
Part of https://github.com/defenseunicorns/uds-core/issues/327

mjnagel (commented):
Not 100% confident we have fixed _all_ the CI issues here but this seems pretty promising given the consistent passing here. Would be great to get a better idea of why this update seems to help, if we can track that in the issue.

chance-coleman (approved):

PR #446, opened 2024-05-29
fix: remove no-tea and update uds version

chance-coleman:
## Description
Make necessary changes for latest uds release v0.11.0

## Related Issue

Fixes #445 

## Type of change

## Checklist before merging


mjnagel (dismissed):
LGTM. I think there are a few more changes we can make with this update but don't have to do them all here:
- drop the zarf tools clear-cache that is in a spot or teo since vendored zarf has resolved that issue in the update
- switch to upstream k3s image for the same reason

I'll leave to you if you want to resolve those other two here or maybe a follow-on PR/issue. Neither are urgent/required but are top of mind for me with the UDS update.

mjnagel (approved):

PR #449, opened 2024-05-30
chore: update loki to 3.1.1

chance-coleman:
## Description
Updates for bringing in loki renovate updates

## Related Issue

Fixes #308 

## Type of change

## Checklist before merging


src/loki/values/values.yaml
```diff
@@ -67,6 +76,7 @@ test:
   enabled: false
 
 # Configuration for the single binary node(s)
+deploymentMode: SingleBinary
```
mjnagel: I think this is a change from the previous? In core before we were defaulting to `SimpleScalable` (although it was done a different way in the past). We likely want to default to scalable still...

src/loki/values/values.yaml
```diff
@@ -25,6 +25,15 @@ loki:
       insecure: false
   commonConfig:
     replication_factor: 1
+  schemaConfig:
+    configs:
+      - from: 2024-04-01
```
mjnagel: I think this should be set to a date in the future based on https://grafana.com/docs/loki/latest/operations/storage/schema/#changing-the-schema, otherwise it will result in logs from 4/1 -> today being unreadable since they are actually v12 schema? I think we could set it to one month+ out explicitly, OR we could look at doing this via zarf var/pepr to mutate in a date in the future.

I'm not certain but we may also need a v12 schema entry here to handle older logs (using a `from` date of something last year to safely handle all possible logs from the first uds-core release)?

mjnagel (commented):
Haven't fully deployed yet but wanted to get these two comments out there.

mjnagel (commented):
We probably also want to switch over to `Secret` for this value - https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml#L112

src/pepr/loki/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+            // Ensure limits_config exists and update allow_structured_metadata to false
+            if (!lokiConfig.limits_config) {
+              lokiConfig.limits_config = {} as LimitsConfig;
+            }
+            lokiConfig.limits_config.allow_structured_metadata = false;
+
+            // Update the secret with the new config.yaml content
+            secret.Raw.data["config.yaml"] = yaml.dump(lokiConfig);
+            log.info(secret.Raw.data["config.yaml"], `Secret config.yaml updated successfully.`);
```
mjnagel: I think there might be some missing logic here - but want to summarize my understanding of the cases and how we should be handling them:
- install/upgrade: the date (as set in this PR) would be a previous date (2024-06-04) and we would want to generate a future date, 2 days in advance as discussed to be safe with the way Loki handles this
- future upgrade: the date coming into the webhook would be our previous date (2024-06-04), we would want to set this to the date we already generated (probably could query the cluster for this?)

The key here is that we want to generate a date in the future once, but once we generate it once always use that date for future mutations (since the baseline applied from the chart will always be an older date). I think in the current state we would end up mutating this date any time it is in the past?

chance-coleman: Yeah that is a good point that i totally misunderstood. I've updated so the date is only generated once and stored as an annotation in the secret for future reference.

mjnagel (dismissed):

src/test/app-tenant.yaml
```diff
@@ -26,7 +26,6 @@ spec:
       - service: test-tenant-app-cidr
         selector:
           app: test-tenant-app
-        remoteCidr: "192.168.0.0/16"
```
chance-coleman, line 29: This [was added ](https://github.com/defenseunicorns/uds-core/pull/719/files#diff-c94db7cc6e40995709784aae6c1a7f05ae2ed175ca1c0eb2553de353ddbd5d5fL25-L26)for testing that the remoteCidr generation wouldn't break anything. If you're going to remove this then the whole netpol should be removed since that was its only usecase.

mjnagel (approved):

PR #450, opened 2024-05-31
chore: add checks before killing pods when updating istio annotations

YrrepNoj:
## Description

As I was deploying multiple Package CRDs to a namespace I noticed that my Pods were cycling whenever I deployed another Package CR to my namespace. When looking through the code (with the help of engineers far more familiar and smarter than me) we realized we might have been too excited about when we were killing the pods.

This PR updates the operator so that it only kills the pods in the namespace if it has actually changed the value of the istio-injection label.

## Type of change

### DIFF
--- a/src/pepr/operator/controllers/istio/injection.ts
+++ b/src/pepr/operator/controllers/istio/injection.ts
@@ -23,11 +23,11 @@ export async function enableInjection(pkg: UDSPackage) {
 
   // Mark the original namespace injection setting for if all packages are removed
   if (!annotations[injectionAnnotation]) {
-    annotations[injectionAnnotation] = labels[injectionLabel] || "non-existent";
+    annotations[injectionAnnotation] = originalInjectionLabel || "non-existent";
   }
 
   // Ensure the namespace is configured
-  if (!annotations[pkgKey] || labels[injectionLabel] !== "enabled") {
+  if (!annotations[pkgKey] || originalInjectionLabel !== "enabled") {
     // Ensure Istio injection is enabled
     labels[injectionLabel] = "enabled";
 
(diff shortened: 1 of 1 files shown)


src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -45,7 +46,10 @@ export async function enableInjection(pkg: UDSPackage) {
       { force: true },
     );
 
-    await killPods(pkg.metadata.namespace, true);
+    // Kill the pods if we changed the value of the istio-injection label
+    if (originalInjectionLabel !== labels[injectionLabel]) {
```
mjnagel, line 50: nit: Since we now have the const for it, using `originalInjectionLabel` on lines 26/30 would clean up/align all the places we reference original state.

src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -88,7 +93,10 @@ export async function cleanupNamespace(pkg: UDSPackage) {
     { force: true },
   );
 
-  await killPods(pkg.metadata.namespace, false);
+  // Kill the pods if we changed the value of the istio-injection label
+  if (originalInjectionLabel !== labels[injectionLabel]) {
+    await killPods(pkg.metadata.namespace, false);
+  }
```
mjnagel, line 99: I think this works and probably makes the most sense to handle the case where the original injection label was already "enabled". Noting mostly for myself, but the problem here was this not having the same `If there are no more UDS Package annotations` check around it (although your check handles everything best I think).

mjnagel (commented):
Overall LGTM - unsure what happened with CI but looks like a registry1 auth issue maybe 🤔 

Edit: identified the auth issue...has to do with this workflow coming from a fork which does not give access to the secrets necessary for registry1 auth.

YrrepNoj:
Closing this PR and reopening as a non-fork.

PR #469, opened 2024-06-09
fix: improve unique naming of netpols

### LINKED ISSUE #466: Operator fails to generate all network policies for all virtual services if they use the same pod selector.

Network policies that are generated for any virtual services that are created don't generate unique enough names.

The below section of code [linked here](https://github.com/defenseunicorns/uds-core/blob/7cf9c4c09dccc1d91edb7acbb2aee8750c5433ed/src/pepr/operator/controllers/network/policies.ts#L55-L39) creates policy descriptions that only use the selector and gateway to differentiate between policies. 

```js
  for (const expose of exposeList.filter(exp => !exp.advancedHTTP?.directResponse)) {
    const { gateway = Gateway.Tenant, port, selector = {}, targetPort } = expose;

    // Create the NetworkPolicy for the VirtualService
    const policy: Allow = {
      direction: Direction.Ingress,
      selector,
      remoteNamespace: `istio-${gateway}-gateway`,
      remoteSelector: {
      },
      // Use the same port as the VirtualService if targetPort is not set
      port: targetPort ?? port,
      description: `${Object.values(selector)} Istio ${gateway} gateway`,
    };
```
When a name is generated in the second snippet of code [linked here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/generate.ts#L119-L136) it uses that description. The problem here is if we need to create multiple virtual services that use the same pod selector but different ports, it will only create a network policy for the last virtual service as the names are not unique.
```js
export function generateName(policy: Allow) {
  const name =
    // Use the description if it exists
    policy.description ||
    // Otherwise use the direction, and combination of remote properties
    [
      Object.values(policy.selector || ["all pods"]),
      policy.remoteGenerated || [
        policy.remoteNamespace,
      ],
    ]
      // Flatten the array
      .flat(1)
      .join("-");

  return `${policy.direction}-${name}`;
}
```
 There is a comment [linked here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/generate.ts#L12-L11) that states how its making the name unique by using values like port, but isn't accurate given the code.
 ```js
  // Create a unique name for the NetworkPolicy based on the package name, index, direction, pod labels, and port
  const name = generateName(policy);
```

We can get it to create all the needed network policies for each virtual service that is created by using a different label selector that is defined in the same pod. Clunky but works although it should work even if using the same selector.

This only creates the network policy for the last virtual service
```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: nexus
spec:
  network:
    expose:
      - service: nexus-nexus-repository-manager
        selector:
          app: nexus-repository-manager
        gateway: tenant
        host: nexus
        port: 8081
      - service: nexus-nexus-repository-manager-docker-5000
        selector:
          app: nexus-repository-manager
        gateway: tenant
        host: nexus-docker
        port: 5000
```
This works and creates all network polices
```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: nexus
spec:
  network:
    expose:
      - service: nexus-nexus-repository-manager
        selector:
          app: nexus-repository-manager
        gateway: tenant
        host: nexus
        port: 8081
      - service: nexus-nexus-repository-manager-docker-5000
        selector:
          app.kubernetes.io/instance: nexus
        gateway: tenant
        host: nexus-docker
        port: 5000
```

anthonywendt:
## Description
Adds port to the generated naming of network policies. This allows network polices to be created using the same pod selector but different ports.
...

## Related Issue

Fixes #466 

## Type of change

## Checklist before merging


src/test/tasks.yaml
```diff
@@ -43,12 +43,19 @@ tasks:
             address: demo.admin.uds.dev/status/410
             code: 410
 
-      - description: Verify the tenant app is accessible
+      - description: Verify the tenant app muli port 8080 is accessible
         wait:
           network:
             protocol: https
-            address: demo.uds.dev/status/202
-            code: 202
```
mjnagel, line 51: This removed the previous tenant demo app wait it looks like. I think having a test for multi-port is valuable but wonder if we could just do it as part of one of our other existing tests so that we aren't adding more pods to the scope of our test (we're not far off from the resource limit for the runner 😢). Maybe we could replace the current tenant app check with this one? Ideally it would also be great to keep this to a single image/container that is listening on two ports but even reducing scale by one less container/pod would be good.

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -132,5 +133,8 @@ export function generateName(policy: Allow) {
       .flat(1)
       .join("-");
 
-  return `${policy.direction}-${name}`;
+  // Add the port if it exists
+  const generatedName = policy.port ? `${policy.port}-${baseName}` : baseName;
+
```
mjnagel, line 138: Open to differing opinion but I think I would prefer to not add the port if description is used - we added description partially to provide a way for short/simple netpol names instead of the generated ones. Maybe we can add port into the `Object.values` flattened array above instead?

mjnagel (commented):
Really appreciate the contribution - two comments. I'm looking into the issue with CI to see what we can do - will follow up on that.

src/test/tasks.yaml
anthonywendt: Ooops. I didn't mean to remove that original one, but it tests the same thing. I originally tried to use httpbin for this multiport test, but it wouldn't work because it only listens on port 80. I needed a container that could setup to listen on multiple ports to force it to need to create multiple netpols for the same pod with the same pod selector. Thats where the bug was that caused netpol collision.

I think http-echo could work for the other app tests? in place of httpbin? I'll try to replace httpbin and see fi that works okay.

src/pepr/operator/controllers/network/generate.ts
anthonywendt: Thats where the problem is though. Since the creation of a VirtualService creates netpols and creates a description, it will just use that description and then run into collision when using the same selector for different ports. I guess I could add the port upstream to the description that the VirtualService logic is using?

src/test/tasks.yaml
anthonywendt: So I re worked this a bit. Its back to just the app-tenant manifests but exercises multi port. I can't recreate the admin tests with http-echo and I can't recreate the mutli-port tests using httpbin. I did cut resource usage in half on everything and that all works? Maybe thats acceptable?

PR #480, opened 2024-06-13
fix: pepr operator derived netpol name collisions

### LINKED ISSUE #466: Operator fails to generate all network policies for all virtual services if they use the same pod selector.

anthonywendt:
## Description
Fixes network policy naming bug that caused name collisions for operator derived network policies. This causes network policies being created for VirtualServices using the same selector and gateway but different ports to be dropped. Multiple legitimate policies would override each other and only the last would survive.

There are still gaps in the fundamentals of naming and validation that I came across on different iterations on solutions for this. I will be creating a follow on issue for a deeper dive into that though. This is a small improvement that fixes a specific issue. 
...

## Related Issue

Fixes #466 

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/operator/controllers/network/generate.ts
+++ b/src/pepr/operator/controllers/network/generate.ts
@@ -8,7 +8,7 @@ import { intraNamespace } from "./generators/intraNamespace";
 import { kubeAPI } from "./generators/kubeAPI";
 
 export function generate(namespace: string, policy: Allow): kind.NetworkPolicy {
-  // Create a unique name for the NetworkPolicy based on the package name, index, direction, pod labels, and port
+  // Generate a unique name for the NetworkPolicy
   const name = generateName(policy);
 
   // Create the NetworkPolicy
@@ -112,9 +112,11 @@ export function generate(namespace: string, policy: Allow): kind.NetworkPolicy {
 }
 
 /**
- * Generates a unique name for the NetworkPolicy based on the description, direction, and combination of remote properties
+ * Generates a unique name for the NetworkPolicy based on the policy.
+ * Will use the description if it exists, otherwise it will use the
+ * direction and combination of remote properties.
  *
- * @param policy the name of the policy
+ * @param policy
  */
 export function generateName(policy: Allow) {
   const name =
--- a/src/pepr/operator/controllers/network/generate.ts
+++ b/src/pepr/operator/controllers/network/generate.ts
@@ -116,7 +116,7 @@ export function generate(namespace: string, policy: Allow): kind.NetworkPolicy {
  * Will use the description if it exists, otherwise it will use the
  * direction and combination of remote properties.
  *
- * @param policy
+ * @param policy The policy to generate a name for
  */
 export function generateName(policy: Allow) {
   const name =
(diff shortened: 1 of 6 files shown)


mjnagel (approved):
LGTM, as discussed we can do a follow on issue/PR to handle any remaining clashes/better validation at admission time.

PR #496, opened 2024-06-19
chore(oscal): begin integration of composed oscal with validations

### LINKED ISSUE #458: Lula Integration in CI
Integrate Lula into the CI process to verify UDS Core configuration against compliance controls. This integration will determine if the controls are satisfied and provide evidence to demonstrate compliance to assessors in an automated and continuous manner.

This approach enables early detection of changes that may impact compliance, offering a mechanism to prevent non-compliant changes if desired. It ensures compliance in a quantitative way, providing assurance and clarity.

- [ ] Integrate Lula Lint Task from uds-common to lula lint new/modified OSCAL files in CI.

- [ ] Integrate Lula Validate Task from uds-common to run lula validate against OSCAL and supporting files to determine pass fail for compliance controls.

- [ ] Integrate Lula Evaluate Task from uds-common to run lula evaluate after upgrade to compare previous assessment-results (OSCAL) to current to determine changes.

- [ ] Update location of current OSCAL files.

Considered GitHub Actions to perform the same functionality but following the Task pattern provides a better mechanism for usage outside of GitHub.

### Additional context
Is dependent on creation of tasks issues being completed first.

brandtkeller:
## Description

This change proposes the integration of [Lula](https://github.com/defenseunicorns/lula) with uds-core for the purposes of compliance assessment on Pull-Request. This Proposal includes:
- all proposed future changes meet the minimum threshold for compliance
- framework established for iterating control body-of-evidence
- process for consuming uds-core compliance for downstream utilization

### How does it work?
Each component that comprises `uds-core` contains an [OSCAL](https://pages.nist.gov/OSCAL/about/) file in the form of the `component-definition` model. This captures  1 -> N controls for 1 -> N standards (control implementations) that pertain to a specific component.

When controls are identified as implemented by a component, we can create or update a `component-definition` artifact through:
```bash
lula generate component -c <catalog-source> -r <controls-comma-delimited> --component "component name>
```

#### Validation
Lula operates on a model where controls are marked as satisfied when presented with programmatic and repeatable processes to to produce evidence that the control is satisfied. 

These are often developed on a control-by-control basis and then composed into a single OSCAL component-definition  in a schema-compliant addition in the back-matter. Reducing required network connectivity and making the compliance information portable.

One control may require evidence from many `validations`. One `validation` is the collection of data from some source and the measurement of that data for required adherence. 

The `lula validate` operation is an objective assessment - determining control `satisfied/not-satisfied` state and writing the results to the `assessment-results` artifact. 

#### Evaluation

The last step for determining greater, equal, or lesser compliance than the previous state is to execute a `lula evaluate`. This consumes the latest `result` in the `assessment-results` artifact and compares it against the `threshold` result. 

The `threshold` result is identified as the `result` containing the following prop:
```yaml
- name: threshold
   ns: https://docs.lula.dev/ns
   value: "true"
```

If the compliance of this latest result is worse than the threshold - the execution fails with a non-zero exit code - allowing for pipeline operations to fail. 

If the compliance is equal - the command passes without any modification required.

If the compliance is greater - Lula will automatically move the threshold identifier to the applicable result item. (this would mean that the updated `assessment-results` artifact should be downloaded and submitted for update to the repository). 

#### Naming

All files are titled with "oscal-" prepended to the OSCAL model type:
- oscal-component.yaml -> component definition
- oscal-assessment-results.yaml -> assessment results

This is primarily to allow for identification for linting workflows to target oscal files for Lula to Lint.

### Metrics

The following are derived from the FedRAMP Moderate baseline:
```
24 controls of *329 total* mapped in OSCAL from Istio (7.29%)
24 controls of *161 total Technical* mapped in OSCAL from Istio (14.90%)
75 Unique controls captured in OSCAL of *329 total* (22.79%)
10 controls marked satisfied in OSCAL of *329 total* (3.03%)
75 Unique controls captured in OSCAL of *161 technical* (46.58%)
10 controls marked satisfied in OSCAL of *161 technical* (6.21%)
```

> [!NOTE]  
> DISA has recently published a new revision to the Cloud Computing Security Requirements Guide which may result in future iterations to a small subset of controls when reconciled. 

[DISA link](https://public.cyber.mil/dccs/dccs-documents/)

### Review

The Lula team is actively working to make the review process better. Composed OSCAL has it's benefits for portability but that comes with the drawbacks of review. 

To assist with the review process:
- Ensure the pipeline additions are sound
- Review the `src/istio/oscal-component.yaml` file. This will map controls to the current state of validations.
- Understand the connection of `compliance/oscal-component.yaml` to the individual components under `src/`

### Troubleshooting

If the pipeline should begin failing in the future - notify the @defenseunicorns/lula-dev team. 

It should be expected that any major revisions to applications could break known assumptions about configuration required for compliance. The Lula team can help navigate failures. 

## Related Issue

Resolves #458 

## Type of change

## Checklist before merging


.github/workflows/lint-oscal.yaml
```diff
@@ -0,0 +1,69 @@
...
+          echo "oscal_files=$CHANGED_FILES" >> "$GITHUB_OUTPUT"
+        shell: bash
+      # checkout for access to the oscal files targeted for linting
+      - name: Checkout the code
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      # setup uds-cli
+      - name: Use Node.js latest
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        with:
+          node-version: 20
```
mjnagel: I don't think we need node setup here unless I'm missing a piece where it's used (typically we need it for pepr things).

Alternatively it might be good to just use the setup action here which would have uds + lula installs (maybe we could make the node install optional, default to true?).

.github/workflows/lint-oscal.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Set up Homebrew
+        uses: Homebrew/actions/setup-homebrew@master
+      - name: Install UDS CLI
+        shell: bash
+        # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
+        run: brew install defenseunicorns/tap/uds@0.11.2
```
mjnagel: If you keep this separate from the install action this should be updated to 0.12.0 (updated by renovate yesterday).

.github/workflows/lint-oscal.yaml
```diff
@@ -0,0 +1,69 @@
+name: Lint OSCAL Files
+
+on:
+  push:
+    paths:
+      - '**/*oscal*.yaml'
```
mjnagel: Is this intentionally on push rather than pull request? I believe this means that linting results can get buried in a PR on subsequent pushes (if my latest push doesn't change those files, this won't run).

.github/workflows/lint-oscal.yaml
```diff
@@ -0,0 +1,69 @@
...
+        run: brew install defenseunicorns/tap/uds@0.11.2
+      # will be pinning this when a tag is released
+      - name: Install Lula
+        uses: defenseunicorns/lula-action/setup@095636b7880051e11b05f10a582fdd911526161c
+        with:
+          # renovate: datasource=github-tags depName=defenseunicorns/lula versioning=semver-coerced
+          version: v0.4.1
+      # lint the oscal files
+      - name: lint-oscal
+        run: uds run lint-oscal --set OSCALFILES=${{ steps.find_changed_files.outputs.oscal_files }}
```
mjnagel: nit: might be helpful to add `--no-progress` here. In the current uds cli state output gets a bit messy without that in CI (see the [previous run](https://github.com/defenseunicorns/uds-core/actions/runs/9746263061/job/26896166351#step:8:8) of this).

src/istio/oscal-component.yaml
```diff
@@ -1,671 +1,1451 @@
...
+                    jsonpath: .data.mesh
+                    type: yaml
+                  group: ""
+                  name: istio
+                  namespaces:
+                  - istio-system
+                  resource: configmaps
+                  version: v1
+            type: kubernetes
+          lula-version: ""
```
mjnagel, line 27: Curious on why this is all part of a `description` string block? Is that a limitation of the oscal schema not supporting anything like Lula natively? It makes yaml validation/highlighting a bit painful 😅

src/istio/oscal-component.yaml
```diff
@@ -1,671 +1,1451 @@
...
+                # Rego policy logic to evaluate if all PeerAuthentications have mtls mode set to STRICT
+                result_all_strict = {"result": true, "msg": msg} {
+                    every peerAuthentication in input.peerAuths {
+                        mode := peerAuthentication.spec.mtls.mode
+                        mode == "STRICT"
+                    }
+                    msg := "All PeerAuthentications have mtls mode set to STRICT."
+                } else = {"result": false, "msg": msg} {
+                    msg := "Not all PeerAuthentications have mtls mode set to STRICT."
+                }
```
mjnagel: One nuance here - peerauthentications can have [port level mtls](https://istio.io/latest/docs/reference/config/security/peer_authentication/) which could be permissive and change the posture. I think this is sufficient for an initial validation, but worth noting there are spots in uds-core and likely SWF that use port level permissive.

src/istio/oscal-component.yaml
```diff
@@ -1,671 +1,1451 @@
...
+            name: communications-terminated-after-inactivity-PLACEHOLDER
+            uuid: 663f5e92-6db4-4042-8b5a-eba3ebe5a622
+          provider:
+            opa-spec:
+              rego: |
+                package validate
+                validate := false
+
+                # Check on destination rule, outlier detection?
+                # -> Doesn't appear that UDS is configured to create destination rules.
```
mjnagel, line 342: Would be helpful for this one/any others if the Lula team could point to the control text / what might be expected for us to get to "compliant" (maybe in issues?).

src/istio/oscal-component.yaml
```diff
@@ -1,671 +1,1451 @@
...
+                \ },\n  \"istio-passthrough-gateway/passthrough-gateway\": [],\n  \"istio-tenant-gateway/tenant-gateway\":
+                {\n    \"keycloak-tenant-remove-private-paths-from-public-gateway\",\n    \"keycloak-tenant-public-auth-access-with-optional-client-certificate\",\n
+                \   \"keycloak-tenant-emulate-gitlab-authorize-endpoint\", \"keycloak-tenant-emulate-gitlab-user-endpoint\",\n
+                \   \"keycloak-tenant-emulate-gitlab-token-endpoint\",\n  },\n}\n\n# Helper
+                to find Gateways\ngateways = {sprintf(\"%s/%s\", [gw.metadata.namespace, gw.metadata.name])
+                |\n  gw := input.gateways[_]\n}\n\n# Helper to find VirtualServices\nvirtual_services
+                = {vs.metadata.name |\n  vs := input.virtualServices[_]\n}\n\n# Helper to list
+                VirtualServices using each Gateway\ngateway_virtual_services := {gw: {vs.metadata.name
+                |\n  vs := input.virtualServices[_]\n  gw_name := vs.spec.gateways[_]\n  gw_name
+                == gw\n} |\n  gw := gateways[_]\n}\n"
```
mjnagel: Not sure if this was a generation issue but this is pretty rough to read 😅

src/istio/oscal-component.yaml
```diff
@@ -1,671 +1,1451 @@
...
+                  check_metrics_enabled.result
+                }
+                msg = check_metrics_enabled.msg
+
+                check_metrics_enabled = { "result": true, "msg": msg } {
+                  input.istioConfig.enablePrometheusMerge
+                  msg := "Metrics logging supported"
+                } else = { "result": false, "msg": msg } {
+                  msg := "Metrics logging not supported"
+                }
```
mjnagel: Can't tell if this already accounts for this, but Istio [enables this by default](https://istio.io/latest/docs/ops/integrations/prometheus/#option-1-metrics-merging) so "not set" or true should be acceptable here.

tasks/test.yaml
```diff
@@ -31,6 +32,15 @@ tasks:
...
       - task: validate-packages
+      - task: compliance:validate
+        with:
+          oscalfile: ./compliance/oscal-component.yaml
+          assessment_results: ./compliance/oscal-assessment-results.yaml
+      - task: compliance:evaluate
+        with:
+          assessment_results: ./compliance/oscal-assessment-results.yaml
+      - description: copy assessment file to log location
+        cmd: cp ./compliance/oscal-assessment-results.yaml /tmp/oscal-assessment-results.yaml
```
mjnagel: I see this being added to the pipeline log artifacts, but wondering if there is a desire/intent to re-commit the result to the repo in a given PR as well?

tasks/test.yaml
```diff
@@ -31,6 +32,15 @@ tasks:
       - task: create:k3d-standard-bundle
       - task: deploy:k3d-standard-bundle
       - task: validate-packages
+      - task: compliance:validate
+        with:
+          oscalfile: ./compliance/oscal-component.yaml
+          assessment_results: ./compliance/oscal-assessment-results.yaml
+      - task: compliance:evaluate
+        with:
+          assessment_results: ./compliance/oscal-assessment-results.yaml
```
mjnagel: I think it would be helpful to have these pulled into a separate step in the workflow, rather than in the same step. As-is the output gets a bit lost in the 3000+ lines of the install/test (see [here](https://github.com/defenseunicorns/uds-core/actions/runs/9779618133/job/26999348205?pr=496#step:5:3497)).

Related - it would also be great if the evaluation were pulled into a separate workflow/job, using the assessment output. That would enable us to not clearly show the test pipeline as passed even if the compliance pipeline fails. Does that seem feasible/useful?

mjnagel (changes requested):
Really appreciate all the work that went into this - sorry for the slow review here. One question for my understanding...I think synchronously you had mentioned there being a control threshold/count that is used for the pass/fail result of evaluate. Is that still accurate/where would I see that in these files?

src/istio/oscal-component.yaml
brandtkeller: Yes this is a constraint of OSCAL currently - as such "composed" OSCAL is when Lula takes validations in standalone files and creates a new OSCAL document that contains everything needed to validate within a single OSCAL file. 

We have an issue for "decomposing" these such that the workflow can be improved for development/review purposes.

src/istio/oscal-component.yaml
brandtkeller: agreed - we'll be refining the current repository where all validation development is done (currently private to the org) -> looking to make it public -> and then using it to capture iterations.

tasks/test.yaml
brandtkeller: Agreed - will revise

tasks/test.yaml
brandtkeller: @CloudBeard had the same thought. 

I'm open that option - largely the guidance at a minimum should be to update your threshold anytime the compliance state is greater (better) than the current threshold. We're working on better ways to identify when that greater compliance state is available. 

That said - wanted to introduce the minimum with regards to workflows.

src/istio/oscal-component.yaml
mjnagel: Awesome, yeah I figured as much just wanted to comment for my learning here.

tasks/test.yaml
mjnagel: Maybe to ask a slightly different question here - I see a committed `compliance/oscal-assessment-results.yaml`. If the one generated in CI differs from what is committed is there an expectation that we would update the committed one? Just want to understand the role of what is committed vs what is in CI.

brandtkeller:
> Really appreciate all the work that went into this - sorry for the slow review here. One question for my understanding...I think synchronously you had mentioned there being a control threshold/count that is used for the pass/fail result of evaluate. Is that still accurate/where would I see that in these files?

The `./compliance/oscal-assessment-results.yaml` artifact contains a list of `results`. For each result there is a `finding` that relates to a given control (such as `ac-4`). Each `finding` contains a `target.status.state` of either `satisfied/not-satisfied`.

The single result currently in that file is the current threshold - all results that follow must meet compliance with this threshold until a better threshold is established. 

if you search `state: satisfied` in the `oscal-assessment-results.yaml` file - you will find the current passing controls. Lula compares new results (output of `lula validate`) to ensure these controls are still passing.

There is a lot loaded into this - and a lot that we are looking to improve. Appreciate the patience.

tasks/test.yaml
brandtkeller: To capture intent: 

1. CI will generate a historical artifact (a new result in the assessment-results artifact) on each run with `lula validate`
2. if you look in the CI artifact that is uploaded - you will now see `two` results (the original threshold committed in this PR and the latest result established in the pipeline)
3. `lula evaluate` will compare the findings and pass the pipeline if the compliance is equal or greater - or fail the pipeline if compliance is lesser
4. The current guidance is that the CI artifact should only need to be re-committed to the repo when the latest `validate` produced a state of greater compliance (IE we need to set this as the new threshold)
5. Lula will automatically set the new threshold identifier in the CI artifact - so it just needs to be re-committed. 

There is no requirement currently that we store every single execution - maybe that changes in the future.

mjnagel: Awesome, that helps my understanding a ton. I'll resolve this thread, I think the way you explained it makes sense that we wouldn't add an auto-commit here right now to simplify the changes.

.github/workflows/lint-oscal.yaml
brandtkeller: Removed in favor of the setup action

.github/workflows/lint-oscal.yaml
brandtkeller: Removed in favor of the setup action

.github/workflows/lint-oscal.yaml
brandtkeller: Updated to `pull_request` with the same path.

.github/workflows/lint-oscal.yaml
brandtkeller: Added `--no-progress`

src/istio/oscal-component.yaml
brandtkeller: Good eye - fixed and updated

tasks/test.yaml
brandtkeller: updated - pulled each into individual steps. Current state is all in the same workflow - as the assessment-results are temporary artifacts currently - which means we'd need to play the upload/download game. 

Not opposed to revising further.

src/istio/oscal-component.yaml
brandtkeller: Created an issue in our backlog to review.

src/istio/oscal-component.yaml
brandtkeller: great point - added as an issue in our backlog to enhance.

tasks/test.yaml
mjnagel: I think this works for now, this does make compliance checks a hard blocker on PR/release. I think this is okay since we have a few paths forward in that scenario depending on the problem (1) update the uds-core code to fix compliance issues, (2) update the validation to check something different, or worst case (3) modify the threshold.

The actions for upload/download artifacts are generally easy to work with and viable here if you are interested in changing more. You'd also need to move evaluate "up a layer" (to [here](https://github.com/defenseunicorns/uds-core/blob/458_lula-integration-istio/.github/workflows/pull-request-conditionals.yaml)) I think to handle the dependencies between jobs as expected. Okay pushing this off, just figured I'd reply here.

compliance/oscal-assessment-results.yaml
```diff
@@ -0,0 +1,1275 @@
...
+                validate.msg_expected: Expected VirtualServices using Gateways: {"istio-admin-gateway/admin-gateway": {"grafana-admin-grafana-80-grafana", "keycloak-admin-admin-access-with-optional-client-certificate", "neuvector-admin-neuvector-8443-neuvector-service-webui"}, "istio-passthrough-gateway/passthrough-gateway": [], "istio-tenant-gateway/tenant-gateway": {"keycloak-tenant-emulate-gitlab-authorize-endpoint", "keycloak-tenant-emulate-gitlab-token-endpoint", "keycloak-tenant-emulate-gitlab-user-endpoint", "keycloak-tenant-public-auth-access-with-optional-client-certificate", "keycloak-tenant-remove-private-paths-from-public-gateway"}}
```
mjnagel: Skimming this output and noticed this one - we recently removed the `keycloak-tenant-emulate-gitlab-*` virtualservices, I think those are the cause of the error here? Likely makes sense to update that list since we removed those entirely but also made me think of some other questions...
- Is that list hardcoded somewhere (just want to understand the effects of updating names/adding new virtualservices and where we may have to change also)?
- Should this validation allow for virtualservices not existing, as long as all the existent ones use the correct gateways?
- Maybe not today's problem but if someone deployed SWF/other tools on top of core would that fail these validations?

brandtkeller: This was a large discussion today actually - how flexible are the validations
- These are hardcoded - some rightfully so and others are still a WIP
- This identified example is one we are going to change now - it is TOO hardcoded

overall we need to get more feedback on what works and doesn't work. This example is detracting from greater compliance by being too specific. So we're updating some of our internal review guidelines (also WIP) to look for validations that cannot "scale" well as we put others apps on uds-core. 

We'll also get a fix in for the istio prometheus item you identified earlier. Thanks for helping us BEGIN to refine this process (and even how we review internally).

tasks/test.yaml
brandtkeller: agreed - I wanted to revisit this. Could this be a good space to implement [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) for the evaluate step? 

End goal would be to accept the constraint of this step blocking PR/release - but I believe we have enough discovery to do to warrant some soak-testing.

mjnagel: I think that would make me more comfortable, and would also resolve the issue with evaluate failing and resulting in the `cp` not happening (still an issue in the branch current state since its a sequential step). Definitely agree with the end goal, don't want to make compliance be an afterthought/"we'll fix that later" but in the near term I think it's the right move to not block on it. 

If we do flip on `continue-on-error` it would be nice to still identify failures clearly (since they won't show up in the action status anymore). What do you think about PR comments on failure? `gh` cli provides an [easy way](https://cli.github.com/manual/gh_pr_comment) to add comments, but there are also [actions that we could leverage](https://github.com/peter-evans/create-or-update-comment) to do more advanced things (i.e. make sure we only comment once + update comment). Don't want to over-engineer on a short-term change (hopefully) but also want to make sure we don't just completely ignore compliance failures because they are out of sight.

brandtkeller: I didn't want to inject too much in the way of dependencies required to maintain - that said something like that comment action would provide an easy interface for tagging the lula team in to explore the problem and evaluate updates in a non-blocking fashion. 

at a minimum it would provide a stop-gap to addressing compliance failures until another solution was identified that reduced dependencies required. 

Next steps:
- `continue-on-error` to allow failure which unblocks the next step of copying the artifacts
- Implement some process to notify when failure occurred and surface it.

compliance/oscal-assessment-results.yaml
brandtkeller: Updated this control to capture higher-level intent of gateways and network flow separation. Easing the restrictions while still collecting evidence. This can be refined more in the future when we can put more guardrails in place. 

Also updated the prometheus validation and subsequently the threshold with 11 controls now satisfied.

tasks/test.yaml
brandtkeller: `continue-on-error` added alongside a `notify-lula` action. I believe this could be made more generic but I don't have evidence to support that being required currently. 

Workflow exists such that:
- evaluate will always continue-on-error
- if failed - a comment will be made on the pr tagging the @defenseunicorns/lula-dev team
- artifact will always be stored for review
- if the next iteration passes - the comment will be updated to indicate passing
- should continue to update the comment as required until merge

brandtkeller:
Currently we have two options for implementation:

## Current Solution
- deploy / test / validate / evaluate as steps in the same job
  - this makes the traceability between validate / evaluate and the corresponding artifacts concise

The results of this are good for traceability but poor for CI feedback outside of a comment from the github actions bot.

## Alternative Solution
- Remove evaluate from current deploy / test / validate workflow
- Upload the artifact using the upload action
- establish a separate job that performs a matrix of download / evaluate / comment 

The results of this should be non-blocking but failing job (Red X) with notification in the comments but requires greater understanding of the artifacts location in CI and how to debug. 

## Proposal

The alternative solution provides more situational awareness in my opinion.

brandtkeller:
Moving to draft to update validations post updates to uds-core

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -85,3 +85,14 @@ jobs:
       flavor: ${{ matrix.flavor }}
       test_type: ${{ matrix.test_type }}
     secrets: inherit # Inherits all secrets from the parent workflow.
+
+  evaluate-package-compliance:
+    needs: run-package-test
+    name: Compliance Evaluation
+    strategy:
+      matrix:
+        flavor: [upstream, registry1]
```
mjnagel: 🫣 this should now also include `unicorn` with the addition of the unicorn flavor

.github/actions/notify-lula/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+      if: ${{ steps.fc.outputs.comment-id != '' }}
+      uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043 # v4.0.0
+      with:
+        comment-id: ${{ steps.fc.outputs.comment-id }}
+        token: ${{ inputs.ghToken }}
+        edit-mode: replace
+        body: |
+          Compliance ${{ inputs.flavor }} Evaluation: ${{ inputs.state }}
+
+          CC: @defenseunicorns/lula-dev
```
mjnagel, line 51: What do you think about making it more clear that this is an update to a previous comment? At first glance [this comment](https://github.com/defenseunicorns/uds-core/pull/496#issuecomment-2226516061) feels weird that the unicorn flavor is called out but the others are not - wondering if a reaction or appending to the comment makes more sense?

Definitely a nit and would defer to whatever helps your team most.

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -85,3 +85,14 @@ jobs:
       flavor: ${{ matrix.flavor }}
       test_type: ${{ matrix.test_type }}
     secrets: inherit # Inherits all secrets from the parent workflow.
+
+  evaluate-package-compliance:
+    needs: run-package-test
+    name: Compliance Evaluation
+    strategy:
+      matrix:
+        flavor: [upstream, registry1, unicorn]
```
mjnagel, line 95:
```suggestion
    strategy:
      matrix:
        flavor: [upstream, registry1, unicorn]
      fail-fast: false
```

Suggestion - I think this makes sense? Would want all compliance evaluated each time vs cancelling parallel flavors on one failing (see https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures).

mjnagel (dismissed):
Two small comments but this LGTM! Thanks for all the hard work here @defenseunicorns/lula-dev

.github/actions/notify-lula/action.yaml
brandtkeller: Yeah I considered more complex workflows for appending. You see the unicorn comment by itself because the other two live persistently up further in the commits. 

I'm open to suggestions here - it could be one comment that gets edited / appended / etc - it could be multiple as it is right now.

Mainly I wanted to balance complexity and confusion - so there still may be work to do here.

Current process:
- Comment only appears after an evaluation has failed
- One comment will be made (per flavor) for the life of a PR
- That comment will be updated on each subsequent run (whether pass or fail)
- This signifies the current state

.github/workflows/pull-request-conditionals.yaml
brandtkeller: Given that the preceding job `run-package-test` is required for this job to execute - the runtime of this job means that everything to this point has passed yes? 

If so, given the overhead of `evaluate` which is merely static file processing (seconds at most) - I could see value in always having all flavors evaluated and available for comparison. One failing might signify a problem with a specific flavor vs all failing might be more of a systemic issue.

Open to thoughts

mjnagel: Yeah I agree with all of the above - which I think means you're aligned with my suggestion? 😅 

As-is `fail-fast` is `true` by default. Specifically for this suggestion I think this means that all evaluations run in parallel but if one fails the others will be cancelled before they complete. In practice I'm not sure I've seen that (ex: [this one](https://github.com/defenseunicorns/uds-core/actions/runs/9942641965) seemed to just run both to completion) - possibly because the jobs take almost identical time to run and there is no time to cancel the others?

What I wouldn't want is a situation where the unicorn flavor evaluation fails and causes the registry1 evaluation to be cancelled - leading to no comparison of results. Would always be beneficial to have all results, as you noted if all flavors are failing there may be a systemic issue vs flavor specific.

brandtkeller: 💯 I have watched these pipelines execute many times and never observed that cancel happening -> but it could certainly be latency and GitHub unable to cancel them fast enough. Agree with the suggestion.

.github/actions/notify-lula/action.yaml
mjnagel: Let's leave it as-is for now and iterate if needed unless you can think of a simple change here.

I think the main thing that feels a bit weird is that comments only appear on failure, but if the checks later pass you end up with a comment that says "Success" without the history of why that comment is there (i.e. because it previously failed).

mjnagel (dismissed):

tasks.yaml
```diff
@@ -110,6 +110,14 @@ tasks:
...
       - task: test:uds-core-upgrade
 
+  - name: test-compliance-validate
+    actions:
+      - task: test:compliance-validate
+
+  - name: test-compliance-evaluate
+    actions:
+      - task: test:compliance-evaluate
+
```
rjferguson21: Can we add descriptions of these commands at this level?

src/istio/oscal-component.yaml
```diff
@@ -1,671 +1,1395 @@
...
-        links:
-          - href: https://defenseunicorns.com
-            rel: website
+  back-matter:
+    resources:
+      - rlinks:
+          - href: https://github.com/istio/istio/
+        title: Istio Operator
+        uuid: 60826461-D279-468C-9E4B-614FAC44A306
+      - description: |
```
rjferguson21, line 8: If we intend for the source of truth for this to be maintained outside of this repo I would love if we were:
* remotely referencing these validations
* tying this to a released version of the validation

Some of that might involve questions about sacrificing consuming this in the airgap but if I'm understanding all of the pieces correctly we could perhaps run `lula tools compose` in CI and capture that as some kind of release artifact.

tasks.yaml
brandtkeller: for sake of clarifying intent - are you requesting the `description` field be added to these?  

IE 
```yaml
- name: test-compliance-evaluate
  description: Evaluation of compliance given an established threshold
  actions:
    - task: test:compliance-evaluate
```

or something else?

src/istio/oscal-component.yaml
brandtkeller: great feedback - I'd like to consider what the implications involved here are as maybe beyond scope of this first iteration. 

I agree fully that we want to tie a given artifact back to a versioned source. We need to re-work our expectations around where OSCAL lives. Current discussion we've been having is removing the middle-man - publishing OSCAL in target locations (uds-core for instance) and then having `compliance-artifacts` be solely for validations. If we do this then yes I agree if others want to introduce "composing" later in the lifecycle.

tasks.yaml
rjferguson21: Yep.

brandtkeller: Added

src/istio/oscal-component.yaml
mjnagel: I think I agree with the sentiment here - current state would make it hard for me as a dev on uds-core to easily identify source of truth for validations/oscal and know where to contribute meaningfully.

I don't think I'm opposed to the source of truth being external, but it would be great if that were referenced here (and versioned as Rob mentioned). I'm okay with this as the first iteration, I think it falls in the same bucket as "how do I identify/triage failures" - it's workable but not ideal. Open to other takes though - probably should be tracked in an issue for further discussion/working if we do punt on it (unsure if that should be on core or compliance repo).

brandtkeller: Added as additional context in a current release issue.

brandtkeller: Other additional context here - we are adding a constraint on network connectivity for every execution (3 flavors) for every validation we develop. These will be growing over-time but we have not stress-tested what this looks like or the overlay of github runners network reliability. 

We can buy down on some of that if we performed a compose in CI -> uploaded the artifact -> and then downloaded the composed version for each execution. More CI steps but we may very well be producing the composed artifact at some point anyway..

rjferguson21: @brandtkeller Can you link to the issue?

brandtkeller: Current placeholder [issue](https://github.com/defenseunicorns/compliance-artifacts/issues/51) for that work with additional thoughts and links.

rjferguson21 (approved):
Approving with intent to track https://github.com/defenseunicorns/compliance-artifacts/issues/51

mjnagel (approved):

PR #497, opened 2024-06-20
feat: identity group auth

chance-coleman:
## Description
Identity Plugin Pepr component for managing group authorization to applications. This is only the implementation, meaning a package (i.e. grafana) will need to be configured for this to work. This is also dependent on a new release of uds-identity-config. 

## How to test
Update the identity config image [here](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/zarf.yaml#L24) and [here](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/values.yaml#L10) to include/change : `ttl.sh/uds-core-config:group-plugin-groups-attribute` 

In addition to the identity config image, update the grafana package to require a specific group to access the application like so:

```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: grafana
spec:
  sso:
    - name: Grafana Dashboard
      redirectUris:
      groups:
        anyOf:
          - /UDS Core/Admin
```

With those changes utilize the `test-uds-core` task to test, no user is created by default so either go into keycloak and setup the user manually or go to `sso.uds.dev` and register a user there, then navigate to the `grafana.admin.uds.dev` and you should be redirected to a login page and depending on if you're in the admin group or not be granted access to the grafana dashboard. 

## Related Issue
Relates to `uds-identity-config` [PR](https://github.com/defenseunicorns/uds-identity-config/pull/107)

## Type of change

## Checklist before merging


src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -89,9 +89,11 @@ async function syncClient(
     // If an existing client is found, update it
     if (token && !isRetry) {
       Log.debug(pkg.metadata, `Found existing token for ${clientReq.clientId}`);
+      handleClientGroups(clientReq);
       client = await apiCall(clientReq, "PUT", token);
     } else {
       Log.debug(pkg.metadata, `Creating new client for ${clientReq.clientId}`);
+      handleClientGroups(clientReq);
       client = await apiCall(clientReq);
     }
```
bburky: This is in both sides of the if/else here. Pull out `handleClientGroups()` and put it above the if/else.

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -142,6 +144,18 @@ async function syncClient(
...
+ * Handles the client groups by converting the groups to attributes.
+ * @param clientReq - The client request object.
+ */
+function handleClientGroups(clientReq: Sso) {
+  if (clientReq.groups) {
+    clientReq.attributes = clientReq.attributes || {};
+    clientReq.attributes["uds.core.groups"] = JSON.stringify(clientReq.groups);
+    delete clientReq.groups;
+  }
+}
```
bburky: I would still like to see us put an actual allowlist on the allowed keys for `attributes`. Technically that can go in a separate PR, but if we don't do it now please create an issue.

bburky (dismissed):

src/pepr/operator/controllers/keycloak/client-sync.ts
chance-coleman: updated

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -142,6 +144,18 @@ async function syncClient(
   }
 }
 
+/**
+ * Handles the client groups by converting the groups to attributes.
+ * @param clientReq - The client request object.
+ */
+function handleClientGroups(clientReq: Sso) {
```
rjferguson21: This needs to handle cases where the `groups:` field is removed.   Might need to test the best way to "unset" the attributes within the Keycloak API.

chance-coleman: updated

src/pepr/operator/controllers/keycloak/client-sync.ts
chance-coleman: issue created: https://github.com/defenseunicorns/uds-core/issues/506

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -303,6 +303,19 @@ const sso = {
           type: "string",
         },
       },
+      groups: {
+        description: "The client sso group type",
+        type: "object",
+        properties: {
+          anyOf: {
+            description: "List of group allowed to access to client",
```
mjnagel:
```suggestion
            description: "List of groups allowed to access to client",
```

Also might be good to add the CRD gen task fix here so we're not carrying around that issue while modifying the CRD type, can steal from my draft PR - https://github.com/defenseunicorns/uds-core/pull/511/files#diff-76783f82b8fa78a37e73d203db3f8168d44cafb1b52fa7dc1df445e68198b7b7

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -142,6 +144,21 @@ async function syncClient(
   }
 }
 
+/**
+ * Handles the client groups by converting the groups to attributes.
+ * @param clientReq - The client request object.
+ */
+export function handleClientGroups(clientReq: Sso) {
+  if (clientReq.groups) {
```
mjnagel: There's two scenarios that might not be the most intuitive experience for end users:

```yaml
      groups:
        anyOf: []
```

Using this block results in a client that is not usable by any users, which I think makes sense. It's probably not a common use case but I could see someone doing this if they had a new app and weren't ready to open it to any users yet.

```yaml
      groups: {}
```

This one also ends up resulting in an unusable client. That feels a bit less intuitive, but again someone would have to explicitly put an empty groups so maybe its okay? The other option I think would be to change our conditional here to something like:

```suggestion
  if (clientReq.groups?.anyOf) {
```

In the future if we add `allOf` or similar we would just need to manage that with `||` conditionals here. What we decide to do here might be related to https://github.com/defenseunicorns/uds-identity-config/issues/116?

src/pepr/operator/crd/sources/package/v1alpha1.ts
chance-coleman: updated the typo and the gen crd task

src/pepr/operator/controllers/keycloak/client-sync.ts
mjnagel: Updating this thread after further testing this evening - the missing piece was also needing to delete `groups` from `clientReq` as part of the else (since `else` could have `groups: {}`). I'm okay with either path for now, but with the change to this:
```ts
export function handleClientGroups(clientReq: Sso) {
  if (clientReq.groups?.anyOf) {
    clientReq.attributes = clientReq.attributes || {};
    clientReq.attributes["uds.core.groups"] = JSON.stringify(clientReq.groups);
    delete clientReq.groups;
  } else {
    clientReq.attributes = clientReq.attributes || {};
    clientReq.attributes["uds.core.groups"] = ""; // Remove groups attribute from client
    delete clientReq.groups; // <--- this was the missing piece
  }
}
```

I was able to run successfully. `groups: {}` does not provide group protection, but `groups.anyOf: []` does.

chance-coleman: updated

mjnagel (approved):

PR #507, opened 2024-06-25
feat: add unicorn flavor to uds-core

mjnagel:
## Description

Adds the unicorn flavor (Chainguard images currently) to uds-core with CI testing/publishing.

Validated the publish workflow changes on a fork. Note that create was skipped due to lack of credentials on the fork so the workflow is running a "dry-run" publish. Take particular note of the `Determine destination repository` job and the dry run publish commands:
- Snapshot: https://github.com/BagelLab/uds-core/actions/runs/9910149555
- Normal Release: https://github.com/BagelLab/uds-core/actions/runs/9910127730

Note: An issue has been opened to switch NeuVector from upstream -> chainguard (https://github.com/defenseunicorns/uds-core/issues/568). Upstream was chosen instead of registry1 to ensure arm64 support for the full unicorn flavor.

## Related Issue

N/A

## Type of change

## Checklist before merging


src/neuvector/chart/templates/_helpers.tpl
```diff
@@ -60,3 +60,19 @@ Create the name of the service account to use
...
+{{- $secretData := (lookup "v1" "Secret" .namespace .secret).data  -}}
+{{- if and $secretData (hasKey $secretData .key) -}}
+  {{- $value = index $secretData .key -}}
+{{- else if .defaultValue -}}
+  {{- $value = .defaultValue | toString | b64enc -}}
+{{- end -}}
+{{- if $value -}}
+{{- printf "%s" $value -}}
+{{- end -}}
+{{- end -}}
```
mjnagel:
```suggestion
```

.github/workflows/publish.yaml
```diff
@@ -70,6 +72,7 @@ jobs:
         if: ${{ always() && !inputs.snapshot }}
         uses: ./.github/actions/debug-output
 
+      # todo: publish chainguard to a private repository
```
mjnagel:
```suggestion
      # todo: publish chainguard to a separate private repository
```

src/neuvector/zarf.yaml
```diff
@@ -48,3 +48,39 @@ components:
...
+        valuesFiles:
+          - values/upstream-monitor-values.yaml
+    images:
+      - docker.io/neuvector/controller:5.3.3
+      - docker.io/neuvector/manager:5.3.3
+      - docker.io/neuvector/updater:latest
+      - docker.io/neuvector/scanner:latest
+      - docker.io/neuvector/enforcer:5.3.3
+      - docker.io/neuvector/prometheus-exporter:5.3.2
+    # charts:
```
mjnagel, line 75:
```suggestion
    # todo: switch to chainguard images once functional
    # charts:
```

mjnagel (commented):
Updating here, still waiting on the chainguard NeuVector manager image to be fully functional. We also need to determine a renovate path forward and make sure that all images are aligned with the latest in `main`.

.github/workflows/publish.yaml
```diff
@@ -70,21 +72,36 @@ jobs:
...
+            fi
+          # Publish snapshots to snapshot repository
+          elif [ "${{ inputs.snapshot }}" = "true" ]; then
+            echo "Publishing snapshot to ghcr.io/defenseunicorns/packages/uds/snapshots"
+            echo "repo=ghcr.io/defenseunicorns/packages/uds/snapshots" >> "$GITHUB_OUTPUT"
+          # Publish all other packages/bundles to uds package repository
+          else
+            echo "Publishing packages and bundles to ghcr.io/defenseunicorns/packages/uds"
+            echo "repo=ghcr.io/defenseunicorns/packages/uds" >> "$GITHUB_OUTPUT"
+          fi
```
rjferguson21: If its not too much of a pain you could build `repo` as a variable and end up with something like:
```
repo=ghcr.io/defenseunicorns/packages

if (unicorn) {
  repo += /private
} 

repo += /uds

if (snapshot) {
  repo += /snapshots
}
```

mjnagel: I like that, will update and validate on my testing fork.

rjferguson21: nit: I would also consider if we should just move this logic to compute the TARGET_REPO into the publish task?  Overall the dumber our GH actions are the better but understand it would be a trade off passing a bunch of inputs around.

mjnagel: Modified the repo construction var and updated the test pipeline links in the PR description. On moving the logic to the publish task I am leaning towards no at least for now:
- Publish should only ever be done in CI/actions so I'm not as concerned with that logic being actions specific
- With the refactor you suggested this logic is a lot simpler and very portable if we did need to move to GitLab or some other CI system

rjferguson21 (approved):

PR #508, opened 2024-06-25
feat(neuvector): auto promote from discover to monitor in 36 hrs

MxNxPx:
## Description

By default, Neuvector is configured to run in ["Discover" mode](https://open-docs.neuvector.com/5.2/policy/modes#discover).  This change toggles the behavior so that after a period of time (36 hours), Neuvector will adjust to run in ["Monitor" mode](https://open-docs.neuvector.com/policy/modes/#monitor).  This occurs as part of the Neuvector [automated promotion of group modes](https://open-docs.neuvector.com/policy/modes/#automated-promotion-of-group-modes
) feature.

## Related Issue

N/A

## Type of change

## Checklist before merging


src/neuvector/values/values.yaml
```diff
@@ -20,6 +20,8 @@ controller:
         no_telemetry_report: true
         scan_config:
           auto_scan: true
+        mode_auto_d2m: true
+        mode_auto_d2m_duration: 3600
```
mjnagel: Do you think 24 hours might be a safer bet for workloads? Maybe overly cautious since it's just monitor mode but I'd suspect some of our workloads/end user workloads might do things on a schedule that is greater than 1 hour (loki indexing comes to mind, unsure if that would trigger different processes/network calls though).

MxNxPx: per our chat, we want to keep the number as low as possible while still trying to capture the various network traffic expected from uds-core pieces.  so we were thinking this probably ought to be at least 24-48 hours.  so planning to just split the difference, and go with 36 hours.

mjnagel (approved):

PR #511, opened 2024-06-26
fix: operator retries and error logging

### LINKED ISSUE #448: Pepr error event/msg on client creation failure can be misleading

Currently we have a catchall for all any Keycloak client creation failures that logs an error "Failed to process client request ... This can occur if a client already exists with the same ID that Pepr isn't tracking.".

This error is often misleading because it can surface for any number of reasons that the client creation was invalid (client scopes, invalid URIs, etc). The result is a confusing error message and the real error message is only visible in the Keycloak logs themselves. It would be good to revisit this catch all and see if we can filter errors or better represent the true failure on the keycloak call.

https://github.com/defenseunicorns/uds-core/blob/434844b827e01808b504abf5ee6af83fba813cb6/src/pepr/operator/controllers/keycloak/client-sync.ts#L128-L142

### LINKED ISSUE #492: `Package` retries can result in `Package`s getting stuck in Pending
### Steps to reproduce
1. Create a package with a known good state.
2. Wait for it to go to `Ready`.
3. Modify the package to a known bad state (ex: use a uri in the sso client id)
4. Observe that the package gets stuck in Pending without doing additional retries.

### Expected result

Package should retry up to 5 times and then either get marked as Ready or Failed.

### Actual Result

Package is stuck in Pending until a restart of the watcher pod.

### Severity/Priority

Medium - this is very confusing to end users on what is happening.

### Additional Context

Our [shouldSkip](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/reconcilers/index.ts#L26-L29) logic filters out pending packages. During a retry we [leave the status in Pending](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/reconcilers/index.ts#L106-L113), resulting in it getting stuck unless it hits other cases (max retries or first time the watcher has seen it).

### LINKED ISSUE #525: `Package` stuck in `Failed` <-> `Pending` loop
### Steps to reproduce

This is related to https://github.com/defenseunicorns/uds-core/issues/492 (a similar issue in the retry/failure logic).

1. Create a package with a known bad state (or introduce problems in your cluster such as deleting keycloak).
1. Observe that the package flips between failed and pending states infinitely.

### Expected result

Package should retry up to 5 times and then either get marked as Ready or Failed.

### Actual Result

Package is infinitely stuck in a loop between failed and pending.

### Severity/Priority

High - this issue results in excessive work on the operator and depending on nature of failure could also result in excessive network traffic against the KubeAPI.

### Additional Context

Our [placement of uidSeen.add](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/reconcilers/package-reconciler.ts#L68-L69) means that a Package will never leave this loop unless the operator has successfully reconciled it once. By moving that "seen" logic to be pre-reconcile this bug should be resolved.

mjnagel:
## Description

This PR has a number of changes, mostly tied to operator behaviors and bug fixes around failure logging. Included:
- URI Encoding of client ID on deletion/updates - when we call updates/deletions on KC clients it gets appended to our URL for our request and must be encoded
- Moves around the try/catch to only wrap the Keycloak API call so that errors are surfaced more accurately in events and modifies the thrown error from the Keycloak response to include the keycloak response status and data (see https://github.com/defenseunicorns/uds-core/issues/448)
- Adds a new Phase to Package for `Retrying` - this differentiates from a `Pending` package that is already being reconciled to allow retries to function as expected (see https://github.com/defenseunicorns/uds-core/issues/492)
- Moves uidSeen addition to patch status to handle infinite failure -> pending -> failure ... loop on first apply (see https://github.com/defenseunicorns/uds-core/issues/525)

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/492

Fixes https://github.com/defenseunicorns/uds-core/issues/448

Fixes https://github.com/defenseunicorns/uds-core/issues/525

## Type of change

## Checklist before merging


src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -78,14 +83,12 @@ async function syncClient(
...
 
   try {
-    // Not including the CR data in the ref because Keycloak client IDs must be unique already
-    const name = `sso-client-${clientReq.clientId}`;
     const token = Store.getItem(name);
-
-    let client: Client;
-
     // If an existing client is found, update it
     if (token && !isRetry) {
```
mjnagel, line 96: note: need to check usecase for retry here and confirm that we not be using the token on a retry.

mjnagel: Removed retries - may need to document the current behavior as it is a slight shift.

mjnagel: Also related I have added a finalizer issue to handle one specific edge case here - https://github.com/defenseunicorns/uds-core/issues/523

rjferguson21 (dismissed):

chance-coleman (dismissed):

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -96,52 +100,59 @@ async function syncClient(
...
+
+  // Create or update the client secret
+  await K8s(kind.Secret).Apply({
+    metadata: {
+      namespace: pkg.metadata!.namespace,
+      // Use the CR secret name if provided, otherwise use the client name
+      name: secretName || name,
+      labels: {
+        "uds/package": pkg.metadata!.name,
+      },
```
MxNxPx: @mjnagel - per our convo, perhaps we ought to do [something similar to what is done with netpols](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/policies.ts#L95)  with generation

mjnagel: If you're okay with it @MxNxPx I'm going to open a separate [issue here](https://github.com/defenseunicorns/uds-core/issues/535) to track this since it's not entirely connected to the intent of the changes here. Would prefer to tackle that in a small follow-on PR since I didn't modify the behavior of this in the current PR. Let me know if that works/resolve this thread if you're okay with it?

chance-coleman (approved):

chance-coleman:
Something of note to capture in a future PR, creating and using a UDS Error type that allows for some better error handling and management. 

- tested deleting admin ui clients and verified they are recreated by reconciling package
- tested deleting pepr store toke and new retry / error handling was expected broken state
- tested deleting both token and admin ui clients and reconcolied package and client was successfully recreated 
- tested a few different package related edge cases like deleting and redeploying a package and verified the creation of a new client or in the case of an existing client a broken state
- tested creating a client with invalid syntax to test that putting that token in the pepr store broke things as well

PR #517, opened 2024-06-27
feat: update to using default scrapeclass for tls config

### LINKED ISSUE #417: Add a way to add podMonitors under the `monitor` key.

Currently some application charts (like SonarQube's) do not directly expose a way to have a a metrics endpoint behind a service to create a service monitor and instead would need resources created outside of the current chart to facilitate integration with UDS core.  It would be nice if the operator supported podMonitors under the `monitor` key so that these resources were not needed.

- **Given** I have a pod with an exposed metrics endpoint
- **When** I specify a pod monitor in the package CR under `monitor`
- **Then** the operator makes a pod monitor for that pod

We could create resources outside of the chart that wire things together but it would be simpler to have this feature built in.

### Additional context
https://github.com/defenseunicorns/uds-package-sonarqube/issues/73

zachariahmiller:
## Description

add pod monitors to uds-core operator automation and UDS package CR monitor spec.
update to using default scrapeClass for tls config in prometheus and "exempt" class to override default tls config
update core components existing pod and service monitor implementations to fit with the new default scrapeClass implementation
migrate pepr over to using the generated helm based implementation to facilitate ability to override and align zarf.yaml composition organization with the other packages.
add authorization to the endpoint configuration options for monitors

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/417

## Type of change

## Checklist before merging


zachariahmiller:
needs docs and there are a couple of things to potentially remove that i left for someone if they wanted to test locally in review. Namely additional resources in the app-tenant test app and commented out values in the pepr chart import components values file that doesnt actually need to be there for things to work. @mjnagel If you want to take a look LMK any feedback. 

After changing the behavior for the scrape classes i verified the changes i made to make the existing uds-core monitors match by deploying both from this branch and from main and doing an yaml diff of the prometheus config yaml of both deployments.

zachariahmiller:
Made an initial attempt at the docs. Not sure how yall want to communicate new vs existing/deprecated behavior format wise especially given that these docs feed into the site.

docs/configuration/uds-monitoring-metrics.md
```diff
@@ -4,10 +4,12 @@ type: docs
 weight: 1
 ---
 
-UDS Core leverages Pepr to handle setup of Prometheus scraping metrics endpoints, with the particular configuration necessary to work in a STRICT mTLS (Istio) environment. We handle this with both mutations of existing service monitors and generation of service monitors via the `Package` CR.
+UDS Core leverages Pepr to handle setup of Prometheus scraping metrics endpoints, with the particular configuration necessary to work in a STRICT mTLS (Istio) environment. We handle this via a default scrapeClass in prometheus to handle the istio certs as well as with both mutations of existing service monitors and generation of service monitors via the `Package` CR when a monitor needs to be exempt from that tlsConfig.
```
mjnagel: Wording here seemed off to me. I'm not sure if this is better but the "generation of service monitors" felt a bit out of place.

```suggestion
UDS Core leverages Pepr to handle setup of Prometheus scraping metrics endpoints, with the particular configuration necessary to work in a STRICT mTLS (Istio) environment. We handle this via a default scrapeClass in prometheus to add the istio certs. When a monitor needs to be exempt from that tlsConfig a mutation is performed to leverage a plain scrape class without istio certs.
```

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -52,10 +53,18 @@ export async function packageReconciler(pkg: UDSPackage) {
...
+      if (pkg.spec?.monitor) {
+        for (const monitor of pkg.spec.monitor) {
+          const monitorKind = monitor.kind || "ServiceMonitor"; // Default to "ServiceMonitor" if kind is undefined
+          if (monitorKind === "PodMonitor") {
+            monitors = await podMonitor(pkg, namespace!);
+          } else if (monitorKind === "ServiceMonitor") {
+            monitors = await serviceMonitor(pkg, namespace!);
+          }
+        }
+      }
```
mjnagel: Two issues here I think...
1. The conditional around this (to ensure `pkg.spec.monitor` is non-null) means that we don't run our "cleanup" functions (for orphaned, previous generation monitors) if a Package doesn't specify any monitors. I think you would see this behavior if you deployed a package with a monitor, then edited it to remove all monitors.
2. The existing `serviceMonitor` function loops over the full monitor list (and looks like the `podMonitor` one does too). We should be safe to drop the loop/conditionals from here I think?

src/pepr/operator/crd/index.ts
```diff
@@ -20,17 +20,18 @@ export {
...
   Endpoint as IstioEndpoint,
+  Location as IstioLocation,
   Port as IstioPort,
+  Resolution as IstioResolution,
+  ServiceEntry as IstioServiceEntry,
 } from "./generated/istio/serviceentry-v1beta1";
 
-export * as Prometheus from "./generated/prometheus/servicemonitor-v1";
+export * as PrometheusPodMonitor from "./generated/prometheus/podmonitor-v1";
+export * as PrometheusServiceMonitor from "./generated/prometheus/servicemonitor-v1";
```
mjnagel: Do we need `*` here or could we take a similar approach as the istio ones above and specifically export `ServiceMonitor as PrometheusServiceMonitor`, `PodMonitor as PrometheusPodMonitor`, etc? From a quick scan it seems like we only use those two in the generation and then Endpoint/Scheme in the mutate. I think it would clean names up and be more readable if we used the same `as` approach.

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -202,6 +236,12 @@ const monitor = {
         description: "HTTP path from which to scrape for metrics, defaults to `/metrics`",
         type: "string",
       },
+      kind: {
+        description:
+          "The type of monitor to create; PodMonitor or ServiceMonitor. ServiceMonitor is the default.",
+        type: "string",
```
mjnagel, line 243: This should be an enum rather than a string. If/when you switch to enum you would be able to use that type in the controller conditionals.

src/pepr/operator/controllers/monitoring/pod-monitor.ts
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @param pkg UDS Package
+ * @param namespace
+ */
+export async function podMonitor(pkg: UDSPackage, namespace: string) {
+  const pkgName = pkg.metadata!.name!;
+  const generation = (pkg.metadata?.generation ?? 0).toString();
+  const ownerRefs = getOwnerRef(pkg);
+
+  Log.debug(`Reconciling PodMonitors for ${pkgName}`);
```
mjnagel: We just switched our logging approach to use our own logger rather than the default Pepr one. That PR might be a good place to see the diffs - https://github.com/defenseunicorns/uds-core/pull/533/files. For you the changes should just be import `Component` and `setupLogger` from our `logger` file (`import { Component, setupLogger } from "../../../logger";` should be right), then add `const log = setupLogger(Component. OPERATOR_MONITORING);`. And then just switch `Log` -> `log` and it should all work the same way.

I think in the other files where there were conflicts I resolved and switched to the new `log`, but might be good to validate.

src/pepr/zarf.yaml
```diff
@@ -0,0 +1,29 @@
...
+          - cmd: ./zarf tools kubectl annotate role -n pepr-system pepr-uds-core-store meta.helm.sh/release-name=module --overwrite || true
+          - cmd: ./zarf tools kubectl annotate rolebinding -n pepr-system pepr-uds-core-store meta.helm.sh/release-name=module --overwrite || true
+          - cmd: ./zarf tools kubectl annotate service -n pepr-system pepr-uds-core meta.helm.sh/release-name=module --overwrite || true
+          - cmd: ./zarf tools kubectl annotate service -n pepr-system pepr-uds-core-watcher meta.helm.sh/release-name=module --overwrite || true
+          - cmd: ./zarf tools kubectl annotate deployment -n pepr-system pepr-uds-core meta.helm.sh/release-name=module --overwrite || true
+          - cmd: ./zarf tools kubectl annotate deployment -n pepr-system pepr-uds-core-watcher meta.helm.sh/release-name=module --overwrite || true
+          - cmd: ./zarf tools kubectl annotate mutatingwebhookconfiguration -n pepr-system pepr-uds-core meta.helm.sh/release-name=module --overwrite || true
+          - cmd: ./zarf tools kubectl annotate validatingwebhookconfiguration -n pepr-system pepr-uds-core meta.helm.sh/release-name=module --overwrite || true
```
mjnagel: I really don't love this, but our hands are a bit tied since our other option with more intelligence (pepr) is the problem here. Making a note mostly for myself, I think this is probably the best path forward, just want to think through other options.

src/test/app-tenant.yaml
```diff
@@ -23,6 +60,47 @@ spec:
...
+      selector:
+        app: "example"
+      targetPort: 8080
+      authorization:
+        credentials:
+          key: "example-key"
+          name: "example-secret"
+          optional: false
+        type: "Bearer"
+      description: Service Monitor with Authorization
```
mjnagel, line 103: I think is fine for now just to run through every case in the operator, we'll want to switch this to different app(s) if/when we actually check the health of the monitor targets.

mjnagel (changes requested):
All targets in core show as healthy on a deploy - hopefully most of these comments are small modifications.

src/pepr/zarf.yaml
zachariahmiller: yeah i dont love it either, but it works. Realistically its a temporary fix and can be removed after a couple of minor releases.

zachariahmiller:
@mjnagel I believe i resolved all your comments. Please take a look again when you get a chance.

mjnagel (dismissed):

src/metrics-server/chart/templates/service-monitor.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:
-  annotation:
+  annotations:
```
mjnagel, line 6: Ty for fixing my problem ❤️ - for awareness to others this was the issue with metrics-server. Zarf/helm do not explode on extra fields in a manifest so this being incorrect was obscured.

mjnagel (dismissed):

tasks/create.yaml
```diff
@@ -63,4 +63,4 @@ tasks:
           CUSTOM_PEPR_IMAGE=$( [ "${FLAVOR}" = "registry1" ] && echo "--custom-image ${REGISTRY1_PEPR_IMAGE}" ) || CUSTOM_PEPR_IMAGE=""
           rm -fr dist
           npm ci
-          npx pepr build $CUSTOM_PEPR_IMAGE
+          npx pepr build $CUSTOM_PEPR_IMAGE -z chart
```
mjnagel: It seems like the chart does not pull our `env` from the `package.json` so we will either need to:
1. Revert this change, just use the manifest for now.
2. Add override values in our zarf.yaml that specify all the default env + our additional ones. This is viable but I'd be worried about us missing changes to the defaults on upgrades.

I think I'd lean towards 1, and we should open a pepr issue regarding the mismatch in behavior between manifest and chart.

zachariahmiller: In the interest of time i'll remove the change to using the chart. Per our other discussion you'll create an issue in pepr for the chart generation to include the extra env things.

mjnagel: FYSA opened https://github.com/defenseunicorns/pepr/issues/951 based on this issue.

mjnagel (approved):
LGTM for real this time.

chance-coleman (approved):

rjferguson21 (approved):

PR #531, opened 2024-07-02
fix(docs): re-ordered small paragraphs, clarified wording, and added links to tech homepages

JoeHCQ1:
# Summary

As a new Unicorn, I made changes which I believe would've increased my comprehension of this document on the first read-through.

## Description

- Re-ordered tools in monitoring to build conceptually in line with the way each enables the next
- Turned first mention of each tool into a link to the relevant docs. Both an ease of reference change and also to disambiguate in the event a name collision exists in the reader's mind.
- Clarified relationship between AuthService and Keycloak as this Unicorn initially saw a redundancy and suspected AuthService was perhaps the service IAM solution while Keycloak was the user IAM solution. We could link out to BigBang's explanation of this relationship but I did not: https://docs-bigbang.dso.mil/2.2.0/docs/understanding-bigbang/package-architecture/authservice/
- Wording changes for precision. Example, in MetricsServer, speaking of "container" metrics is more precise (and I believe more accurate) than "cluster" metrics. If it reports on nodes and containers I would (personally) speak of both over the more nebulous "cluster" (which has many more primitives (notably servics and pods) which could be but are not (I think) being reported on).
- Most likely errors needing correction:
  - Metrics Server may be misunderstood
  - AuthService may be better characterized but I didn't want to plagarize the BigBang docs, and am new to this service.

## Type of change

## Checklist before merging


docs/application-baseline.md
```diff
@@ -16,10 +16,10 @@ For optimal deployment and operational efficiency, it is important to deliver a
...
+| **Security and Compliance**        | **[NeuVector](https://open-docs.neuvector.com/):** Offers container-native security, protecting applications against threats and vulnerabilities.<br><br> **[Pepr](https://pepr.dev/):** UDS policy engine and operator for enhanced security and compliance.|
+| **Identity and Access Management** | **[Keycloak](https://www.keycloak.org/):** A robust open-source Identity and Access Management solution, providing centralized authentication, authorization, and user management for enhanced security and control over access to mission-critical resources.|
+| **Backup and Restore**             | **[Velero](https://velero.io/):** Provides backup and restore capabilities for Kubernetes clusters, ensuring data protection and disaster recovery.|
+| **Authorization**                  | **[AuthService](https://github.com/istio-ecosystem/authservice):** Offers centralized authorization services, managing access control and permissions within the Istio mesh. AuthService plays a supporting role to Keycloak as it handles part of the OIDC redirect flow.
```
mjnagel:
```suggestion
| **Authorization**                  | **[AuthService](https://github.com/istio-ecosystem/authservice):** Offers centralized authorization services, managing access control and permissions within the Istio mesh. AuthService plays a supporting role to Keycloak as it handles part of the OIDC redirect flow.|
```

docs/application-baseline.md
```diff
@@ -16,10 +16,10 @@ For optimal deployment and operational efficiency, it is important to deliver a
...
+| **Service Mesh**                   | **[Istio](https://istio.io/):** A powerful service mesh that provides traffic management, load balancing, security, and observability features.                                                                                                                                                                                                                       |
+| **Monitoring**                     | **[Metrics Server](https://kubernetes-sigs.github.io/metrics-server/):** Provides container resource utilization metrics API for Kubernetes clusters.<br><br>**[Prometheus](https://prometheus.io/):** Scrapes Metrics Server API and stores the data in a time-series database for insights into container health and performance.<br><br> **[Grafana](https://grafana.com/grafana/):** Provides visualization and alerting capabilities based on Prometheus's time-series database of metrics. |
```
mjnagel: I think the prometheus description should also capture application metrics (not just metrics server api) since we use it to grab more than just the metrics-server? The before state might not be perfect but does capture that part for prometheus a bit, maybe can steal some wording from there.

mjnagel (commented):
Appreciate the contribution - few small comments! @UnicornChance may be good to validate for me that nothing changes here with the expected format on the docs site side.

docs/application-baseline.md
JoeHCQ1: @mjnagel want to re-check my wording? I added a reference to application metrics, and the goal of these data points from "container health" back to "application health" as it had been.

mjnagel: LGTM!

mjnagel (approved):

chance-coleman (approved):

PR #533, opened 2024-07-02
fix: initial creation of child logging

### LINKED ISSUE #526: Separate uds-core log stream from pepr-core log stream

See context on https://github.com/defenseunicorns/pepr/issues/927

By implementing a child log we can control our own log levels without needing to increase/decrease the pepr-core log level.

chance-coleman:
## Description
The main goal is to be able to have pepr maintain one level of logging ( info ) and uds logs at another level ( debug ). 

There are some log changes in the package output to decrease extra noise, as well as a few logs moved from debug to trace level. 

This change also addresses a comment on another PR about JWT's being logged, which was part of pepr logs at the debug level. 

Definitely open to discussions about at what level we want to log the childLog subproject. 

## Related Issue

Fixes #526
Relates to #201

## Type of change

## Checklist before merging


src/pepr/logger.ts
```diff
@@ -0,0 +1,15 @@
+import { Log } from "pepr";
+
+export function childLog(subproject: string) {
```
mjnagel: It might be good to make this function named something like `setupLogger` or `getLogger`?

src/pepr/logger.ts
```diff
@@ -0,0 +1,15 @@
...
+  const childLog = Log.child({ subproject: subproject });
```
mjnagel: Not in love with `subproject`, maybe `submodule` or `component` instead?

src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -1,7 +1,11 @@
-import { K8s, Log, kind } from "pepr";
+import { K8s, kind } from "pepr";
 
+import { childLog } from "../../../logger";
 import { UDSPackage } from "../../crd";
 
+// configure subproject logger
+const log = childLog("operator.controllers.istio");
+
```
mjnagel: I'm leaning towards shorter, slightly more generic titles here...maybe like `operator.istio`? Do you think that would be more confusing? Part of my thinking is that the end user viewing these logs probably doesn't need to know the nitty-gritty of the directory/file structure, but it's probably helpful to capture:
- Is this policy code, operator code (`Package` CR related), or "other"?
- If operator related - what piece of the setup is this (at minimum istio vs sso vs monitoring vs networkpolicies?)
- If other - what is it? (probably come up with some standard of how we name these)
- If policy related - maybe just policy is sufficient or we differentiate between exemption related pieces and validation pieces?

src/pepr/logger.ts
chance-coleman: agreed, i think setupLogger() makes more sense to me.

src/pepr/logger.ts
chance-coleman: changing to component

src/pepr/operator/controllers/istio/injection.ts
chance-coleman: I'm thinking the naming scheme would be the top level directory name ( currently these: `istio`, `operator`, `policies`, `prometheus` ) followed by the directory name where the file is found seprated by '.' ( ex. `operator.istio` or `policies.exemptions` ). Essentially what you have and addresses the "other" usecases.

mjnagel: I like that yeah. Should be enough information without being too fine-grained, and we can always adjust these or add additional fields as needed.

src/pepr/policies/index.ts
```diff
@@ -1,12 +1,16 @@
...
+import { setupLogger } from "../logger";
 import { ExemptionStore } from "../operator/controllers/exemptions/exemption-store";
 import { processExemptions } from "../operator/controllers/exemptions/exemptions";
 import { Matcher, Policy, UDSExemption } from "../operator/crd";
 import "./networking";
 import "./security";
 import "./storage";
 
+// configure subproject logger
+const log = setupLogger("policies");
```
mjnagel: It definitely matches the structure for the naming we chose but watching the logs around the test exemptions they end up looking like this:
```
{"level":20,"time":1720454522447,"pid":17,"hostname":"pepr-uds-core-774bfd4b86-v4lt7","component":"policies","msg":"Processing exemption podinfo3, watch phase: ADDED"}
{"level":20,"time":1720454522447,"pid":17,"hostname":"pepr-uds-core-774bfd4b86-v4lt7","component":"operator.exemptions","msg":"No existing exemption for owner 84cb0536-7a78-487e-8b3d-d872c9a21a9a"}
{"level":20,"time":1720454522447,"pid":17,"hostname":"pepr-uds-core-774bfd4b86-v4lt7","component":"operator.exemptions","msg":"Added exemption to DropAllCapabilities: {\"kind\":\"pod\",\"name\":\"^podinfo.*\",\"namespace\":\"podinfo\"}"}
...
{"level":20,"time":1720454523025,"pid":17,"hostname":"pepr-uds-core-774bfd4b86-v4lt7","component":"policies.exemptions","msg":"Checking podinfo-dd8b569b- against DropAllCapabilities exemptions: [{\"kind\":\"pod\",\"name\":\"^neuvector-enforcer-pod.*\",\"namespace\":\"neuvector\",\"owner\":\"c114993e-6004-48ff-8b6a-dfa7617c8966\"},{\"kind\":\"pod\",\"name\":\"^neuvector-controller-pod.*\",\"namespace\":\"neuvector\",\"owner\":\"c114993e-6004-48ff-8b6a-dfa7617c8966\"},{\"kind\":\"pod\",\"name\":\"^podinfo.*\",\"namespace\":\"podinfo\",\"owner\":\"84cb0536-7a78-487e-8b3d-d872c9a21a9a\"}]"}
{"level":30,"time":1720454523025,"pid":17,"hostname":"pepr-uds-core-774bfd4b86-v4lt7","component":"policies.exemptions","msg":"podinfo-dd8b569b- is exempt from DropAllCapabilities"}
...
{"level":20,"time":1720454530256,"pid":17,"hostname":"pepr-uds-core-774bfd4b86-v4lt7","component":"policies","msg":"Processing exemption podinfo3, watch phase: DELETED"}
{"level":20,"time":1720454530256,"pid":17,"hostname":"pepr-uds-core-774bfd4b86-v4lt7","component":"operator.exemptions","msg":"Removed all policy exemptions for 84cb0536-7a78-487e-8b3d-d872c9a21a9a"}
```

Basically we end up with a mix of logs labelled as `policies`, `operator.exemptions`, and `policies.exemptions`. I think this is okay if we want to stick with the rigid naming based on location (maybe it's a good sign we need to move around some of our code 🤷)? Just thought I'd call it out as a bit of a weird result, if I wanted to easily filter on all exemption logs that wouldn't be the easiest to do?

src/pepr/operator/reconcilers/index.ts
```diff
@@ -47,7 +51,7 @@ export function shouldSkip(cr: UDSPackage) {
  * @param status The new status
  */
 export async function updateStatus(cr: UDSPackage, status: PkgStatus) {
-  Log.debug(cr.metadata, `Updating status to ${status.phase}`);
+  log.debug(cr.metadata, `Updating status to ${status.phase}`);
```
mjnagel: I think here we can/should find a way to clean up this log (it looks a bit messy currently logging out the full metadata). We can probably just log out the name/namespace?

src/pepr/operator/reconcilers/index.ts
```diff
@@ -70,7 +74,7 @@ export async function updateStatus(cr: UDSPackage, status: PkgStatus) {
  * @param type The type of event to write
  */
 export async function writeEvent(cr: UDSPackage, event: Partial<kind.CoreEvent>) {
-  Log.debug(cr.metadata, `Writing event: ${event.message}`);
+  log.debug(cr.metadata, `Writing event: ${event.message}`);
```
mjnagel: I think here we can/should find a way to clean up this log (it looks a bit messy currently logging out the full metadata). We can probably just log out the name/namespace?

src/pepr/policies/index.ts
mjnagel: I think for now this is okay - we'll probably revisit some of the metadata around these logs in the future and can evaluate moving logs around/changing names at that point. Thinking through the existing 3 log types for exemptions:
- `policies.exemptions`: these are the actual validation checks against the existing exemptions
- `operator.exemptions`: these are the `Exemption` CR lifecycle handling, adding/removing exemptions from our "store"
- `policies`: these are related to the `Exemption` CR watch setup/events

Really it's just that `policies` one that feels a bit out of place, might be good to re-evaluate in the future if this code should move to the operator folders since it's closer to those pieces. For now don't want to propose any changes since this is primarily meant to reduce the verbosity of pepr's logs, with the `component` addition being an added bonus and start towards better structured core logs.

rjferguson21:
Adding this [issue](https://github.com/defenseunicorns/uds-core/issues/549) as a potential follow on to work on creating more consistent structured data for our log lines so we could potentially go a step further to help users answer the question "Show me the operator logs just for _my_ Package." and just in general keep ourselves consistent with which what kind of data we are attaching to our log lines.

This data contract (the types we use) would be useful for anyone consuming our logs like UDS Engine.

src/pepr/config.ts
```diff
@@ -16,10 +16,12 @@ export const UDSConfig = {
   allowAllNSExemptions: process.env.UDS_ALLOW_ALL_NS_EXEMPTIONS === "true",
 };
 
-Log.info(UDSConfig, "Loaded UDS Config");
+// configure subproject logger
+const log = setupLogger("config");
```
rjferguson21: Should we define these components as a string enum?

src/pepr/operator/controllers/network/generators/kubeAPI.ts
```diff
@@ -1,9 +1,13 @@
...
-import { K8s, kind, Log, R } from "pepr";
+import { K8s, kind, R } from "pepr";
 
+import { setupLogger } from "../../../../logger";
 import { RemoteGenerated } from "../../../crd";
 import { anywhere } from "./anywhere";
 
+// configure subproject logger
+const log = setupLogger("operator.generators");
+
```
mjnagel: Should this just be `operator.network`? Or does that violate the structure we were trying to set? This might be the only place where there's something 2 layers deep in `operator/controllers`...

chance-coleman: yeah based on the initial structure it would be different since we went with top most directory followed by bottom most directory. Kind of why initially i went with something like `operator.network.generators` but does start to get long.

mjnagel (dismissed):

mjnagel (approved):

rjferguson21 (approved):

PR #547, opened 2024-07-05
feat!: set istio passthrough gateway as optional component

### LINKED ISSUE #510: Make the Istio Passthrough Gateway Optional
As a bundle deployer who doesn't need the passthrough gateway, I would like the ability to opt out

- **Given** a deployment of UDS Core bundle
- **When** I create / deploy the bundle
- **Then** I have the ability to opt out of the passthrough gateway that isn't needed

UDS Task to `zarf package remove uds-core --components istio-passthrough-gateway --confirm`

joelmccoy:
BREAKING CHANGE: Istio passthrough gateway is no longer deployed by default

## Description
- Sets the `istio-passthrough-gateway` as not required in the `slim-dev` and `standard` package.  
- Added `istio-passthrough-gateway` as an included `optionalComponent` in the `k3d-standard` bundle so that it can still be deployed and tested.  
- Added some documentation around the istio gateways and use cases
...

## Related Issue

Fixes #510

## Type of change

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: set istio passthrough gateway as optional component (https://github.com/defenseunicorns/uds-core/pull/547)

Release-As: 0.24.0
END_COMMIT_OVERRIDE


joelmccoy:
~~The passthrough gateway test fails when testing the single source package based on the [test](https://github.com/defenseunicorns/uds-core/blob/dcdadb49255a5052dcb3fe079335976b758b32f9/src/istio/tasks.yaml#L11-L16) currently defined.~~

~~I could remove this test, but then we wouldn't test for the passthrough gateway even when we include it in the standard bundle.~~

~~Is there a way to conditionally check after deployment to see what components were deployed in a zarf package?  Then I could rewrite the test to check first if the passthrough gateway was supposed to be deployed.~~

I ended up editing the single package deploy task to automatically include all optional components so that the validation task doesn't fail.

tasks/deploy.yaml
```diff
@@ -23,14 +23,14 @@ tasks:
...
-            uds zarf package deploy build/zarf-package-uds-core-istio-${UDS_ARCH}.tar.zst --confirm --no-progress
+            uds zarf package deploy build/zarf-package-uds-core-istio-${UDS_ARCH}.tar.zst --confirm --no-progress --components '*'
           fi
       - description: "Deploy the Pepr Module"
         cmd: |
           PEPR_VERSION=$(npm pkg get version | tr -d '"')
           uds zarf package deploy build/zarf-package-pepr-uds-core-${UDS_ARCH}-${PEPR_VERSION}.tar.zst --confirm --no-progress --set UDS_SINGLE_TEST=true
       - description: "Deploy the requested Zarf Package (must set UDS_PKG environment variable)"
-        cmd: uds zarf package deploy build/zarf-package-uds-core-${UDS_PKG}-${UDS_ARCH}.tar.zst --confirm --no-progress
+        cmd: uds zarf package deploy build/zarf-package-uds-core-${UDS_PKG}-${UDS_ARCH}.tar.zst --confirm --no-progress --components '*'
```
joelmccoy, line 33: Added this in so that the optional components get automatically deployed for single package tests.  This command is only used in testing single packages. I think it is safe to assume when we run the validations/tests we want all components to be deployed.

mjnagel: This makes a ton of sense, very clever approach to make sure we always test all components.

src/istio/README.md
```diff
@@ -1 +1,11 @@
...
+
+1. **(Required)** Tenant Gateway - This gateway provides ingress to typical end-user applications. By default, UDS Core deploys a few services on this gateway, such as the Keycloak SSO portal. This gateway is typically exposed to end users of the applications deployed on top of UDS Core.
+2. **(Required)** Admin Gateway - This gateway provides ingress to admin-related applications that are not for use by the default end user. By default, UDS Core deploys a few services on this gateway, such as the Admin Keycloak interface. This gateway is typically accessible to admins of the applications deployed on top of UDS Core. *Since the Admin and Tenant Gateways are logically separated, it is possible to have different security controls on each gateway.*
+3. **(Optional)** Passthrough Gateway - This gateway allows mesh ingress without TLS termination performed by Istio. This could be useful for applications that need to (or currently) handle their own TLS termination. This gateway used to be a default component of UDS Core but is no longer deployed by default. To deploy this gateway, you must specify `istio-passthrough-gateway` as an `optionalComponent` in your UDS Bundle configuration.
```
mjnagel: This readme doesn't get picked up on our docs site (https://uds.defenseunicorns.com/docs/) since that only grabs from the `docs/` subfolder. I wonder if it would make sense to add a new file under the [configuration folder](https://uds.defenseunicorns.com/core/configuration/) for Istio/"ingress"? We could start that file in this PR by listing this gateway information + example for the passthrough gateway optionalComponent, and I think it would also make sense to list out some of the config + examples for domain and cert values either in this PR or in follow on updates.

mjnagel (commented):
Overall LGTM, tested deploys with and without the passthrough gateway to validate and saw no issues. Pending that one doc thread this should be g2g.

src/istio/README.md
joelmccoy: I can definitely port this information over into the docs site.  I will also add an example for adding istio passthrough gateway as an optional component.  

@mjnagel As for overriding the cert and domain values... at what layer is the end user recommended to set these?  Is this possible via a uds-config file or do you literally override the helm values set in a file like [this](https://github.com/defenseunicorns/uds-core/blob/main/src/istio/values/config-admin.yaml) one?

mjnagel: Yeah for domain/cert the path we typically recommend is with bundle overrides + uds-config file. [This](https://github.com/defenseunicorns/uds-core/blob/main/bundles/k3d-standard/uds-bundle.yaml#L76-L93) is a good example of the override section needed to expose the TLS cert/key pairs as config, then someone could set `ADMIN_TLS_CERT` and the others via config file or the [other methods for variables](https://uds.defenseunicorns.com/cli/overrides/#variables).

For domain, it is exposed as a zarf variable ([here](https://github.com/defenseunicorns/uds-core/blob/main/src/istio/chart/values.yaml#L5)) and typically is set via the `shared` key in your uds config (see example [here](https://uds.defenseunicorns.com/cli/quickstart-and-usage/#variables-and-configuration)).

mjnagel: Also just to point you in the right direction if it wasn't clear, [this is the config folder](https://github.com/defenseunicorns/uds-core/tree/main/docs/configuration) from the docs site and all docs should be markdown with a hugo flavor (ex: [frontmatter](https://gohugo.io/content-management/front-matter/) for title, etc).

joelmccoy: Updated.  Will need a thorough read through to make sure I captured everything correctly.

docs/configuration/istio/ingress.md
```diff
@@ -0,0 +1,83 @@
...
+              path: tls.key
+```
+The environment variables `ADMIN_TLS_CERT`, `ADMIN_TLS_KEY`, `TENANT_TLS_CERT`, and `TENANT_TLS_KEY` should be base64 encoded strings of the TLS certificate and key for the admin and tenant gateways respectively.
+
+
+You can then set the domain via [uds-config](https://uds.defenseunicorns.com/cli/quickstart-and-usage/#variables-and-configuration) file using shared key to override the Zarf Domain Variable (see below `uds-config.yaml`).
+```yaml
+shared:
+   domain: yourawesomedomain.com # shared across all packages in a bundle
+```
```
mjnagel: Might be useful to provide the TLS cert/keys as config file examples as well? If specified as environment variables they would need to be prefixed with `UDS_` for uds cli to actually pick them up.

docs/configuration/istio/ingress.md
```diff
@@ -0,0 +1,83 @@
...
+    optionalComponents:
+      - istio-passthrough-gateway
+```
+
+### Configure Domain Name and TLS for Istio Gateways
+
+By default, the UDS Core Istio Gateways are set up to use the `uds.dev` domain and have a valid TLS certificate packaged.  You will want to change the domain name for your environment and provide a valid TLS certificate for this domain.
+
+You can set the TLS certs via overrides in a [UDS Bundle](https://uds.defenseunicorns.com/bundles/) (see below).
+```yaml
```
mjnagel: Formatting, should add a newline here (don't recall if the docs site will enforce this during lint, but probably good to add regardless.

docs/configuration/istio/ingress.md
```diff
@@ -0,0 +1,83 @@
...
+        uds-istio-config:
+          variables:
+            - name: TENANT_TLS_CERT
+              description: "The TLS cert for the tenant gateway (must be base64 encoded)"
+              path: tls.cert
+            - name: TENANT_TLS_KEY
+              description: "The TLS key for the tenant gateway (must be base64 encoded)"
+              path: tls.key
+```
+The environment variables `ADMIN_TLS_CERT`, `ADMIN_TLS_KEY`, `TENANT_TLS_CERT`, and `TENANT_TLS_KEY` should be base64 encoded strings of the TLS certificate and key for the admin and tenant gateways respectively.
```
mjnagel, line 75: Formatting, should add a newline here (don't recall if the docs site will enforce this during lint, but probably good to add regardless.

docs/configuration/istio/ingress.md
```diff
@@ -0,0 +1,83 @@
...
+
```
mjnagel: Formatting, should add a newline here after the codeblock and delete the extra one at the end of the paragraph.

src/istio/README.md
mjnagel: LGTM overall, added 2 formatting comments and one on the content (envs + config file). Resolving this thread in favor of the new ones.

docs/configuration/istio/ingress.md
mjnagel:
```suggestion

You can then either use environment variables (`UDS_ADMIN_TLS_CERT`, `UDS_ADMIN_TLS_KEY`, `UDS_TENANT_TLS_CERT`, and `UDS_TENANT_TLS_KEY`) or a config file to configure the certs for each gateway. These values should be base64 encoded strings of the TLS certificate and key for the admin and tenant gateways respectively.

Domain should be set via your [uds-config](https://uds.defenseunicorns.com/cli/quickstart-and-usage/#variables-and-configuration) file using the shared key to override the Zarf Domain Variable (see example `uds-config.yaml` below).

```yaml
shared:
  domain: yourawesomedomain.com # shared across all packages in a bundle

# TLS Certs/Keys if not provided via environment variables
variables: 
  core:
    admin_tls_cert: # base64 encoded admin cert here
    admin_tls_key: # base64 encoded admin key here
    tenant_tls_cert: # base64 encoded tenant cert here
    tenant_tls_key: # base64 encoded tenant key here
```
```

mjnagel: I think the suggestion trimmed off a trailing end of code block, but that's what I'd envision for the updated content.

joelmccoy: added suggestions! thanks for explaining this and I'm glad we will have easy snippets to use now (instead of trying to navigate all the overrides and naming conventions 😂 ).

docs/configuration/istio/ingress.md
joelmccoy: fixed

docs/configuration/istio/ingress.md
joelmccoy: fixed

mjnagel (approved):

PR #548, opened 2024-07-08
feat: added standalone dns service for loki

### LINKED ISSUE #68: Loki gateway does not deploy on RKE2
### Environment
Device and OS: ec2 instance, amazon linux 2023
App version: 0.6.2
Kubernetes distro being used: RKE2 1.27.x

### Steps to reproduce
1. Attempt to deploy the example uds-core bundle on an RKE2 cluster.
2. Loki gateway fails to become healthy with the error:
    ```
    nginx: [emerg] host not found in resolver "kube-dns.kube-system.svc.cluster.local."
    ```

### Expected result

Happy deploy!

### Actual Result

The loki gateway is configured to point at the DNS service in kube-system. The [default svc name](https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml#L10) it expects is `kube-dns` and RKE2 has a different name (`rke2-coredns-rke2-coredns`).

### Severity/Priority

Medium? Completely blocks deployment after loki if the dns service is not named `kube-dns`, but with bundle overrides it could be changed.

### Additional Context

While this could be done with a variable it would be annoying to have to pass in an environment specific value like this. Ideally a standard service could be created that maps to the DNS pods in cluster already?

joelmccoy:
## Description

Creates a standard service in the loki package so that the loki gateway can rely on the name of this service instead of nonstandard names for the DNS service in certain k8s distros (i.e. rke2). 
...

## Related Issue

Fixes #68 

## Type of change

## Checklist before merging


mjnagel (commented):
Changes LGTM, deploying quickly here to validate all looks as expected.

mjnagel (approved):

PR #554, opened 2024-07-09
fix: decouple `devMode` and postgres egress

### LINKED ISSUE #489: egress rule for postgres should not be tied to devMode
### Overview
Currently, [the egress rule for postgres](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/templates/uds-package.yaml#L55-L62) is conditioned on `devMode: false`.
DevMode provides increased logging (among other things), and there are many reasons why you might want both. In any case, if an external database is configured, the egress rule should be created accordingly.

### Environment
App version: `0.22.1-registry1`

### Steps to reproduce
1. `devMode: true`
2. configure an external postgres database
3. roll the dice

### Expected result
If a postgres database is configured, the egress rule should be created accordingly

### Actual Result
The rule is only created when devMode is false

### Additional Context
This may be a case where the race ([see related gitlab issue](https://github.com/defenseunicorns/uds-package-gitlab/issues/152))between pepr creating the netpol and the keycloak pods starting up may have masked the behavior on initial install.. but now that pepr is more consistent in applying rules, upgrades in our long lived environments are failing to restart keycloak pods with JDBC connection timeouts.

noahpb:
## Description
Updates the internal `keycloak` helm chart to be more explicit about using an external postgres database connection. 
Notable changes are:
- Configures egress rule and `keycloak` env vars for postgres based on `postgresql` being populated
- Defaults `postgresql.username`, `postgresql.password`, `postgresql.database`, and `postgresql.host` to an empty string
- Adds option to enable debug logging via `debugMode: true`
- Adds a `fail` case when `devMode` is true and `postgresql` has values defined
- Adds fail cases when users do not supply required values for `postgresql` when `devMode` is `false`

## Related Issue

Fixes #489 
## Type of change

## Checklist before merging


noahpb:
In the codecentric helm chart for keycloak, the `postgres` values are used to configure the `postgres` dependency. I considered renaming `postgres` to something like `database` or `externalPostgres` to reflect that we're no longer using the `postgres` subchart, but opted not to as this would be a breaking change.

src/keycloak/chart/values.yaml
```diff
@@ -57,6 +57,9 @@ clusterDomain: cluster.local
 # Sets development mode for Keycloak. This disables caching, Postgres and HPAs and should only be used for testing
 devMode: true
 
+# Enable debug logging for keycloak and quarkus
+debug: false
+
```
mjnagel: Would it be better to make this `debugMode` to align with existing `devMode` flag?

src/keycloak/chart/values.yaml
```diff
@@ -149,16 +152,17 @@ service:
...
-  # PostgreSQL port
+  # The username of the database user
+  username: ""
+  # The password of the database user
+  password: ""
+  # Database name of the default JDBC URL
+  database: ""
+  # Hostname of the default JDBC URL
+  host: ""
+  # Port of the default JDBC URL
```
mjnagel: Did these new comments come from somewhere else? The username/password ones make sense, the inclusion of "JDBC URL" in the other 3 might be confusing though to an average end user?

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -113,15 +113,17 @@ spec:
...
             - name: KC_LOG_LEVEL
               value: DEBUG
             - name: QUARKUS_LOG_CATEGORY__ORG_APACHE_HTTP__LEVEL
               value: DEBUG
             - name: QUARKUS_LOG_CATEGORY__ORG_KEYCLOAK_SERVICES_X509__LEVEL
               value: TRACE
-          {{- else }}
+          {{- end }}    
+          {{- if .Values.postgresql.host }}
+          {{- if .Values.devMode -}}{{fail "Cannot use devMode when an external PostgreSQL Database (postgres.host) is configured."}}{{- end -}}
```
mjnagel: Would it make sense to add a helper that checks if host, user, pass, database, and port are all set? This would generally work, but maybe is better to strictly enforce all 5 values being passed in for this config?

Another scenario I just thought of...if someone disabled devMode but did not configure postgres that may also be worth failing on?

src/keycloak/chart/values.yaml
noahpb: Yeah, good callout. I copy/pasted directly from the Keycloak [docs](https://www.keycloak.org/server/all-config). I think simplifying them makes sense.

src/keycloak/chart/templates/statefulset.yaml
noahpb: Agreed - a helper function that checks for all of the cases mentioned would be the best solution.

noahpb: @mjnagel just pushed up some changes that moves all of the logic into a helper function (`keycloak.postgresql.config`). The function makes all config for `postgresql` conditional if all of the requirements mentioned are satisfied. LMK if you want me to walk you through the test cases I ran.

src/keycloak/chart/values.yaml
noahpb: Updated descriptions

src/keycloak/chart/values.yaml
noahpb: Switched to `debugMode`

src/keycloak/chart/templates/_helpers.tpl
```diff
@@ -74,3 +74,23 @@ Create the service DNS name.
...
+{{- if .Values.postgresql -}}
+{{ $requiredKeys := list "username" "password" "database" "host" "port" }}
+{{- range $k := $requiredKeys -}}
+{{ if empty (get $.Values.postgresql $k) }}{{- fail (printf "Missing value for \"postgresql.%s\"." $k ) -}}{{- end }}
+{{- end }}
+{{- else -}}{{fail "You must define \"username\", \"password\", \"database\", \"host\", and \"port\" for \"postgresql\"."}}
+{{- end -}}
+{{- default "true" "" }}
+{{- else if not (empty (compact (values .Values.postgresql))) -}}
+{{ fail "Cannot use an external PostgreSQL Database when devMode is enabled." -}}
```
noahpb, line 93: Adding a check for `.Values.postgresql.enabled` here would allow users to easily toggle between `devMode` and using `postgresql` if desired. Currently, users have to omit all values under `.Values.postgresql` to successfully enable `devMode`.

Thoughts on adding `postgresql.enabled`?

src/keycloak/chart/README.md
```diff
@@ -10,7 +10,11 @@ For more information on Keycloak and its capabilities, see its [documentation](h
 
 ### Dev Mode
 
-When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off. Devmode also leverages PVCs by default for `data` and `themes`.
+When `devMode: true` is set, the chart will deploy a single Keycloak Pod with an in-memory database and scaling turned off. Dev Mode also leverages PVCs by default for `data` and `themes`. 
+
+Using an external database with Dev Mode enabled is not supported.
+
+Dev Mode enables debug logging for Keycloak. To configure debug logging outside of Dev Mode, set `debug: true` in your values.
```
mjnagel: (my fault) the update to `debugMode` no longer matches the doc here 🙃

src/keycloak/chart/templates/_helpers.tpl
mjnagel: Hm, I think that's viable. Can we easily do this in a non-breaking change way, that wouldn't be confusing 😆?

In the past to "enable postgres" I just needed to set `devMode: false`, technically (although I would also obviously need the actual postgres values specified for it to work. If we default `postgresql.enabled` to false, then my upgrade would result in these values:

```yaml
devMode: false
postgresql:
  enabled: false
```

If this still enabled postgres then it would be non-breaking, but that might be confusing? I think I'd be happy with leaving off the explicit `enabled` key and using our multi-key check/logic, as long as we have those clear `fail` messages. Open to other opinions though, I think to some extent our `devMode` key is already the inverse of `postgresql.enabled` so having both feels duplicative to me (if we wanted to add an `enabled` key it might make sense to deprecate/remove the `devMode` one).

src/keycloak/chart/README.md
noahpb: Ah whoops, good catch.

src/keycloak/chart/templates/_helpers.tpl
noahpb: Chatted offline with @mjnagel, we decide to leave this out for now.

src/keycloak/chart/values.yaml
```diff
@@ -149,17 +153,18 @@ service:
...
+  # The username of the database user
+  username: ""
+  # The password of the database user
+  password: ""
+  # Database name
+  database: ""
+  # URL for the database
+  host: ""
+  # Port the database is listening on (default 5432)
+  port: ""
```
mjnagel: I think I missed this one before, how do you feel about leaving 5432 here as the default and removing port from the helper checks? (if not we should probably remove the "default 5432" part of the comment)

noahpb: I'm good with that.

mjnagel (approved):

src/keycloak/chart/templates/_helpers.tpl
```diff
@@ -74,3 +74,24 @@ Create the service DNS name.
...
 {{- end }}
+
+{{/*
+Check external PostgreSQL connection information. Fails when required values are missing or if PostgreSQL is configured when devMode is enabled.
+*/}}
+
+{{- define "keycloak.postgresql.config" -}}
+{{- if not .Values.devMode -}}
+{{- if .Values.postgresql -}}
+{{ $requiredKeys := list "username" "password" "database" "host" "port" }}
```
mjnagel, line 85: Just thought of an edge case here - username/password might not be required in IRSA setups. Maybe we should only check the other 3 here?

mjnagel: Reviewed and not sure IRSA was supported in the past - opened an issue to track that support https://github.com/defenseunicorns/uds-core/issues/570

PR #555, opened 2024-07-09
feat: initial creation of allOf group auth

### LINKED ISSUE #119: allOf group auth
### Description

Follow on to PR: https://github.com/defenseunicorns/uds-identity-config/pull/111

The current implementation of group authz is `anyOf`. Which requires a user belong to a single group of a list of groups. There is a use case for requiring a user belong to all groups in a list of groups. Using something like `allOf`. 

Look into the implementation of this and what if any issues arise.

### Questions
Can `anyOf` be combined with `allOf`, for example when there is two different list of groups?

chance-coleman:
## Description
Building out functionality for incorporating an allOf group authentication option type. This group auth is a list of groups defined by an application and for a user to authenticate to that app must have all of the groups in that list.

This new feature also makes sure that the anyOf type can be used with allOf for more complicated use cases.

Unit tests have been created and possibly a follow on PR for e2e tests. Docs also updated to describe the use case.

## Related Issue

Relates to [this uds-identity PR](https://github.com/defenseunicorns/uds-identity-config/pull/135)

## Type of change

## Checklist before merging


mjnagel:
Closing this for now - we aren't seeing any user requirements/requests for it. This could be reevaluated in the future.

PR #559, opened 2024-07-10
fix: grafana logout not working in some environments

### LINKED ISSUE #515: Grafana logout not working in some environments
### Environment
Device and OS: MBP macOS
App version: 0.22.2
Kubernetes distro being used: k3d

### Steps to reproduce
1. Deploy uds-core (without IDP)
2. Login to Grafana
3. Click logout button

### Expected result

I am logged out of Grafana and my session is terminated in Keycloak.

### Actual Result

I see Grafana refresh but remain logged in. Even with an IDP it appears the session is kept around?

chance-coleman:
## Description
Updated grafana values to include a proper signout url that manages removing the session from both grafana and keycloak.

Also added the group auth definition to grafana to require that users have `/UDS Core/Admin` membership ( kind of redundant since it already requires users be an admin but this will just reinforce that a bit more ).

## Related Issue

Fixes #515 

## Type of change

## Checklist before merging


src/grafana/chart/templates/uds-package.yaml
```diff
@@ -9,6 +9,9 @@ spec:
       clientId: uds-core-admin-grafana
       redirectUris:
         - "https://grafana.admin.{{ .Values.domain }}/login/generic_oauth"
+      groups:
+        anyOf:
+          - /UDS Core/Admin
```
mjnagel: Do we want to include this as well? If so we should make sure Auditor is allowed.

chance-coleman: ahh yeah i didnt push auditor, i know we discussed this but maybe we didnt come to a decision on if we want to put this in the baseline or not.

mjnagel (approved):

rjferguson21 (approved):

PR #564, opened 2024-07-11
fix: address network policy generation inter-namespace bug

### LINKED ISSUE #528: Network policies generated by operator separate remote namespace and pod selectors
### Environment

N/A

### Steps to reproduce

When combining `remoteNamespace` and `remoteSelector` in a single `network.allow` policy the generated netpol results in two separate peers. This effectively allows anything in the namespace ingress or egress rather than locking down to the specific pod.

https://github.com/defenseunicorns/uds-core/blob/de60e252526d73e439f5665b27f84e8773c24949/src/pepr/operator/controllers/network/generate.ts#L42-L63

### Severity/Priority

High - this results in incorrect policies, that tend to work for the end user so they may not notice the reduced security posture.

### Additional Context

Fixing this will likely be a breaking change, although it is meant to be a bugfix (behavior should've been this already).

MxNxPx:
## Description

A bug exists when combining `remoteNamespace` and `remoteSelector` in a single network.allow policy.  The generated network policy results in two separate peers.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/528

## Type of change

## Checklist before merging

- [] Test, docs, adr added or updated as needed


src/pepr/operator/controllers/network/generate.ts
```diff
@@ -7,6 +7,50 @@ import { cloudMetadata } from "./generators/cloudMetadata";
...
+        break;
+    }
+  } else if (policy.remoteNamespace !== undefined || policy.remoteSelector !== undefined) {
+    const peer: V1NetworkPolicyPeer = {};
+
+    if (policy.remoteNamespace !== "") {
+      peer.namespaceSelector = {
+        matchLabels: { "kubernetes.io/metadata.name": policy.remoteNamespace || "" },
+      };
+    } else if (policy.remoteNamespace === "") {
```
rjferguson21: @MxNxPx Added this check because as it turns out we rely on this behavior for Prometheus - https://github.com/defenseunicorns/uds-core/blob/main/src/prometheus-stack/chart/templates/uds-package.yaml#L50-L54

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -7,6 +7,50 @@ import { cloudMetadata } from "./generators/cloudMetadata";
 import { intraNamespace } from "./generators/intraNamespace";
 import { kubeAPI } from "./generators/kubeAPI";
 
+function getPeers(policy: Allow): V1NetworkPolicyPeer[] {
```
rjferguson21, line 14: I refactored this a bit to reflect the assumption that you either use remoteGenerated or (remoteNamespace and/or remoteSelector).

mjnagel (approved):

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -7,6 +7,56 @@ import { cloudMetadata } from "./generators/cloudMetadata";
...
+
+      case RemoteGenerated.IntraNamespace:
+        peers = [intraNamespace];
+        break;
+
+      case RemoteGenerated.Anywhere:
+        peers = [anywhere];
+        break;
+    }
+  } else if (policy.remoteNamespace !== undefined || policy.remoteSelector !== undefined) {
```
mjnagel, line 35: Just noting this is technically a change in behavior - previously someone could chain together both remoteGenerated and remoteNamespace/Selector and end up with a policy with 2 peers I think?

I think this approach is more sane, require each list item to have one or the other? We may want to add to the [schema](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/crd/sources/package/v1alpha1.ts#L39) though if that's feasible to ensure it gets blocked at the schema validation?

mjnagel: Missed [this thread](https://github.com/defenseunicorns/uds-core/pull/564#discussion_r1692181399) earlier - resolving that one to continue any discussion here.

mjnagel: Forgot about the validation we are already doing [here](https://github.com/defenseunicorns/uds-core/blob/fix-inter-ns-netpol/src/pepr/operator/crd/validators/package-validator.ts#L61-L64) so this is not actually a shift in functionality and is already being caught in the validator.

PR #565, opened 2024-07-11
chore: add util function for purging orphans

rjferguson21:
## Description
Uses common function for purging orphan resources

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/operator/controllers/utils.ts
+++ b/src/pepr/operator/controllers/utils.ts
@@ -1,5 +1,7 @@
 import { V1OwnerReference } from "@kubernetes/client-node";
-import { GenericKind } from "kubernetes-fluent-client";
+import { GenericClass, GenericKind } from "kubernetes-fluent-client";
+import { K8s } from "pepr";
+import { Logger } from "pino";
 
 /**
  * Sanitize a resource name to make it a valid Kubernetes resource name.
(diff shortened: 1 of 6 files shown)


src/pepr/operator/controllers/utils.ts
```diff
@@ -38,3 +40,20 @@ export function getOwnerRef(cr: GenericKind): V1OwnerReference[] {
...
+  log: Logger,
+) {
+  const resources = await K8s(kind).InNamespace(namespace).WithLabel("uds/package", pkgName).Get();
+
+  for (const resource of resources.items) {
+    if (resource.metadata?.labels?.["uds/generation"] !== generation) {
+      log.debug(resource, `Deleting orphaned ${resource.kind!} ${resource.metadata!.name}`);
+      await K8s(kind).Delete(resource);
+    }
+  }
```
mjnagel, line 58: This seems fine, just curious if there was a particular intent behind the switch from filter -> loop over only items to delete vs loop over all -> conditional delete?

rjferguson21: It was not for any particular reason, I copied the code from the most recent example in front of me which was the authservice code. 

I think technically this is ever so slightly better, avoiding creating another array, but I could go either way if you think the other way is more idiomatic/readable.

mjnagel (approved):

PR #566, opened 2024-07-12
fix: podmonitor mTLS mutations

mjnagel:
## Description

This PR has a few follow-on changes for https://github.com/defenseunicorns/uds-core/pull/517:
- Slightly reworded/ordered documentation to focus on current setup rather than deprecated setup
- Updated code to support `podSelector` or `selector` for pod monitor generation
- Updated mutation logic for servicemonitors to:
  - Combine conditionals for exemption
  - Add a new `uds/skip-mutate` annotation for exemption
  - Clarify deprecation of tlsConfig mutation (not scheme)
- Added mutation logic for podmonitors to:
  - Allow for exempting podmonitors from default scrape class via annotation or injection detection
  - Mutate `scheme` to `https` if not exempted
- Switched core podmonitors to use `uds/skip-mutate` annotation rather than explicit scrape class
- Deleted example test monitoring from other test-apps and enabled pod + svc monitor for podinfo (since it exposes a real metrics endpoint)

## Related Issue

No issue opened as this PR was just merged.

## Type of change

## Checklist before merging

### DIFF
--- a/src/grafana/tasks.yaml
+++ b/src/grafana/tasks.yaml
@@ -8,12 +8,13 @@ tasks:
             name: "app.kubernetes.io/instance=grafana"
             namespace: grafana
             condition: Ready
-      - description: Validate grafana servicemonitor exists
-        wait:
-          cluster:
-            kind: ServiceMonitor
-            name: grafana-metrics
-            namespace: grafana
+      # This test is only functional on full uds-core deploys
+      # - description: Validate grafana servicemonitor exists
+      #   wait:
+      #     cluster:
+      #       kind: ServiceMonitor
+      #       name: grafana-metrics
+      #       namespace: grafana
       - description: Validate grafana virtual service
         cmd: |
           if [ "$(curl -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}')" = "302" ]; then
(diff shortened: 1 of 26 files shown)


src/grafana/tasks.yaml
```diff
@@ -8,6 +8,13 @@ tasks:
             name: "app.kubernetes.io/instance=grafana"
             namespace: grafana
             condition: Ready
+      # This test is only functional on full uds-core deploys
```
rjferguson21: Can you explain more about this test? It fails on single package tests? 

Are we testing this functionality somewhere else?

mjnagel: Single package tests do not deploy the prometheus stack (or CRDs) and we [filter out](https://github.com/defenseunicorns/uds-core/blob/podmonitor-exemption/src/pepr/operator/reconcilers/package-reconciler.ts#L60) the generation of those resources in the pepr code. So running this validation in CI would only work on the full core testing we do.

This generation functionality is unit tested in the pepr code, but no e2e test exists currently (the same is true for service monitors and netpols I believe).

mjnagel: @rjferguson21 I'm going to add a new test, tacking it onto the podinfo checks (to cover both pod and service monitors).

Edit: going to resolve this thread, let me know if that works for you.

rjferguson21 (approved):

PR #572, opened 2024-07-12
fix: only allow istio gateways to set x509 client certificate header

bburky:
## Description

We terminate TLS at the Istio ingress gateway, and then we  In the Keycloak VirtualService, we securely extract the client certificate, urlencode it, and copy it to the `istio-mtls-client-certificate` header. Keycloak then uses this as the x509 client identity. 
https://github.com/defenseunicorns/uds-core/blob/e505dc936d16c763be0f7a487f727f91cf9d00b3/src/keycloak/chart/templates/uds-package.yaml#L87-L92

When Istio sets the header, we also delete any existing `istio-mtls-client-certificate` header. This ensures that all users through the ingress gateway cannot forge a `istio-mtls-client-certificate` when they do _not_ use TLS client authentication.

However... we only do this at the gateway and we allow in-cluster mesh communication to Keycloak to non-admin URLs.

This PR adds a DENY AuthorizationPolicy rule to the Keycloak Istio sidecar, to deny any incoming requests with a `istio-mtls-client-certificate` header not originating from the admin or tenant gateway.

Credit to @rjferguson21 for asking great questions and catching this. 

## Type of change

## Checklist before merging


jeff-mccoy (approved):
Discussed in slack. This covers more cases, except the unlikely attack from the exempted namespaces. At that point you likely have the ability to get the db secrets anyway for keycloak.

mjnagel (approved):

PR #579, opened 2024-07-16
chore(loki): default query settings, config as secret

mjnagel:
## Description

- Updates to the same defaults for loki as are used on our staging test clusters (better experience querying logs)
- Switches to a secret for the config rather than configmap (note that aws access keys end up in this configmap today if not using IRSA)
- Cleans up some unnecessary values on tokengen (disabling sidecar, legacy BB values based on job termination issue)

## Related Issue

N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/loki/values/values.yaml
+++ b/src/loki/values/values.yaml
@@ -15,6 +15,7 @@ memberlist:
     publishNotReadyAddresses: true
 
 loki:
+  configStorageType: Secret
   storage:
     bucketNames:
       chunks: uds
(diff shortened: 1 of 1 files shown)


MxNxPx:
> Description
> * Updates to the same defaults for loki as are used on our staging test clusters (better experience querying logs)
> * Switches to a secret for the config rather than configmap (note that aws access keys end up in this configmap today if not using IRSA)
> * Cleans up some unnecessary values on tokengen (disabling sidecar, legacy BB values based on job termination issue)

Can confirm that these changes to the loki values fixed query issues we were experiencing.

```yaml
            # Ensure we don't hit query limits
            - path: loki.limits_config.split_queries_by_interval
              value: "30m"
            - path: loki.query_scheduler.max_outstanding_requests_per_tenant
              value: 32000 # This is the new default in Loki 3.0
              ```

MxNxPx (approved):

PR #582, opened 2024-07-16
fix: update monitor mutation to not overwrite explicitly defined scrape class

zachariahmiller:
## Description

- Updates the behavior of the monitor mutation to not overwrite and explicitly defined scrapeClass

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/prometheus/index.ts
+++ b/src/pepr/prometheus/index.ts
@@ -1,14 +1,13 @@
 import { Capability, K8s, kind } from "pepr";
 import { Component, setupLogger } from "../logger";
 import {
-  PodMonitorEndpoint,
-  PodMonitorScheme,
-  PrometheusPodMonitor,
-  PrometheusServiceMonitor,
-  ServiceMonitorEndpoint,
-  ServiceMonitorScheme,
+    PodMonitorEndpoint,
+    PodMonitorScheme,
+    PrometheusPodMonitor,
+    PrometheusServiceMonitor,
+    ServiceMonitorEndpoint,
+    ServiceMonitorScheme,
 } from "../operator/crd";
-
 // configure subproject logger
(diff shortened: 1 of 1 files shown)


src/pepr/prometheus/index.ts
```diff
@@ -35,10 +34,14 @@ When(PrometheusServiceMonitor)
       sm.Raw.metadata?.annotations?.["uds/skip-sm-mutate"] ||
       !(await isIstioInjected(sm))
     ) {
-      log.info(
-        `Mutating scrapeClass to exempt ServiceMonitor ${sm.Raw.metadata?.name} from default scrapeClass mTLS config`,
-      );
-      sm.Raw.spec.scrapeClass = "exempt";
+      // if scrapeClass is already set, don't override it
+      if (sm.Raw.spec.scrapeClass === undefined) {
```
mjnagel: The placement of this means that we still mutate things _for_ Istio mTLS even if a scrape class is set (i.e. the `else` condition here doesn't take a pre-defined `scrapeClass` into account).

Real example of where this could be problematic - if the [metrics-server servicemonitor](https://github.com/defenseunicorns/uds-core/blob/main/src/metrics-server/chart/templates/service-monitor.yaml#L6-L7) set `scrapeClass: exempt` rather than the annotation for skipping, `endpoint.scheme` would still end up getting mutated.

Am I thinking about this right? Wondering if we should just add an initial check in both for `if (sm.Raw.spec.scrapeClass != undefined)` rather after/OR-ed with `if (sm.Raw.spec === undefined)`?

zachariahmiller: sounds good will do

chance-coleman (approved):
lgtm

PR #586, opened 2024-07-17
docs: add uds-core functional layers adrs

jeff-mccoy:
Introduces a concept of functional layers that changes how UDS Core packages are published / consumed.


docs/adrs/0002-uds-core-functional-layers.md
```diff
@@ -0,0 +1,73 @@
...
+
+Date: 2024-07-17
+
+## Status
+
+Accepted
+
+## Context
+
+Historically, UDS Core intentionally only published a single official Zarf Package to ensure a standard baseline across all UDS environments. We learned form the complex buffet menu-style of Big Bang that it is better to form opinions and test them than to allow unlimited configuration possibilities. We have also found the monorepo structure of UDS Core to be be much simpler for developing UDS Core. However, through user feedback, we have seen cases where more optionality is warranted, but having to modify UDS CLI to accommodate disabling portions of UDS Core is not ideal. We also have a need for an edge-compatible version of UDS Core that is more lightweight and does not include all the components of the full UDS Core package.
```
Racer159:
```suggestion
Historically, UDS Core intentionally only published a single official Zarf Package to ensure a standard baseline across all UDS environments. We learned from the complex buffet menu-style of Big Bang that it is better to form opinions and test them than to allow unlimited configuration possibilities. We have also found the monorepo structure of UDS Core to be be much simpler for developing UDS Core. However, through user feedback, we have seen cases where more optionality is warranted, but having to modify UDS CLI to accommodate disabling portions of UDS Core is not ideal. We also have a need for an edge-compatible version of UDS Core that is more lightweight and does not include all the components of the full UDS Core package.
```

docs/adrs/0002-uds-core-functional-layers.md
```diff
@@ -0,0 +1,73 @@
...
+3. UDS Core Metrics (Metrics Server)
+4. UDS Core Monitoring (Prometheus + Grafana)
+5. UDS Core Logging (Promtail + Loki)
+6. UDS Core Runtime Security (NeuVector)
+7. UDS Core Backup and Restore (Velero)
+8. UDS Core (combination of all layers)
+
+These layers can be combined as needed, with UDS Core Base serving as the foundation for all deployments. The UDS Core package will continue to offer the complete suite of tools. The only change in the UDS Core repo will be the addition of new package definitions for each functional layer. Versioning will be managed at the monorepo level, with each layer being versioned in lockstep with the rest of UDS Core.
+
+We will still publish the UDS Slim Dev bundle for testing UDS Packages. This will be a combination of UDS Core Base, UDS Core Identity and Authorization, and UDS-K3d. We will not publish other bundles beyond UDS Slim Dev and UDS Core Demo. Only the UDS Core zarf package will publish combined layers.
```
Racer159: For this will there still be the CRs for Prometheus things? (to allow the monitor key to work / create resources from the operator)

docs/adrs/0002-uds-core-functional-layers.md
```diff
@@ -0,0 +1,73 @@
...
+Today, we publish both the official UDS Core package and a UDS Slim Dev package, in addition to dev and demo UDS Bundles. The UDS Slim Dev package is a stripped-down version of UDS Core that includes limited services meant for testing UDS Packages. We have also learned that some teams outside of UDS need similar functionality, but slightly different parts of UDS Core, such as keycloak + authsvc.
+
+## Decision
+
+We will replace the existing 2-package (one dev and one prod) structure with a layered set of packages within the existing monorepo. This will allow users to choose the components they need for their specific use case, while maintaining a single source of truth for all UDS Core components. We will not break these out by applications as Big Bang did, but instead by layers of functionality for related/integrated components. This allows more flexibility for which functionality is provided by UDS Core and which is provided by other external packages.
+
+The functional layers will be:
+
+1. UDS Core Base (Istio + Pepr w/UDS Operator & UDS Policies)
+2. UDS Core Identity and Authorization (Keycloak + AuthService)
```
Racer159, line 22: Do we have timing metrics on auth service (not opposed to its inclusion though wanted to bring up the concerns @zachariahmiller had had on it being on by default/inseparable)

jeff-mccoy: I think we can work to optimize it (as far as I know keycloak is actually the slow boot), but the reality is functional-based authsvc on its own it not super useful and not something we want to support (realize SWF doesnt need authsvc rn)

docs/adrs/0002-uds-core-functional-layers.md
jeff-mccoy: tbd I think, want to do whatever makes sense here, realize rn we have to push this at the beginning of UDS Core, so maybe it stays in the base layer, but needs more dialog

docs/adrs/0002-uds-core-functional-layers.md
Racer159: yeah KC is def the slowdown today (~60s of the 5 mins to deploy uds-core) - if authservice is only a few seconds and a few milli-CPUs I can see it being useful even for us - we have talked about making podinfo (in uds-common) a better example and that would need authservice.

docs/adrs/0002-uds-core-functional-layers.md
Racer159: For my 2 pennies I'd advocate having that in there (or having an "operator dev" layer) that allows the full functionality of the UDS operator without needing everything deployed (which could push up against resources)

jeff-mccoy: the operator should be smart enough to handle this imo

docs/adrs/0002-uds-core-functional-layers.md
```diff
@@ -0,0 +1,73 @@
...
+3. UDS Core Metrics (Metrics Server)
+4. UDS Core Monitoring (Prometheus + Grafana)
+5. UDS Core Logging (Promtail + Loki)
+6. UDS Core Runtime Security (NeuVector)
+7. UDS Core Backup and Restore (Velero)
+8. UDS Core (combination of all layers)
+
+These layers can be combined as needed, with UDS Core Base serving as the foundation for all deployments. The UDS Core package will continue to offer the complete suite of tools. The only change in the UDS Core repo will be the addition of new package definitions for each functional layer. Versioning will be managed at the monorepo level, with each layer being versioned in lockstep with the rest of UDS Core.
```
UncleGedd: > each layer being versioned in lockstep with the rest of UDS Core

I'm assuming each layer will be published to GHCR?

jeff-mccoy: Correct, we will publish each as their own zarf package, like we do for uds-slim-dev and uds-core today.

docs/adrs/0002-uds-core-functional-layers.md
zachariahmiller: We're certainly on the edge of having to go off free runners for CI on a couple of packages, so theoretically even though authservice only takes ~10 seconds to come up it could push us over that threshold, but if switching to the large runners in that event is an acceptable cost then its fine.

docs/adrs/0002-uds-core-functional-layers.md
rjferguson21: Given the optionality here how would the handle the user experience for folks deploying a Package that might contain an `sso:` block but not have Identity and Authorization deployed?

zachariahmiller: I think the onus is on the package maintainer there to deploy the correct layers of uds core to test their package and/or provide an escape hatch with bundle override in CI for sso. In at least most of the SWF packages, we template out `{{- if .Values.sso.enabled }}` on the sso config in the Package CR. 

Adding these options is at the expense of our opinionation too though so comes with additional overhead as well as complexity (our gitlab package CR template is almost 450 lines long for example) and are technically exposed to the bundle creators (which is fine since its not the deployers), so while it has inevitably happened a lot, we should aim to limit this as much as possible especially for scenarios like this vs a legitimate need from an end-user.

mjnagel: I think the short answer is that Pepr/UDS Operator would have the knowledge of the cluster state and handle/ignore the relevant sections. So in the case you mentioned the `sso` block would just be ignored. I don't think we want to shift the complexity to the end user/bundle creator to have to toggle on/off all the pieces.

The challenge here becomes the app configuration. We would also need to find a way to dynamically handle things like [this](https://github.com/defenseunicorns/uds-core/blob/main/src/grafana/values/values.yaml#L23-L25) where the app has to have SSO toggled on/off.

docs/adrs/0002-uds-core-functional-layers.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+Historically, UDS Core intentionally only published a single official Zarf Package to ensure a standard baseline across all UDS environments. We learned from the complex buffet menu-style of Big Bang that it is better to form opinions and test them than to allow unlimited configuration possibilities. We have also found the monorepo structure of UDS Core to be be much simpler for developing UDS Core. However, through user feedback, we have seen cases where more optionality is warranted, but having to modify UDS CLI to accommodate disabling portions of UDS Core is not ideal. We also have a need for an edge-compatible version of UDS Core that is more lightweight and does not include all the components of the full UDS Core package.
+
+Today, we publish both the official UDS Core package and a UDS Slim Dev package, in addition to dev and demo UDS Bundles. The UDS Slim Dev package is a stripped-down version of UDS Core that includes limited services meant for testing UDS Packages. We have also learned that some teams outside of UDS need similar functionality, but slightly different parts of UDS Core, such as keycloak + authsvc.
```
mjnagel: nit: today we only publish core as a package. We build the slim-dev package but only publish it as part of the dev bundle. Not sure if its worth rewording/how we should.

docs/adrs/0002-uds-core-functional-layers.md
rjferguson21: I'm think that all sounds technically feasible but trying to consider users who might not know what layers are installed on a cluster.  How do they know that when they deploy their Package certain pieces would be ignored?

docs/adrs/0002-uds-core-functional-layers.md
mjnagel: Yeah I think in most cases, including `monitor` keys/pieces, we should just have the operator logic be more intelligent about looking up current cluster state and skipping over pieces that aren't applicable. Imo there is little benefit in having the CRs created if nothing is "using them" (i.e. the CRD exists but the prometheus-operator isn't deployed). I recognize the desire to CI test bits of functionality like monitoring but if you only have the CRDs deployed you're only testing one aspect (ex: servicemonitor creation) not the full picture (ex: servicemonitor actually works).

docs/adrs/0002-uds-core-functional-layers.md
```diff
@@ -0,0 +1,73 @@
...
+- Simplified versioning and dependency management by keeping everything in one repo
+- Easier coordination of changes across layers
+- Maintains a single source of truth for all UDS Core components
+- Enables both testing environments and production edge deployments
+
+### Negative
+
+- Need for more comprehensive documentation on layer combinations and usage
+- Testing needs to be expanded to cover common layer combinations
+- Need to address dependencies and configurations shared across layers
```
mjnagel, line 48: Maybe doesn't need to be in this ADR since it's a proposal for CLI, but I think it's worth considering if uds-cli should handle dependencies/ordering at the bundle layer rather than core needing to think about that aspect. For example if a bundle creator made a bundle like this:

```yaml
packages:
  - core-identity
  - core-monitoring
  - core-base
```

CLI should be smart enough to either error out on the ordering (core-base should come first) or reorder the bundle so that it would work. Any thoughts on this @UncleGedd? Very connected to other discussions on the idea of coupling CLI closer with Core.

docs/adrs/0002-uds-core-functional-layers.md
```diff
@@ -0,0 +1,73 @@
...
+The functional layers will be:
+
+1. UDS Core Base (Istio + Pepr w/UDS Operator & UDS Policies)
+2. UDS Core Identity and Authorization (Keycloak + AuthService)
+3. UDS Core Metrics (Metrics Server)
+4. UDS Core Monitoring (Prometheus + Grafana)
+5. UDS Core Logging (Promtail + Loki)
+6. UDS Core Runtime Security (NeuVector)
+7. UDS Core Backup and Restore (Velero)
+8. UDS Core (combination of all layers)
```
rjferguson21, line 28: I think we need to be careful about the terminology we use here.  I might consider saving the 'Core' moniker to represent when users have all functional layers.

docs/adrs/0002-uds-core-functional-layers.md
mjnagel: I think it should at least partially be pretty transparent based on what they picked to deploy 😅 - you deployed without monitoring in your bundle, it shouldn't be overly surprising that you didn't get monitoring on your other components. But definitely documentation around what each piece of the stack includes/provides will be important, and figuring out good ways to represent the UDS config/deployment state. Maybe this is also a good spot where CLI coupling/integration with Core could warn the end user that their chosen stack is missing identity/ask for confirmation this was intentional.

Definitely want to balance this against the expectation that some users will provide their own alternatives for pieces of core or intentionally opt out of pieces (i.e. they want a lightweight version). We don't want the information/warnings to be aggressive or confusing to users that intentionally are skipping pieces.

docs/adrs/0002-uds-core-functional-layers.md
mjnagel, line 28: It may be worth considering a future where UDS Core does not have all the layers by default, or has all layers but some are optional. Maybe could be saved for a future ADR but two cases I am thinking of:
1. Metrics-server: Most distros either provide metrics-server out of the box (rke2, k3s/k3d) or provide an easy/standard addon for it (EKS). To deploy core today someone would have to explicitly disable RKE2/k3d's metrics-server ([ex in uds-k3d](https://github.com/defenseunicorns/uds-k3d/blob/main/zarf.yaml#L48)). This might be a good case for making it in core but optional.
2. Kiali/Tempo: Both have come up in the past and could be useful but there are varied opinions on them. If neither provides any additional compliance value then they may not necessary as defaults in core, but could still be valuable as a `service-mesh-addons` type package, optional in core.

docs/adrs/0002-uds-core-functional-layers.md
rjferguson21: Some kind of warning might ease my mind but in general I don't think we should be assuming folks who are authoring Package CRs are always the same people who built/deployed the UDS Core bundle.

mjnagel: Yeah I wasn't thinking they are the same persona - in my mind the `Package` CR spec would stay the same the whole time, the only thing changing is the operator having some logic be conditional. So the bundle creator doesn't necessarily know/care about what was in the `Package` CR - they only know that they did or didn't include monitoring/sso in their bundle and the effects of that.

mjnagel: To pull the bottom line out of this thread - I think the nontrivial problem here is we need to figure out how to dynamically modify upstream helm charts based on the "uds cluster state". Two things discussed synchronously (neither of which feel great):
1. Pepr could mutate all the things we need to dynamically configure - this could be any number of resources, and would have to be specific per resource/package (i.e. a specific mutate for Grafana's configuration to add/remove the keycloak pieces). The downside of this is that each specific mutate must be baked in - which isn't a good solve for SWF, etc, on top of core.
2. UDS-CLI could handle this _somehow_ - maybe this looks like dynamic configuration/referencing of which values to use based on a cluster state resource/lookup. This might mean baking business logic into CLI for what capabilities necessitate a change in configuration (or at least having individual packages produce metadata that could be used by CLI)?

Would be curious to hear @jeff-mccoy or other's thoughts here.

docs/adrs/0002-uds-core-functional-layers.md
Racer159: Yeah I could get behind that too - we mostly assume this will #work today but it definitely makes more sense to validate in fuller contexts

docs/adrs/0002-uds-core-functional-layers.md
jeff-mccoy: lol I thought that and then looked at the packages tab and only looked at the first word "packages" for slim-dev....obviously ignoring the fact they all have that first word...

docs/adrs/0002-uds-core-functional-layers.md
jeff-mccoy: A key point we need to make going forward, and maybe adjust this doc to state that more plainly, it is assumed you will deploy UDS Core with UDS CLI as a bundle in the end. Whether that is only the complete uds-core package or otherwise. The idea being here at bundle create time you already know what you have/what you need. My thinking on those that required SSO--let's just keep that as a required assumption for now. Start with it in the docs, later on we can add a field of some sort that says "hey my guy, this package needs this this package"...not saying full-on DAG, but one-layer deep for these functional layers would not be hard. So that if someone tried to create a bundle without uds-core-base, for example, it would yell at them.

docs/adrs/0002-uds-core-functional-layers.md
jeff-mccoy: Yeah just mentioned in another comment--we need a way at the bundle level to verify these very simple dependencies, I'd be fine if that was hard-coded in UDS CLI for the time being.

docs/adrs/0002-uds-core-functional-layers.md
jeff-mccoy: This is intentional because the packages will be `uds-core-*` and `uds-core`, the dependency tracking discussed above will prevent bad behavior

ntwkninja: > It may be worth considering a future where UDS Core does not have all the layers by default, or has all layers but some are optional. Maybe could be saved for a future ADR but two cases I am thinking of:
> 
>     1. Metrics-server: Most distros either provide metrics-server out of the box (rke2, k3s/k3d) or provide an easy/standard addon for it (EKS). To deploy core today someone would have to explicitly disable RKE2/k3d's metrics-server ([ex in uds-k3d](https://github.com/defenseunicorns/uds-k3d/blob/main/zarf.yaml#L48)). This might be a good case for making it in core but optional.

Are there any good patterns we could leverage to conditionally deploy metrics server (i.e. if the capability exists on the cluster, skip this component install). That may help reduce optional levers for bundle creators.

mjnagel: Yeah absolutely - with zarf this isn't really possible (conditional components based on a cluster check), but with uds-cli we can build it to make it work for us. As the other active PR indicates, to start we'll likely start with a manual optional, but move to building out an automated method of detection in CLI.

bdegeeter: Add Azure AKS to that list.  We're having to work around the native metrics server.  I'm sure we can come up with some consistent patterns to skip for pre-existing services.

docs/adrs/0002-uds-core-functional-layers.md
mjnagel: Resolving this - it appears we are aligned/okay with not including CRDs in slim-dev although this is easy enough to change in the future if it causes issues. For specific package tests slim-dev + monitoring could be deployed if there is a desire to test that full functionality.

docs/adrs/0002-uds-core-functional-layers.md
mjnagel: Updated ADR to reflect docs for dependencies + future CLI changes to codify dependencies. For initial implementation we can document SSO as a requirement for any core packages that use it.

docs/adrs/0002-uds-core-functional-layers.md
mjnagel: Updated to reflect current state.

docs/adrs/0002-uds-core-functional-layers.md
mjnagel: Resolving - @rjferguson21 let me know if there are any other concerns here. I think it's good to attribute these layers as part of core rather than generically calling them "UDS Logging". If that does cause confusion in the future I think we should always feel comfortable revisiting decisions and assumptions.

mjnagel: Did not make any specific changes to the ADR based on this thread - I think it still makes sense for Core itself to contain all the layers, even if some are optional. The specific convo on metrics-server here will be addressed separately as a part of https://github.com/defenseunicorns/uds-core/pull/611 and the corresponding issue.

corang (approved):

ntwkninja (approved):

rjferguson21 (approved):

chance-coleman (approved):

PR #591, opened 2024-07-19
chore(deps): update uds-identity-config to v0.5.1

chance-coleman:
## Description
Update uds-identity-config version. Currently renovate has this update captured inside of the keycloak grouping and is blocked by updating keycloak to v25

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/values.yaml
+++ b/src/keycloak/chart/values.yaml
@@ -7,7 +7,7 @@ image:
   pullPolicy: IfNotPresent
 
 # renovate: datasource=github-tags depName=defenseunicorns/uds-identity-config versioning=semver
-configImage: ghcr.io/defenseunicorns/uds/identity-config:0.5.0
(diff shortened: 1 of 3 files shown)


mjnagel (approved):
LGTM - are there any notes we should make for the upgrade around the [realm changes](https://github.com/defenseunicorns/uds-identity-config/compare/v0.5.0...v0.5.1#diff-b6f0825346f8666903af5ca9412171d2aa42befc790cb8eba8c14ee45ded44d3)? Not sure if anything would be strictly required for functionality, but suspect some of the attribute changes/password policy/etc would be desired. Don't know if we'd want to document a similar "upgrade path" in the identity-config readme?

chance-coleman:
yeah that's a good call. None of the changes are breaking changes, but would be good to document the stig changes and where they came from as well as the keycloak managed user attributes. Both would probably be best to be put in the docs.

PR #602, opened 2024-07-24
chore: add debug logs for istio injection logic

rjferguson21:
## Description
Adds debug logs to better understand circumstances where istio injection is not occurring

## Related Issue
https://github.com/defenseunicorns/uds-core/issues/604

## Type of change

## Checklist before merging


mjnagel (approved):
LGTM - some of these feel slightly closer to trace logs just because of the noise. Could be interesting in the future to have CI run with trace logging enabled and change the level of these/others.

PR #609, opened 2024-07-25
feat!: switch from promtail to grafana alloy

### LINKED ISSUE #377: Ship Logs to Multiple Destinations
As a platform admin of UDS-Core, I need to ship logs to multiple locations

- **Given** a uds-core deployment
- **When** configuring logging
- **Then** I have the ability to ship logs to select multiple destinations (i.e. grafana + SIEM, grafana + S3, etc.)

adding & integrating an additional logging capability

### Additional context
This is a requirement for most environments we support

mjnagel:
## Description

Closed in favor of Vector - https://github.com/defenseunicorns/uds-core/pull/724

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/377

## Type of change

## Checklist before merging


src/alloy/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Grafana Alloy has been chosen as our replacement for Promtail. Primary motivations include:
+1. It is positioned as the "successor"/more feature-rich alternative to Promtail, to include [migration documentation](https://grafana.com/docs/alloy/latest/set-up/migrate/from-promtail/) focused on the switch. This makes it very easy to switch from Promtail to Alloy with little end user impact.
+1. As Alloy is part of the Grafana ecosystem it has good integrations with Loki to provide the enriched log data we have come to expect from Promtail (simple approach to labelling logs, etc).
+1. Through Alloy's S3 integration we can export logs to an additional storage location in generic formats (raw logs, OTel JSON) without any modification - no need to provide Alloy with edit/delete permissions.
+1. By choosing a distribution of the OTel Collector we have flexibility in the future to switch to a different distribution of the OTel collector and/or easily swap out our logging backend (Loki) for something else that fits in the OTel Framework without needing to ensure specific tool compatibility. OTel is part of the CNCF (Incubating) and seems to be on a good trajectory.
```
rjferguson21, line 34: Not as it stands today your recommendation would be continue to rely on alloy shipping logs to Loki in the Loki native format?  I don't think it is a huge sticking point but curious with Loki 3.0 supporting otel native format we _could_ use that but it's probably an inferior experience at this point?

src/alloy/README.md
```diff
@@ -0,0 +1,37 @@
...
+1. Choose a tool that does not require major changes in our logging stack, but is flexible for future adjustments to the stack: As we do have active users of our product we want to be careful in switching tools, so ideally we would like a tool that is a "drop-in" replacement. However, we don't want to rule out future changes to other pieces of the stack (i.e. Loki) so choosing a tool that doesn't lock us into Loki is important.
+1. Focus on the log collection/shipping problem: While there are a number of tools that offer far more than just logging pipelines (metrics, traces, etc), we don't currently see a need to focus on these tools. These features are seen as a nice to have, but not being evaluated as the focus here.
```
rjferguson21, line 21: Maybe not for this document but I am really curious how the metrics side of all of this could play out. Some questions:
* Do we need to export our metrics out of cluster as well? Is alloy a good fit there? 
* Is it conceivable we could drop other uds-core components?  Just glancing at alloy it supports metrics and surprising to me at least, things like ServiceMonitors?

src/alloy/README.md
mjnagel: Yeah I think that is something we could look at and evaluate as a follow on here - don't think it's worth pursuing as part of this initial switch over. A few considerations here:
- We aren't on Loki 3.0 yet, although there is a pending PR for that update. One of the main challenges in that upgrade is the schema change which is required for OTel support.
- (unsure, question) would switching to OTel actually change the way Loki stores logs? I think this would be the main/only real benefit - if it could enable us to switch to a different backend easier in the future.
- Would we lose anything by switching (similar to your note of "inferior experience") - might make more sense just to stay with the native integration if its an option.

src/alloy/README.md
mjnagel: Yeah I intentionally left off metrics (and traces) from here. One key distinction/point to make - OTel Collector/Alloy is as the name suggests just a collector. My understanding is that ultimately you still need something else to handle storage and something to query (ex: still need Loki + Grafana for log storage/querying). For metrics I believe the setup would end up looking like Alloy scraping metrics then pushing them to Prometheus, so you're really just ending up with an extra middleman.

Definitely could be evaluated, but in our current state I don't think it would reduce components/complexity and haven't heard any ask/need to export metrics. It does provide an option for that if we determine that is necessary in the future.

src/alloy/chart/templates/alloy-config.yaml
```diff
@@ -0,0 +1,213 @@
...
+        __path__  = "/var/log/kubernetes/**/*.log",
+        job       = "kubernetes-logs",
+        host = env("HOSTNAME"),
+      }]
+    }
+
+    // Forward all logs to Destinations
+    loki.source.file "node_logs" {
+      targets    = local.file_match.node_logs.targets
+      legacy_positions_file = "/run/promtail/positions.yaml"
```
mjnagel, line 128: Pushed up to make sure this was tracked but it does not appear to be working currently - seeing duplicate logs for node/pods and no indication it used the promtail position file.

mjnagel: In further testing logs did appear to properly be using the positions file across the upgrade. I did not note any changes in the latest patch version of Alloy so unsure if I just had faulty testing previously or if something else was at play.

mjnagel:
Closed in favor of Vector - https://github.com/defenseunicorns/uds-core/pull/724

Issue #522, opened 2024-06-28
Configure Keycloak audit log


rjferguson21:
We intend to satisfy [AC 2.1](https://csf.tools/reference/nist-sp-800-53/r5/ac/ac-2/ac-2-1/) with [auditing/events from Keycloak](https://www.keycloak.org/docs/latest/server_admin/#configuring-auditing-to-track-events) but these events are currently not configured to ship anywhere.

The easiest thing to do would probably be audit and ensure:
* all relevant login/admin events are configured correctly
* verify they are being logged by configuring the Logging Event Listener

See: https://www.keycloak.org/docs/latest/server_admin/#configuring-auditing-to-track-events

rjferguson21:
Just adding a note here @austenbryan would love metrics showing 'active users over time'.   I believe this would be necessary to provide that kind of view.

I also noticed https://github.com/aerogear/keycloak-metrics-spi which would might be worth looking into but would probably be less useful for auditing and satisfying controls.

rjferguson21:
Poked at this for a second and realized we are already logging events, but they were not JSON so created this PR https://github.com/defenseunicorns/uds-core/pull/610 to enable JSON logging.

Unfortunately the default event handler prints the event information as part of its string "message" so if we actually wanted to leverage this for auditing I think we'd probably want to create a custom event listener that printed the event information as JSON.  There might be existing implementations out there.

PR #610, opened 2024-07-25
feat: add json logging for keycloak

rjferguson21:
## Description
Adds optional flag to configure Keycloak to log JSON formatted logs.  Defaults to `true`.   

JSON formatted log lines are preferable when doing log aggregation or parsing and allows users to filter more sanely when viewing logs in Grafana.  This would also be a step in the direction of satisfying auditing and reporting needs based on existing log output as mentioned in https://github.com/defenseunicorns/uds-core/issues/522.   

Note: This falls short of addressing our auditing/reporting needs for Keycloak as event log information is surfaced in the the log line as a `message` field which ends up being a string regardless of this PR.

## Related Issue
Relates to https://github.com/defenseunicorns/uds-core/issues/522 

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -72,9 +72,12 @@ spec:
             # This will only import the realm if it does not exist
             - "--import-realm"
             - "--features=preview"
+            {{- if .Values.jsonLogFormat }}
+            - "--log-console-output=json"
+            {{- end }}
           envFrom:
             - secretRef:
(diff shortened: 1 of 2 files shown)


chance-coleman (approved):
lgtm

mjnagel (approved):

mjnagel (commented):
No concerns here but might be helpful to add context on the "why" (in this PR directly) since this is a default switch in behavior rather than just a new option.

PR #619, opened 2024-07-31
fix: account for keycloak HA ports

### LINKED ISSUE #620: Keycloak HA ports are not routed
Testing locally with an HA deployment of Keycloak (See [keycloak-ha-dev](https://github.com/defenseunicorns/uds-core/tree/keycloak-ha-dev)) logging in yields this error:

Inspecting Keycloak pod istio-proxy logs, JGroups traffic shows up like:

Which indicates Istio could not find a matching service to route the traffic too which would explain why clustering is not functioning as intended.

rjferguson21:
## Description
Fixes a number of issues to fix support Keycloak for HA deployments
* scopes AuthorizationPolicy to port 8080 (http) to avoid denying JGroups traffic
* adds 7800 and 57800 to headless service and pod to ensure Istio routes traffic correctly.
* adds network policy to allow internamespace traffic on 7800 + 57800

Note: It was not immediately obvious that port 57800 is used by Keycloak's HA deployment. I noticed traffic on 57800 when debugging and did some research and determined it was related to used for Infinispan's "failure discovery protocol".  The number is computed based on the [port offset](https://infinispan.org/docs/stable/titles/server/server.html#jgroups-system-properties_cluster-transport), so port offset (50000) + 7800 = 57800. 
 
## Related Issue
Fixes https://github.com/defenseunicorns/uds-core/issues/620

## Type of change

## Checklist before merging


src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -59,6 +59,17 @@ spec:
...
+      - direction: Ingress
+        remoteGenerated: IntraNamespace
+        ports:
+          - 7800
+          - 57800
+      - direction: Egress
+        remoteGenerated: IntraNamespace
+        ports:
+          - 7800
+          - 57800
```
mjnagel, line 74: nit: this could/should be wrapped in a conditional on `autoscaling.enabled`

rjferguson21: Updated

mjnagel (dismissed):
LGTM - validated with an HA setup that I did not see any session issues and all traffic appears to be working between replicas.

mjnagel (approved):

Issue #305, opened 2024-03-27
Keycloak client mappers configuration


ericwyles:
### Is your feature request related to a problem? Please describe.
When using SSO Secret Templating (#263) to make a new client in keycloak, we need a way to be able to configure mappers in the new client. Mattermost requires the following mappers to be present in order to work with keycloak masquerading as gitlab:

![image](https://github.com/defenseunicorns/uds-core/assets/23637493/c85130d7-4d66-494a-a271-497ae99d322e)

### Describe the solution you'd like

    Given I have an application configured to integrate with a UDSPackage CRD
    When I add a new secret template field to the sso section with mappers specified in the template
    And I apply that CRD to my cluster
    The mappers are created in the client in keycloak in my cluster.

### Additional context
I ran onto this integrating mattermost with keycloak and solved it by creating the mappers manually

![image](https://github.com/defenseunicorns/uds-core/assets/23637493/00efca5a-5acc-4b82-a5b0-1c94020865cc)

ericwyles:
@mjnagel -- There is an error in the screenshot for the mapper configuration. In the first mapper the user attribute needs to be 'mattermostid' (all lower case). Without mixed case, mattermost fails to create the user after initial authentication.

mjnagel:
@rjferguson21 @UnicornChance I _think_ we might be able to bake these mappers into the realm somehow so that we don't have to expose full mapper configuration in the Package spec? That would be my preference to start with, if possible - adding full mapper config sounds like a lot of extra complexity to expose. I'm not sure how often mappers might be a need though - if its a common requirement then maybe we should expose this. From Big Bang history in the past...Mattermost is the only place I recall mappers being used?

ericwyles:
Further data point on this, I'm just getting started integrating SonarQube and looking at [these instructions](https://docs.sonarsource.com/sonarqube/latest/instance-administration/authentication/saml/how-to-set-up-keycloak/) and it looks like it will require some mappers as well.

EDIT: There are more options on the sonarqube side, I'm going to sync with the team and see if it's already determined which way to go. If we go with SAML it requires the mappers.

ericwyles:
@rjferguson21 Following up here based on our discussion in slack. I created this draft PR just to show an example and kick off discussion.

https://github.com/defenseunicorns/uds-core/pull/621

This just adds the ability to specify keycloak protocolMappers in the sso object. 

I tested that with mattermost. Before that change mattermost uds-package looks like this:

```yaml
[...]

PR #621, opened 2024-07-31
feat: protocol mappers

ericwyles:
## Description
This is just here to demonstrate POC of protocol mappers to kick off discussion. At this point no intention to merge.

## Related Issue

Relates to #305


chance-coleman:
Example package crd:

```yaml
spec:
  sso:
    - name: Grafana Dashboard
      redirectUris:
      protocolMappers:
        - name: username
          protocol: "openid-connect"
          protocolMapper: "oidc-usermodel-property-mapper"
          config:
            user.attribute: "username"
            claim.name: "username"
            userinfo.token.claim: "true"
        - name: email
          protocol: "openid-connect"
          protocolMapper: "oidc-usermodel-property-mapper"
          config:
            user.attribute: "email"
            claim.name: "email"
            userinfo.token.claim: "true"
        - name: mattermostid
          protocol: "openid-connect"
          protocolMapper: "oidc-usermodel-attribute-mapper"
          config:
            user.attribute: "mattermostid"
            claim.name: "id"
            jsonType.label: "long"
            userinfo.token.claim: "true"

```

Updated to have more defined set of config options when defining a `protcolMapper`, I decided on the config options based on the current protocolMappers we have defined in the realms.json. Can be updated to add more if needed in the crd.

rjferguson21:
@UnicornChance Based our discussion, I updated this to permit arbitrary protocolMapper `config`.

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -303,6 +303,42 @@ const sso = {
           type: "string",
         },
       },
+      protocolMappers: {
+        description: "Protocol Mappers to configure on the client",
+        type: "array",
+        default: [],
```
rjferguson21, line 309: @UnicornChance @mjnagel @ericwyles @ntwkninja I added the additional `default: []` to ensure if you remove the `protocolMappers` key from the Package's sso configuration it will remove any existing protocolMappers from the dedicated scope. 

Alternatively we could not specify a default here and it would not 'delete' any previously created `protocolMappers` unless you explicitly set `protocolMappers: []`.

mjnagel: I think this approach makes sense - we should lean into declarative nature here and if protocol mappers get added/removed from Package CR I would expect the same to happen on the keycloak side.

mjnagel (approved):

PR #624, opened 2024-07-31
feat(istio): add configurable TLS version

### LINKED ISSUE #599: Ability to configure UDS Core to accept TLS 1.2
Currently UDS Core requires TLS 1.3+. This is frustrating in a variety of situations due to compatibility issues with clients that don't support 1.3 yet.

Requesting that minimum TLS version be configurable in UDS Core such that I can make it accept 1.2 when I am deploying it.

Proposed solution:

- Make minimum TLS version configurable in UDS Core, with allowed values of 1.2 or 1.3
- When 1.2 is EOLed, exact date unknown estimate 2026, remove 1.2 as valid minimum TLS version

FAQ:

- **Why only 1.2 or 1.3 as allowed values?** v1.1 is already EOLed. 1.2 is not EOLed yet.
- **Why is this needed?** Many client devices still don't support 1.3. 1.3 was released 6 years ago in 2018. It took 10+ years for 1.2 to achieve widespread adoption.

jamestexas:
## Description
- Introduced support for configuring the minimum TLS version for UDS Core.
- Added `supportTLSV1_2` boolean flag in values file to allow selection between TLS 1.2 and TLS 1.3.
- Updated Helm templates to conditionally set `minProtocolVersion` based on `supportTLSV1_2` value.
- Ensured backward compatibility by defaulting to TLS 1.3.
- Commented and documented the new configuration option for clarity.

## Related Issue

Fixes #599

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/src/istio/chart/templates/gateway.yaml
+++ b/src/istio/chart/templates/gateway.yaml
@@ -32,7 +32,8 @@ spec:
         mode: {{ $server.mode }}
         {{- if ne $server.mode "PASSTHROUGH" }}
         credentialName: gateway-tls
-        minProtocolVersion: TLSV1_3
+        # if supportTLSV1_2 is both defined and true, use TLSV1_2, otherwise use TLSV1_3
+        minProtocolVersion: {{ if $.Values.tls.supportTLSV1_2 }}TLSV1_2{{ else }}TLSV1_3{{ end }}
         {{- end }}
     {{ end }}
 {{ end }}
--- a/src/istio/chart/values.yaml
+++ b/src/istio/chart/values.yaml
@@ -24,3 +24,5 @@ domain: "###ZARF_VAR_DOMAIN###"
 #       # Hosts to add to this gateway
 #       hosts:
 #         - "*"
+#   # Whether to support TLS 1.2 (if false, only TLS 1.3 will be supported)
+#   supportTLSV1_2: true


jamestexas:
I originally ran into issues with this (before using the ternary) where the lack of `tls.supportTLSV1_2` define made it error out when attempting to deploy it. This change added logic to where "if `supportTLSV1_2` is both defined and true" it'll use TLS 1.2. It being set as `false` ([like it is here](https://github.com/defenseunicorns/uds-core/blob/d67d0223bc3852db1f5f2d5d0cadbffe6edbb556/src/istio/values/config-admin.yaml#L4)) retains the existing behavior of defaulting to 1.3, and I feel it makes it self documenting.  Let me know if it's preferred to remove it.

Additionally I wasn't sure if more tests need-be added here, but if they do let me know!

src/istio/chart/templates/gateway.yaml
```diff
@@ -32,7 +32,8 @@ spec:
         mode: {{ $server.mode }}
         {{- if ne $server.mode "PASSTHROUGH" }}
         credentialName: gateway-tls
-        minProtocolVersion: TLSV1_3
+        # if supportTLSV1_2 is both defined and true, use TLSV1_2, otherwise use TLSV1_3
+        minProtocolVersion: {{ if (and $.Values.tls.supportTLSV1_2 (hasKey $.Values.tls "supportTLSV1_2")) }}TLSV1_2{{ else }}TLSV1_3{{ end }}
```
mjnagel: Do you happen to have a set of values that I could use to reproduce the issue mentioned [here](https://github.com/defenseunicorns/uds-core/pull/624#issuecomment-2261567370)? Typically with helm you are safe to just use a conditional as long as the parent key(s) are defined - if the specific key (`supportTLSV1_2`) doesn't exist it will evaluate as false, which should work in our favor here.

It does fail with a helm `ternary` but with just if/else it appears to work even after dropping all reference to `supportTLSV1_2` in the values files (validated `helm template` against this chart, passing in each of the config- values files. Ternary would be nicer but this is slightly easier to handle the conditional with I think.

```suggestion
        minProtocolVersion: {{ if $.Values.tls.supportTLSV1_2 }}TLSV1_2{{ else }}TLSV1_3{{ end }}
```

src/istio/values/config-passthrough.yaml
```diff
@@ -1,6 +1,7 @@
 name: passthrough
 
 tls:
+  supportTLSV1_2: false # set to true to enable TLSv1.2 support
```
mjnagel: This addition made me realize our conditional around `minProtocolVersion` (checking that its not passthrough) probably shouldn't be there - I believe Istio allows the protocol version to be set on passthrough gateways as well.

src/istio/values/config-admin.yaml
```diff
@@ -1,6 +1,7 @@
 name: admin
 domain: "admin.###ZARF_VAR_DOMAIN###"
 tls:
+  supportTLSV1_2: false # set to true to enable TLSv1.2 support
```
mjnagel: nit: since this aligns with the default value (`false`) we don't really need it in all these config values files.

src/istio/chart/templates/gateway.yaml
jamestexas: @mjnagel so I very well could be wrong w/ how this behaved. It was my first time setting up the dev stuff in the repo and I ran into a few problems along the way. Let experiment again with the above and i'll report back!

jamestexas: So this was silly - I had originally tried to use:

```
minProtocolVersion: {{ if .Values.tls.supportTLSV1_2 }}TLSV1_2{{ else }}TLSV1_3{{ end }}
```

And I got this error with it:

```
      WARNING  Attempt (1/3) failed with: template: uds-istio-config/templates/gateway.yaml:36:41:
               executing "uds-istio-config/templates/gateway.yaml" at <.Values.tls.supportTLSV1_2>: nil
               pointer evaluating interface {}.tls
```
Easy mistake! Just went and re-read docs for it too haha.  Anyway this is all pushed up and 🤞 that tests go through and such!

mjnagel (approved):
LGTM - thanks for the contribution @jamestexas !

PR #627, opened 2024-08-01
chore: generate a schema for keycloak helm chart

### LINKED ISSUE #563: add `values.schema.json` to `keycloak` custom chart
To help avoid misconfigurations, add a `values.schema.json` file to the `keycloak` helm chart.

samayer12:
## Description

Using [remarshal](https://github.com/remarshal-project/remarshal) and [generate-json-schema](https://github.com/Nijikokun/generate-schema), I locally created a `values.schema.json` file for keycloak's chart. I also referenced [this post](https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/) to build context on schemas & helm charts.

## Related Issue

Fixes #563 

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/values.schema.json
+++ b/src/keycloak/chart/values.schema.json
@@ -1,0 +1,410 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "properties": {
+    "affinity": {
+      "type": "string"
+    },
+    "autoscaling": {
+      "type": "object",
+      "properties": {
+        "behavior": {
+          "type": "object",
+          "properties": {
+            "scaleDown": {
+              "type": "object",
+              "properties": {
+                "policies": {
+                  "type": "array",
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "periodSeconds": {
+                        "type": "number"
+                      },
+                      "type": {
+                        "type": "string"
+                      },
+                      "value": {
+                        "type": "number"
+                      }
+                    }
+                  }
+                },
+                "stabilizationWindowSeconds": {
+                  "type": "number"
+                }
+              }
+            },
+            "scaleUp": {
+              "type": "object",
+              "properties": {
+                "policies": {
+                  "type": "array",
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "periodSeconds": {
+                        "type": "number"
+                      },
+                      "type": {
+                        "type": "string"
+                      },
+                      "value": {
+                        "type": "number"
+                      }
+                    }
+                  }
+                },
+                "stabilizationWindowSeconds": {
+                  "type": "number"
+                }
+              }
+            }
+          }
+        },
+        "enabled": {
+          "type": "boolean"
+        },
+        "labels": {
+          "type": "object",
+          "properties": {}
+        },
+        "maxReplicas": {
+          "type": "number"
+        },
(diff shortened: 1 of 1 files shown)


lemonprogis:
Hey @samayer12 you'll want to get that commit verified.

samayer12:
@lemonprogis, done! It looks like I rebased through the web UI after I opened the PR which then didn't apply a signature. Seems like I'm not the only one who has encountered [that issue](https://github.com/orgs/community/discussions/11639).

Web commit signing is supported in [Enterprise Server 3.13](https://docs.github.com/en/enterprise-server@3.13/authentication/managing-commit-signature-verification/about-commit-signature-verification), but apparently not on [Free, Pro, or Team](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) instances.

noahpb (dismissed):
LGTM. Thanks for contributing @samayer12!

src/keycloak/chart/values.schema.json
```diff
@@ -0,0 +1,384 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "properties": {
```
mjnagel, line 5: It would be good to add `"additionalProperties": false,` here I think to make sure that people aren't adding extra values that are invalid.

I think stretch goal would be add that under some of the other properties - but obviously we want to be more careful with that because some of these values might support more than we know about.

src/keycloak/chart/values.schema.json
```diff
@@ -0,0 +1,384 @@
...
+    "affinity": {
+      "type": "string"
+    },
+    "autoscaling": {
+      "type": "object",
+      "properties": {
+        "behavior": {
+          "type": "object",
+          "properties": {
+            "scaleDown": {
```
mjnagel, line 31: Somewhat relevant to previous comment, this object for `behavior` is dropped directly into the [HPA template](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/templates/hpa.yaml#L22). Behavior in [that spec](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#configurable-scaling-behavior) actually supports a few more fields, notably `scaleUp` is missing here. This doesn't cause any immediate issues because we allow additionalProperties but might be good to update this one to add `scaleUp` at least.

As a sidenote - it would be nice to try and pull in the upstream k8s resource definitions/schema for things like this so that we aren't reinventing the wheel.

mjnagel (commented):
A few comments on how we approach the schema here. I think this is a good first pass, noting that it was primarily auto-generated. Might be good to take a look at a couple spots where we can add some missing fields/more strictly enforce the schema.

src/keycloak/chart/values.schema.json
samayer12: Added!

src/keycloak/chart/values.schema.json
samayer12: Added `scaleUp` following the same definition as `scaleDown` to this file. I agree that we shouldn't have to reinvent the wheel on schema definitions, hopefully this is a good enough start to have _a_ schema that we use.

samayer12 (commented):
Added changes from @mjnagel

samayer12:
Rebased with `main`, as of 77ea6f5

mjnagel (approved):

chance-coleman (approved):

PR #630, opened 2024-08-02
feat: add support for public clients and disabling standard auth flow

### LINKED ISSUE #509: Keycloak support for public clients and device flow (enables sigstore)

As Kay I want to have Sigstore integrated with Keycloak so that I can sign images, git commits and more without needing to manage keys

- **Given** I have an image I want to sign
- **And** the Sigstore stack is available in my environment
- **When** I run something like `cosign sign` with the fulcio, rekor, and oidc-issuer urls
- **Then** I am taken through the Keycloak signin process (device flow)
- **And** the action completes without needing a client secret

We could not integrate with Keycloak directly and instead use something like Dex to issue the id tokens: https://dexidp.io/

### Additional context

Currently the uds-core operator does not allow the creation of clients without a client secret.

Racer159:
## Description

This enables support for public clients and allows disabling the standard flow in a client to fully support the OAuth device flow.

You can see an example of how this would be implemented here: https://github.com/defenseunicorns/uds-package-sigstore/pull/19/files#diff-84cee30d5e16b64372e9b71dc11e8d2643001f0fdd881c71f76768372ed18edeR10

## Related Issue

Fixes #509

## Type of change

## Checklist before merging


Racer159:
(note - though not directly required this also makes `redirectUris` optional since they are not needed for device flow clients and are optional in the KC api: https://www.keycloak.org/docs-api/21.1.1/rest-api/#_clientrepresentation)
image

Racer159:
here is what this looks like for `sigstore` when tested in that repo: 
image

src/pepr/operator/crd/generated/package-v1alpha1.ts
```diff
@@ -549,11 +549,15 @@ export interface Sso {
...
+  /**
+   * Defines whether the client requires a client secret for authentication
+   */
+  publicClient?: boolean;
   /**
    * Valid URI pattern a browser can redirect to after a successful login. Simple wildcards
    * are allowed such as 'https://unicorns.uds.dev/*'
    */
-  redirectUris: string[];
+  redirectUris?: string[];
```
mjnagel, line 560: I think I would feel better about this if we were only making it optional _if_ standardFlowEnabled is false. Not sure if we have the flexibility to do that with the version of schema we are using, but we could put the logic in the validator?

Racer159: done 1 min ago...

mjnagel: You read my mind @Racer159 !

bburky:
This allows disabling `standardFlowEnabled`, but doesn't seem to enable the device flow? Or how are you doing that?

`publicClient` is a risky but does make sense for certain scenarios like device flow. I'd kind of like to protect our usual configurations to prevent a misconfiguration of incorrectly using `publicClient`, but I'm not sure how to do that. If nothing else we should have a validation rule that you don't use `publicClient: true` with `secretName`, `secretTemplate`, `secret`, `enableAuthserviceSelector`, `protocol: saml` (might be meaningless on SAML, unsure). 

Ideally using a public client is like a policy violation that somehow has to actually be marked as allowed. I don't think uds-operator's own policies really map well to this though. If Lula(?) was ready, I guess we could use it to detect all Packages with public clients.

For the device flow, I think `redirectUris` and `webOrigins` don't matter/aren't used? I think they're always ignored if `standardFlowEnabled: false`? I just want to make sure that there's no way to use the normal flow on the client intended for device flow, if we need to throw an error on setting `redirectUris` and `webOrigins` with `standardFlowEnabled: false` too.

Racer159:
Enabling the device flow is an attribute you add to the client: https://github.com/defenseunicorns/uds-package-sigstore/pull/19/files#diff-84cee30d5e16b64372e9b71dc11e8d2643001f0fdd881c71f76768372ed18edeR13

The redirectUris, post logout redirectUris, and webOrigins are disabled when standard flow is disabled.

Screenshot from 2024-08-02 11-17-07
Screenshot from 2024-08-02 11-17-24

Racer159:
We could disallow a public client for anything but device flow attributed standardAuthFlow disabled clients

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -92,6 +92,32 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+      client.publicClient &&
+      (client.standardFlowEnabled === undefined ||
+        client.standardFlowEnabled ||
+        client.secret ||
+        client.secretName ||
+        client.secretTemplate ||
+        client.enableAuthserviceSelector ||
+        client.protocol === "saml" ||
+        !client.attributes ||
+        client.attributes["oauth2.device.authorization.grant.enabled"] != "true")
```
bburky: I think this should work to block misconfiguring a client to be public but still attempt use a secret. That scenario is actually an error.

It's really a separate _policy_ thing though to block all public clients with the standard flow though. All other policy things in UDS usually have a way to say "please bypass this policy please". I guess we could check an annotation on the Package CR? Seems weird though. There are other plausible use cases for public clients, but we UDS doesn't have them currently, I'm unsure what kind of weird Delivery use case might need this.

For now, here's some cleanup around the tests below. Notably, in JS the empty string is falsy, I think we need `!== undefined` to ensure things are actually entirely unset (setting them to the empty string should be disallowed too).
```suggestion
    if (
      client.publicClient &&
      (client.standardFlowEnabled !== false ||
        client.secret !== undefined ||
        client.secretName !== undefined ||
        client.secretTemplate !== undefined ||
        client.enableAuthserviceSelector !== undefined ||
        client.protocol === "saml" ||
        client.attributes?.["oauth2.device.authorization.grant.enabled"] !== "true")
```

Racer159: applied your suggestion - we could cross that bridge when we get there and just not allow it for now until someone needs it - I can't really think of anything where you'd want a public client for a normal web app - the only time you might want the standard auth flow and a public client would be the default flow for cosign / witness but in all of the use cases I can think of you can use the device flow instead - it does slightly hurt UX since you have to have step or add a command flag depending on the client but I think that is largely ok (and we could wrap things for them in our own stuff too))

bburky: Are there unit tests in UDS operator? Do we usually do them for things like this?
Can we test with each bad field set to both the "wrong" value (true/false) and also to `""`?
Or just do manual testing I guess.

I just don't actually trust myself (or the insanity that is JS) to write complex `if` statements. I want to make sure we've actually tested this.

Racer159: I can add some in there - would likely be similar to the exempt validator

Racer159: Added tests covering the package validator

bburky (dismissed):
This is fine I guess, the allowed configurations are limited enough that it should never be possible to use with a web client.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -92,6 +92,27 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+        `The client ID "${client.clientId}" must specify redirectUris if standardFlowEnabled is turned on`,
+      );
+    }
+    // If this is a public client ensure that it only sets itself up as an OAuth Device Flow client
+    if (
+      client.publicClient &&
+      (client.standardFlowEnabled !== false ||
+        client.secret !== undefined ||
+        client.secretName !== undefined ||
+        client.secretTemplate !== undefined ||
```
mjnagel, line 107: If we are enforcing no ability to name/template out the secret should we also not be creating the secret? I might be missing the full justification here, but this would still result in a default secret being generated in cluster.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -92,6 +92,27 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+        client.enableAuthserviceSelector !== undefined ||
+        client.protocol === "saml" ||
+        client.attributes?.["oauth2.device.authorization.grant.enabled"] !== "true")
+    ) {
+      return req.Deny(
+        `The client ID "${client.clientId}" must _only_ configure the OAuth Device Flow as a public client`,
```
mjnagel, line 113: This denial message is relatively clear but it might be useful to document the intended usage of Device Flow (and the corresponding toggles for `publicClient` and `standardFlowEnabled`). Hopefully at least that way it is documented what the intent for these toggles are and people are less likely to get through most a deploy before it bombs out.

src/pepr/operator/crd/validators/package-validator.ts
Racer159: yeah that makes sense I can add an `if` to that

src/pepr/operator/crd/validators/package-validator.ts
Racer159: Why do _I_ have to write all the docs and tests :sob: :sweat_smile:

mjnagel: Please see checklist on PR description: `Test, docs, adr added or updated as needed`

mjnagel (dismissed):

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -92,6 +92,27 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
         `The client ID "${client.clientId}" uses an invalid secret name ${client.secretName}`,
       );
     }
+    // If standardFlowEnabled is undefined (defaults to `true`) or explicitly true and there are no redirectUris set, deny the req
+    if (client.standardFlowEnabled !== false && !client.redirectUris) {
+      return req.Deny(
+        `The client ID "${client.clientId}" must specify redirectUris if standardFlowEnabled is turned on`,
```
rjferguson21: nit: Will this message confuse someone (probably most), who will not know they have `standardFlowEnabled: true` because they did not explicitly set it to `true`?

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -92,6 +92,27 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+    }
+    // If this is a public client ensure that it only sets itself up as an OAuth Device Flow client
+    if (
+      client.publicClient &&
+      (client.standardFlowEnabled !== false ||
+        client.secret !== undefined ||
+        client.secretName !== undefined ||
+        client.secretTemplate !== undefined ||
+        client.enableAuthserviceSelector !== undefined ||
+        client.protocol === "saml" ||
```
mjnagel:
```suggestion
        client.protocol === Protocol.Saml ||
```
nittiest of nits, but could check this against the type rather than string directly (would also need the import)

rjferguson21 (dismissed):

src/pepr/operator/crd/validators/package-validator.ts
Racer159: added a default note

src/pepr/operator/crd/validators/package-validator.ts
Racer159: done

mjnagel (approved):

PR #634, opened 2024-08-02
feat(operator): add finalizer to package CRs to await removals properly

### LINKED ISSUE #523: Leverage `finalizer` for `Package` deletions

In the current state of things deletions happen in a watch in our pepr code. As we manage additional "external resources" (such as keycloak clients), blocking on that deletion is critical to prevent errors when a rapid delete + create process happens. Kubernetes finalizers feel like the right solution to ensure that someone cannot reuse a `Package` name before the previous one is fully cleaned up.

This block of code is the most relevant - note that these are two separate flows and there is currently no "block on delete" in here:

https://github.com/defenseunicorns/uds-core/blob/d71d83ed4d6e6a35724e70fc5a27cb7ff6e1adaa/src/pepr/operator/index.ts#L43-L60

### Additional context

https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/

mjnagel:
## Description

Adds a [k8s finalizer](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) to all `Package` CR resources during initial reconciliation. This will ensure that all packages end up with the finalizer, even on an upgrade.

Switching to a finalizer removes the usage/need for Pepr's `.IsDeleted` since we are actually watching for updates now (when a resource is deleted it will have an update with a `deletionTimestamp`). The deletion logic was moved to the `packageReconciler` to trigger when a `deletionTimestamp` is present. The first step in the deletion logic is patching the Package to a new status of "Deleting".

A few notes on the new behavior:
- Packages with SSO clients will take a few seconds to full delete their client and remove the token from the store. During this time they will be in the `Deleting` phase.
- Packages without SSO clients will delete nearly instantly in most cases.
- Since this is now in the same queue as other events, deletions may be slightly delayed depending on other events in the queue. This is mostly a good thing since it ensures that deletions are processed before new creations/changes - but it does result in slightly odd behavior for the end user since status does not change while a package is waiting in the queue.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/523

## Type of change

## Checklist before merging


mjnagel:
Closing for now, pending upstream Pepr changes to natively support this (see issue).

PR #644, opened 2024-08-06
feat(azure): azure blob storage support for velero

### LINKED ISSUE #643: Velero Azure Blob Support
When deploying UDS Core on Azure AKS it would be nice to have Velero support Azure Blob storage

- **Given** UDS Core is deployed on AKS
- **When** Velero backups are taken
- **Then** They are saved to Azure Blob storage

rwh85:
## Description

This PR aim to add minimal support for Azure AKS cluster. 

Changes include: 

- Add Azure Blob support to velero
- [ ] https://support.chainguard.dev/hc/en-us/requests/1077 for the chainguard Azure Plugin Image

- Configure neuvector to use ContainerD

## Related Issue

Relates to # https://github.com/defenseunicorns/uds-core/issues/643

## Type of change

## Checklist before merging


src/velero/zarf.yaml
```diff
@@ -19,6 +19,7 @@ components:
       - velero/velero:v1.14.0
       - velero/velero-plugin-for-aws:v1.10.0
       - docker.io/bitnami/kubectl:1.31.0
+      - velero/velero-plugin-for-microsoft-azure:v1.1.0
```
lemonprogis: This should match the version like on registry1 (v1.10.0)

lemonprogis (dismissed):

src/velero/zarf.yaml
rwh85: :) thanks!

src/velero/README.md
```diff
@@ -44,9 +44,47 @@ By overriding the velero values in the bundle as follows:
...
+            - path: configuration.backupStorageLocation
+              value:
+                - name: default
+                  provider: azure
+                  bucket: ${VERLERO_BUCKET_NAME}
+                  config: 
+                    storageAccount=${VELERO_STORAGE_ACCOUNT}
+                    resourceGroup=${VELERO_RESOURCE_GROUP}
+                    storageAccountKeyEnvVar:${VELERO_STORAGE_ACCOUNT_ACCESS_KEY}
+```
```
mjnagel: This section reads a bit confusing to me - I think the intent is that someone would just use bundle (helm) overrides to setup the azure connection, is that accurate? The way this reads it almost seems like I'd need to make my own secret first which I don't think is true? I think there's 2 separate options (existing secret someone manages on their own OR bundle overrides)?

Maybe not a part of this PR, but might be good down the road to re-evaluate where this doc belongs since it doesn't surface on the docs site if its just in this readme (cc @defenseunicorns/uds-core).

src/velero/zarf.yaml
```diff
@@ -34,6 +35,7 @@ components:
       - registry1.dso.mil/ironbank/opensource/velero/velero:v1.14.0
       - registry1.dso.mil/ironbank/opensource/velero/velero-plugin-for-aws:v1.10.0
       - registry1.dso.mil/ironbank/big-bang/base:2.1.0
+      - registry1.dso.mil/ironbank/opensource/velero/velero-plugin-for-microsoft-azure:v1.10.0
 
   - name: velero
     required: true
```
mjnagel, line 41: I think it's _okay_ to leave azure plugin off of the unicorn flavor for now but would like to at least track that with an issue. The alternative is for us to add the upstream image here but it has several medium/high/cricital CVEs 😒.

src/velero/README.md
rwh85: @mjnagel You are correct, it's either or. I've updated it to just provide the bundle override method.

src/velero/zarf.yaml
rwh85: Copy that. I'll create a separate issue that tracks the updates needed to the unicorn flavor (tied to the ticket opened with Chainguard to provide the image).

src/velero/values/upstream-values.yaml
```diff
@@ -14,3 +14,9 @@ initContainers:
     volumeMounts:
       - mountPath: /target
         name: plugins
+  - name: velero-plugin-for-azure
+    image: velero/velero-plugin-for-microsoft-azure:v1.1.0
```
mjnagel:
```suggestion
    image: velero/velero-plugin-for-microsoft-azure:v1.10.0
```

Looks like this is failing CI currently - I think just because of the image ref not matching what is in the zarf.yaml.

src/velero/zarf.yaml
mjnagel: Tracking [this issue](https://github.com/defenseunicorns/uds-core/issues/674) to resolve this. Also there is are pending PRs for the azure plugin that would bump the go version and azure sdk version which would resolve all CVEs. If that does get merged it may make sense to add the upstream image here to start.

mjnagel (approved):

PR #648, opened 2024-08-06
chore: allow for extra keycloak gateway usage with client certs

mjnagel:
## Description

End users may be using different gateways/extra gateways for keycloak. To support this usage, this value would allow adding other namespaces that are allowed to set client certs.

## Related Issue

N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/istio-admin.yaml
+++ b/src/keycloak/chart/templates/istio-admin.yaml
@@ -56,7 +56,7 @@ spec:
             notNamespaces:
             - istio-tenant-gateway
             - istio-admin-gateway
-            {{- range .Values.additionalIstioGatewayNamespaces }}
+            {{- range .Values.additionalGatewayNamespaces }}
             - {{ . }}
             {{- end }}
 {{- end }}
--- a/src/keycloak/chart/templates/istio-admin.yaml
+++ b/src/keycloak/chart/templates/istio-admin.yaml
@@ -1,4 +1,4 @@
-{{- if .Capabilities.APIVersions.Has "security.istio.io/v1beta1" }}
+{{- if .Capabilities..Has "security.istio.io/v1beta1" }}
 apiVersion: security.istio.io/v1beta1
 kind: AuthorizationPolicy
 metadata:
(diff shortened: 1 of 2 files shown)


src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -56,4 +56,7 @@ spec:
             notNamespaces:
             - istio-tenant-gateway
             - istio-admin-gateway
+            {{- range .Values.additionalGatewayNamespaces }}
+            - {{ . }}
+            {{- end }}
```
mjnagel: I think we could optionally enforce that this name starts with `istio-` as a small safeguard. But that does make the values override a bit more confusing/not just the namespace.

mjnagel: Two options with this - helm's `fail` + a check on the string, or just add `istio-` to this template.

rjferguson21: I'd consider not doing anything as an option too, and just accepting the additional namespace.

rjferguson21:
Additional context in https://github.com/defenseunicorns/uds-core/pull/572

src/keycloak/chart/templates/istio-admin.yaml
mjnagel: Yeah I do think current state of the PR is viable as well. This really should only be used by gateway namespaces but we don't strictly need to require a naming convention on them I suppose.

chance-coleman: I'm in favor of a naming convention to set a standard for cluster cleanness/organization. If in the future we needed to accommodate other types of gateways, not just `istio-`, we could use a whitelist of allowed prefixes that's used in the validation.

mjnagel: Updated to use a helm `fail` to enforce this prefix. We can always re-evaluate this decision but it should work for the only known consumer needing this.

rjferguson21 (approved):

PR #650, opened 2024-08-07
fix: add backoff to operator retry mechanism

### LINKED ISSUE #649: Package fails re-reconcile on upgrade
### Environment
App version: `0.25.0`
Kubernetes distro being used: `k3d`
Other: `uds-k3d-demo`

We've seen intermittent issues where on upgrades Packages fail to reconcile because Keycloak is temporarily unresponsive. 

### Steps to reproduce
1.  Install uds-k3d-demo
2.  Delete the keycloak pod
3.  Immediately delete the pepr-watcher pod

### Expected result
The operator would re-reconcile all Packages, and they would end up in a `status=Ready` state.

### Actual Result
The operator tries to reconcile the Packages, triggering an HTTP request to Keycloak as part of the reconciliation process for packages that leverage SSO.  This request fails and the operator immediately retries 5 times in short succession and subsequently sets the Package to `status=Failed`. 

### Severity/Priority
Critical

### Additional Context
We debated a couple of solutions:
* Adding a retry mechanism in the operator, specially for the Keycloak requests.
* Adding a check to ensure Keycloak is healthy prior to making requests.
* Adding a global operator delay on retries.

rjferguson21:
## Description
Adds delay to operator retries to fix issues with temporary failures causing `status=Failed` for Packages

Open to suggestions/thoughts on the delay time but this is the current breakdown:
| Retry Attempt (c) | backOffSeconds (3^c) |
|-------------------|----------------------|
|         1         |           3          |
|         2         |           9          |
|         3         |          27          |
|         4         |          81          |

It is possible it would make more sense to retry more frequently with less delay (maybe it being exponential doesn't make sense here).  

Example of what this looks like from the logs:
```bash
(⎈|k3d-uds:default) uds-core (backoff) kubectl logs -l pepr.dev/controller=watcher -n pepr-system --tail -1 | grep "seconds before" | jq '.msg'
"Waiting 3 seconds before processing Package grafana/grafana, status.phase: Retrying, observedGeneration: 1, retryAttempt: 1"
"Waiting 3 seconds before processing Package neuvector/neuvector, status.phase: Retrying, observedGeneration: 1, retryAttempt: 1"
"Waiting 9 seconds before processing Package grafana/grafana, status.phase: Retrying, observedGeneration: 1, retryAttempt: 2"
"Waiting 9 seconds before processing Package neuvector/neuvector, status.phase: Retrying, observedGeneration: 1, retryAttempt: 2"
"Waiting 27 seconds before processing Package grafana/grafana, status.phase: Retrying, observedGeneration: 1, retryAttempt: 3"
"Waiting 27 seconds before processing Package neuvector/neuvector, status.phase: Retrying, observedGeneration: 1, retryAttempt: 3"
```

## Related Issue
Fixes https://github.com/defenseunicorns/uds-core/issues/649

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/operator/reconcilers/package-reconciler.ts
+++ b/src/pepr/operator/reconcilers/package-reconciler.ts
@@ -35,6 +35,22 @@ export async function packageReconciler(pkg: UDSPackage) {
     return;
   }
 
+  if (pkg.status?.retryAttempt && pkg.status?.retryAttempt > 0) {
+    // calculate exponential backoff where backoff = 3s * 2^retryAttempt
+    const backOff = 3000 * (2 ** (pkg.status?.retryAttempt ?? 0));
+
+    log.info(
+      metadata,
+      `Waiting ${
+        backOff / 1000
+      } seconds before processing Package ${namespace}/${name}, status.phase: ${pkg.status
+        ?.phase}, observedGeneration: ${pkg.status?.observedGeneration}, retryAttempt: ${pkg.status
+        ?.retryAttempt}`,
+    );
+    // wait for backOff seconds before retrying
+    await new Promise(resolve => setTimeout(resolve, backOff));
+  }
+
   // Migrate the package to the latest version
   migrate(pkg);
 
--- a/src/pepr/operator/reconcilers/package-reconciler.ts
+++ b/src/pepr/operator/reconcilers/package-reconciler.ts
@@ -1,4 +1,4 @@
-import { handleFailure, shouldSkip, updateStatus } from ".";
+import { handleFailure, shouldSkip, updateStatus, writeEvent } from ".";
 import { UDSConfig } from "../../config";
 import { Component, setupLogger } from "../../logger";
 import { enableInjection } from "../controllers/istio/injection";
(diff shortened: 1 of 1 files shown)


src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -35,6 +35,18 @@ export async function packageReconciler(pkg: UDSPackage) {
...
   }
 
+  if (pkg.status?.retryAttempt && pkg.status?.retryAttempt > 0) {
+    // calculate exponential backoff where backoffSeconds = 3^retryAttempt
+    const backOffSeconds = 3 ** pkg.status?.retryAttempt;
+
+    log.info(
+      metadata,
+      `Waiting ${backOffSeconds} seconds before processing Package ${namespace}/${name}, status.phase: ${pkg.status?.phase}, observedGeneration: ${pkg.status?.observedGeneration}, retryAttempt: ${pkg.status?.retryAttempt}`,
+    );
```
mjnagel: Might be nice to drop an event on the Package as well when we kick off this backoff.

rjferguson21: Updated to include an event

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -35,6 +35,23 @@ export async function packageReconciler(pkg: UDSPackage) {
...
+    // calculate exponential backoff where backoffSeconds = 3^retryAttempt
+    const backOffSeconds = 3 ** pkg.status?.retryAttempt;
+
+    log.info(
+      metadata,
+      `Waiting ${backOffSeconds} seconds before processing package ${namespace}/${name}, status.phase: ${pkg.status?.phase}, observedGeneration: ${pkg.status?.observedGeneration}, retryAttempt: ${pkg.status?.retryAttempt}`,
+    );
+
+    await writeEvent(pkg, {
+      message: `Waiting ${backOffSeconds} seconds before retrying package ${namespace}/${name}`,
```
mjnagel: very nit, but the namespace/name here feels redundant since its on that specific resource.

rjferguson21: You're right. Updated.

mjnagel (approved):
LGTM in testing. Noting that the only real downside here is that all Package resources are in the same queue, so a backoff on a single Package delays any others from being reconciled. This would be a better experience if/when [this pepr request](https://github.com/defenseunicorns/pepr/issues/1019) is resolved.

PR #652, opened 2024-08-07
fix: network allows for core netpols

mjnagel:
## Description

Noted an unknown field warning when deploying core this morning. Impact of this change for promtail is that network policies previously were open to the full promtail namespace rather than just the promtail pods (nuance since nothing else runs in the promtail namespace, but a safeguard).

For context on this - we have a deprecated `podLabels` field that gets migrated to `selector` (see [migrate function](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/crd/migrate.ts#L25-L28)). I did not dig further but unsure if `podSelector` was just an accidental addition or if it was valid at some point in time.

I also migrated two other spots using the deprecated `podLabels` to `selector`.

## Related Issue

N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/authservice/chart/templates/uds-package.yaml
+++ b/src/authservice/chart/templates/uds-package.yaml
@@ -19,7 +19,7 @@ spec:
         description: "SSO Provider"
 
       - direction: Ingress
-        podLabels:
+        selector:
           app.kubernetes.io/name: authservice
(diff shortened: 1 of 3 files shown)


Racer159 (approved):
lgtm - the selector for promtail matches the others and that should be the only logic change.

rjferguson21 (approved):

PR #658, opened 2024-08-08
feat: support authservice with redis, switch to pepr helm chart

### LINKED ISSUE #518: Configure authservice to use Redis
The existing authservice chart uses an in memory session store which makes it impossible to deploy in an HA mode.

rjferguson21:
## Description
* Updates Pepr to 0.34.0
* Uses Pepr Helm Chart
* Adds support to configure redis as authservice memory store

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/518

## Type of change

## Checklist before merging


src/uds-core-config/chart/templates/secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: uds-core-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "uds-core-config.labels" . | nindent 4 }}
+type: Opaque
+data:
+{{- range $key, $value := .Values.operator }}
+  {{ $key }}: {{ $value | b64enc | quote }}
+{{- end }}
```
mjnagel: nit: newline

src/istio/common/manifests/pepr-istio-config.yaml
```diff
@@ -21,3 +21,18 @@ spec:
...
+  namespace: pepr-system
+spec:
+  selector:
+    matchLabels:
+      pepr.dev/controller: watcher
+  mtls:
+    mode: STRICT
+  portLevelMtls:
+    "3000":
+      mode: PERMISSIVE
```
mjnagel, line 38: Just for my memory, is this primarily to handle metrics?

mjnagel (commented):
LGTM overall - still need to deploy locally to validate metrics/redis setup.

src/istio/common/manifests/pepr-istio-config.yaml
rjferguson21: Yes

packages/slim-dev/zarf.yaml
```diff
@@ -46,6 +46,27 @@ components:
...
       path: ../../dist
       name: module
+    charts:
+      - name: module
+        valuesFiles:
+          - ../../src/pepr/values.yaml
+    actions:
+      onDeploy:
+        before:
+          - cmd: ./zarf tools kubectl annotate secret -n pepr-system pepr-uds-core-api-token meta.helm.sh/release-name=module --overwrite || true
```
mjnagel: As discussed synchronously, I think we can pull this into a single `src/pepr/zarf.yaml`, potentially along with the `uds-operator-config` chart. Then we can simplify this import a bit and de-dupe the action portion.

src/pepr/zarf.yaml
```diff
@@ -0,0 +1,69 @@
+kind: ZarfPackageConfig
+metadata:
+  name: pepr-uds-core
+  description: 'Pepr Module: A collection of capabilities for UDS Core'
+  url: https://github.com/defenseunicorns/pepr
+  version: 0.5.0
```
mjnagel:
```suggestion
```

I think we can/should drop this version for simplifying the create/deploy lifecycle.

tasks/deploy.yaml
```diff
@@ -31,7 +31,7 @@ tasks:
       - description: "Deploy the Pepr Module"
         cmd: |
           PEPR_VERSION=$(npm pkg get version | tr -d '"')
-          uds zarf package deploy build/zarf-package-pepr-uds-core-${UDS_ARCH}-${PEPR_VERSION}.tar.zst --confirm --no-progress --set UDS_SINGLE_TEST=true
+          uds zarf package deploy build/zarf-package-pepr-uds-core-${UDS_ARCH}-${PEPR_VERSION}.tar.zst --confirm --no-progress --set UDS_SINGLE_TEST=true --set PEPR_SERVICE_MONITORS=false
```
mjnagel: If you go with the above suggestions we should be able to drop `PEPR_VERSION` from here.

mjnagel (commented):
LGTM otherwise!

src/pepr/zarf.yaml
rjferguson21: Updated

mjnagel (approved):

mjnagel (changes requested):
Two quick asks:
- PR title should probably indicate more than just the pepr update, maybe `feat: support authservice with redis, switch to pepr helm chart`
- Might be good to document the var for redis?

mjnagel (approved):

PR #666, opened 2024-08-13
fix: validate unique names for monitors

### LINKED ISSUE #656: Package CR fails to properly create servicemonitor resources when looping through monitor array
### Environment
Device and OS: Mac (M3)
App version: 0.22.2
Kubernetes distro being used: k3d (core-slim-dev)
Other: 

As part of the effort to [package ArgoCD](https://github.com/defenseunicorns/uds-package-argocd) so that it is "Made for UDS", I encountered an issue with the way the pepr operator creates servicemonitor resources by looping through the `monitor:` field as defined in the uds-argocd-config [chart template](https://github.com/defenseunicorns/uds-package-argocd/blob/main/chart/templates/uds-package.yaml#L27).

### Steps to reproduce
1. Spin up a cluster with uds core-core-slim dev
2. Install the prometheus-stack package from uds-core
    - I could not find a ghcr package for this, I ran a `zarf dev deploy` from [this dir](https://github.com/defenseunicorns/uds-core/tree/main/src/prometheus-stack) to install. 
3. Deploy the ArgoCD zarf package from [this repo](https://github.com/defenseunicorns/uds-package-argocd/tree/main)

### Expected result

Given 4 items in the `monitor:` array of a Package CR manifest, I expect to see 4 service monitors. 

### Actual Result

I only see a single serviceMonitor resource with spec field values associated with the last item in my monitor array. Despite the Package resource indicating 4 "monitors" in the package resource:

image

image

### Severity/Priority

Low considering we can use the upstream package in this case to generate serviceMonitors for each ArgoCD component as well as the `network` field to generate the proper ingress rules. It might not be common for an upstream product to expose multiple metrics endpoints like this.

### Additional Context

I think the issue might be the use of the `pkgname` in `generageMonitorName()` [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/monitoring/service-monitor.ts#L57) , potentially overriding the previous service monitor by creating a new one with the same name.

andygodish:
## Description

Added a new piece to the pepr operator's validator function. The new functionality loops through the Package CR's monitor array, adding each description field (or a dummy string in the event of no provided description) to a set, failing given a duplicate. 

The sanitizeResourceName() function was used in the validator since its also used downstream to create the ServiceMonitor resource name based on the string from the monitor.description field. 

...

## Related Issue

Fixes #656 

## Type of change

## Checklist before merging


mjnagel (commented):
Haven't had a chance to fully review the code yet but to fix the one CI failure you'll want to rename your PR title to follow a conventional commit syntax, something like `fix: validate unique names for monitors`

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -115,5 +115,20 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
     }
   }
 
+  const monitors = pkg.spec?.monitor ?? [];
+
+  // Ensure the descriptions are generating unique names
+  const monitorDescriptions = new Set<string>();
+
+  for (const monitor of monitors) {
+    const description = sanitizeResourceName(monitor.description || "null");
```
mjnagel: I think this actually causes some unnecessary denials. Take this example:
```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: podinfo
  namespace: podinfo
spec:
  monitor:
    - selector:
        app.kubernetes.io/name: foo
      targetPort: 9000
      portName: http-foo
    - selector:
        app.kubernetes.io/name: bar
      targetPort: 1000
      portName: http-bar
```

Because of the `|| "null"` this will end up with overlapping `null` values. I think we could do
```typescript
    const description = sanitizeResourceName(monitor.description || `${Object.values(monitor.selector)}-${monitor.portName}`);
```

This would match the [name generation](https://github.com/defenseunicorns/uds-core/blob/monitor-validate/src/pepr/operator/controllers/monitoring/common.ts#L8) and allow something like the above to work. May also want to adjust the denial message if we did change the logic here to reflect "generated name" is not unique or something to that effect (maybe with a suggestion to use a unique description).

andygodish: This has me thinking that a combination of selector.value + targetPort might be better suited for generating a unique name altogether. My inclination is to use the description field for a long descriptive sentence that I wouldn't my k8s resources to be named after. Curious why description was chosen as the primary over the other fields with generateMonitorName().

portName could be used as well, but imo is more likely to be named the same across services (ex: http) in charts containing multiple.

mjnagel: Good question. I think it's worth evaluating again - when we implemented this initially we intended for it to be a way to name resources more descriptively rather than getting an autogenerated name. I think we would want that functionality regardless, but maybe `name` would've been more appropriate, with a description occupying an annotation instead. We do trim/sanitize regardless so a long sentence would fit into the resource name, but it might end up looking weird if description is set to something like `This service monitor is used to monitor the application metrics for foobar app`.

I would say we keep the current naming logic in this PR but open to a follow-on ticket/PR to evaluate a different precedence/naming convention.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -141,5 +141,24 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
 
+  const monitors = pkg.spec?.monitor ?? [];
+
+  // Ensure serviceMonitors use a unique description or selector/portName used for generating the resource name
+  const monitorNames = new Set<string>();
+
+  for (const monitor of monitors) {
+    const monitorName = sanitizeResourceName(
+      monitor.description || `${Object.values(monitor.selector)}-${monitor.portName}`,
+    );
```
mjnagel: Realized in retrospect we could just import [`generateMonitorName`](https://github.com/defenseunicorns/uds-core/blob/monitor-validate/src/pepr/operator/controllers/monitoring/common.ts#L4) and use that here to ensure we keep this logic in sync with our name generation rather than duplicating.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -141,5 +141,24 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
     }
   }
 
+  const monitors = pkg.spec?.monitor ?? [];
+
+  // Ensure serviceMonitors use a unique description or selector/portName used for generating the resource name
+  const monitorNames = new Set<string>();
+
+  for (const monitor of monitors) {
```
mjnagel, line 197: Didn't think about this before but I think there's another edge case where podmonitors and servicemonitors _should_ be allowed to have overlapping names, but this would deny them. We may need separate sets per type and check the `kind` in this loop.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -141,5 +141,24 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+    const monitorName = sanitizeResourceName(
+      monitor.description || `${Object.values(monitor.selector)}-${monitor.portName}`,
+    );
+
+    if (monitorNames.has(monitorName)) {
+      return req.Deny(
+        `A serviceMonitor resource generated from the Package, ${pkg.metadata?.name} contains a duplicate name, please provide a unique description for each item in the monitor array`,
+      );
```
mjnagel: nit: might be good to align this with the wording we use for the [VirtualService check](https://github.com/defenseunicorns/uds-core/blob/monitor-validate/src/pepr/operator/crd/validators/package-validator.ts#L44-L48), and make sure we are capturing `kind` accurately here (pod vs svc monitor).

chance-coleman (approved):
lgtm

PR #676, opened 2024-08-15
fix!: client attribute allow list

### LINKED ISSUE #188: Increase the "ssoSessionIdleTimeout" to 15 mins
Change the "ssoSessionIdleTimeout" from 600 secs to 900 secs which would still conform to:
V-222389: “Design and configure the application to terminate the non-privileged users session after 15 minutes of inactivity”

- Using uds-core keycloak deployment
- when a user session is idle for 15 mins
- Then keycloak will automatically expire that user's token and require a new login/token

I configure and maintain my own identity-config image that changes that value
Manually update the value after uds-core keycloak deployment

### LINKED ISSUE #506: Allow list for SSO client attributes
Currently we do not restrict which Keycloak client attributes can be added to clients created through the Package custom resource. 
We should enumerate explicitly which attributes we pass through to actual Keycloak client update/create.

### Additional context
See https://www.keycloak.org/docs-api/22.0.1/rest-api/index.html#ClientRepresentation `client.attributes`

chance-coleman:
## Description
Creation of the allow list for SSO Client Attributes in the Pepr CRD.

The added fields are the only fields we are actively using in our realms. All attributes that are not present will use default keycloak values. 

This is the list of current attributes and how the grafana uds-package.yaml file can be modified to test their functionality
```yaml
  sso:
    - name: Grafana Dashboard
      redirectUris:
      attributes:
        "oidc.ciba.grant.enabled": "true"
        "backchannel.logout.session.required": "false"
        "backchannel.logout.revoke.offline.tokens": "true"
        "post.logout.redirect.uris": "+##*"
        "oauth2.device.authorization.grant.enabled": "true"
        "pkce.code.challenge.method": "S256"
        "client.session.idle.timeout": "1500"
```

## Related Issue

Fixes #506 [#188](https://github.com/defenseunicorns/uds-identity-config/issues/188)

## Type of change

## Checklist before merging


src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -299,8 +299,32 @@ const sso = {
...
+            type: "string",
+          },
+          "oauth2.device.authorization.grant.enabled": {
+            type: "string",
+          },
+          "pkce.code.challenge.method": {
+            type: "string",
+            enum: ["S256", "plain"],
+          },
+          "uds.core.groups": {
```
rjferguson21: I don't think this should ever be specified in the Package as a client attribute.  You might have done this because we use the type, but we probably don't want it to seem like this is a recommended way of setting the groups on a client.

mjnagel (commented):
LGTM overall pending resolution of the one comment - not sure if we should mark this as a "breaking change" just to be extra careful here since previously attributes was the wild west?

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -299,8 +299,32 @@ const sso = {
...
+          },
+          "backchannel.logout.session.required": {
+            type: "string",
+          },
+          "backchannel.logout.revoke.offline.tokens": {
+            type: "string",
+          },
+          "post.logout.redirect.uris": {
+            type: "string",
+          },
```
mjnagel: Maybe just for my understanding - how does this differ from our current `redirectUris` in the sso spec/do we know of places this is being used?

chance-coleman: The existing `redirectUris` is used for post *login* redirect, the attributes is used for post *logout* redirects. I don't believe any of our current clients are utilizing this field outside of the default `*` but could be useful for certain applications that don't have the redirect configuration.

mjnagel: Ah that's helpful. I think okay adding this here, no real risk exposing it and seems like it could be useful.

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -299,8 +299,29 @@ const sso = {
...
+          "oauth2.device.authorization.grant.enabled": {
+            type: "string",
+          },
+          "pkce.code.challenge.method": {
+            type: "string",
+            enum: ["S256", "plain"],
+          },
+          "client.session.idle.timeout": {
+            type: "string",
+          },
```
mjnagel: I'm thinking there might be more value in having this in our Package validator, rather than in the schema. Current behavior with extra properties in `attributes` is that they just get silently dropped (or maybe a warning log message, but still dropped). If we put this in the validate instead we could send a clear message back to the user that certain attributes are not allowed. This would mean we keep the schema/type loose (any string properties allowed).

Either way we should probably find a location to document the allowed attributes so that it's not hidden in the code only.

chance-coleman: updated, no official documentation made though.

mjnagel (dismissed):

docs/configuration/uds-operator.md
```diff
@@ -191,6 +190,26 @@ spec:
...
+    - name: Sigstore Login
+      clientId: sigstore
+      standardFlowEnabled: false
+      publicClient: true
+      attributes:
+        oauth2.device.authorization.grant.enabled: "true"
+```
+
+The [package-validator.ts](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/crd/validators/package-validator.ts#L84) will verify that the attributes being used are allowed. If attributes are used that are not allowed the package will be denied and not installed.
+
```
mjnagel: I don't know if this would become hard to maintain but should we just enumerate the supported attributes here? I think that might be more valuable than the example + link to code?

mjnagel:
```suggestion
The SSO spec supports a subset of the Keycloak attributes for clients, but does not support all of them. The current supported attributes are:
- oidc.ciba.grant.enabled
- backchannel.logout.session.required
- backchannel.logout.revoke.offline.tokens
- post.logout.redirect.uris
- oauth2.device.authorization.grant.enabled
- pkce.code.challenge.method
- client.session.idle.timeout

If you have a use-case for an additional attribute please open an issue for us to explore it.

```

Not sure on the last line there but this is more what I was thinking, quick note on supported attributes w/ the list.

mjnagel (approved):

rjferguson21 (approved):

PR #682, opened 2024-08-20
chore: add watch config to exemption watch

mjnagel:
## Description

Adds the watch config to exemption watch. Minimal change but ensures that the watch is configurable for all places we are using watches. `PEPR_LAST_SEEN_LIMIT_SECONDS` is also added to the default envs for the config chart, intending to set this to 60s to shrink the window/tolerance for failure in staging environment.

## Related Issue

N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/uds-operator-config/values.yaml
+++ b/src/pepr/uds-operator-config/values.yaml
@@ -5,3 +5,5 @@ operator:
   UDS_SINGLE_TEST: "###ZARF_VAR_UDS_SINGLE_TEST###"
   UDS_LOG_LEVEL: "###ZARF_VAR_UDS_LOG_LEVEL###"
   AUTHSERVICE_REDIS_URI: "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
+  # Tweak Pepr watch config to react to dropped connections faster
+  PEPR_LAST_SEEN_LIMIT_SECONDS: "60"
--- a/src/pepr/uds-operator-config/values.yaml
+++ b/src/pepr/uds-operator-config/values.yaml
@@ -6,4 +6,4 @@ operator:
   UDS_LOG_LEVEL: "###ZARF_VAR_UDS_LOG_LEVEL###"
   AUTHSERVICE_REDIS_URI: "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
   # Tweak Pepr watch config to react to dropped connections faster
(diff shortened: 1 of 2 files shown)


src/pepr/uds-operator-config/values.yaml
```diff
@@ -5,3 +5,5 @@ operator:
   UDS_SINGLE_TEST: "###ZARF_VAR_UDS_SINGLE_TEST###"
   UDS_LOG_LEVEL: "###ZARF_VAR_UDS_LOG_LEVEL###"
   AUTHSERVICE_REDIS_URI: "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
+  # Tweak Pepr watch config to react to dropped connections faster
+  PEPR_LAST_SEEN_LIMIT_SECONDS: "60"
```
mjnagel: Open question - I think it makes sense to update the default here but we could make this zarf var and update the default or just make it zarf var and update the deploy time config in other environments where we have experienced this issue.

mjnagel:
```suggestion
  # Tweak Pepr watch config to react to dropped connections faster
  PEPR_LAST_SEEN_LIMIT_SECONDS: "300"
```

mjnagel: Updating to default to the default pepr values, this can be helm overridden at the bundle layer for testing in staging.

rjferguson21 (dismissed):

src/pepr/uds-operator-config/values.yaml
```diff
@@ -5,3 +5,5 @@ operator:
   UDS_SINGLE_TEST: "###ZARF_VAR_UDS_SINGLE_TEST###"
   UDS_LOG_LEVEL: "###ZARF_VAR_UDS_LOG_LEVEL###"
   AUTHSERVICE_REDIS_URI: "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
+  # Tweak Pepr watch config to react to dropped connections faster
+  PEPR_LAST_SEEN_LIMIT_SECONDS: "300"
```
mjnagel:
```suggestion
  # Allow Pepr watch to be configurable to react to dropped connections faster
  PEPR_LAST_SEEN_LIMIT_SECONDS: "300"
```

rjferguson21 (approved):

chance-coleman (approved):

PR #698, opened 2024-08-23
fix: default `ctx.allowPrivilegeEscalation` to `false` if `undefined`

### LINKED ISSUE #527: Pepr policies do not trigger if `allowPrivilegeEscalation` is null
### Environment
Kubernetes distro being used: EKS

### Steps to reproduce
1. Without any Exemptions create a pod without setting `allowPrivilegeEscalation`.

### Expected result
`allowPrivilegeEscalation` defaults to true when not set, so I would expect to be unable to apply the manifest file.
### Actual Result
You are allowed to apply the manifest files without Pepr triggering.
image
### Severity/Priority
Medium
### Additional Context
Issue is caused by the filtering allowing both undefined and false, when it should only allow false for allowPrivilegeEscalation.
https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/policies/security.ts#L34-L36

noahpb:
## Description
The default behavior when admitting pod containers that do not have `securityContext.allowPrivilegeEscalation` explicitly defined is to admit the request. As noted in #527, if not included, `allowPrivilegeEscalation` defaults to `true`. 

This PR updates the `DisallowPrivileged` policy to match any containers that do not have a `securityContext` and/or `allowPrivilegeEscalation` defined and mutates to explicitly set to `false`.

Configuring the policy to deny resources that do not have `allowPrivilegeEscalation` explicitly defined could break existing deployments. Adding a mutation is a safe bet, assuming existing workloads are not already taking advantage of privilege escalation.

Read more about the default behavior:
https://medium.com/pareture/how-allowprivilegeescalation-works-in-kubernetes-ce696494f87b

## Related Issue
Fixes #527 

## Type of change

## Checklist before merging


noahpb:
See nested comments for validations performed.

noahpb:
Validation below. 

Example manifest:
```yaml
---
apiVersion: v1
kind: Pod
metadata:
  name: init-demo
spec:
  containers:
    - name: nginx
      image: nginx
      ports:
        - containerPort: 80
      resources:
        limits:
          memory: '1Gi'
          cpu: '800m'
        requests:
          memory: '700Mi'
          cpu: '400m'
```

Pepr logs during `kubectl apply` of above manifest:
```console
...
[11:54:00.680] DEBUG (/init-demo/19701): Patches generated
    uid: "be95ef99-43ed-4f06-bc79-9d16a2fa6510"
    namespace: "default"
    patches: [
...
      {
        "op": "add",
        "path": "/spec/containers/0/securityContext",
        "value": {
          "allowPrivilegeEscalation": false,
          "capabilities": {
            "drop": [
              "ALL"
            ]
          }
        }
...        
```

Resulting pod yaml after applied:
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    uds-core.pepr.dev/mutated: '["disallow-privileged","require-non-root-user","drop-all-capabilities"]'
    uds-core.pepr.dev/uds-core-policies: succeeded
  name: init-demo
  namespace: default
spec:
  containers:
  - image: nginx
    name: nginx
    ports:
    - containerPort: 80
    resources:
      limits:
        cpu: 800m
        memory: 1Gi
      requests:
        cpu: 400m
        memory: 700Mi
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
    volumeMounts:
    ...
```

noahpb:
Applying a manifest where `securityContext.privileged` is `false`, but `allowPrivilegeEscalation` is not present (scenario from #527):

```yaml
---
apiVersion: v1
kind: Pod
metadata:
  name: init-demo
spec:
  containers:
    - name: nginx
      securityContext:
        privileged: false
        #allowPrivilegeEscalation: false
      image: nginx
      ports:
        - containerPort: 80
      resources:
        limits:
          memory: '1Gi'
          cpu: '800m'
        requests:
          memory: '700Mi'
          cpu: '400m'
```

Resulting mutation:
```console
... 
     {
        "op": "add",
        "path": "/spec/containers/0/securityContext/allowPrivilegeEscalation",
        "value": false
      },
..
```
Resulting pod yaml after applied:
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    uds-core.pepr.dev/mutated: '["disallow-privileged","require-non-root-user","drop-all-capabilities"]'
    uds-core.pepr.dev/uds-core-policies: succeeded
  name: init-demo
  namespace: default
spec:
  containers:
  - image: nginx
    name: nginx
    ports:
    - containerPort: 80
    resources:
      limits:
        cpu: 800m
        memory: 1Gi
      requests:
        cpu: 400m
        memory: 700Mi
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      privileged: false
    volumeMounts:
```

src/pepr/policies/security.ts
```diff
@@ -25,7 +25,24 @@ import { exemptionAnnotationPrefix, isExempt, markExemption } from "./exemptions
...
+    if (request.HasAnnotation(`${exemptionAnnotationPrefix}.${Policy.DisallowPrivileged}`)) {
+      return;
+    }
+
+    // Check if any containers defined in the pod do not have the `allowPrivilegeEscalation` field present. If not, include it and set to false.
+    for (const container of containers(request)) {
+      container.securityContext = container.securityContext || {};
+      if (
+        container.securityContext.allowPrivilegeEscalation === undefined &&
+        !container.securityContext.privileged
```
mjnagel: Did you happen to check the case where `CAP_SYS_ADMIN` is added as a capability? Re-reading the [docs here](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) it seems like that might fall in a similar bucket of "requested privileges elsewhere" and maybe should opt-out of this mutation rather than requiring the end user to do it themselves?

Assuming we do the validate mentioned in my comment below the end user would likely need to have an exemption for this case anyways so this might reduce the things they need to add/do by one.

src/pepr/policies/security.ts
```diff
@@ -25,7 +25,24 @@ import { exemptionAnnotationPrefix, isExempt, markExemption } from "./exemptions
...
+      ) {
+        container.securityContext.allowPrivilegeEscalation = false;
+      }
+    }
+    annotateMutation(request, Policy.DisallowPrivileged);
+  })
   .Validate(request => {
```
mjnagel, line 53: We typically still validate a given condition, even if our mutation should guarantee it (see [drop all capabilities](https://github.com/defenseunicorns/uds-core/blob/fix/527-ctx-priv-esc/src/pepr/policies/security.ts#L330-L367) for example - we both mutate and validate that all capabilities are dropped).

I think we would want to check for this undefined case and deny workloads that have it undefined in our Validate (in addition to the mutation you already have). This keeps us from just assuming the mutation worked and ensures that our validate is strictly blocking the scenario, rather than assuming well-formed pods are already coming through it.

noahpb: I see what you mean. I'll update `violations` to be more specific.

src/pepr/policies/security.ts
noahpb: @mjnagel not sure I'm following. Are you saying to add a condition to skip the mutation if the `securityContext` has the `CAP_SYS_ADMIN` capability?

mjnagel: I think so - or at least validate the experience when that capability is added. I need to deploy and test this but based on the docs:
```
allowPrivilegeEscalation is always true when the container:
- is run as privileged, or
- has CAP_SYS_ADMIN
```

It would seem to me if we are checking for `privileged` here and skipping the mutation when the workload is privileged we may want to do the same based on the capability? Or maybe I misunderstood the intent of this check here `!container.securityContext.privileged`?

noahpb: That's a good callout. I'll add a check to see if the `CAP_SYS_ADMIN` capability is included alongside the existing check for `privileged`.

noahpb:
Deleted outdated comments around validation.

src/pepr/policies/security.ts
```diff
@@ -25,14 +25,38 @@ import { exemptionAnnotationPrefix, isExempt, markExemption } from "./exemptions
...
+      if (
+        container.securityContext.allowPrivilegeEscalation === undefined &&
+        //Skip mutation if privileged
+        !container.securityContext.privileged &&
+        !container.securityContext.capabilities?.add?.includes("CAP_SYS_ADMIN")
+      ) {
+        container.securityContext.allowPrivilegeEscalation = false;
+      }
+    }
+    annotateMutation(request, Policy.DisallowPrivileged);
```
mjnagel: nit: the placement of `annotateMutation` could mean that something gets annotated even though it didn't pass the conditional for any of its containers. We should probably have some bool `wasMutated` that is only set to true inside of that conditional, and check that before annotating?

noahpb: Good catch! What if we put the call to `annotateMutation` inside the conditional? Right below line 43. What that achieve the same effect?

mjnagel: I think that may work, didn't dive into the code to validate that calling it multiple times would be okay here though (since multiple containers could pass that check and result in calling that func 2x, 3x, etc on a single pod).

noahpb: Fair point. In my testing, I have found that the `annotateMutation` function will add a policy name to the value of the annotation even if that value already exists in the list.  I will implement your suggestion and call `annotateMutation` the way you suggested 👍.

noahpb:
Talked offline with @mjnagel and ultimately decided on the following:
- skip the mutation if workload is privileged or has CAP_SYS_ADMIN added
- validate + deny if allowPrivilegeEscalation = undefined

src/pepr/policies/security.ts
```diff
@@ -25,14 +25,43 @@ import { exemptionAnnotationPrefix, isExempt, markExemption } from "./exemptions
...
     }
 
     const violations = securityContextContainers(request).filter(
-      c => c.ctx.allowPrivilegeEscalation || c.ctx.privileged,
+      // Checking if allowPrivilegeEscalation is undefined. If yes, fallback to true as the default behavior in k8s is to allow if undefined.
+      // Checks the three different ways a container could escalate to admin privs
+      c =>
+        (c.ctx.allowPrivilegeEscalation ?? true) ||
+        c.ctx.privileged ||
+        c.ctx.capabilities?.add?.includes("CAP_SYS_ADMIN"),
```
mjnagel: Might advocate we drop this last conditional - I may be forgetting some pieces of our last convo but I think it could be confusing for some to get blocked by this privileged policy when they added a capability separately (and they would be separately denied by the RestrictCapabilities policy).

Would be curious on other opinions here, @MxNxPx any thoughts? For added context this is one of our policies that is focused on "privileged" workloads. Per [upstream k8s](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) using the `CAP_SYS_ADMIN` capability is effectively the same as enabling `allowPrivilegeEscalation`. My understanding however is that pods will fail to schedule if `allowPrivilegeEscalation` is disabled and that capability is added. Happy to discuss offline as well.

MxNxPx: @mjnagel - that makes sense to me.  despite the `CAP_SYS_ADMIN` essentially being a form of allowing privileged escalation, it would get caught by the capabilities policy.  i made some minor adjustments to @noahpb 's code and it seemed to work as desired.  going to add as code suggestions.

src/pepr/policies/security.ts
```diff
@@ -25,14 +25,43 @@ import { exemptionAnnotationPrefix, isExempt, markExemption } from "./exemptions
...
     );
```
MxNxPx:
```suggestion
      c => (c.ctx.allowPrivilegeEscalation ?? true) || c.ctx.privileged,
    );
```

src/pepr/policies/security.ts
```diff
@@ -25,14 +25,43 @@ import { exemptionAnnotationPrefix, isExempt, markExemption } from "./exemptions
...
+        container.securityContext.allowPrivilegeEscalation === undefined,
+        !container.securityContext.privileged,
+        !container.securityContext.capabilities?.add?.includes("CAP_SYS_ADMIN"),
+      ];
+      //We are only mutating if the conditions above are all satisifed
+      if (mutateCriteria.every(priv => priv === true)) {
+        container.securityContext.allowPrivilegeEscalation = false;
+        wasMutated = true;
+      }
+    }
```
MxNxPx:
```suggestion
    // Check if any containers defined in the pod do not have the `allowPrivilegeEscalation` field present. If not, include it and set to false.
    for (const container of containers(request)) {
      container.securityContext = container.securityContext || {};
      const mutateCriteria = [
        container.securityContext.allowPrivilegeEscalation === undefined,
        !container.securityContext.privileged,
      ];
      // We are only mutating if the conditions above are all satisfied
      if (mutateCriteria.every(priv => priv === true)) {
        container.securityContext.allowPrivilegeEscalation = false;
        wasMutated = true;
      }
    }
```

MxNxPx: going to abandon this suggestion

src/pepr/policies/security.ts
```diff
@@ -25,14 +25,43 @@ import { exemptionAnnotationPrefix, isExempt, markExemption } from "./exemptions
...
+
+    // Check if any containers defined in the pod do not have the `allowPrivilegeEscalation` field present. If not, include it and set to false.
+    for (const container of containers(request)) {
+      container.securityContext = container.securityContext || {};
+      const mutateCriteria = [
+        container.securityContext.allowPrivilegeEscalation === undefined,
+        !container.securityContext.privileged,
+        !container.securityContext.capabilities?.add?.includes("CAP_SYS_ADMIN"),
+      ];
+      //We are only mutating if the conditions above are all satisifed
```
MxNxPx:
```suggestion
      // We are only mutating if the conditions above are all satisfied
```

mjnagel (approved):

PR #699, opened 2024-08-23
feat: add istio ambient

### LINKED ISSUE #681: Add Istio ambient mesh support
### Overview

I would like the ability as a UDS Core user to opt into Istio's ambient mode (no sidecar) when deploying UDS Core. Ambient mode is still being developed, but appears to be gaining traction and promises resource and security advantages over the existing sidecar mesh model:

https://istio.io/latest/blog/2022/introducing-ambient-mesh/
https://istio.io/latest/blog/2022/ambient-security/
https://istio.io/latest/docs/ambient/architecture/traffic-redirection/

- A deployment variable to allow opting in to Istio's ambient mode, using `--set`.
- Alternatively, a UDS Core "flavor" which uses ambient mode by default.

- Istio is usable today w/ the existing sidecar mode, though in large-scale deployments like ours, we reserve signficant resources for the sidecar container, multiplied by hundreds (!) of pods. 

### Additional context

- The `ztunnel` agent might be necessary as a separate container image within. There's a container available in IronBank for `ztunnel` but I need to do more research to determine how the `ztunnel` agent is intended to be deployed.
- `ztunnel` has FIPS mode support: https://github.com/istio/ztunnel?tab=readme-ov-file#boring-fips

rjferguson21:
## Description

Enables Istio Ambient for UDS Core with optionality in the Package CR to use ambient or sidecars.

[Short doc on ambient and the experience with it
](https://github.com/defenseunicorns/uds-core/blob/istio-ambient/src/istio/Ambient.md)

## Related Issue

Probably Fixes https://github.com/defenseunicorns/uds-core/issues/681

## Type of change

## Checklist before merging


mjnagel:
Closing - this can be used for reference but we plan to attack ambient at a more granular level with smaller PRs.

PR #700, opened 2024-08-26
fix: update ci workflows for docs shim

### LINKED ISSUE #589: Skip full testing for documentation only changes
### Feature Request

When a PR is created with documentation only changes - only linting/docs validation should run. In order to ensure that the required PR checks still pass we may need to build out a "docs shim" to run workflows with the same name (see [this](https://github.com/defenseunicorns/uds-common/blob/cgr-setup/.github/workflows/ci-docs-shim.yaml) as an example).

### LINKED ISSUE #628: Apply more granular checks before triggering actions in CI

The CI/CD pipeline runs around 17 unique checks for all PRs with the existing conditional logic.
In some cases (e.g., updating a `.md` documentation file) it is probably not necessary to run all 17 checks.

For example, we may choose to skip running tests on PRs that only change `.md` files.
This could encourage contributors focused on just documentation updates to be more active.

For an even faster feedback loop, we can consider using more granular trigger rules in GitHub actions.

* [dorny/paths-filter](https://github.com/dorny/paths-filter)
* Current implementation of conditional checks: [pull-request-conditionals.yaml](https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/pull-request-conditionals.yaml)
* Filters are defined in [filters.yaml](https://github.com/defenseunicorns/uds-core/blob/main/.github/filters.yaml)

### Additional context

This is a good issue to work on so that you become familiar with how the `uds-core` CI/CD pipeline is configured.

chance-coleman:
## Description
We don't want to run CI jobs on documentation only changes.

Testing was done in a personal repo.

## Related Issue

Fixes #589 
Fixes #628 

## Type of change

## Checklist before merging


noahpb:
I'm not super familiar with the `Select Jobs` workflow, but would we also have to exclude docs related directories and file extensions in `.github/filters.yaml` to avoid `all` packages being [selected](https://github.com/defenseunicorns/uds-core/actions/runs/10495965282/job/29075610748#step:3:23) in test runs where docs changes and code changes are made on the same branch? I.E. if I submit a PR for the istio package but also update `README.md` in the same PR, would that trigger the pipeline to test all packages (given the [current](https://github.com/defenseunicorns/uds-core/blob/269ae8be86b2434ea26b2f8d34a44945002e71a9/.github/filters.yaml#L2) filter for `all`)?

chance-coleman:
@noahpb yeah that's a good call out

chance-coleman:
links to personal repo where you can see the PR's for different usecases. hopefully public visibility 

- [project readme.md ( only doc test )](https://github.com/UnicornChance/uds-core/pull/36)
- [authservice readme.md and deployment.yaml ( doc and yaml in same directory test )](https://github.com/UnicornChance/uds-core/pull/38)
- [keycloak readme.md and deployment.yaml ( slim dev test )](https://github.com/UnicornChance/uds-core/pull/37)
- [grafana readme and authservice deployment.yaml ( doc plus yaml in separate directory test )](https://github.com/UnicornChance/uds-core/pull/39)
- [keycloak readme.md only ( only doc test for slim dev )](https://github.com/UnicornChance/uds-core/pull/40)


Also found that the slim-dev job only ran when changes were made to the first level of the directories it was watching. Needed to be updated to `/**` to pick up changes lower than first level.

noahpb (dismissed):

.github/workflows/docs-shim.yaml
```diff
@@ -0,0 +1,85 @@
...
+        run: echo "Simulating check-paths job - Success!"
+
+  run-package-test:
+    needs: check-only-docs-changes
+    if: ${{ needs.check-only-docs-changes.outputs.only_docs_changed == 'false' }}
+    name: Package Test(Simulated)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run Package Test
+        run: echo "Simulating run-package-test job - Success!"
```
mjnagel: Discussed offline, but I'm not sure this actually handles the original issue in the way we'd want. Our repo has required checks for `Schedule (all, <flavor>, <install/upgrade>) / Test`, so we'd need jobs named identically here (with the matrix) in order to satisfy those required checks.

Also think we can ignore slim-dev and single-package for now since those aren't required and just always run lint-check (at some point that may encompass docs linting/spellcheck).

chance-coleman:
latest commit ensures that expected behavior for docs only changes passes repo merge requirements. As well as adds eks concurrency limits so that previous commit eks tests are cancelled before new eks tests are run ( hopefully limiting the number of orphaned eks clusters in AWS )

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -24,26 +38,8 @@ concurrency:
...
-      - name: Install UDS CLI
-        shell: bash
-        # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
-        run: brew install defenseunicorns/tap/uds@0.14.2
-      - name: Run Formatting Checks
-        run: uds run lint-check --no-progress
-
   # This job checks if there are changes in specific paths source packages.
   check-paths:
-    needs: lint-check
```
mjnagel: Preferably would like to keep this dependent on linting for better fast feedback/cancelling of workflows when something fails lint. As discussed we could pull the lint into a shareable workflow/action to let us call it in both the shim and this workflow.

mjnagel (approved):

PR #703, opened 2024-08-26
fix: pre-commit linting

chance-coleman:
## Description
numerous times lately I have pushed commits that break our linting ( typically the end of file new line ) but on occasion also spacing changes. I've noticed others occasionally have the same issue. Setup a pre-commit check that runs the same linting command that `uds run lint` uses.

If lint is successful the commit is made, if lint failed commit is not made and user is notified of what failed lint so they can fix. This also only tests staged files.

Also could be built out a bit more to add restrictions to commit messages and other things if we wanted, or even logic to run the lint-fix ( which doesn't always work ). 

## Type of change

## Checklist before merging


package.json
```diff
@@ -33,7 +33,9 @@
   },
   "devDependencies": {
     "@jest/globals": "29.7.0",
+    "husky": "^8.0.0",
     "jest": "29.7.0",
+    "lint-staged": "^15.2.9",
```
mjnagel: Noting that at the time being including these in devDependencies ends up baking them into our final pepr artifact/what runs in cluster, see https://github.com/defenseunicorns/pepr/issues/1103

chance-coleman: @mjnagel the linked issue doesn't seem to be going anywhere

mjnagel: Walked through validating what was stated in that ticket. It seems valid that the only effect of adding this here is that it shows up in a copy of the package.json, nothing is in the actual code/accessible from what I can tell. Resolving this.

.lintstagedrc.json
```diff
@@ -0,0 +1,7 @@
+{
+    "*": [
+        "npm ci",
+        "npx pepr format --validate-only",
```
mjnagel: Would suggest we check if we can change to:
```suggestion
        "npx pepr format",
```

So that things get formatted pre-commit rather than just failing the commit.

.lintstagedrc.json
```diff
@@ -0,0 +1,7 @@
...
+        "yamllint -c .yamllint --no-warnings"
+    ]
```
mjnagel, line 10: Also would be neat to add https://github.com/codespell-project/codespell here. It should generate few/no false-positives because of how it works, but might be good to setup an ignore file anyways just in case?

.codespellrc
```diff
@@ -0,0 +1,6 @@
+# Lint Codespell configurations
+[codespell]
+skip = .codespellrc,.git,node_modules,build,dist,*.zst,CHANGELOG.md
+ignore-words-list = NotIn,AKS
+enable-colors =
+count =
```
mjnagel: nit: the count was confusing me in the output - we can leave if you think it's valuable but my vote would be to drop this line.

.lintstagedrc.json
```diff
@@ -0,0 +1,11 @@
+{
+    "*": [
+        "bash -c 'codespell'"
```
mjnagel: You may have explained to me already but wondering why this is wrapped in `bash -c`? I changed it locally to just `codespell` and it seemed to work as expected still.

docs/configuration/uds-operator.md
```diff
@@ -169,6 +169,17 @@ The UDS Operator uses the first `redirectUris` to populate the `match.prefix` ho
...
+If you wish to scale Authservice horizontally, Authservice supports using an [external redis session store](https://docs.tetrate.io/istio-authservice/configuration/oidc#session-store-configuration) which can be configured by setting [UDS_AUTHSERVICE_REDIS_URI](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/zarf.yaml#L20-L22).
+
+You can also specify the `AUTHSERVICE_REDIS_URI` variable in your `uds-config.yaml`:
+
+```yaml
+variables:
+  core:
+    AUTHSERVICE_REDIS_URI: redis://redis.redis.svc.cluster.local:6379
+```
+
```
mjnagel: Suspect this was added from some conflicts - I think we moved this doc to [here](https://github.com/defenseunicorns/uds-core/blob/pre-commit-lint/docs/configuration/resource-configuration-and-ha.md#authservice) so may be able to drop this section?

tasks/lint.yaml
```diff
@@ -5,9 +5,14 @@ tasks:
...
     actions:
+      - description: install codespell deps
+        cmd: CMD=pip && which $CMD || CMD=pip3 && $CMD install codespell
       - description: "Pepr Format"
         cmd: npx pepr format
-
+      - description: Show codespell lint issues
+        cmd: codespell
+      - description: Fix codespell lint issues
+        cmd: codespell -w
```
mjnagel: I kind of like the UX of squishing these two together in a single action (removes the `uds run` boilerplate between them).
```suggestion
      - description: Fix codespell lint issues
        cmd: |
          codespell || true
          codespell -w
```

Okay with either but we will want to add the `|| true` to the `codespell` "show" to ensure that it passes and allows the fix to run.

.husky/pre-commit
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+OUTPUT=$(npx lint-staged > /dev/null && echo $? || echo $?)
+
+if [ $OUTPUT -eq 0 ]; then
+    echo "\n\n✅ Lint Check Passed. ✅\n\n"
+    exit 0
+else
+    echo "\n\n❌ Lint Check failed... Use the uds task: lint-fix to resolve issues and re-commit. ❌\n\n"
```
mjnagel:
```suggestion
    echo "\n\n❌ Lint Check failed... Run \`uds run lint-fix\`  to resolve issues and re-commit. ❌\n\n"
```
This might be a more "beginner friendly" output (just give the exact command)?

mjnagel (commented):
LGTM overall - few small nits/UX comments.

.lintstagedrc.json
chance-coleman: was needed before the bash pieces we put in the pre-commit. no longer needed.

mjnagel (approved):

PR #710, opened 2024-08-28
fix: switch secret `data` to `stringData`

noahpb:
## Description
Fixes deployment errors with the operator helm chart not properly handling integer helm values.

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/uds-operator-config/templates/secret.yaml
+++ b/src/pepr/uds-operator-config/templates/secret.yaml
@@ -8,5 +8,5 @@ metadata:
 type: Opaque
 data:
 {{- range $key, $value := .Values.operator }}
-  {{ $key }}: {{ $value | b64enc | quote }}
+  {{ $key }}: {{ $value | toString | b64enc | quote }}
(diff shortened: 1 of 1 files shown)


mjnagel (dismissed):

chance-coleman (dismissed):

src/pepr/uds-operator-config/templates/secret.yaml
```diff
@@ -8,5 +8,5 @@ metadata:
 type: Opaque
 data:
 {{- range $key, $value := .Values.operator }}
-  {{ $key }}: {{ $value | b64enc | quote }}
+  {{ $key }}: {{ $value | toString | b64enc | quote }}
```
mjnagel: Calling out an alternative here that might be a simpler template if we wanted to go that route:
```yaml
stringData:
{{- range $key, $value := .Values.operator }}
  {{ $key }}: {{ $value | quote }}
```

mjnagel (approved):

PR #719, opened 2024-09-03
feat: investigate and restrict network policies

### LINKED ISSUE #558: Lock down "egress anywhere" policies for known external services

Our package should operate under a "least privilege" type model for network access, and specifically egress network access should be limited to specific services/addresses rather than "anywhere". All [current `remoteGenerated: Anywhere` netpols](https://github.com/search?q=repo%3Adefenseunicorns%2Fuds-core%20remoteGenerated%3A%20anywhere&type=code) should be evaluated for their necessity. There are a few changes we could make to reduce the scope of these:
- For egress to SSO we should lock down to the tenant/admin gateway since our service entries will handle this and not require true egress outside the cluster.
- For egress to specific external services like S3, RDS, we should explore istio service entries + a locked down outbound traffic policy
- For egress to "unknown things" (may not have a hostname) we should consider configurable CIDR values for the end user to pass in. This could also be a short-term solve instead of service entries if that is more challenging to implement.

https://github.com/search?q=repo%3Adefenseunicorns%2Fuds-core%20remoteGenerated%3A%20anywhere&type=code

### Additional context

This would also be valuable to ensure we are providing the correct network access in our CI testing and identify places where apps may have phone home telemetry or other external network requirements (related to https://github.com/defenseunicorns/uds-core/issues/487).

chance-coleman:
## Description
Our package should operate under a "least privilege" type model for network access, and specifically egress network access should be limited to specific services/addresses rather than "anywhere".

Investigated current `anywhere` policies, updated restrictions where necessary. Added a new package CR field `remoteCidr` for defining a custom cidr to be used in place of the anywhere cidr. Add some validations to verify the use the `remoteGenerated`, `remoteSelector`, `remoteNamespace`, and `remoteCidr` don't overlap or break each other. They should be used individually except `remoteSelector` and `remoteNamespace` being used together.

Potentially follow on issues for _KubeAPI ingress relation network policy management, as well as utilizing service entries for known things like S3 buckets.

## Related Issue

Fixes #558

## Type of change

## Checklist before merging


src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -26,14 +26,15 @@ spec:
...
-      # @todo: remove this once cluster pods is a remote generated target
+      # todo: remove this once cluster pods is a remote generated target
       - description: "Keycloak backchannel access"
         direction: Ingress
         selector:
           app.kubernetes.io/name: keycloak
         remoteGenerated: Anywhere
         port: 8080
 
+      # Keycloak OCSP to check certs cannot gurantee a static IP
```
mjnagel:
```suggestion
      # Keycloak OCSP to check certs cannot guarantee a static IP
```

src/authservice/chart/templates/uds-package.yaml
```diff
@@ -14,9 +14,17 @@ spec:
...
+          app: tenant-ingressgateway
+        remoteNamespace: istio-tenant-gateway
+        description: "SSO Provider"
+
+      {{- if .Values.redisUri }}
       - direction: Egress
         remoteGenerated: Anywhere
-        description: "SSO Provider & Redis Session Store"
+        description: "Redis Session Store"
+      {{- end }}
```
mjnagel: nit: could we do a similar template here to what was done with S3? Example:
```yaml
      {{- if .Values.redis.uri }}
      - direction: Egress
        description: Redis Session Store
        {{- if .Values.redis.internal.enabled }}
        remoteSelector: {{ .Values.redis.internal.remoteSelector }}
        remoteNamespace: {{ .Values.redis.internal.remoteNamespace }}
        {{- else if .Values.redis.cidr }}
        remoteCidr: {{ .Values.redis.cidr }}
        {{- else }}
        remoteGenerated: Anywhere
        {{- end }}
      {{- end }}
```

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -58,9 +58,33 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+        );
+      }
+    } else if (policy.remoteNamespace) {
+      if (policy.remoteGenerated || policy.remoteCidr) {
+        return req.Deny("remoteNamespace cannot be combined with remoteGenerated or remoteCidr");
+      }
+    } else if (policy.remoteSelector) {
+      if (policy.remoteGenerated || policy.remoteCidr) {
+        return req.Deny("remoteSelector cannot be combined with remoteGenerated or remoteCidr");
+      }
```
mjnagel: nit: I'd be in favor of collapsing these 3 for code simplicity...
```suggestion
    } else if (policy.remoteNamespace || policy.remoteSelector) {
      if (policy.remoteGenerated || policy.remoteCidr) {
        return req.Deny(
          "remoteNamespace and remoteSelector cannot be combined with remoteGenerated or remoteCidr",
        );
      }
```

If we wanted to make the deny message super explicit about which one was used that clashed I think we could probably conditionally modify the deny message.

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -26,14 +26,15 @@ spec:
...
 
       # Temp workaround for any cluster pod
-      # @todo: remove this once cluster pods is a remote generated target
+      # todo: remove this once cluster pods is a remote generated target
       - description: "Keycloak backchannel access"
         direction: Ingress
         selector:
           app.kubernetes.io/name: keycloak
         remoteGenerated: Anywhere
         port: 8080
```
mjnagel, line 36: Not commenting on the right lines here, but might suggest we lockdown the "PostgreSQL Database access" policy with conditionals for internal, cidr, or anywhere (similar to storage conditionals for velero/loki).

src/grafana/chart/templates/uds-package.yaml
```diff
@@ -28,23 +28,20 @@ spec:
...
         selector:
           app.kubernetes.io/name: grafana
-        remoteNamespace: tempo
+        remoteNamespace: loki
         remoteSelector:
-          app.kubernetes.io/name: tempo
-        port: 9090
-        description: "Tempo Datasource"
+          app.kubernetes.io/name: loki
+        description: "for loki"
```
mjnagel: very nit, but might prefer this description to be "Loki Datasource"
```suggestion
        description: "Loki Datasource"
```

Current netpol name for reference:
```
allow-grafana-egress-for-loki                            app.kubernetes.io/name=grafana   8s
```

src/grafana/chart/templates/uds-package.yaml
```diff
@@ -28,23 +28,20 @@ spec:
...
-
-      - direction: Egress
-        remoteNamespace: tempo
+        remoteNamespace: monitoring
         remoteSelector:
-          app.kubernetes.io/name: tempo
-        port: 9411
-        description: "Tempo"
+          app.kubernetes.io/name: prometheus
+        description: "for monitoring"
```
mjnagel: very nit, but might prefer this description to be "Prometheus Datasource"
```suggestion
        description: "Prometheus Datasource"
```

Current netpol name for reference:
```
allow-grafana-egress-for-monitoring                      app.kubernetes.io/name=grafana   8s
```

(on this one, if you don't agree with this suggestion would suggest at least changing to `prometheus` instead of `monitoring`)

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -52,6 +53,8 @@ function getPeers(policy: Allow): V1NetworkPolicyPeer[] {
     }
 
     peers.push(peer);
+  } else if (policy.remoteCidr !== undefined) {
+    peers = [remoteCidr(policy.remoteCidr)];
```
mjnagel, line 57: Could we toss a remoteCidr policy in one of our test packages? Doesn't need to be "tested" beyond the package being ready, just would like to make sure we have one somewhere in our code.

mjnagel (commented):
A ton of great work here - thanks for locking this down @UnicornChance . Couple of other notes beyond my specific line comments here:
- Might be good to document the new `storage`, etc values _somewhere_ and their effect on netpols
- @MxNxPx tagging for awareness on the "document package CR" side - remoteCidr would be a good new item to include
- I'm unsure if some of the keycloak access is working as expected. I deleted the uds-k3d coredns override and am seeing some connection refused errors from Grafana as well as Neuvector (nothing clear on this one for the actual error.

src/grafana/chart/templates/uds-package.yaml
mjnagel: I think we could also lock this policy down to a specific port - potentially port 8080 (https://github.com/defenseunicorns/uds-core/blob/lock-down-egress/src/promtail/chart/templates/uds-package.yaml#L34-L36)

src/grafana/chart/templates/uds-package.yaml
mjnagel: I think we could also lock this policy down to a specific port - potentially port 9090 (https://github.com/defenseunicorns/uds-core/blob/lock-down-egress/src/grafana/chart/templates/datasources.yaml#L16)

src/pepr/operator/controllers/network/generate.ts
chance-coleman: Added generate.spec.ts tests as well as a new netpol for the test-tenant-app that utilizes the remoteCidr.

src/pepr/operator/crd/validators/package-validator.ts
chance-coleman: accepting your proposed changes here, we can come back later and add more explicit logging if it's an issue. IMO

mjnagel (dismissed):

src/grafana/chart/templates/uds-package.yaml
```diff
@@ -28,23 +28,31 @@ spec:
...
-        remoteNamespace: tempo
+        selector:
+          app.kubernetes.io/name: grafana
+        remoteNamespace: keycloak
         remoteSelector:
-          app.kubernetes.io/name: tempo
-        port: 9411
-        description: "Tempo"
+          app.kubernetes.io/name: keycloak
+        description: "Keycloak Datasource"
```
mjnagel: very nit - technically keycloak isn't a grafana datasource so this could be SSO Provider as the description maybe?

mjnagel:
```suggestion
        description: "SSO Provider"
```

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -58,9 +58,25 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
   const networkPolicyNames = new Set<string>();
 
   for (const policy of networkPolicy) {
-    // remoteGenerated cannot be combined with remoteNamespace or remoteSelector
-    if (policy.remoteGenerated && (policy.remoteNamespace || policy.remoteSelector)) {
-      return req.Deny("remoteGenerated cannot be combined with remoteNamespace or remoteSelector");
+    // Ensure only valid combinations of remoteGenerated, remoteNamespace, remoteSelector, and remoteCidr
+    if (policy.remoteGenerated) {
```
rjferguson21: A bit of a nit but I would probably avoid each nested if statement in this block and just combine them all into a single conditional.

src/velero/chart/values.yaml
```diff
@@ -0,0 +1,6 @@
+storage:
+  internal:
+    enabled: false
+    remoteSelector: {}
+    remoteNamespace: ""
+  cidr: ""
```
rjferguson21: I think it would be helpful to call this something that makes it more clear it is used for network policy generation.

mjnagel (approved):

rjferguson21 (approved):

Issue #685, opened 2024-08-21
Update CI to support release candidate releases


rjferguson21:
### Is your feature request related to a problem? Please describe.

When deploying a release to my environment I would like a higher level of confidence that it has been tested in a production like environment, ideally for longer than a CI run.

### Describe the solution you'd like

When a release is ready to be cut (i.e. within a day or two), then a maintainer of the project can create a tag with the `-rc.0` suffix, designating it as a release candidate. This release candidate tagging should then trigger CI to publish artifact(s) for the release that could be consumed in a long-lived testing environment.

May need some evaluation on which artifacts should be published for RCs:
- Slim-dev bundle: This is of minimal value for the stated use case of prod testing so might advocate we skip publishing RCs for it
- Demo bundle: This is of minimal value for the stated use case of prod testing so might advocate we skip publishing RCs for it
- Core package (all flavors): Definitely need to publish at least one flavor, I think it makes sense to publish all

### Describe alternatives you've considered

I could deploy a nightly-unstable release to my test cluster but it is not guaranteed to be the same as the release and also makes it hard to track versioning/what I am testing.

### Additional context

The current [workflow run on main](https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/tag-and-release.yaml) is used to trigger the [publish workflow](https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/publish.yaml). I think we could add an additional trigger that calls the publish workflow when a tag/release is created, specifically matching a regex of `[0-9]*.[0-9]*.[0-9]*-rc.[0-9]*`. We would also need to make sure that versions in the package are updated to the correct tag version (could copy [what was done for snapshot](https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/publish.yaml#L36-L50)).

We will want to be very careful to ensure that we do not write "real" release artifacts through this process. We may also want to modify/create tag restrictions in this repo to lock down who is able to create these RC releases.

mjnagel:
Closing as not planned with our current snapshot workflows providing sufficient coverage for now.

PR #724, opened 2024-09-04
feat!: switch from promtail to vector

### LINKED ISSUE #377: Ship Logs to Multiple Destinations

mjnagel:
## Description

BREAKING CHANGE: Promtail has been removed from UDS Core and replaced by Vector. If using overrides to setup additional log targets/endpoints this configuration will need to be updated to Vector's chart/config formats. See Vector's [Sources and Sinks](https://vector.dev/components/) as well as the [helm chart values](https://github.com/defenseunicorns/uds-core/blob/1bf29582f9c5b1fe01763e86e56c19b6e17aef85/src/vector/values/values.yaml#L4) for guidance in configuration. 

Primary docs on rationale, decision, and impact of this switch are [here](https://github.com/defenseunicorns/uds-core/blob/vector-add/src/vector/README.md).

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/377

## Type of change

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: switch from promtail to vector (https://github.com/defenseunicorns/uds-core/pull/724)

BREAKING CHANGE: Promtail has been removed from UDS Core and replaced by Vector. If you were previously using overrides to setup additional log targets/endpoints for Promtail this configuration will need to be updated to Vector's chart/config formats. See Vector's [Sources and Sinks](https://vector.dev/components/) as well as the [helm chart values](https://github.com/defenseunicorns/uds-core/blob/1bf29582f9c5b1fe01763e86e56c19b6e17aef85/src/vector/values/values.yaml#L4) for guidance in configuration. 
END_COMMIT_OVERRIDE


docs/application-baseline.md
```diff
@@ -18,7 +18,7 @@ For optimal deployment and operational efficiency, it is important to deliver a
...
-| **Logging**                        | **[Promtail](https://grafana.com/docs/loki/latest/send-data/promtail/):** A companion agent that efficiently gathers and sends container logs to Loki, simplifying log monitoring, troubleshooting, and compliance auditing, enhancing the overall observability of the mission environment.<br><br> **[Loki](https://grafana.com/docs/loki/latest/):** A log aggregation system that allows users to store, search, and analyze logs across their applications. |
+| **Logging**                        | **[Vector](https://vector.dev/):** A companion agent that efficiently gathers and sends container logs to Loki, simplifying log monitoring, troubleshooting, and compliance auditing, enhancing the overall observability of the mission environment.<br><br> **[Loki](https://grafana.com/docs/loki/latest/):** A log aggregation system that allows users to store, search, and analyze logs across their applications. |
```
MxNxPx: Would it make sense here to mention the ability to send logs to Loki and/or other destinations?

src/vector/oscal-component.yaml
```diff
@@ -78,9 +78,9 @@ component-definition:
...
 
                 # Control Implementation
-                  Logs are captured by promtail from the node. The node logs will contain the necessary log data from all pods/applications inside the selected nodes.
-                  Validating `logfmt` as the config.logFormat would be the goal. This is currently a secret mounted to /etc/promtail/promtail.yaml in the promtail container. We will ensure the promtail.yaml file is at a minimum the target config.
-                https://grafana.com/docs/loki/latest/send-data/promtail/stages/logfmt/
+                  Logs are captured by vector from the node. The node logs will contain the necessary log data from all pods/applications inside the selected nodes.
+                  Validating `logfmt` as the config.logFormat would be the goal. This is currently a secret mounted to /etc/vector/vector.yaml in the vector container. We will ensure the vector.yaml file is at a minimum the target config.
+                https://grafana.com/docs/loki/latest/send-data/vector/stages/logfmt/
```
MxNxPx: remove this link?

mjnagel: Yeah absolutely - I did an initial find+replace of promtail -> vector, but expect that @CloudBeard might have more changes on this one. Didn't spend too much effort on the OSCAL in particular because of that.

docs/application-baseline.md
mjnagel: Definitely, good call - similar to the other I most just find+replace here, but I think expanding on this would be good.

src/vector/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+As with any decisions of tooling in core this can always be reevaluated in the future as different tools or factors affect how we look at our logging stack.
+
+### Upgrade Considerations
+
+During the upgrade there may be some duplication/overlap of log lines shipped to Loki due to the transition from Promtail's "position" file to Vector's "checkpoint" file (both used for tracking the last log line scraped/shipped). Grafana provides a built in feature to de-duplicate log entries when querying Loki, but this does not consistently work with all log lines due to the approach used by Grafana for de-duplication.
+
+To ensure easy querying of logs across the upgrade, all logs shipped by Vector also have a `collector` label (with the value of `vector`). This can be used to filter down any logs to those collected by either Vector or Promtail (using the `=` and `!=` operators). In general you can use these filters along with tracking your upgrade timing to properly ignore duplicate logs for the short upgrade period.
```
mjnagel, line 42: Open to other thoughts on this - wanted to specifically note it as a known "limitation" here during the upgrade. Also unsure if we should relocate this doc/this portion of the doc to the `docs/` folder and docs site - or if linking this in the release notes would be sufficient?

chance-coleman: Also unsure if bringing this into the docs IF its a long lived doc... I think if we bring it into the docs we should add some upgrade steps / troubleshooting / expectations to the docs for the upgrade.

src/vector/README.md
```diff
@@ -0,0 +1,42 @@
...
+### Motivations
+
+Promtail has historically been the tool of choice for log collection/shipping when using Loki. It provides a very lightweight layer to scrape logs from pods and hosts, label them with additional metadata, and ship them to Loki.
+
+One of the main issues that has arisen with Promtail is its limited output/export options. Promtail only supports sending logs to one or more Loki instances. A common requirement in production environments is to ship logs to a secondary destination for collection/analysis by security teams and SIEM tools. Promtail is currently listed as [feature complete](https://grafana.com/docs/loki/latest/send-data/promtail/) so there is no expectation that additional export functionality would be added.
+
+### Goals and Options
+
+In choosing an alternative to Promtail we have a few primary objectives:
+1. Chosen tool must be capable of gathering host and pod logs: This has been our primary usage of Promtail in the past - gathering pods logs and host logs (to include k8s audit logs, controlplane logs, etc).
```
chance-coleman: nit: Im partial to using bullets here if there isn't any sequence to how it should be read.

mjnagel: Will make that change - might wait on more reviews to not burn CI on a small docs modification but taking note of this.

src/vector/values/values.yaml
```diff
@@ -0,0 +1,118 @@
...
+    node_logs_labelled:
+      type: remap
+      inputs: ["node_logs"]
+      source: |
+        .node_name = "${NODE_HOSTNAME}"
+        if contains(string!(.file), "/var/log/kubernetes/") {
+          .job = "kubernetes-logs"
+        } else {
+          .job = "varlogs"
+        }
```
rjferguson21, line 52: This might be the 'recommended way' but curious if there is an easy way to just have separate sources for each `.job` type instead of running this VRL one every line?

src/vector/Dockerfile
rjferguson21, line 1: Is this used somewhere?

rjferguson21:
Open to discussion on other options to accomplish the same end but I think ideally this could be deployed to staging pre-release to build our confidence.

src/vector/Dockerfile
mjnagel: Good catch - this was the first draft of the IB image before it was in registry1. Deleting...

src/vector/values/values.yaml
mjnagel: We could remove the VRL here, but I think it's more DRY to do it this way. Without the VRL you need:
- 2 sources: ~3 lines duplicated
- 2 sinks: ~14 lines duplicated

It was really the duplication of the sinks that pushed me towards the remap with VRL - since labels cannot be added on the file source itself you end up needing 2 sinks if you want different labels from the two different sources. Are there particular concerns with this path? I can split out into 2 sources and sinks if desired, but wasn't overly concerned with the addition of this check based on upstream [VRL performance notes](https://vector.dev/docs/reference/vrl/#performance). Happy to chat more and see what makes the most sense here.

mjnagel: I think another alternative - we could just rethink the job label here and use `varlogs` for both, and lean into Loki querying for someone to identify specific k8s audit logs. That would be a slight shift and probably good to callout, but it would remove the VRL usage and keep this DRY.

rjferguson21: Hmm.  I think based on that I'm good with just leaving it.  If it becomes an issue we can revisit but you're probably right in that it will be fine performance wise.

mjnagel:
> Open to discussion on other options to accomplish the same end but I think ideally this could be deployed to staging pre-release to build our confidence.

@rjferguson21 Are you thinking post-merge, pre-release? I think I agree - may require us to figure out https://github.com/defenseunicorns/uds-core/issues/685 unless we just want to cut a nightly and use that?

rjferguson21:
Yea, I think post-merge / pre-release makes the most sense. I'd be fine with a nightly assuming it is possible.

rjferguson21 (approved):

PR #734, opened 2024-09-09
chore: update loki

rjferguson21:
## Description

...

## Related Issue

Fixes #

Relates to #

## Type of change

## Checklist before merging


src/loki/chart/templates/loki-schema.yaml
```diff
@@ -0,0 +1,15 @@
...
+data:
+  # v13schemadate defaults to two days in the future
+  {{- $configmap := (lookup "v1" "ConfigMap" "loki" "loki-schema") }}
+  {{- if .Values.v13schemadate }}
+  v13schemadate: {{ .Values.v13schemadate }}
+  {{- else if $configmap }}
+  v13schemadate: {{ $configmap.data.v13schemadate }}
+  {{- else }}
+  v13schemadate: {{ (now | dateModify "+48h" | date "2006-01-02") | quote }}
+  {{- end }}
```
mjnagel: (realize that I suggested some of this, but a few thoughts)

- Not really sure we need `v13schemadate` value - if someone wants to override they could directly modify the loki schema values, although they'd have to override the full block
- If this is a fresh install we can actually put the date in the past - we don't have to, but it's one less thing for new installs to have to include. Probably could do with [`.Release.IsInstall`](https://helm.sh/docs/chart_template_guide/builtin_objects/) + dateModify minus 48h?
- Controlling this schema config with helm templating makes me a bit scared, but it may be the right place? Can see both sides to this...

mjnagel (approved):

mjnagel (commented):
approving since this is just a PR into another branch/PR

PR #735, opened 2024-09-09
fix: batch authservice checksum updates

### LINKED ISSUE #561: Batch authservice checksum updates

Created from [thread](https://github.com/defenseunicorns/uds-core/pull/201#discussion_r1672863840). Currently we apply a new checksum to the authservice deployment for each chain addition, update, or deletion. This results in multiple applies, sometimes for a single package.

We should find a way to batch the checksum applies so that there is not excessive applies or cycling of the authservice pod. This could be done in two ways:
- Simple: batch per packages so that a package with multiple authservice clients only results in one checksum. This would be easy to move the checksum functionality to the end of the authservice reconciliation
- Harder: batch on some time interval to handle multiple packages at once.

chance-coleman:
## Description

Currently we update the authservice secret and apply a checksum each time a chain is added, updated or deleted. This ends up resulting in unnecessary churn on Authservice and also can be problematic when we switch to the `kindNsName` based queues for Pepr (multiple authservice clients can reconcile concurrently).

The changes in this PR switch to using an in-memory secret for tracking chain changes (rather than pulling from the cluster on each change) as well as debouncing updates so that concurrent chain creations are batched together.

## Related Issue

Fixes #561 

## Type of change

## Checklist before merging


rjferguson21:
@UnicornChance Did you take a stab at addressing at batching authservice updates across multiple packages? 

IMO I don't think this is worth doing unless we address that scenario.  I don't think we have anyone who currently using more than one sso client per Package.

I would try to debounce whatever function is actually checksuming the deployment and that would probably work.

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -1,29 +1,48 @@
...
+const pendingPackages: Map<AuthserviceConfig, () => void> = new Map();
+
+// Backup for the last known successful state of the secret
+let lastSuccessfulSecret: AuthserviceConfig | null = null;
+
+// Timer for debouncing updates to the secret
+let debounceTimer: NodeJS.Timeout | null = null;
+
+// Debounce duration (12 seconds) to reduce excessive updates, configurable via environment variable
+const DEBOUNCE_DURATION = parseInt(process.env.DEBONCE_DURATION || "12000", 10);
```
rjferguson21: IMO this should be way lower.  We need to balance the responsiveness of reconciling a single package.  

I would probably test by applying `n` packages at the same time.  If those packages are updating as part of the same secret update, lower the debounce time.  Repeat until the logs indicate there are multiple updates happening and then dial it back.

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -94,47 +120,122 @@ function buildInitialSecret(): AuthserviceConfig {
...
+
+        // Resolve the promises for all pending packages after the secret update
+        pendingPackages.forEach(resolveFunc => {
+          resolveFunc();
+        });
+      } catch (e) {
+        log.error(e, `Failed to write authservice secret`);
+
+        // Rollback to the last known successful state if the update fails
+        inMemorySecret = lastSuccessfulSecret;
```
rjferguson21, line 219: It looks like this is reverting back to `lastSuccessfulSecret` but that is only ever set/updated on initial secret fetch?

src/pepr/operator/controllers/keycloak/authservice/config.ts
chance-coleman: @rjferguson21 yeah good call out. I came to this number because i was testing with 8 packages and this is how long the debounce needed for it to be a single update vs two updates. Not sure what the best number of packages would be to use maybe half that? 4-5

src/pepr/operator/controllers/keycloak/authservice/config.ts
chance-coleman: @rjferguson21 iirc my thought process here was that the `lastSuccessfulSecret` is global in memory state meaning it would be the original secret until a successful secret update at which time it would update, regardless if it was current thread or another thread, that way if one of the threads did fail it would still have the most current `successfulSecret` ( in the event there was or wasnt updates to the secret in the time it took to fail ).

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -4,31 +4,55 @@
...
+> = new Map();
+
+// Backup for the last known successful state of the secret
+let lastSuccessfulSecret: AuthserviceConfig | null = null;
+
+// Timer for debouncing updates to the secret
+let debounceTimer: NodeJS.Timeout | null = null;
+
+// Debounce duration (12 seconds) to reduce excessive updates, configurable via environment variable
+const DEBOUNCE_DURATION = parseInt(process.env.DEBONCE_DURATION || "1000", 10);
```
chance-coleman: Probably from my original code, typo here:

```suggestion
const DEBOUNCE_DURATION = parseInt(process.env.DEBOUNCE_DURATION || "1000", 10);
```

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -99,47 +130,112 @@ function buildInitialSecret(): AuthserviceConfig {
...
+ * @returns {Promise<AuthserviceConfig>} - The authservice configuration.
+ */
+export async function getAuthserviceConfig(): Promise<AuthserviceConfig> {
+  if (inMemorySecret) {
+    log.info("Returning in-memory authservice secret");
+    return Promise.resolve(inMemorySecret);
+  }
+
+  // Fetch the authservice secret from Kubernetes if not in cache
+  pendingSecretFetch = K8s(kind.Secret)
```
chance-coleman: In the event of multiple simultaneous calls to this function, i think we could end up with multiple calls to fetch the secret if there isn't one in the cache already. Checking before could be beneficial **i think**. 
```suggestion
if (pendingSecretFetch) {
  return pendingSecretFetch;
}
pendingSecretFetch = K8s(kind.Secret)
...
```

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -99,47 +130,112 @@ function buildInitialSecret(): AuthserviceConfig {
...
     .InNamespace(operatorConfig.namespace)
-    .Get(operatorConfig.secretName);
-  return JSON.parse(atob(authSvcSecret!.data!["config.json"])) as AuthserviceConfig;
+    .Get(operatorConfig.secretName)
+    .then(secret => {
+      const config = JSON.parse(atob(secret.data!["config.json"])) as AuthserviceConfig;
```
chance-coleman: Is it possible that `secret.data` is null? should we have a null check before this `JSON.parse`?

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -156,3 +252,43 @@ async function checksumDeployment(checksum: string) {
...
+          },
+          data: {
+            "config.json": base64EncodedConfig,
+          },
+        },
+        { force: true },
+      )
+      .then(secret => JSON.parse(atob(secret.data!["config.json"])) as AuthserviceConfig);
+  } catch (e) {
+    log.error(`Failed to apply the authservice config secret: ${e.data?.message}`);
```
chance-coleman: No guarantee that the error has the `data` structure, should we have default message here for better error handling? This might be applicable in a couple other places as well.

src/pepr/operator/controllers/keycloak/authservice/config.ts
rjferguson21: Fixed

src/pepr/operator/controllers/keycloak/authservice/config.ts
rjferguson21: Updated

src/pepr/operator/controllers/keycloak/authservice/config.ts
rjferguson21: I could check and throw an explicit error here but I think I'm ok with it throwing all the way back to the main `handleFailure` in the package reconciler.   I think that would be the only benefit to checking, vs just letting it throw an error but let me know if you see it differently.

mjnagel (approved):
Reviewed synchronously with @rjferguson21 - this behaves as expected in testing and LGTM. Definitely a good feature/change to track with users - we may want to capture the reconcile strategy change in the description and/or PR title as well.

PR #758, opened 2024-09-13
docs: bump colima memory recommendation

bushong1:
## Description

Running a core install with colima will exhaust memory right as velero is being deployed.  Bumping memory suggestion from 6 to 7 will get past this.

## Type of change

## Checklist before merging


bushong1:
Disregard.  I had the wrong settings used.

PR #761, opened 2024-09-13
chore: custom pepr image for ci testing

### LINKED ISSUE #760: allow custom dev image for pepr for release candidate testing against UDS Core

We need to be able to run uds packages with a pepr release candidate image in CI.

**Background:** 
When Pepr updates functionality, although they are heavily tested, we don't always know how they will affect UDS Core. Pepr wants to run a compatibility test in CI. 

- **Given** a PR in issued to Pepr that may affect UDS Core
- **When** the Pepr image is built, a uds run command of some sort is run in Pepr CI
- **Then** the packages are fetched to make sure they all reconciled successfully. 

(optional) A clear and concise description of any alternative solutions or features you've considered.

cmwylie19:
## Description

This allows a custom release candidate image of pepr to be run with UDS Core for CI testing 

## Related Issue

Fixes #760 

Relates to #

## Type of change

## Checklist before merging

### DIFF
--- a/tasks/create.yaml
+++ b/tasks/create.yaml
@@ -60,7 +60,17 @@ tasks:
     actions:
       - description: "Build the UDS Core Pepr Module"
         cmd: |
-          CUSTOM_PEPR_IMAGE=$( [ "${FLAVOR}" = "registry1" ] && echo "--custom-image ${REGISTRY1_PEPR_IMAGE}" ) || CUSTOM_PEPR_IMAGE=""
+          if [ "${FLAVOR}" = "registry1" ] ; then
+            PEPR_CUSTOM_IMAGE="--custom-image ${REGISTRY1_PEPR_IMAGE}"
+
+          elif [ -n "${PEPR_CUSTOM_IMAGE}" ] ; then
+            # e.g. PEPR_CUSTOM_IMAGE="pepr:dev uds run slim-dev"
+            PEPR_CUSTOM_IMAGE="--custom-image ${PEPR_CUSTOM_IMAGE}"
+
+          else
+            PEPR_CUSTOM_IMAGE=""
+          fi
+
           rm -fr dist
           npm ci
-          npx pepr build -z chart $CUSTOM_PEPR_IMAGE
+          npx pepr build -z chart $PEPR_CUSTOM_IMAGE
--- a/tasks/create.yaml
+++ b/tasks/create.yaml
@@ -60,13 +60,11 @@ tasks:
     actions:
       - description: "Build the UDS Core Pepr Module"
         cmd: |
-          if [ "${FLAVOR}" = "registry1" ] ; then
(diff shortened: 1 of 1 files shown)


tasks/create.yaml
```diff
@@ -60,7 +60,17 @@ tasks:
     actions:
       - description: "Build the UDS Core Pepr Module"
         cmd: |
-          CUSTOM_PEPR_IMAGE=$( [ "${FLAVOR}" = "registry1" ] && echo "--custom-image ${REGISTRY1_PEPR_IMAGE}" ) || CUSTOM_PEPR_IMAGE=""
+          if [ "${FLAVOR}" = "registry1" ] ; then
```
rjferguson21: IMO we should let `PEPR_CUSTOM_IMAGE` override the pepr image even if the flavor is registry1.

btlghrants: We don't feel too strongly about what should override what, so deferring to you!

PR updated to allow PEPR_CUSTOM_IMAGE to override FLAVOR below. :+1:

tasks/create.yaml
```diff
@@ -60,7 +60,17 @@ tasks:
...
+            PEPR_OVERRIDE_IMAGE="--custom-image ${REGISTRY1_PEPR_IMAGE}"
+
+          if [ -n "${PEPR_CUSTOM_IMAGE}" ] ; then
+            # e.g. PEPR_CUSTOM_IMAGE="pepr:dev uds run slim-dev"
+            PEPR_OVERRIDE_IMAGE="--custom-image ${PEPR_CUSTOM_IMAGE}"
+
+          else
+            PEPR_OVERRIDE_IMAGE=""
+          fi
```
mjnagel:
```suggestion
          if [ -n "${PEPR_CUSTOM_IMAGE}" ] ; then
            # e.g. PEPR_CUSTOM_IMAGE="pepr:dev uds run slim-dev"
            PEPR_OVERRIDE_IMAGE="--custom-image ${PEPR_CUSTOM_IMAGE}"
          elif [ "${FLAVOR}" = "registry1" ] ; then
            PEPR_OVERRIDE_IMAGE="--custom-image ${REGISTRY1_PEPR_IMAGE}"
          else
            PEPR_OVERRIDE_IMAGE=""
          fi
```

I think this fixes the syntax issues and sets the proper precedence.

mjnagel: Might also request that we add `PEPR_OVERRIDE_IMAGE` as a variable to the top of this task, with a description of its usage if possible.

btlghrants: Pfft, yeah... was rushed when pushing the update to this -- meant to make those 3 independent if/fi blocks that all just overrode each other top-down -- dunno what I was thinking with that incomplete condition.  Oof.

I think what you have works just fine too though.

mjnagel (approved):

Issue #748, opened 2024-09-10
Document/examples for additional ingress gateways / non-HTTP traffic


mjnagel:
UDS Core and the UDS Operator out of the box only support HTTP traffic. In some scenarios it may be necessary/desirable to expose a non-HTTP service through gateways.

To do this an end user needs to:
- Provision a new ingress gateway with the necessary ports exposed/forwarded (using the upstream istio helm chart) OR override the ports exposed on one of the existing gateways (tenant/admin) 
- Create a `Gateway` custom resource selecting that ingress gateway, that also specifies the necessary port
- Create a `VirtualService` with the necessary spec (commonly `tcp`)
- Create a `NetworkPolicy` to allow traffic between all the pieces (ingressgateway -> service)

One example of this is Gitlab/SSH:
- Extra manifests: https://github.com/defenseunicorns/uds-bundle-software-factory-nutanix/tree/main/packages/additional-manifests/gitlab
- Bundle overrides to add port to ingress gateway: https://github.com/defenseunicorns/uds-bundle-software-factory-nutanix/blob/main/bundles/uds-core-swf/uds-bundle.yaml#L346-L366

We would like to provide a doc explaining how to do this (separate zarf package with these pieces most likely) and potentially an example of what this would look like. We should also add a warning about why this is not directly enabled in Core and should only be used if required. Primary reasons include:
- simplifying the network boundary, "everything goes through an HTTPS Istio ingress gateway" (allowing non-HTTPs through istio is quite different)
- reducing attack vectors (non-http/s ports may expose additional vulnerabilities)

mjnagel:
This may also help with https://github.com/defenseunicorns/uds-core/issues/785

PR #765, opened 2024-09-16
feat: allow service ports to be overridden in test bundles

Racer159:
## Description

This adds a variable to override service ports in the uds-core test bundles (useful for adding additional ports to expose)

## Related Issue

Relates to https://github.com/defenseunicorns/uds-package-gitlab/pull/196

## Type of change

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
chore: allow service ports to be overridden in test bundles (#765)
END_COMMIT_OVERRIDE

### DIFF
--- a/bundles/k3d-slim-dev/uds-bundle.yaml
+++ b/bundles/k3d-slim-dev/uds-bundle.yaml
@@ -55,6 +55,24 @@ packages:
             - name: TENANT_TLS_KEY
               description: "The TLS key for the tenant gateway (must be base64 encoded)"
               path: tls.key
+        gateway:
+          variables:
+            - name: TENANT_SERVICE_PORTS
+              description: "The ports that are exposed from the tenant gateway LoadBalancer (useful for non-HTTP(S) traffic)"
+              path: "service.ports"
+              default:
+                - name: status-port
+                  port: 15021
+                  protocol: TCP
+                  targetPort: 15021
+                - name: http2
+                  port: 80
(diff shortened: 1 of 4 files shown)


bundles/k3d-slim-dev/uds-bundle.yaml
```diff
@@ -55,6 +55,24 @@ packages:
...
+                  protocol: TCP
+                  targetPort: 15021
+                - name: http2
+                  port: 80
+                  protocol: TCP
+                  targetPort: 80
+                - name: https
+                  port: 443
+                  protocol: TCP
+                  targetPort: 443
```
mjnagel: On the fence on the default - while it is a nice example, it's also something extra to maintain/keep in parity with the upstream defaults. Looking upstream it doesn't look like the ports have changed since the chart was created 3+ years ago, so this might be okay.

Edit: the alternative I might suggest is no default - similar to the TLS pieces. We could focus on docs/examples to add alongside this in https://github.com/defenseunicorns/uds-core/issues/748.

Racer159: Removed it - my initial fear was that it would replace with an empty list but tested that and it does not - I think if people are asking for how to do this we could just point to GL as an example.

mjnagel (approved):
Thanks for the docs updates as well!

rjferguson21 (approved):

PR #768, opened 2024-09-17
chore: remove unused neuvector exporter

mjnagel:
## Description

Delete, delete, delete.

Exporter is not used and cannot be used in its current form due to auth limitations with accessing the metrics when SSO sign in is enabled.

## Related Issue

N/A

## Type of change

## Checklist before merging


rjferguson21 (approved):

chance-coleman (approved):
lgtm, based on our chat i think this makes sense

PR #788, opened 2024-09-19
chore: add uds-runtime as an optional component in core

mjnagel:
## Description

Adds UDS Runtime as an optional component in core. Denoted as a chore since this is optional and we would like to cut a patch release including this for fast consumption.

## Related Issue

Closes https://github.com/defenseunicorns/uds-runtime/issues/346

## Type of change

## Checklist before merging


TristanHoladay:
LGTM

UncleGedd (dismissed):
🚀

src/runtime/zarf.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+components:
+  - name: uds-runtime
+    required: false
+    images:
+      - ghcr.io/defenseunicorns/uds-runtime:0.4.0
+    charts:
+      - name: uds-runtime
+        namespace: uds-runtime
+        version: "v0.4.0"
```
rjferguson21, line 15: Should this not be 0.1.0? Aligned with https://github.com/defenseunicorns/uds-runtime/blob/main/chart/Chart.yaml#L18 ?

mjnagel: This is actually the git tag/version for the repo. Slightly confusing, but makes sure that we are pulling the chart for the same image. Might be good for runtime to synchronize the version of the chart as well - cc @UncleGedd ?

I do believe this works as is, even if not the most obvious - it does deploy in CI and locally as expected.

src/runtime/zarf.yaml
```diff
@@ -0,0 +1,28 @@
+kind: ZarfPackageConfig
+metadata:
+  name: uds-core-runtime
+  description: "UDS Core Runtime"
+  url: "https://github.com/defenseunicorns/uds-runtime"
+
+components:
+  - name: uds-runtime
```
rjferguson21, line 8: Could/Should we be importing this from the published runtime package?

mjnagel: Unfortunately that's only possible with skeleton I believe - and I'm not sure runtime is publishing/should be publishing a skeleton package? I think that pattern is generally being discouraged?

UncleGedd: I think you can only import remote components as skeleton packages, which we don't currently publish

UncleGedd: I should refresh more lol. Just deleted my comment saying ^^. I was also under the impression 💀 packages are discouraged

src/runtime/zarf.yaml
UncleGedd: Made an issue to address: https://github.com/defenseunicorns/uds-runtime/issues/355

rjferguson21 (approved):

PR #804, opened 2024-09-23
feat: add default deny pre-authservice/keycloak

### LINKED ISSUE #519: Evaluate resource ordering to ensure authn/authz cannot  be circumvented in failure modes
Blake pointed out that as we add authn/authz functionality within Istio via authservice we need to be mindful of how unexpected failures and in general resource ordering could potentially leave applications exposed.

A straightforward way we might be able to somewhat alleviate this would be applying authorization policies (keycloak + authservice functionality) before the Istio resource generation.  This would likely be an improvement for first reconciliation but we'd have to probably be more creative for cases where the `VirtualService` already existed.

"There is a potential security issue where this can fail open:

1. Create a Package without groups, and wait for UDS operator to create the VirtualService
2. Update the Package to have authservice (and optionally groups).
   * If the operator has an error during processing of either of these SSO functions, these functions will throw an error and halt the reconcile... but the old VirtualService will stay around. This means that it fails open.

(similarly you could go from Package+authservice → Package+authservice+groups and get an error in group application and get similar bad fail-open)

I'm unsure what the best solution is. Should we maybe catch errors in the authservice code and tear down the virtual services? That would mean the operator fails closed? Perhaps move `istioResources()` below this code and pass it a bool `destroy` option that triggers teardown (this allows better code reuse)?"

_Originally posted by @bburky in https://github.com/defenseunicorns/uds-core/pull/201#discussion_r1657844035_

rjferguson21:
## Description
Apply default deny AuthorizationPolicy for authservice enabled packages pre-authservice/keycloak

## Related Issue
Fixes https://github.com/defenseunicorns/uds-core/issues/519

## Type of change

## Checklist before merging


src/pepr/operator/controllers/utils.ts
```diff
@@ -68,3 +69,36 @@ export async function purgeOrphans<T extends GenericClass>(
...
+  log.debug(`Deleting children of ${pkgName} in ${namespace} with generation ${generation}`);
+  const resources = await K8s(kind).InNamespace(namespace).WithLabel("uds/package", pkgName).Get();
+
+  for (const resource of resources.items) {
+    if (resource.metadata?.labels?.["uds/generation"] == generation) {
+      log.debug(resource, `Deleting child ${resource.kind!} ${resource.metadata!.name}`);
+      await K8s(kind).Delete(resource);
+    }
+  }
+}
```
mjnagel: I think this is no longer used? I'm _more_ in favor of the current path so I think we can drop this delete func.

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -66,8 +68,15 @@ export async function packageReconciler(pkg: UDSPackage) {
     await enableInjection(pkg);
 
     // Configure SSO
+    const authserviceClients = getAuthserviceClients(pkg);
+    if (authserviceClients.length > 0) {
+      // for each authservice client, create a default deny policy
+      for (const sso of authserviceClients) {
+        await preApplyDefaultDeny(sso.enableAuthserviceSelector!, pkg, sso.clientId);
+      }
+    }
```
mjnagel: I'm good with this as is. I'm still not really sure I see this being common enough of an ask that we should make it the default behavior but I'm happy to see what shakes out with the user experience around this.

I do wonder if we should document this expectation for users though - if authservice/keycloak things failed they should expect to be locked out of their app until they resolve issues with it?

rjferguson21:
Closing for now, see https://github.com/defenseunicorns/uds-core/issues/519

PR #806, opened 2024-09-23
feat: add support for keycloak saml attributes

### LINKED ISSUE #805: Add support for Keycloak SAML attributes
In order to support NiFi SSO via SAML there are several `Fine Grain SAML Endpoint Configuration` options that need to be set in addition to `Assertion Consumer Service POST Binding URL`

* `Assertion Consumer Service Redirect Binding URL`
* `Logout Service POST Binding URL`
* `Logout Service Redirect Binding URL`

- **Given** a UDS package with a SAML dependency
- **When** a SSO block is defined with attributes as specified above
- **Then** the SAML client configuration includes the values as configured

Can use OIDC but given the preference to use SAML where possible these attributes are required

### Additional context
NiFi requires the use of SSO to support user auth, otherwise client certificate auth is required.

nfoucha:
## Description

Adds SAML fine grained attributes:
`saml_assertion_consumer_url_redirect`
`saml_single_logout_service_url_post`
`saml_single_logout_service_url_redirect`

## Related Issue

Fixes #805 

## Type of change

## Checklist before merging


chance-coleman:
Unverified commit

@nfoucha looks like you had an unverified commit that will need to be resolved before we can merge this in. Definitely think this PR is valid and ready once you get that commit fixed.

nfoucha:
> @nfoucha looks like you had an unverified commit that will need to be resolved before we can merge this in. Definitely think this PR is valid and ready once you get that commit fixed.

@UnicornChance Guessing that happened when I did the rebase via UI, oops.

chance-coleman (approved):
LGTM, thanks for the PR!

Issue #729, opened 2024-09-05
Run validation tasks against nightly deployments


mjnagel:
Related to https://github.com/defenseunicorns/uds-core/issues/213

Currently our nightly testing simply deploys core on top of a given infrastructure and then removes + destroys it (see [EKS example](https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/test-eks.yaml#L80)). To ensure proper coverage of testing here we should at minimum run our [existing validation tasks](https://github.com/defenseunicorns/uds-core/blob/main/tasks/test.yaml#L18-L26). Note that this may require some additional work to ensure endpoints are reachable by the github runner (potentially via `/etc/hosts` modifications + public load balancers?).

Any expansion to validations/testing is out of scope for this ticket. While some validations may need updates to function in the nightly testing, this ticket is focused on running the existing validations rather than expanding testing (that will be covered in other work).

Definition of done: Validation tasks are run as part of nightly testing (and passing).

mjnagel:
Added context: Current tests run with a combination of uds tasks (basic validations) and jest/playwright (e2e). We would want to account for DNS/properly resolving for both of these.

Helpful ref links for ref-arch testing:
- https://github.com/defenseunicorns/uds-aws-eks-reference-deployment/blob/ec595344f42ca6c3273da3f7532f9bda513a8a28/test/e2e/e2e_test.go#L30
- https://github.com/defenseunicorns/uds-reference-deployments/blob/main/aws/eks/core/tasks/test.yaml

PR #808, opened 2024-09-23
feat: add nightly testing for rke2

### LINKED ISSUE #726: Add nightly CI testing on RKE2
Related to https://github.com/defenseunicorns/uds-core/issues/213

Currently we do nightly testing on EKS. In order to expand our testing posture we should also test on RKE2. The EKS testing is contained in these locations:
- Workflow: https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/nightly-testing.yaml
- Nightly bundle: https://github.com/defenseunicorns/uds-core/tree/main/.github/bundles
- Test infra: https://github.com/defenseunicorns/uds-core/tree/main/.github/test-infra/buckets-iac
- IAC task file: https://github.com/defenseunicorns/uds-core/blob/main/tasks/iac.yaml

Definition of done: Nightly test added to spin up RKE2 and deploy core (all flavors, matrixed) on it. This should use external S3 buckets where required. For this initial iteration we should keep it as simple as possible:
1. Leverage images built by the image builder: https://github.com/defenseunicorns/uds-rke2-image-builder (these are published in the CI AWS account)
2. Leverage basic terraform to spin up the cluster: Since this is ephemeral we should focus on MVP for now. Potentially could reuse some of the terraform from the [image builder](https://github.com/defenseunicorns/uds-rke2-image-builder/tree/main/.github/test-infra/rke2-cluster).
3. Use as few AWS dependencies as possible to start. Example: rather than setting up IRSA, just use key based auth for buckets. These may be future enhancements to the RKE2 test but should not be in scope for this initial setup.

noahpb:
## Description
This pull request introduces a new nightly test of uds-core on RKE2. The pipeline runs in parallel alongside our existing EKS nightly tests. Included in this PR are additional IaC resources to deploy RKE2, an additional bundle for RKE2 and updated task files/workflows.

Details:
- Changes `test-infra` directory layout to support multiple k8s distributions (merges `test-infra/buckets-iac` and `test-infra/rds-iac` into `test-infra/aws/eks` and adds `test-infra/aws/rke2`)
- Adds IaC for a minimal deployment of RKE2 under `test-infra/aws/rke2`
- Introduces new test bundle for uds-core on RKE2 (`.github/bundles/rke2`), changes `.github/bundles` directory layout to support additional distros (moves `.github/bundles/*.yaml` to `.github/bundles/eks`)
- Introduces new workflow for nightly tests of uds-core on rke2 (`.github/workflows/test-rke2.yaml`)
- Updates workflows and task files to support testing across multiple K8s distros.

Callouts:
- uses existing vpc and subnets in ci account
- the `rke2_version` var will be used to query the latest ami for the given minor version
- uses rhel amis built from uds-rke2-image-builder repo
- includes support for irsa
- no cluster-autoscaler deployed, currently hardcoded to provision 4 `m5.2xlarge` nodes
- does not deploy any external dependencies apart from s3 buckets for loki and velero

Example pipeline run with new RKE2 tests can be seen [here](https://github.com/defenseunicorns/uds-core/actions/runs/11150004170/).

## Related Issue

Fixes #726 

## Type of change

## Checklist before merging


tasks/create.yaml
```diff
@@ -16,7 +17,7 @@ tasks:
       - task: pepr-build
 
       - description: "Create the UDS Core Standard Zarf Package"
-        cmd: "uds zarf package create packages/standard --confirm --no-progress --flavor ${FLAVOR}"
+        cmd: "uds zarf package create packages/standard --confirm --output .github/build/${K8S_DISTRO} --no-progress --flavor ${FLAVOR}"
```
mjnagel: I suspect this addition is what is causing CI failures at the moment - our other tasks for k3d CI assume that created packages are under `build/` at the root. Is there a reason these have to be moved to `.github` and placed per distro? If so, we could probably use maru `if` lines in the task here and conditionally add the `--output` flag.

noahpb: Ah good callout. I'll fix that.

noahpb: Added a fix in https://github.com/defenseunicorns/uds-core/pull/808/commits/69255b7060062fd13c84c3faf68e91e57624fa37. LMK if that jives. I figured a default empty value was easier to implement than a conditional.

When refactoring, I forgot to update the `path` to the core zarf package in each bundle, hence hardcoding `.github/build`. Thanks for catching this :)

mjnagel: I think that works - I'm still a bit unclear on why a different build folder is needed for eks vs rke2? The core package layer should be standard between them (and k3d) so couldn't we just continue to point to `../../../build` and not pass in a specific output path? In other words, the only unique part between rke2, eks, and k3d is the bundle, but all should be able to import the same package for core.

noahpb: Ahh, I see. You're 100% right. I'll back those changes out.

mjnagel: Np - just figured we could potentially reduce the added var/flag here, but wasn't sure if I missed something.

.github/workflows/nightly-testing.yaml
```diff
@@ -7,17 +7,19 @@ on:
...
       - tasks/iac.yaml
       - .github/workflows/test-eks.yaml
-      - .github/bundles/*
-      - .github/test-infra/buckets-iac/*
-      - .github/test-infra/rds-iac/*
+      - .github/workflows/test-rke2.yaml
+      - .github/bundles/eks/*
+      - .github/bundles/rke2/*
+      - .github/test-infra/aws/eks/*
+      - .github/test-infra/aws/rke2/*
```
mjnagel, line 14: More of a question/proposal - would it be valuable to move the `on.pull_request` triggers down a layer, into `test-eks` and `test-rke2`, rather than at this top level? If we did that PRs would be able to run just RKE2 or just EKS depending on what specific files were changed, reducing a bit of the load on some PRs if they only modified one or the other.

Not sure if there would be pieces where we might still want to run both based on though?

mjnagel: We can save this for a future todo - I'll get an issue opened for it.

noahpb: Agree with the ask, can tackle this as a quick follow-up PR.

.github/test-infra/aws/rke2/terraform.tfvars
```diff
@@ -0,0 +1,8 @@
+default_user = "ec2-user"
+ssh_key_name = "packer-rke2-rhel-key"
+os_distro    = "rhel"
+# Need to allow in from internet for github runner to connect to node
+allowed_in_cidrs = ["0.0.0.0/0"]
+
+ami_id       = "ami-0699e51c9b070e3d5"
+rke2_version = "1.30"
```
mjnagel: Food for thought - might be good to think through a way to grab the ami id with a lookup to auto-update to the latest AMI for a given RKE2 version. I think `ami-06510b3e0431a5ad3` is the newest one, published a few days ago [here](https://github.com/defenseunicorns/uds-rke2-image-builder/actions/runs/11119553607/job/30894933270#step:7:2957). This current one is from August if I'm reading the list right.

mjnagel: Via the web console I filtered on AMI name containing `uds-rhel-rke2-v1.30` and then was able to sort by creation date. I think we could do similar with TF and use the rke2_version var here as part of the filter.

.github/workflows/test-eks.yaml
```diff
@@ -57,43 +57,43 @@ jobs:
       - name: Setup Tofu
         uses: opentofu/setup-opentofu@12f4debbf681675350b6cd1f0ff8ecfbda62027b # v1.0.4
         with:
-          tofu_version: 1.8.1
+          tofu_version: 1.8.2
           tofu_wrapper: false
 
       - name: Create UDS Core Package
-        run: ZARF_ARCHITECTURE=amd64 uds run -f tasks/create.yaml standard-package --no-progress --set FLAVOR=${{ inputs.flavor }}
+        run: ZARF_ARCHITECTURE=amd64 uds run -f tasks/create.yaml standard-package --no-progress --set FLAVOR=${{ inputs.flavor }} --set K8S_DISTRO=eks
```
mjnagel: Think we dropped the K8S_DISTRO var so could remove here as well.

.github/workflows/test-rke2.yaml
```diff
@@ -0,0 +1,108 @@
...
+      #   run: uds run -f tasks/iac.yaml install-eksctl --no-progress
+
+      - name: Setup Tofu
+        uses: opentofu/setup-opentofu@12f4debbf681675350b6cd1f0ff8ecfbda62027b # v1.0.4
+        with:
+          tofu_version: 1.8.2
+          tofu_wrapper: false
+
+      - name: Create UDS Core Package
+        run: ZARF_ARCHITECTURE=amd64 uds run -f tasks/create.yaml standard-package --no-progress --set FLAVOR=${{ inputs.flavor }} --set K8S_DISTRO=rke2
```
mjnagel: Think we dropped the K8S_DISTRO var so could remove here as well.

.github/workflows/test-rke2.yaml
```diff
@@ -0,0 +1,108 @@
...
+        if: always()
+        run: uds run -f tasks/iac.yaml destroy-iac --no-progress --set K8S_DISTRO=rke2
+        timeout-minutes: 10
+        continue-on-error: true
+
+      # - name: Teardown RKE2 cluster
+      #   if: always()
+      #   run: uds run -f tasks/iac.yaml destroy-cluster --no-progress --set K8S_DISTRO=rke2
+      #   timeout-minutes: 30
+      #   continue-on-error: true
```
mjnagel: might be good to cleanup this commented out code - at first glance I got concerned thinking we were missing a teardown 😅

mjnagel (commented):
Few last comments here, LGTM overall. Tracking a few potential follow-on pieces of work:
- Swap EKS to TF deploy so that we can reuse more and parallelize more IAC for EKS
- Add RDS to RKE2 (probably?) to support HA deployment of Keycloak/Grafana
- [Run validations against all nightly clusters](https://github.com/defenseunicorns/uds-core/issues/729) (may need some rework of metallb depending on how tests are run)
- [Modify PR triggers](https://github.com/defenseunicorns/uds-core/issues/856)

.github/test-infra/aws/rke2/terraform.tfvars
noahpb: Added in https://github.com/defenseunicorns/uds-core/pull/808/commits/062a88b70eb10087620d1ec181bf2e83974e1759

noahpb:
Moving to draft temporarily while I validate irsa functionality.

mjnagel (approved):

PR #812, opened 2024-09-23
fix: eks iac issues, document storage class pre-reqs

### LINKED ISSUE #718: Unable to upgrade Keycloak after Velero Backup / Restore
### Environment
Device and OS: RHEL 8
App version: [v0.24.1]
Kubernetes distro being used: zarf k3s 1.28
Other:

### Steps to reproduce
1. Deploy UDS-core (With velero configured for EnableCSI, in this case I was using aws-ebs-csi-driver)
2. trigger backup `k -n velero exec -it deploy/velero -- velero backup create --from-schedule=velero-udsbackup`
3. stand up another cluster
4. Deploy UDS-core (Ideally you would just install Velero, but I didn't have access to just that zarf package)
5. trigger restore `k -n velero exec -it deploy/velero -- velero restore create --from-backup=<reference> --existing-resource-policy=update --include-namespaces=keycloak`
6. try to install UDS-core to a newer version (or the same)

### Expected result
Successful upgrade of UDS-core on a cluster with velero restored UDS-core components.

### Actual Result
PVCs that are restored have these fields in the spec due to the restoration options:
```yaml
spec:
  dataSource:
    apiGroup: snapshot.storage.k8s.io
    kind: VolumeSnapshot
    name: velero-keycloak-data-12345
  dataSourceRef:
    apiGroup: snapshot.storage.k8s.io
    kind: VolumeSnapshot
    name: velero-keycloak-data-12345
```

When you try to re-install or upgrade to a new version, the Keycloak deployment fails saying that these fields are immutable.

N/A

### Severity/Priority
High - I vote this is pretty high since Velero is supposed to be the disaster recovery mechanism if an entire cluster goes down.  This bug would prevent you from ever doing updates/maintenance on UDS-core

### Additional Context
N/A

mjnagel:
## Description

EBS impose a 1Gi size limitation on restored PVCs. This adds a short note to pre-reqs about checking CSI limitations.

While testing with our EKS IAC I also discovered a few other issues:
- IRSA annotations were not correct
- Config did not properly variablize region
- Config had an unmatched `"` around one of the values
- Gitignore did not exclude terraform/tfstate files that shouldn't be committed

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/718

## Type of change

## Checklist before merging


rjferguson21 (approved):
Looks good.

PR #818, opened 2024-09-25
feat: add checkpoint uds-core slim package

Racer159:
## Description

This adds a ~75% faster way to deploy or reset a full uds-core cluster (theoretically would work for other preloaded things like testing GitLab Runner w/GitLab too).

Normal:
image

Checkpoint:
image

Tradeoffs:

 - Requires `sudo` - not sure of a great way around this without mangling volume permissions for containerd
 - May become unwieldy with more permutations (i.e. with layers work)
 - The cluster would be fully published (so all credentials are reused)

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


Racer159:
[Checkpoint task passed in this PR](https://github.com/defenseunicorns/uds-core/actions/runs/11079493867/job/30788639020) (except for the actual publish task)
image

packages/checkpoint-dev/README.md
```diff
@@ -0,0 +1,6 @@
+# K3d + UDS Core Slim Dev Checkpoint
+
+This is a special modified version of UDS Core that rehydrates K3d + UDS Core Slim Dev from a committed container and volumes.
+
+> [!IMPORTANT]
+> In order to `zarf package create` this package you must have a running UDS k3d cluster with the UDS Core Slim dev installed in it!  This package also requires `sudo` to create and deploy currently - if you see a prompt and it seems stalled it is waiting for password input (hidden by the spinner)
```
catsby: Can we expand this README with more detail on using this package? It's clear I need a core slim dev installed to create the package, but what about using it? Do I need core slim running to deploy it? I assume so for the "reset", but once it's created can I destroy my k3d cluster and just do `zarf package deploy`? 

Also, I typically deploy core (slim or not) via the `uds` CLI, can I use this package with UDS or do I need to use zarf directly?

catsby (commented):
I'm not an approver but the code does look good to me. I would like to see more information on how to use this package though so it's more clear on how/why/when someone would want to use it.

packages/checkpoint-dev/zarf.yaml
```diff
@@ -0,0 +1,83 @@
...
+                --image ghcr.io/defenseunicorns/uds-core/checkpoint:latest ${ZARF_VAR_K3D_EXTRA_ARGS} \
+                ${ZARF_VAR_CLUSTER_NAME}
+            description: "Create the cluster"
+          - description: Keycloak to be Healthy
+            wait:
+              cluster:
+                kind: Pod
+                name: app.kubernetes.io/name=keycloak
+                namespace: keycloak
+                condition: Ready
```
mjnagel, line 95: Is there a reason this particular pod is called out here?

tasks.yaml
```diff
@@ -125,6 +125,14 @@ tasks:
     actions:
       - task: test:compliance-evaluate
 
+  - name: test-slim-dev
+    description: "Run validate for the components contained in the slim dev bundle"
+    actions:
+      - cmd: |
+          uds run -f src/istio/tasks.yaml validate-slim --no-progress
+          uds run -f src/keycloak/tasks.yaml validate --no-progress
+          uds run -f src/pepr/tasks.yaml validate --no-progress
```
mjnagel: nit: this probably belongs in `tasks/test.yaml`, named `slim-dev` and could be called as `test:slim-dev`. Also might make more sense there to include the tasks files rather than `uds run` them?

.github/workflows/checkpoint.yaml
```diff
@@ -0,0 +1,56 @@
...
+          uds run -f tasks/deploy.yaml checkpoint-package --no-progress
+          npm ci
+          uds run test-slim-dev --no-progress
+
+      - name: Debug Output
+        if: always()
+        uses: ./.github/actions/debug-output
+
+      - name: Publish Checkpoint Package
+        run: uds run -f tasks/publish.yaml checkpoint-package --no-progress
```
mjnagel: nits on all of these but I think we could/should call these with the included task names rather than filepaths where possible (i.e. `uds run deploy:checkpoint-package`). Tracking that this is not the current pattern elsewhere but we came across this while working on functional layers and tried to lean into the included task model more. But also happy to be told if that's a bad pattern since you are likely closer to Maru work.

.github/actions/setup/action.yaml
```diff
@@ -14,6 +14,10 @@ inputs:
   chainguardIdentity:
     description: "ID for Chainguard Identity"
     required: true
+  udsCliVersion:
+    description: The uds-cli version to install
+    # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
+    default: 0.16.0
```
mjnagel: No objection to this change, but was there a specific need to pull the version out of the `run` command directly?

packages/checkpoint-dev/zarf.yaml
```diff
@@ -0,0 +1,83 @@
...
+    actions:
+      onDeploy:
+        before:
+          - cmd: |
+              echo "This package requires elevated permissions to create - requesting sudo (if paused enter password)"
+              sudo echo "got sudo! success!"
+          - cmd: k3d cluster delete ${ZARF_VAR_CLUSTER_NAME}
+            description: "Destroy the cluster"
+          - cmd: |
+              sudo rm -rf data
```
mjnagel, line 39: This slightly worries me, since it would be relative to the local path someone runs their deploy from I think - and would be a bit worried that someone could have something else in their `data` folder locally? A few thoughts/options:
- Maybe make the path `uds-checkpoint/data` to be a bit more specific?
- Error if the path exists, rather than deleting?
- "Sniff" the data and see if it matches the expected k3d/k3s data (kubelet, k3s paths)?

Similar to other comment, we can ignore this if we only designed this for creating in CI, but it's nice to run locally even just for validation of this PR.

packages/checkpoint-dev/checkpoint.sh
```diff
@@ -0,0 +1,69 @@
...
+
+echo "Saving image to data/uds-k3d-checkpoint-latest.tar..."
+docker save -o data/uds-k3d-checkpoint-latest.tar "$IMAGE_NAME"
+
+echo "Container image saved to data/uds-k3d-checkpoint-latest.tar"
+
+# Step 6: Create a tarball from the data contents
+echo "Creating a final tarball to include in the package"
+sudo tar --blocking-factor=64 -cpf uds-checkpoint.tar data
+sudo chown -R $(whoami):$(whoami) "$(pwd)/uds-checkpoint.tar"
```
mjnagel: On my mac machine I am not in a group with my username so this chown fails:
```
chown: mjnagel: illegal group name
```

I think the `groups` command is valid on both mac and linux, but not sure if order is guaranteed to be in a helpful order to identify the primary user group. `id` is also an option on both distros. We could also ignore this if we only anticipate it being run/created in CI, but would be nice to be able to run locally even just for validation of this PR.

packages/checkpoint-dev/zarf.yaml
Racer159: This is really to wait for any Pod from the statefulset to be ready (since the Statefulset itself doesn't have a status) - KC is the last thing that will come up healthy and this helps the package mimic the Helm install so that downstream packages can interact with things immediately (without this wait there would be ~30s or so that you'd be waiting for KC to wake up)

Racer159: Most packages wouldn't _really_ need that and pepr should retry (and should be the only thing hitting KC directly off the jump) but I did it to be safe - I can remove it or add a comment if you'd like

tasks.yaml
Racer159: wilco

.github/workflows/checkpoint.yaml
Racer159: :nerd_face: awkshually - there is a bug / enhancement needed to Maru for this to work right. Maru only reads variables in from the root task file (the one it called first) so the FLAVOR and VERSION defaults in publish won't get set if the root isn't the publish yaml :/

.github/actions/setup/action.yaml
Racer159: The main reason was to handle it more easily with renovate (since the version is in the tag and also the filename)

Racer159: (and this change was to make arm64 work since brew doesn't play nice with that in linux land)

packages/checkpoint-dev/zarf.yaml
Racer159: gonna try to move it to a temp dir - will help prevent things from accidentally being committed locally too

packages/checkpoint-dev/checkpoint.sh
Racer159: cool swapping to id shouldn't be too bad

packages/checkpoint-dev/README.md
Racer159: Added some more info on it - you have to use zarf but it can be `uds zarf` - I added that as the stated command in the README and added some notes to answer those q's

packages/checkpoint-dev/zarf.yaml
mjnagel: Comment might be nice - I assumed it was some sort of "readiness" but might not have understood that this is used as it's the last thing to come up (thinking in the future slim-dev will also include authservice, although Keycloak will likely still be the slowest piece).

.github/workflows/checkpoint.yaml
mjnagel: Yeah tracking that issue, we hit the same thing earlier today 😢. I think we could still change the create/test ones here to use included tasks (that was the veiled intent of my "where possible" qualifier). Didn't look carefully though, I was hoping the deploys didn't use version but they do. For flavor we could just pass that in easy enough but that might start to erode the argument here for going this direction.

Okay with staying the course for now.

tasks/deploy.yaml
```diff
@@ -51,7 +51,21 @@ tasks:
       - description: "Deploy the latest UDS Core package release"
         cmd: uds zarf package deploy oci://${TARGET_REPO}/core:${LATEST_VERSION} --confirm --no-progress --components '*'
 
+  - name: latest-slim-bundle-release
+    actions:
+      - description: "Get latest tag version from OCI"
+        cmd: uds zarf tools registry ls ghcr.io/defenseunicorns/packages/uds/bundles/k3d-core-slim-dev | sort -V | tail -1
```
mjnagel: This always returns `latest` for me - I think that's okay but wanted to call it out. If we're okay with just using the `latest` tag, we could drop the need for this cmd.

packages/checkpoint-dev/zarf.yaml
```diff
@@ -0,0 +1,95 @@
...
+                --image ghcr.io/defenseunicorns/uds-core/checkpoint:latest ${ZARF_VAR_K3D_EXTRA_ARGS} \
+                ${ZARF_VAR_CLUSTER_NAME}
+            description: "Create the cluster"
+          - description: Keycloak to be Healthy
+            wait:
+              cluster:
+                kind: Pod
+                name: app.kubernetes.io/name=keycloak
+                namespace: keycloak
+                condition: Ready
```
mjnagel, line 95:
```suggestion
          # This action waits on Keycloak since it is the slowest pod to start after cluster creation. By waiting on it, we guarantee the cluster is healthy and usable after deployment.
          - description: Keycloak to be Healthy
            wait:
              cluster:
                kind: Pod
                name: app.kubernetes.io/name=keycloak
                namespace: keycloak
                condition: Ready
```

Something like this.

packages/checkpoint-dev/checkpoint.sh
```diff
@@ -0,0 +1,75 @@
...
+
+echo "Saving image to ${DATA_DIR}/uds-k3d-checkpoint-latest.tar..."
+docker save -o "${DATA_DIR}/uds-k3d-checkpoint-latest.tar" "$IMAGE_NAME"
+
+echo "Container image saved to ${DATA_DIR}/uds-k3d-checkpoint-latest.tar"
+
+# Step 6: Create a tarball from the data contents
+echo "Creating a final tarball to include in the package"
+sudo tar --blocking-factor=64 -cpf uds-checkpoint.tar -C "$DATA_DIR" .
+sudo chown -R $(id -u):$(id -g) "$(pwd)/uds-checkpoint.tar"
```
mjnagel: This might be a naive question, but I was trying to wrap my head around this a bit more. If this is run in CI to publish the checkpoint... ownership would be by the github runner user/group I assume? Then when I go to deploy from the OCI published checkpoint is some combination of sudo/docker magic making permissions work out?

packages/checkpoint-dev/checkpoint.sh
```diff
@@ -0,0 +1,75 @@
...
+
+    case "$DESTINATION" in
+        "/var/lib/kubelet")
+            echo "Copying $SOURCE to ${DATA_DIR}/kubelet_data/"
+            sudo cp -a "$SOURCE"/. "${DATA_DIR}/kubelet_data/"
+            ;;
+        "/var/lib/rancher/k3s")
+            echo "Copying $SOURCE to ${DATA_DIR}/k3s_data/"
+            sudo cp -a "$SOURCE"/. "${DATA_DIR}/k3s_data/"
+            ;;
```
mjnagel, line 54: During creation I see these errors (which cause the deploy to fail later):
```
     cp: /var/lib/docker/volumes/c0d8ea4ead46f3c6649218be409e19d1cd63bfcc68f32d548a116c7924d7a793/_data/.: No such file or directory
     cp: /var/lib/docker/volumes/822e843b8cf644f9c4c9118671f6014d32ad84a062d690e69b07d5c6fdfcfbe2/_data/.: No such file or directory
```

I think pretty much universally on macOS docker is run inside of a VM, in my case the VM can be accessed with `colima ssh` but docker desktop, rancher desktop, etc would likely have similar issues and ways to access the VM.

I was able to rewrite a portion of this script to use [`docker cp`](https://docs.docker.com/reference/cli/docker/container/cp/) instead and got closer (at least didn't get errors with the volumes). I think this is probably a better, more agnostic option here and simplifies a lot of this logic (no looping through volumes, just copy the two paths we need explicitly). I was hoping it might also remove the need for `sudo` but in my case one of the paths gave some permission errors still until I added sudo. I'm sure there's some efficiency loss here, but since it's create time I think it's worth it to make this work across distros? In my run locally it took less than a minute still to run which still seems decently performant (granted I couldn't get it to run successfully previously so unsure of the real comparison).

Would be curious your thoughts on this - I dropped the script changes into a gist since there were a handful of changes across the entirety of the file: https://gist.github.com/mjnagel/6d681678df83067169c4e652466f704f

I also had to add `--no-xattrs` to the final tar command, I got warnings/errors without this (suspect that's some macOS <> Linux stuff). This got me much closer but I hit some issues with the token:
```
time="2024-10-02T15:19:18Z" level=fatal msg="starting kubernetes: preparing server: bootstrap data already found and encrypted with different token"
```

I tried to tweak the commands around startup (using the k3d `--token` option rather than k3s arg) and validated the token exists after extraction but couldn't figure this one out. Would be curious if you hit the same issue with my modified script and can figure out what's wrong?

packages/checkpoint-dev/checkpoint.sh
Racer159: technically we are just chowning the wrapping tar - we technically don't need to do this but I was wary about leaving files owned by user `0` in the current directory - I can remove it though if that sounds like being too careful.

Racer159: removed this

tasks/deploy.yaml
Racer159: swapped to latest

packages/checkpoint-dev/zarf.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+components:
+  - name: destroy-cluster
+    required: true
+    description: "Optionally destroy the cluster before creating it"
+    actions:
+      onDeploy:
+        before:
+          - cmd: |
+              echo "This package requires elevated permissions to create - requesting sudo (if paused enter password)"
```
mjnagel: nit
```suggestion
              echo "This package requires elevated permissions to deploy - requesting sudo (if paused enter password)"
```

bburky:
Probably ignore all of the following, I tried testing CRIU (`docker checkpoint` or `podman container checkpoint`). It almost works with podman... except for not supporting nested containers). Recording my notes here anyway.

* * *

Did you try `docker checkpoint` which uses CRIU and is somewhat meant for this purpose? ...It is still experimental though and requires `"experimental": "true"` in your `/etc/docker/daemon.json`, and install a CRIU package into your linux distro...
https://docs.docker.com/reference/cli/docker/checkpoint/
https://criu.org/Docker

If you use `--checkpoint-dir` you can save the checkpoint to disk and restore it after recreating the container (possibly on a different machine). There seems to be a bug during restore, but there's a workaround, see below.

```bash
docker rm -f count
sudo rm -rf /tmp/checkpoint

docker run -d --name=count busybox /bin/sh -c 'for i in $(seq 9999999); do echo "$i" && sleep 1; done'
docker checkpoint create --checkpoint-dir=/tmp/checkpoint count checkpoint1
docker rm count

docker create --name count busybox
# Apparently `docker start --checkpoint-dir` is broken, use workaround: https://github.com/moby/moby/issues/37344#issuecomment-450782189
# docker start --checkpoint-dir /tmp/checkpoint --checkpoint checkpoint1 count
sudo mv /tmp/checkpoint/checkpoint1 "/var/lib/docker/containers/$(docker ps -aq --no-trunc --filter name=count)/checkpoints/"
docker start --checkpoint=checkpoint1 count

docker ps
docker logs -f count
```

The biggest downside would be this is near impossible to use with Docker Desktop. A big advantage is the cluster never actually "stops", it's magically paused and resumed elsewhere.

Podman seems to support this too, and seems to be a bit more fully supported. k3d (somewhat) supports Podman too. Unlike docker, Podman's CRIU support includes volumes, and capturing multiple containers at once. It can apparently pack the checkpoint into an OCI image too (useful for publishing to GHCR?)

Except... this whole idea may be useless because don't think CRIU supports checkpointing nested namespaces (which is how k3d works to embed sub containers inside it's parent docker container for the k8s node)
https://github.com/checkpoint-restore/criu/blob/v4.0/criu/include/namespaces.h#L47-L48

```bash
limactl start template://podman-rootful
export DOCKER_HOST=unix://$HOME/.lima/podman-rootful/sock/podman.sock
k3d cluster create

limactl shell podman-rootful sudo podman container checkpoint --export=/tmp/lima/checkpoint.tgz k3d-k3s-default-server-0 k3d-k3s-default-serverlb
# Error:
#   Can't dump nested pid namespace for 4663
```

packages/checkpoint-dev/checkpoint.sh
Racer159: per discussion ignoring macOS for now - we can revisit at a later time - but would like to use this to speed up CI while we think about macOS support

corang (dismissed):

.github/workflows/checkpoint.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+on:
+  pull_request:
+    # milestoned is added here as a workaround for release-please not triggering PR workflows (PRs should be added to a milestone to trigger the workflow).
+    types: [milestoned, opened, reopened, synchronize]
+  # triggered by tag-and-release.yaml
+  workflow_call:
+
+jobs:
+  publish-uds-core:
```
mjnagel: Can we change the name here to accurately reflect what's happening in this job? It's not always publishing, and probably should have checkpoint in here?

.github/workflows/checkpoint.yaml
```diff
@@ -0,0 +1,63 @@
...
+    strategy:
+      matrix:
+        architecture: [amd64, arm64]
+    runs-on: ${{ matrix.architecture == 'arm64' && 'uds-ubuntu-arm64-4-core' || 'uds-ubuntu-big-boy-4-core' }}
```
mjnagel, line 27: Do we have confidence the arm variant of this works? I _think_ all of the issues are macOS/container sockets on mac related but want to confirm we have somewhere tried a deploy of the arm variant (maybe CI is the best way?).

.github/workflows/checkpoint.yaml
```diff
@@ -0,0 +1,63 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+name: Checkpoint UDS Core
+
+on:
+  pull_request:
+    # milestoned is added here as a workaround for release-please not triggering PR workflows (PRs should be added to a milestone to trigger the workflow).
+    types: [milestoned, opened, reopened, synchronize]
```
mjnagel, line 9: Might be good to conditionally run this with filters/file paths similar to the slim-dev-test workflow? Rather than running always.

mjnagel: I'm also hesitant on running this on PR since its not a good indication that the PR is working for checkpoint? The tasks here just deploy the latest slim-dev so there's no clear benefit to running this on PR outside of this immediate PR or workflow/task changes?

In which case two thoughts:
- modify the tasks to run against the active branch OR a release, but then you have to build
- run on workflow/task changes only, so most PRs wouldn't run this

.gitignore
```diff
@@ -21,6 +21,9 @@ extract-terraform.sh
 **/.terraform*
 cluster-config.yaml
 **.tfstate
+
+packages/checkpoint-dev/uds-checkpoint.tar
+
```
mjnagel: nit: probably safe to make this `**.tar` but should be fine either way.

packages/checkpoint-dev/README.md
```diff
@@ -0,0 +1,34 @@
...
+> This package does not currently work on macOS due to some limitations with filesystem permissions.
+
+## Creating this package
+
+In order to create this package you must follow the following:
+
+1. Setup a K3d cluster (named `uds`) containing the contents you'd like to checkpoint
+
+> [!NOTE]
+> The intent for this package is that those contents are the `uds dev stack`, `zarf init` and `uds core slim`
```
mjnagel: nit: 
```suggestion
> The intent for this package is that those contents are the `uds dev stack`, `zarf init` and the `core-slim-dev` package (`core-base` and `core-identity-authorization`).
```

mjnagel (commented):
Few more comments + need an update to the release-please config to ensure the checkpoint zarf.yaml is versioned properly: https://github.com/defenseunicorns/uds-core/blob/gotta-go-fast/release-please-config.json#L14

.github/workflows/checkpoint.yaml
```diff
@@ -0,0 +1,63 @@
...
+  # triggered by tag-and-release.yaml
+  workflow_call:
+
+jobs:
+  publish-uds-core:
+    strategy:
+      matrix:
+        architecture: [amd64, arm64]
+    runs-on: ${{ matrix.architecture == 'arm64' && 'uds-ubuntu-arm64-4-core' || 'uds-ubuntu-big-boy-4-core' }}
+    name: Publish checkpoint
```
mjnagel: Similar name comment here if this stays as a test job as well (and not always publish)

.github/workflows/checkpoint.yaml
Racer159: I think #2 is reasonable - I wouldn't expect it to really fail with a change in the rest of UDS Core

.github/workflows/checkpoint.yaml
Racer159: the arm variant is tested in CI  - if you want I can dig out my Nintendo Switch :sweat_smile:

mjnagel (approved):
LGTM, thanks for the work on this!!! Would be great to revisit the macOS support at some point and look at other places we could checkpoint things as well.

mjnagel (changes requested):
Oops missed this one - need an update to the release-please config to ensure the checkpoint zarf.yaml is versioned properly: https://github.com/defenseunicorns/uds-core/blob/gotta-go-fast/release-please-config.json#L14

mjnagel (dismissed):

mjnagel (approved):

Issue #839, opened 2024-09-27
Pepr watch limit too high


docandrew:
### Environment
Device and OS: Ubuntu 22.04
App version: UDS Core 0.27.2-registry1
Kubernetes distro being used: RKE2 1.30.3

### Steps to reproduce
1. Install a UDS Package with Exemptions
2. Remove the UDS Package w/ `zarf package remove`
3. Redeploy the UDS Package immediately thereafter (less than 5 minutes/300s)

### Expected result

Everything comes up normally and the UDS Operator re-provisions necessary resources for the re-installed UDS Package, and recognizes that Exemptions are in place for the Pod.

### Actual Result

UDS Operator fails to recognize that Exemptions are in place for the re-installed UDS Package, and the Pods are blocked.

### Visual Proof (screenshots, videos, text, etc)

n/a

### Severity/Priority

Low. This can be overridden with the PEPR_LAST_SEEN_LIMIT_SECONDS environment variable in the running Pepr Pod (and restarted). Or the Helm value can be overridden here:

https://github.com/defenseunicorns/uds-core/blob/edc068d38f47ba373877593b854d21c4e4fc39ca/src/pepr/uds-operator-config/values.yaml#L9

docandrew:
PR: https://github.com/defenseunicorns/uds-core/pull/840

PR #840, opened 2024-09-27
chore: updated pepr watch limit to 60s

docandrew:
## Description

Adjust Pepr watch limit to a lower value to help UDS Package delete/reinstalls succeed more often.

## Related Issue

Fixes # https://github.com/defenseunicorns/uds-core/issues/839

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/uds-operator-config/values.yaml
+++ b/src/pepr/uds-operator-config/values.yaml
@@ -6,4 +6,4 @@ operator:
   UDS_LOG_LEVEL: "###ZARF_VAR_UDS_LOG_LEVEL###"
   AUTHSERVICE_REDIS_URI: "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
   # Allow Pepr watch to be configurable to react to dropped connections faster
-  PEPR_LAST_SEEN_LIMIT_SECONDS: "300"
+  PEPR_LAST_SEEN_LIMIT_SECONDS: "60"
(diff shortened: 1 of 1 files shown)


src/pepr/uds-operator-config/values.yaml
```diff
@@ -6,4 +6,4 @@ operator:
   UDS_LOG_LEVEL: "###ZARF_VAR_UDS_LOG_LEVEL###"
   AUTHSERVICE_REDIS_URI: "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
   # Allow Pepr watch to be configurable to react to dropped connections faster
-  PEPR_LAST_SEEN_LIMIT_SECONDS: "300"
+  PEPR_LAST_SEEN_LIMIT_SECONDS: "60"
```
mjnagel, line 9: In testing this actually hasn't yielded much better results. We may want to also include this override:
```suggestion
  PEPR_LAST_SEEN_LIMIT_SECONDS: "60"
  PEPR_RELIST_INTERVAL_SECONDS: "60"
```

This should effectively make pepr poll every 60s with a GET call on the k8s api.

docandrew: Added - thanks!

mjnagel (approved):

rjferguson21 (approved):

PR #844, opened 2024-09-30
chore: add e2e playwright tests for grafana

### LINKED ISSUE #764: E2E Testing for Grafana
E2E testing for Grafana should only be run on the all package test only since it focuses on interactions between components - see https://github.com/defenseunicorns/uds-core/issues/211 for some thoughts on this.

Initial target scope:
- Validate datasource connectivity/availability: Confirm that loki + prometheus are visible under datasources in Grafana and connection is active/working
- Validate dashboards are provisioned/populated (check that the "custom" dashboards we deploy via configmap exist in Grafana)
- Validate SSO login

mjnagel:
## Description

Adds e2e testing structure and specific e2e test for Grafana using playwright. This test:
- Validates existence and successful connection to datasources (Loki and Prometheus)
- Validates two custom dashboards exist and dropdowns populate for ns selection (resources and loki quicksearch)
- Validates SSO login success

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/764

## Type of change

## Checklist before merging


src/keycloak/chart/values.schema.json
```diff
@@ -154,7 +154,7 @@
       "type": "object",
       "properties": {
         "enabled": {
-          "type": "boolean"
+          "type": ["boolean", "string"]
```
mjnagel: Not sure how I feel about this but we need a way to enable this for our CI tests so that we can create a test user, using the admin user. If we make a zarf helm override (as done in this PR) it ends up as a string rather than a boolean ([upstream issue](https://www.github.com/zarf-dev/zarf/issues/2783)). To allow this to template and deploy I adjusted the schema to allow strings, but also adjusted the template to handle strings in a more sane way.

e2e/playwright/playwright.config.ts
```diff
@@ -0,0 +1,40 @@
...
+  reporter: [
+    // Reporter to use. See https://playwright.dev/docs/test-reporters
+    ['html', { outputFolder: `${playwrightDir}/reports`, open: 'never' }]
+  ],
+
+  outputDir: `${playwrightDir}/output`,
+
+  use: {
+    trace: 'retain-on-failure', // save trace for failed tests. See https://playwright.dev/docs/trace-viewer#opening-the-trace
+  },
```
mjnagel, line 29: todo: capture these traces in the github artifacts

src/keycloak/zarf.yaml
```diff
@@ -7,6 +7,9 @@ variables:
   - name: DOMAIN
     description: "Cluster domain"
     default: "uds.dev"
+  - name: INSECURE_ADMIN_PASSWORD_GENERATION
+    description: "Generate an insecure admin password for dev/test"
+    default: "false"
```
mjnagel: todo: test alternative methods here to avoid exposing this with a zarf var

mjnagel: Resolved this by using a port forward and tasking to automate admin user creation.

src/grafana/tasks.yaml
```diff
@@ -26,3 +29,20 @@ tasks:
...
+      - description: "Setup the Doug User for testing"
+        # Self-reference this task file to avoid https://github.com/defenseunicorns/maru-runner/issues/144
+        cmd: uds run -f tasks/test.yaml common-setup:create-doug-user --set KEYCLOAK_GROUP="/UDS Core/Admin" # Adds the test doug user
+      - description: E2E Test for Grafana, optionally set FULL_CORE=true to test integrations with Loki
+        cmd: |
+          # renovate: datasource=docker depName=mcr.microsoft.com/playwright versioning=docker
+          docker run --rm --ipc=host -e FULL_CORE="${FULL_CORE}" --net=host --mount type=bind,source="$(pwd)",target=/app mcr.microsoft.com/playwright:v1.48.1-noble sh -c " \
+            cd app && \
+            npm ci && \
+            npx playwright test grafana.test.ts \
```
rjferguson21, line 46: Curious if you evaluated any other mechanisms to include/exclude tests like ---grep or [projects](https://playwright.dev/docs/test-projects).

rjferguson21 (dismissed):

src/grafana/tasks.yaml
mjnagel: I did not - but I like the project approach. Maybe we can get this initial structure merged and then split up into projects? That might also help with some of the [config overrides](https://github.com/defenseunicorns/uds-core/pull/844/files#diff-9b1502a856cbae10e041e6674e36cea314630139fccab39acef9c2bfb51979bfR9-R10) per project.

chance-coleman (dismissed):

e2e/playwright/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "target": "es2016", /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
```
noahpb: nit: we're using `ES2022` in root of this repo. should we align?

rjferguson21 (approved):

chance-coleman (approved):
lgtm

PR #852, opened 2024-10-02
feat: add service accounts options to sso

### LINKED ISSUE #851: Keycloak support for client_credentials grant via Service Account Roles
We are deploying an API that has a multi-tenancy issue to solve, but the users of the API are automated processes / servers.

- **Given** An automated process needs to obtain a claim to a broad resource pool for an API
- **When** The automated process is triggered via an upload or webhook
- **Then** It hits the API to either access files in S3 or kick off workflows
- **And** I want to limit one automated process to only its resources even though the API has access to all of them

We considered deploying the API to multiple namespaces (one per client) but thought that was hacky.  We looked at using device auth flow but don't have a user to interact with any of the flows.

### Additional context
[Service Account Roles](https://www.keycloak.org/docs/latest/server_admin/#_service_accounts) keycloak docs
[Client Credentials Grant](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4) oauth docs

ldgriswold:
## Description
This enables support for service account roles in keycloak for client credentials type grants
...

## Related Issue

Fixes #851

## Type of change

## Checklist before merging


ldgriswold:
image

From https://www.keycloak.org/docs-api/21.1.1/rest-api/#_clientrepresentation

ldgriswold:
Example config: https://github.com/defenseunicorns-partnerships/wfapi/blob/feature/sso/chart/templates/uds-package.yaml

Resulting keycloak client:

image

ldgriswold:
Pushed a fix due to a trick in the system:
```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: {{ include "wfapi.fullname" . }}
spec:
  sso:
    - name: wfapi-creds
      clientId: wfapi
      standardFlowEnabled: false
      publicClient: true
      serviceAccountsEnabled: true #TODO: Add this to UDS Core
      attributes:
        oauth2.device.authorization.grant.enabled: "true"
```
Was allowing a public client to be created, now pepr catches the error:

image

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -139,10 +139,17 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
         `The client ID "${client.clientId}" must specify redirectUris if standardFlowEnabled is turned on (it is enabled by default)`,
       );
     }
+    // If serviceAccountsEnabled is true, do not allow standard flow
+    if (client.serviceAccountsEnabled && client.standardFlowEnabled) {
+      return req.Deny(
+        `The client ID "${client.clientId}" cannot support standard flow or authservice if using service accounts`,
+      );
+    }
     // If this is a public client ensure that it only sets itself up as an OAuth Device Flow client
```
bburky, line 149: This comment is out of date now

ldgriswold: If you're referring to line 148 (old line 142), I think this should still be true.

It will deny creation of the client if `publicClient` is set to `false` and `serviceAccountsEnabled` is not set to `false`

bburky: I think it's now something like?
~If this is a public client ensure that it only sets itself up as an OAuth Device Flow client or service account (grant=client_credentials)~

It also seems like the current code allows making something both a service account (client credentials grant) and device authorization grant at the same time... but I guess that's fine, those are both public clients

bburky: I'm also still not tracking what the use case for this PRs changes are yet, I'm happy to chat and get some more context about the use case

ldgriswold: Pepr will deny creation of the client if it has `publicClient` set to `true` and `serviceAccountsEnabled` set to `true`.  In the above comment I showed a case where I was trying to set:
```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: {{ include "wfapi.fullname" . }}
spec:
  sso:
    - name: wfapi-creds
      clientId: wfapi
      standardFlowEnabled: false
      publicClient: true
      serviceAccountsEnabled: true #TODO: Add this to UDS Core
      attributes:
        oauth2.device.authorization.grant.enabled: "true"
```

Before I added line 152 this was allowed, but line 152 now denies the creation of a `publicClient` if `serviceAccountsEnabled` is not set to `false`.

This pattern is for machine-to-machine authZ to obtain access to resources via an API with a multi-tenancy problem to solve (The API needs to assess the claims of the other machine without any user interactions).  I'll schedule some time to talk through this idea in more detail.

bburky: Oh, I think I follow now. You're trying to make sure someone doesn't set `publicClient: true` and `serviceAccountsEnabled: true` together.

`serviceAccountsEnabled: true` is always a confidential client, correct? You use a secret to do grant=client_credentials.

Ok, this makes sense then.

ldgriswold: Correct, I think you can create a public client that uses client_credentials grant, but that seems like a really terrible idea so I forbade it.

bburky:
I need to test this, but I think this is PR is fine.

We (@rjferguson21 ?) can do a follow up PR to implement a hypothetical `authz:` or something that is the same as `authserviceServiceSelector:` but only Istio JWT authn/zuthz, no authservice. This field would sit beside these other options.

This PR should work today for **grant=client_credentials auth to a (mission app) workload that does internal JWT validation**. I really don't want to encourage apps to do their own JWT validation and crypto though, so adding `authz:` would let Istio do the JWT auth and more safely implement **grant=client_credentials auth workload with automatic Istio JWT validation**. I would argue Keycloak service accounts primarily make sense for **out-of-cluster machine-to-machine clients connecting to an in-cluster workload**.

Very similarly, we could support istio authz on users using **OAuth Device Flow to connect to an in-cluster workload** with the same `authz:` field (generates the same istio resources as above, just a different client config).

However, if you need to **authenticate in-cluster workloads to each other**, I would really suggest instead using **Kubernetes service account, Istio mTLS and a `principals:` AuthorizationPolicy**. I _think_ the service account name is in the x-forwarded-client-cert header. Alternately, you could use something like Kubernetes projected service account tokens or GitLab CI JWTs. If any of these options are chosen, no changes to the Package CR need to be made, just deploy some AuthorizationPolicy resources from the Helm chart. I don't think Keycloak should be needed for two in-cluster workloads communicating.

Anyways, that's all TODO and can be done as a followup of this PR.

bburky:
I did a bunch of testing today with @ldgriswold , and we realized that you don't get an `aud` audience claim by default, you need a mapper. 

But if you're using a mapper anyway, you can customize the claim. If you pick the same audience as your authservice client, this lets your service accounts access authservice protected apps with no other configuration than defining a custom `protocolMapper`. This implements "grant=client_credentials auth to workload with automatic Istio JWT validation".

I updated the docs to include this and cleaned up some other code

docs/configuration/uds-operator.md
```diff
@@ -207,6 +207,48 @@ spec:
...
+          protocolMapper: "oidc-audience-mapper"
+          config:
+            included.client.audience: "uds-core-httpbin" # Set this to match the app's authservice client id
+            access.token.claim: "true"
+            introspection.token.claim: "true"
+            id.token.claim: "false"
+            lightweight.claim: "false"
+            userinfo.token.claim: "false"
+
+      # Optionally, store the client secret a Kubernetes secret
```
mjnagel:
```suggestion
      # Optionally, store the client secret as a Kubernetes secret
```

I think with our current operator code the secret is actually non-optional, although customizing the output of that secret is optional. Also more generally not following why this would be optional - wouldn't you necessarily want the client secret saved off for use? Might be missing how this client is used though.

docs/configuration/uds-operator.md
```diff
@@ -207,6 +207,48 @@ spec:
...
+      secretName: httpbin-api-client
+      secretTemplate:
+        clientId: "clientField(clientId)"
+        clientSecret: "clientField(secret)"
+```
+Setting `serviceAccountsEnabled: true` requires `standardFlowEnabled: false` and is incompatible with `publicClient: true`.
+
+If needed, multiple clients can be added to the same application: an authservice client, an device flow client and as many service account clients as required.
```
mjnagel:
```suggestion
If needed, multiple clients can be added to the same application: an AuthService client, a device flow client, and as many service account clients as required.
```

docs/configuration/uds-operator.md
```diff
@@ -207,6 +207,48 @@ spec:
...
+
+      # By default, Keycloak will not set the audience `aud` claim for service account access token JWTs.
+      # You can optionally add a protocolMapper to set the audience.
+      # If you map the audience to the same client used for authservice, you can enable access to authservice protected apps with a service account JWT.
+      protocolMappers:
+        - name: audience
+          protocol: "openid-connect"
+          protocolMapper: "oidc-audience-mapper"
+          config:
+            included.client.audience: "uds-core-httpbin" # Set this to match the app's authservice client id
```
mjnagel, line 235: This is the one part of this setup that slightly worries me and I want to make sure I understand. By adding this protocol mapper to a `serviceAccountsEnabled` client, someone can effectively set the audience here to any client. Our current istio/authservice protections would not identify JWTs from this client as being any different from the "real" authservice client since we effectively just check the issuer and audience (which would be 1:1 with the authservice client). I think that is the intent here of course, effectively enabling machine auth on authservice protected apps.

Are there any concerns with this though, especially thinking about multi-tenant clusters? Tenant A could have an authservice protected app, and Tenant B could use a `serviceAccountsEnabled` client to access that app (I think?). Are there things we could/should recommend that the application do to prevent this? Or modifications we could make to our Istio resources to conditionally allow this type of behavior, only if the app is expecting machine auth?

I may also just be missing things here but want to make sure I understand the intent with this flow and any risks/mitigations that are or could be present.

docs/configuration/uds-operator.md
bburky: We have made it required that this client will be `publicClient: false` (has a secret), but this comment refers to `secretName` and `secretTemplate`. As far as I can tell `secretName` is optional but will default to `secretName` if unset.

docs/configuration/uds-operator.md
ldgriswold: What you're describing is exactly the use-case.  I want to use the `client_id` (or a group / role etc associated with the client) as an authZ key to solve a multi-tenancy issue.  BUT, all of the different tenants need to use the same API.  I originally had this set up so the API did the JWT verification and would accept multiple "audiences" but we always had a goal of getting authservice to take over this role.

We looked at manually supplying the `AuthorizationPolicy` and `RequestAuthentication` CRs in the chart for my project to allow multiple audiences, but figured it's probably safer to let the operator handle this.

I think this is the most intuitive way:  The API is the "audience", which makes sense, but the machines that access the API are doing so on behalf of different companies and development teams, hence the different clients that have different permissions.

docs/configuration/uds-operator.md
ldgriswold: It would make sense to me if it wasn't optional.  One use case where it might not matter as much is if the client is outside the cluster.  Then you can find a means to pass them the client secret and either grab it from a kubernetes secret or as the keycloak admin.

docs/configuration/uds-operator.md
bburky: > Are there any concerns with this though, especially thinking about multi-tenant clusters?

We already don't do any real multi tenant isolation at configuration time, the person deploying a Package CR is basically assumed to be cluster admin I think? Or is that the place of Exemptions, only those are privileged configuration? We already don't isolate Keycloak clients: it turns out two Package CRs can reference the same Keycloak client, and pepr will fetch it out of the store. You can also do other things to hurt multi-tenant isolation, like add any NetworkPolicy you want.

We could consider each custom protocolMappers to be an exemption and require documentation? This is really why I was recommending an allowlist from the beginning for this feature, with the right config you can make your client do anything (probably even takeover Keycloak by overriding claims).

(I really would like to sketch out a threat model and what implied permissions are associated with each CR)

I'm doing a little bit more testing, but as far as I can tell a Keycloak service account JWT can be distinguished by a username prefix of `service-account-` and a new claim called `client_id`. I also think this audience mapper should work with Device Flow, is that maybe worth documenting too?

ldgriswold: Also, I think this isn't intuitive, but a `serviceAccount` type client does not automatically come with an `audience` field in keycloak.  Since a lot of people use `aud` to do validation we needed to test a code pattern to add the field.

docs/configuration/uds-operator.md
mjnagel: Yeah I guess to clarify my comment on the comment 😄: The kubernetes secret will always be created unless the client is a public client. So regardless of whether secret name and template are specified here, a k8s secret will be created - it will just end up with a default name and format if you don't provide these (ref code [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/client-sync.ts#L186-L205)). Not to get hung up on one line here in a docs comment but "store the client secret as a Kubernetes secret" is not really optional/conditional on setting these fields was my main point. Comment might be better stated as "optionally set the name and template for the generated kubernetes secret"?

docs/configuration/uds-operator.md
mjnagel: Yeah sorry @ldgriswold was tracking that the intent of these clients was exactly as described - authenticating using one client against a different authservice app (one that is protected by a different client). My concern was more around potential for two app teams to be working in the same cluster and one to maliciously gain access to the others application using a service account client. Talked async with @bburky about this a bit and I think it's probably not worth addressing at this point since there are other multi-tenant considerations like this that we haven't addressed and don't have a great story around how we want this to look. For now I think we can make the assumption that `Package` CR resources are likely managed by a cluster administrator and/or created via "approved" UDS packages that have been vetted. There might be future things to evaluate with this like restrictions around the audience protocol mapper or service account clients to better ensure they are "cluster admin" functions.

I do think it may be worth adding a quick note around what @bburky commented above...
> a Keycloak service account JWT can be distinguished by a username prefix of `service-account-` and a new claim called `client_id`

Not sure we need to document the device flow + service account piece at this point though.

docs/configuration/uds-operator.md
```diff
@@ -207,6 +207,48 @@ spec:
...
+            introspection.token.claim: "true"
+            id.token.claim: "false"
+            lightweight.claim: "false"
+            userinfo.token.claim: "false"
+
+      # Optionally, store the client secret as a Kubernetes secret
+      secretName: httpbin-api-client
+      secretTemplate:
+        clientId: "clientField(clientId)"
+        clientSecret: "clientField(secret)"
```
mjnagel: Follow up to the previous thread and suggestion: https://github.com/defenseunicorns/uds-core/pull/852#discussion_r1792162389

Two paths I think we could take here:
1. Just delete this since we capture the SSO secret templating separately [here](https://github.com/defenseunicorns/uds-core/blob/main/docs/configuration/uds-operator.md#example-uds-package-cr-with-sso-templating). I don't _think_ we need to call it out here again?
2. Update the comment to be a bit more accurate:
```suggestion

      # Optionally, provide a custom secret name and template for the secret contents
      secretName: httpbin-api-client
      secretTemplate:
        clientId: "clientField(clientId)"
        clientSecret: "clientField(secret)"
```

ldgriswold: If it's captured elsewhere it may make more sense to delete it so we're not crossing the streams too much?

docs/configuration/uds-operator.md
```diff
@@ -207,6 +207,42 @@ spec:
...
+            access.token.claim: "true"
+            introspection.token.claim: "true"
+            id.token.claim: "false"
+            lightweight.claim: "false"
+            userinfo.token.claim: "false"
+```
+Setting `serviceAccountsEnabled: true` requires `standardFlowEnabled: false` and is incompatible with `publicClient: true`.
+
+If needed, multiple clients can be added to the same application: an AuthService client, a device flow client, and as many service account clients as required.
+
```
ldgriswold, line 245:
```suggestion

A keycloak service account JWT can be distinguished by a username prefix of `service-account-` and a new claim called `client_id`.  Note that the `aud` field is not set by default, hence the mapper in the example.
```

docs/configuration/uds-operator.md
```diff
@@ -207,6 +207,43 @@ spec:
...
+A keycloak service account JWT can be distinguished by a username prefix of `service-account-` and a new claim called `client_id`.  Note that the `aud` field is not set by default, hence the mapper in the example.
```
mjnagel, line 246:
```suggestion
A keycloak service account JWT can be distinguished by a username prefix of `service-account-` and a new claim called `client_id`.  Note that the `aud` field is not set by default, hence the mapper in the example.

```

mjnagel (approved):

chance-coleman (approved):

PR #877, opened 2024-10-08
fix: snapshot ci version modification and tasks for publish

mjnagel:
## Description

Snapshot CI has been failing due to our changes to the slim-dev package/bundle: https://github.com/defenseunicorns/uds-core/actions/workflows/snapshot-release.yaml. This fixes our yq commands to update the new layer packages and bundle locations as expected.

This same change was added on the pepr branch and validated: https://github.com/defenseunicorns/uds-core/actions/runs/11244510863

## Related Issue

N/A

## Type of change

## Checklist before merging


tasks/create.yaml
```diff
@@ -29,6 +29,16 @@ tasks:
...
+      - description: "Create base package"
+        task: single-layer
+        with:
+          layer: base
+
+      - description: "Create identity-authorization package"
+        task: single-layer
+        with:
+          layer: identity-authorization
+
```
mjnagel, line 43: Moved this here rather than keeping the package separate since we have no use-case for the "slim-dev" package to be created on its own outside of the bundle, and if someone wanted to do that they could create the individual layers on their own anyways.

noahpb (approved):

PR #884, opened 2024-10-09
fix: support for anywhere network policies in cilium

### LINKED ISSUE #871: IP based netpols can't select in cluster resources with Cilium CNI
### Environment
Device and OS: RHEL 8
App version: 0.28.0-registry1
Kubernetes distro being used: eks-d
Other: Using Cilium for CNI

### Steps to reproduce
1. Deploy uds core to cluster using Cilium for CNI
2. Attempt to log into Grafana using SSO

### Expected result
User is successfully logged into Grafana using keycloak

### Actual Result
Grafana throws error trying to get a token from Keycloak

`logger=authn.service t=2024-10-07T17:41:18.145089994Z level=error msg="Failed to authenticate request" client=auth.client.generic_oauth error="[auth.oauth.token.exchange] failed to exchange code to token: Post \"http://keycloak-http.keycloak.svc.cluster.local:8080/realms/uds/protocol/openid-connect/token\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"`

### Severity/Priority
High

### Cause of the issue identified
Cilium can not select in cluster resources using CIDR based selectors in netpols. Currently the `allow-keycloak-ingress-keycloak-backchannel-access` netpol is using a cidr based policy to match 0.0.0.0/0, which means Grafana (and any other apps that attempt to access keycloak directly via its internal service hostname) will timeout when running in an environment using Cilium for CNI.

This should be updated to include a namespace selector that matches all namespaces instead of or in addition to the 0.0.0.0/0 policy. Have confirmed that adding `- namespaceSelector: {}` to the from selector resolves the issue.

### Additional context
This issue will affect anything relying on network policies that use an IP selector to allow ingress or egress to in cluster resources, not just keycloak backchannel access. So there may be other netpols that should be updated in the same way. The issue was just identified with Grafana SSO.

mjnagel:
## Description

Reference [this doc](https://github.com/cilium/cilium/blob/v1.16.2/Documentation/network/kubernetes/policy.rst#networkpolicy) for the limitations of Cilium with `ipBlock` based netpols.

Two changes included to support this behavior:
- Modifies the keycloak backchannel policy to include all namespaces instead of using the `Anywhere` generated target. This was intended to be anywhere in cluster anyways (see the deleted TODO comment in the diff).
- Modifies `Anywhere` target to include both the `0.0.0.0/0` CIDR and an empty namespace selector. For any non-Cilium CNIs `0.0.0.0/0` would've already covered any in-cluster endpoints, so this only changes the behavior for Cilium.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/871

## Type of change

## Checklist before merging


MxNxPx (approved):
thanks for the speedy fix @mjnagel !!!

chance-coleman (approved):
lgtm

PR #895, opened 2024-10-10
chore(docs): update doc structure for site refresh

pauldipietro:
## Description

This PR changes the docs structure to reflect the updated site architecture.

- ADRs moved to a separate folder
- All docs are served out of `docs/reference` (or possibly `docs/troubleshooting` if added)
- Remove unnecessary front matter as needed

## Related Issue

N/A

## Type of change

## Checklist before merging


docs/reference/configuration/ingress.md
```diff
@@ -38,7 +36,7 @@ packages:
 
 By default, the UDS Core Istio Gateways are set up to use the `uds.dev` domain and have a valid TLS certificate packaged.  You will want to change the domain name for your environment and provide a valid TLS certificate for this domain.
 
-You can set the TLS certs via overrides in a [UDS Bundle](https://uds.defenseunicorns.com/bundles/) (see below). UDS Core Istio Gateways default to only supporting TLS v1.3, but this can also be overridden per gateway if clients use TLS 1.2 (as seen in the tenant gateway example `value` below).
+You can set the TLS certs via overrides in a [UDS Bundle](https://uds.defenseunicorns.com/bundles/) (see below).
```
mjnagel: Can we re-add this comment and the example in the block below? Unless there was a reason for deleting it?

docs/reference/UDS Core/distribution-support.md
```diff
@@ -12,8 +10,8 @@ UDS Core is a versatile software baseline designed to operate effectively across
...
-| ------------------- | ---------------------- | --------------------------------------------------------------------------------------------------------- |
-| K3d/K3s, Amazon EKS | Tested                 | Supported Kubernetes distributions undergoing testing in CI environments.                                 |
-| RKE2                | Tested                 | Supported Kubernetes distribution tested in production environments other than CI.                        |
-| Other               | Untested/Unknown state | Compatible Kubernetes distributions that are not explicitly tested, documented, or supported by UDS Core. |
+| Distribution    | Category               | Support Level                                                                                             |
+| --------------- | ---------------------- | --------------------------------------------------------------------------------------------------------- |
+| K3d, Amazon EKS | Tested                 | Supported Kubernetes distributions undergoing testing in CI environments.                                 |
```
mjnagel: Was k3s intentionally removed from here? It's probably okay if it was, but technically k3s is the distro, k3d is the way we run it.

docs/application-baseline.md
```diff
@@ -1,26 +0,0 @@
----
```
mjnagel, line 1: Was this doc relocated somewhere (docs repo)? Since this is specific to the contents of core I would think we want to keep it here, but might just be missing why it was removed/moved.

mjnagel (commented):
Maybe a more generic structure comment now that I'm reviewing the actual diffs - the deployment and config sections are all specific to uds-core (not general "uds things"). Does it still make sense for them to be under top level "configuration" and "deployment"? Or should they be subsections under core itself?

docs/reference/configuration/ingress.md
pauldipietro: I don't remember removing this but I can add it back.

docs/reference/UDS Core/distribution-support.md
pauldipietro: Same as above.

docs/application-baseline.md
pauldipietro: Most of the content was moved here: https://uds.defenseunicorns.com/overview/uds-structure/#uds-core-capabilities. Some of the other non-table stuff might not exist so perhaps I just put this back. My bad if it's not anywhere else.

docs/reference/configuration/ingress.md
```diff
@@ -63,9 +61,6 @@ packages:
               path: tls.key
       istio-tenant-gateway:
         uds-istio-config:
-          values:
-            - path: tls.supportTLSV1_2
-              value: true # Add support for TLS 1.2 on this gateway, can be specified via variables if needed at deploy time
```
mjnagel, line 68: re-add as well for example I think?

chance-coleman (dismissed):
lgtm

noahpb (approved):

chance-coleman (approved):

PR #896, opened 2024-10-10
chore: handle upgrade path for functional layers, add doc for usage

### LINKED ISSUE #868: Provide an example `uds-bundle.yaml` that uses functional layers
With the introduction of functional layers, we should provide an example configuration with best practices for end users to reference. A starting point can be found at `bundles/k3d-slim-dev/uds-bundle.yaml`. A proper example should include:
 - the base layer at a minimum (required for all layers)
 - references to the layers at their published location
 - indicators that all layers should be versioned lockstep with one another
 - example overrides

The example configuration should not demonstrate all variations of which layers can be referenced in a bundle, rather, it should serve as a starting point for consumers looking to use functional layers.

### LINKED ISSUE #900: bad link in uds-deploy documentation
### Environment
Device and OS: x86 Windows
App version: v0.28.0
Kubernetes distro being used: N/A
Other: 

### Steps to reproduce
1. on page: https://github.com/defenseunicorns/uds-core/blob/main/docs/deployment/uds-deploy.md
2. click "bundles" (on line 24 in the .md)

### Expected result

Not 404.

### Actual Result

404

404
Page not found. Check the URL or try using the search bar.
### Additional Context
https://github.com/defenseunicorns/uds-core/blob/main/docs/deployment/uds-deploy.md
references: https://uds.defenseunicorns.com/bundles/

mjnagel:
## Description

Changes included:
1. Functional layers upgrade was not successful when I tested initially. There were ownership issues with the `manifests` we use for istio. To resolve that problem I moved the manifests into a chart and added an action to update the ownership before the upgrade. I also removed the pepr action that we have had for several releases since it was needed as a one-time upgrade step (similar to this one).
2. Added a lightweight doc on the usage of functional layers with a very brief explanation and warning as well as a full example of a bundle pulling in the layers.
3. A few misc link fixes and other things to follow the astro docs switch.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/868

Fixes https://github.com/defenseunicorns/uds-core/issues/900

## Type of change

## Checklist before merging


mjnagel:
Tentatively moving this into review, but would benefit from @pauldipietro's thoughts/awareness on this as we are in the middle of moving docs around.

docs/deployment/functional-layers.md
```diff
@@ -0,0 +1,53 @@
+---
+title: UDS Core Functional Layers
+type: docs
+weight: 9
```
chance-coleman: Do we think this doc should come before the [flavors deployment](https://github.com/defenseunicorns/uds-core/blob/main/docs/development/flavor-specific-dev.md?plain=1#L4) doc? Another idea is referencing this doc in the main getting started doc for the [doc site.](https://github.com/defenseunicorns/uds-docs-refresh/blob/main/src/content/docs/getting-started/install-and-deploy-uds.md)

docs/deployment/functional-layers.md
```diff
@@ -0,0 +1,53 @@
...
+---
+
+## Background
+
+Context on the inclusion of "functional layers" can be viewed in our [ADR](https://github.com/defenseunicorns/uds-core/blob/main/docs/adrs/0002-uds-core-functional-layers.md). In short, UDS Core publishes smaller zarf packages that contain subsets of core's capabilities, grouped by their function (such as monitoring, logging, backup/restore, etc) to allow more flexibility in deployment. This helps to support resource constrained environments (edge deployments) and other situations where an environment has different needs than the default core stack.
```
noahpb:
```suggestion
Context on the inclusion of "functional layers" can be viewed in our [ADR](https://github.com/defenseunicorns/uds-core/blob/main/docs/adrs/0002-uds-core-functional-layers.md). In short, UDS Core publishes smaller Zarf packages that contain subsets of core's capabilities, grouped by their function (such as monitoring, logging, backup/restore, etc) to allow more flexibility in deployment. This helps to support resource constrained environments (edge deployments) and other situations where an environment has different needs than the default core stack.
```

docs/deployment/functional-layers.md
mjnagel: I intentionally shoved it lower in the priority order and did not directly link it anywhere else since I don't think we want to generally encourage this path (ref [this comment](https://github.com/defenseunicorns/uds-core/issues/868#issuecomment-2401713123)). This ideally should just be a reference "just in case" + to provide the warning about usage.

Open to thoughts here if there's something minimal we can do to make it more usable without overly "advertising" it.

chance-coleman: Yeah that's a good reminder, forgot about that. Should probably leave that were it is.

chance-coleman (dismissed):

noahpb:
While it's implied in the ADR, is it worth calling out somewhere that we recommend pulling layers at the same version?

mjnagel:
Yeah good callout @noahpb - will update when I rebase this.

noahpb (approved):

chance-coleman (approved):

PR #902, opened 2024-10-11
chore(docs): custom resource docs generation

### LINKED ISSUE #382: Documentation for UDS Package CR / Exemption CR

Currently there is some documentation surrounding the UDS Package/Exemption CRs but it is hidden in the [src/pepr/operator folder](https://github.com/defenseunicorns/uds-core/tree/main/src/pepr/operator) making it not highly visible to end users. It is also currently missing documentation for `monitor`, which exists in a separate [doc](https://github.com/defenseunicorns/uds-core/blob/main/docs/MONITOR.md). UDS exemptions are also briefly documented [here](https://github.com/defenseunicorns/uds-core/blob/main/docs/CONFIGURE_POLICY_EXEMPTIONS.md) with other documentation under the pepr folder.

It would be good to consistently document the operator/CRs under the `docs/` folder. This will especially be helpful as we intend to surface documentation in a docs site _soon_.

An ideal state in my head:
- high level "operator" doc in `docs/` that provides some overview of what the operator is and how it is used
- individual docs for pieces of the spec (`expose`, `monitor`, `allow`, `sso`) with more in depth explanations of options/architecture
- other operator docs as necessary (nerd notes?)

### Additional context

There are probably plenty more docs considerations to help with clarifying things, but the operator seems like a minimal effort high value place to start.

chance-coleman:
## Description
Create some custom docs generation for custom resources. Dynamic to work in the future if new custom resources are created. Also tied into running with the `gen-crds` task that is part of the src/pepr tasks.

## Related Issue

Fixes #382

## Type of change

## Checklist before merging


src/pepr/docs-gen/main.ts
```diff
@@ -0,0 +1,207 @@
...
+  field: string,
+  currentDepth: number,
+): { type: string; markdown: string } {
+  let type = "";
+  let markdown = "";
+
+  if (schema.items?.enum) {
+    type = `Policy[] (enum):<ul>${schema.items.enum.map((value: string) => `<li><code>${value}</code></li>`).join("")}</ul>`;
+  } else if (schema.items?.properties) {
+    const capitalizedField = capitalizeFirstLetter(field);
```
mjnagel:
```suggestion
  const capitalizedField = capitalizeFirstLetter(field);
  if (schema.items?.enum) {
    type = `${capitalizedField}[] (enum):<ul>${schema.items.enum.map((value: string) => `<li><code>${value}</code></li>`).join("")}</ul>`;
  } else if (schema.items?.properties) {
```

Might suggest this to future proof the logic here, it will update from `Policy` to `Policies` but otherwise does the same thing.

mjnagel (commented):
LGTM otherwise!

mjnagel (dismissed):

mjnagel (approved):

Issue #856, opened 2024-10-03
Separate our PR triggers for RKE2/EKS


mjnagel:
_Originally posted by @mjnagel in https://github.com/defenseunicorns/uds-core/pull/808#discussion_r1786666773_

As of the linked PR we are adding RKE2 as an additional nightly test. In order to make PR CI runs more targeted and not run as many jobs when not necessary we should consider moving the [PR triggers](https://github.com/defenseunicorns/uds-core/blob/5d586eac03c49319a2aa522770ecc28558b6823b/.github/workflows/nightly-testing.yaml#L6-L14) from the top level nightly testing into the individual test-eks and test-rke2 workflow files. There may still be some file changes that affect both which we could duplicate in each file.

mjnagel:
@noahpb good to close?

Issue #727, opened 2024-09-05
Add nightly CI testing on Azure (AKS)


MxNxPx:
Related to https://github.com/defenseunicorns/uds-core/issues/213

Currently we do nightly testing on EKS. In order to expand our testing posture we should also test on AKS in Azure. The EKS testing is contained in these locations:

- Workflow: https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/nightly-testing.yaml
- Nightly bundle: https://github.com/defenseunicorns/uds-core/tree/main/.github/bundles
- Test infra: https://github.com/defenseunicorns/uds-core/tree/main/.github/test-infra/buckets-iac
- IAC task file: https://github.com/defenseunicorns/uds-core/blob/main/tasks/iac.yaml

Definition of done: Nightly test added to spin up AKS and deploy core (all flavors, matrixed) on it. There are multiple pre-reqs for this work including...

- Setup an Azure (Govcloud?) subscription for CI and plumb it up to GitHub
- Reuse and/or template out an AKS OpenTofu (using the internally built example) and tweak for fully automated create/destroy

mjnagel:
Work ongoing to handle organizational setup for Azure account. Existing example of AKS deployment exists but nothing in CI yet.

noahpb:
Validated functionality in a private repo. Working through final steps to get https://github.com/defenseunicorns/uds-core/pull/908 ready for review.

noahpb:
All validations passing, switching to In Review.

mjnagel:
@noahpb good to close?

PR #908, opened 2024-10-15
chore: add nightly testing for AKS

noahpb:
## Description
Adds nightly testing for uds-core on AKS

Callouts:
- Removes `nightly-testing.yaml` in favor of a single workflow for each distribution
- Deploys storage account and containers for velero and loki and configures uds-core to use them
- Deploys postgresql database and configures grafana to use it for HA configuration
- adds `uds-config.tf` file and writes `uds-config.yaml` using terraform `local_sensitive_file` instead of `tf output xyz >> uds-config.yaml` pattern used in the past

## Related Issues
Fixes:
  - https://github.com/defenseunicorns/uds-core/issues/727
  - https://github.com/defenseunicorns/uds-core/issues/856

## Type of change

## Checklist before merging


.github/bundles/aks/uds-bundle.yaml
```diff
@@ -0,0 +1,143 @@
...
+              description: "Primary access Key for the Storage Account"
+              path: "loki.storage_config.azure.account_key"
+            - name: AZURE_LOKI_STORAGE_ACCOUNT_CONTAINER
+              description: "The destination container in the Storage Account where logs will be saved"
+              path: "loki.storage_config.azure.container_name"
+          values:
+            - path: loki
+              value:
+                storage:
+                  type: azure
```
mjnagel: nit: I think this works fine but is confusing at a glance on what value is being overridden. I think this suggestion should work and be a bit more direct on the path?
```suggestion
            - path: loki.storage.type
              value: "azure"
```

.github/bundles/aks/uds-bundle.yaml
```diff
@@ -0,0 +1,143 @@
...
+
+      neuvector:
+        core:
+          values:
+            - path: k3s.enabled
+              value: false
+            - path: containerd.enabled
+              value: true
+            - path: containerd.path
+              value: /run/containerd/containerd.sock
```
mjnagel: The upstream chart [deprecated all of these options](https://github.com/neuvector/neuvector-helm/blob/master/charts/core/values.yaml#L609-L610) - did you find that they were necessary to deploy? If so, maybe the better value to override is `runtimePath` and set to `/run/containerd/containerd.sock`?

.github/workflows/test-aks.yaml
```diff
@@ -0,0 +1,93 @@
...
+        run: uds create .github/bundles/aks --confirm
+
+      - name: Create IAC
+        run: uds run -f tasks/iac.yaml apply-tofu --no-progress --set K8S_DISTRO=aks --set CLOUD=azure
+
+      - name: Deploy Core Bundle
+        env:
+          UDS_CONFIG: .github/bundles/aks/uds-config.yaml
+        run: uds deploy .github/bundles/aks/uds-bundle-uds-core-aks-nightly-*.tar.zst --confirm
+        timeout-minutes: 30
```
mjnagel, line 83: It would be nice after the core deployment (in each cloud/distro) to run our [debug output action](https://github.com/defenseunicorns/uds-core/blob/main/.github/actions/debug-output/action.yaml). That would give a bit more confidence without seeing the cluster that everything is running as expected, and provide a quick debug sanity check if this fails in the future. You should be able to just drop in this snippet, before the remove:

```yaml
      - name: Debug Output
        if: ${{ always() }}
        uses: ./.github/actions/debug-output
```

mjnagel: (noting that this isn't directly connected to the AKS work, but given my lack of visibility into AKS it would be helpful in reviewing this to get a quick scan of everything up and running)

mjnagel (commented):
Overall this looks great - few small comments. May also want to sync on how the other devs can get access to spin up clusters outside of CI.

.github/bundles/aks/uds-bundle.yaml
noahpb: Interesting, I wouldn't have realized this as I copy-pasted this from another example. I'll move towards your suggestion.

noahpb: fixed in https://github.com/defenseunicorns/uds-core/pull/908/commits/d06b443cc464842ab2b01793f8e777a31d7414c6

.github/workflows/test-aks.yaml
noahpb: added in https://github.com/defenseunicorns/uds-core/pull/908/commits/a5af9d3936803dec6d68cf9c74cf7c14ec3c29ae

.github/test-infra/aws/rke2/scripts/get-kubeconfig.sh
```diff
@@ -32,7 +32,9 @@ scp -o StrictHostKeyChecking=no -i key.pem ${node_user}@${bootstrap_ip}:/home/${
 
 # Replace the loopback address with the cluster hostname
 sed -i "s/127.0.0.1/${bootstrap_ip}/g" ./rke2-config > /dev/null
-export KUBECONFIG=$(pwd)/rke2-config
+mkdir -p /home/runner/.kube
+mv ./rke2-config /home/runner/.kube/config
+#export KUBECONFIG=$(pwd)/rke2-config
```
mjnagel, line 37: I think this is fine, just noting that it would no longer work seamlessly locally I think, unless someone exported KUBECONFIG on their own.

noahpb: Good callout - I am doing the same thing here: https://github.com/defenseunicorns/uds-core/blob/78e09d917a4b340eaa136d693f1c301286ebc626/.github/test-infra/azure/aks/uds-config.tf#L32

Considering that there's multiple perquisites to deploying any of this TF locally, we can capture this in a separate document. Opened this [issue](https://github.com/defenseunicorns/uds-core/issues/923) to capture that.

mjnagel: Agreed - thanks for capturing that!

mjnagel (approved):

PR #925, opened 2024-10-17
feat: adds registry1 flavor of uds runtime

UncleGedd:
## Description

adds registry1 flavor of uds runtime


src/runtime/zarf.yaml
```diff
@@ -34,3 +36,29 @@ components:
...
+      cluster:
+        architecture: amd64
+    images:
+      - registry1.dso.mil/ironbank/opensource/defenseunicorns/uds/runtime:v0.6.1
+    charts:
+      - name: uds-runtime
+        namespace: uds-runtime
+        version: "v0.6.1"
+        url: https://github.com/defenseunicorns/uds-runtime.git
+        gitPath: chart
```
mjnagel: Likely missing a values file here to override the image?

src/runtime/zarf.yaml
```diff
@@ -34,3 +36,29 @@ components:
                 name: uds-runtime
                 namespace: uds-runtime
                 condition: "'{.status.phase}'=Ready"
```
mjnagel: Currently missing upstream flavor. It might make sense to abstract out the common pieces (chart, actions) into a common zarf.yaml like [this](https://github.com/defenseunicorns/uds-core/blob/3c5d3416ffb3141f40b83601b94699e532a572f0/src/grafana/common/zarf.yaml) and then import with flavors and the specific values + image overrides like [this](https://github.com/defenseunicorns/uds-core/blob/3c5d3416ffb3141f40b83601b94699e532a572f0/src/grafana/zarf.yaml).

packages/standard/zarf.yaml
```diff
@@ -99,7 +99,7 @@ components:
 
   # UDS Runtime
   - name: uds-runtime
-    required: false
+    required: true
```
mjnagel, line 102: Should be able to also drop from [the bundle](https://github.com/defenseunicorns/uds-core/blob/3c5d3416ffb3141f40b83601b94699e532a572f0/bundles/k3d-standard/uds-bundle.yaml#L45) now.

src/runtime/zarf.yaml
```diff
@@ -15,6 +15,8 @@ variables:
 components:
   - name: uds-runtime
     required: false
```
mjnagel: Could change at this layer too - shouldn't matter but good to keep aligned.

packages/standard/zarf.yaml
UncleGedd: done!

src/runtime/zarf.yaml
UncleGedd: done!

src/runtime/zarf.yaml
UncleGedd: good catch, updated

src/runtime/zarf.yaml
UncleGedd: Added `upstream` but its a dup of `unicorn`, let me know if it should be something different

mjnagel: Yep this is what I'd expect - we can DRY it up in the future with a common zarf.yaml but tracking that unicorn = upstream for this one.

mjnagel (approved):

Issue #823, opened 2024-09-25
E2E Testing for Vector


mjnagel:
E2E testing for Vector should only be run on the all package test only since it focuses on interactions between components - see https://github.com/defenseunicorns/uds-core/issues/211 for some thoughts on this.

Initial thoughts on testing areas:
- Input handling: Ensure that Vector is picking up the expected files (pod + node logs), possibly via the API?
- Output handling: Ensure that Vector is able to ship to the destinations, possibly via the API or a test with Vector connected to MinIO and checking bucket contents?
- Other testing: This is not super well defined but we should evaluate if there are other health endpoints or metrics that would be helpful to look at

[Vector's API](https://vector.dev/docs/reference/api/) is not enabled by default so we may need to enable it in CI at minimum and use it to access some of these data points.

noahpb:
Evaluating the following:
- GraphQL API should be healthy 
- Vector should be collecting node logs
- Vector should be collecting pod logs
- Loki Host sink should be receiving node logs
- Loki Pod sink should be receiving pod logs
- Vector should be able to ship logs to Loki

Noting that I will have to deploy a privileged pod to generate node logs as there are no node logs being generated by the time vector is deployed in CI. I will also have to expose a new zarf var on the vector package to toggle enabling the API.

mjnagel:
> Noting that I will have to deploy a privileged pod to generate node logs as there are no node logs being generated by the time vector is deployed in CI

This is surprising to me - there should be a very short log file for k3d startup?

noahpb:
> > Noting that I will have to deploy a privileged pod to generate node logs as there are no node logs being generated by the time vector is deployed in CI
> 
> This is surprising to me - there should be a very short log file for k3d startup?

There is - about a dozen lines. But from what I found in testing, vector only collects logs that were generated after it starts up. I could be missing something though. Maybe there's a parameter I can add that allows me to search back in time before it was deployed.

noahpb:
Discussed offline - found there is no need to manually generate node logs at this time.

mjnagel:
Should also note - log generation for testing could be a good path if we want to test additional log scrape targets in the future, I've definitely seen similar tests in other projects where a pod will just generate some noise to ensure there is always something happening.

noahpb:
[...]

PR #939, opened 2024-10-18
chore: add prometheus, loki, and vector e2e testing

### LINKED ISSUE #822: E2E Testing for Loki
E2E testing for Loki should only be run on the all package test only since it focuses on interactions between components - see https://github.com/defenseunicorns/uds-core/issues/211 for some thoughts on this.

Initial thoughts on testing areas:
- Log ingestion: Use Loki's API endpoints to confirm that logs have been ingested from promtail/vector
- Log storage: View loki's logs or MinIO endpoints to confirm that logs have been stored to the storage backend
- Query test: Using Loki's apis query for specific logs/labels
- Health/Clustering: Unsure if this is viable to check, but the [Loki services endpoint](https://grafana.com/docs/loki/latest/reference/loki-http-api/#list-running-services) should give us a list of services running for Loki that we could use to confirm read/write/backend pods are running and healthy?

The Loki api reference will likely be helpful here: https://grafana.com/docs/loki/latest/reference/loki-http-api/

### LINKED ISSUE #824: E2E Testing for Prometheus Stack/Monitoring
E2E testing for Prometheus should only be run on the all package test only since it focuses on interactions between components - see https://github.com/defenseunicorns/uds-core/issues/211 for some thoughts on this.

Initial thoughts on testing areas:
- Alertmanager alerts firing: At minimum there is a watchdog alert that should be firing when the connection between prometheus and alertmanager is functional (this may require a UI interaction or setting up a receiver in CI for alerts?)
- Prometheus metrics endpoints healthy: This will touch on other apps being properly configured as well, but we should validate that all metrics targets are healthy in prometheus (likely easy to do via [API](https://prometheus.io/docs/prometheus/latest/querying/api/#targets) call)
- Validate UI accessibility: Both prometheus and alertmanager have UIs that are not exposed via Istio, but can be useful for debugging. We should validate that these are accessible in cluster (via svc address/port-forward)

chance-coleman:
## Description
Add E2E prometheus, loki, and vector testing. 

## Related Issues

Fixes #824 #823 #822 

## Type of change

## Checklist before merging


noahpb (dismissed):
lgtm

rjferguson21 (dismissed):
Can we chat about the approach here.  I'd be curious if we'd be better off just building these type of tests into a node script with a test framework.  

Might be some trade offs with needing to build another image but I think long term it will make some of these tests easier to author.

mjnagel (commented):
Still going through the test contents but 3 higher level comments:
1. Can we rename this PR and update the description to capture the current contents (prometheus and loki tests, and vector indirectly)?
2. Let's go forward with renaming/structuring the filepaths here to have a top level `test` directory and subdirectories for `jest` and `playwright`.
3. @rjferguson21 mentioned this one to me, but it would be good to format all of this code the same way, possibly using the same eslint formatting as we use for the pepr code.

e2e/test/prometheus.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  test('all prometheus targets should be up', async () => {
+    const response = await fetch(`${prometheusProxy.url}/api/v1/targets`);
+    expect(response.status).toBe(200);
+
+    const body = await response.json() as unknown as { data: { activeTargets: any[] } };
+    expect(body.data).toBeDefined();
+    expect(body.data.activeTargets.length).toBeGreaterThan(0);
+
+    // filter out unknown because test runner pod is unknown, probably can filter out specific pod
+    const failedTargets = body.data.activeTargets.filter(target => target.health !== 'up' && target.health !== 'unknown');
```
mjnagel: I think `unknown` refers to the target not having been scraped yet ([ref](https://training.promlabs.com/training/monitoring-and-debugging-prometheus/web-status-pages/target-scrape-status/)). I think this might be okay but would be a bit worried we'd miss something by filtering those out, if something _would fail_ once scraped? Not really sure what to do here, I don't think we strictly want to fail on unknown targets, but if we could throw a warning message with any unknown targets?

src/vector/tasks.yaml
```diff
@@ -11,3 +11,9 @@ tasks:
             name: app.kubernetes.io/name=vector
             namespace: vector
             condition: Ready
+  - name: e2e-test
+    actions:
+      - description: "Run Vector integration tests"
+        cmd: |
+            npm ci && npx jest --testPathPattern vector* || echo "No tests found for vector, skipping..."
```
mjnagel: Should the other stubbed out tests follow this pattern as well?

src/vector/values/values.yaml
```diff
@@ -5,6 +5,10 @@
 role: "Agent"
 
 customConfig:
+  api:
+    enabled: true
+    address: 0.0.0.0:8686
+    playground: false
```
mjnagel: Do we still want this enabled by default? I don't think it is being used by the tests anymore?

tasks/test.yaml
```diff
@@ -70,6 +70,12 @@ tasks:
             npm ci && \
             npx playwright test \
             "
+      - description: "Run E2E Tests"
+      - cmd: |
```
mjnagel:
```suggestion
      - description: "Run E2E Tests"
        cmd: |
```

tasks/test.yaml
```diff
@@ -70,6 +70,12 @@ tasks:
...
+          for package in src/*; do
+            uds run -f ${package}/tasks.yaml e2e-test --no-progress
+          done
```
mjnagel: In the case of grafana (or other playwright tests if we had them) this ends up re-running the tests. The previous lines here will run any playwright tests, is there a reason this doesn't just run `npm ci && npx jest` against all the jest tests under `e2e/test`?

With Grafana I had structured it so that this task would run all playwright tests, with the assumption that everything in core is installed. Then the task specifically in the grafana task file ran assuming only the functional layer of monitoring/dependencies was installed, so it actually tests a few less integrations. Not sure if that will be helpful/apply with any others? (ref for grafana running [first time](https://github.com/defenseunicorns/uds-core/actions/runs/11670065040/job/32493527275?pr=939#step:5:2949) and [second time](https://github.com/defenseunicorns/uds-core/actions/runs/11670065040/job/32493527275?pr=939#step:5:3047) with less tests)

src/vector/tasks.yaml
chance-coleman: We can add this to the other stubbed tests if we make the assumption they will be jest tests.

mjnagel: Ah tracking, maybe we just stub this one out the same as the others for now since we don't have any vector-only tests anymore? Definitely a nit, we could leave this as-is too.

src/vector/values/values.yaml
noahpb: Agree, let's disable it.

e2e/test/prometheus.spec.ts
rjferguson21: Updated to warn if unknown targets are encountered.  I think the only real fix for this would be adding a synthetic delay to wait for all targets to be healthy.

mjnagel (approved):

PR #948, opened 2024-10-21
feat: secret copy #741

### LINKED ISSUE #741: Secrets manipulation using the UDS Operator

Frequently when "wiring" two components together, component A will use credentials or a path, etc. stored in a secret, and another component B needs to reference this secret to know how to connect to A. In Kyverno, policies can be used to mutate resources on create, and one use of this capability is to copy secrets between namespaces. I propose adding a similar capability to UDS Operator to allow for secrets to be copied between namespaces.

1. Helm lookups can be used here - when a chart is applied a manifest with a "blank" secret can be inserted where the contents of the secret are obtained from the `lookup` function as described [here](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function)

2. A custom Pepr capability can be used for this purpose: 

```typescript
/**
 *  Copy secrets between namespaces
 *  To test this capability you run `pepr dev`and then run the following command:
 *  `kubectl apply -f capabilities/hello-pepr.samples.yaml`
 */
export const CopySecrets = new Capability({
  name: "copy-secrets",
  description: "Copy secrets between namespaces w/ Pepr",
  namespaces: ["minio", "cti-updater"],
});

// Use the 'When' function to create a new action, use 'Store' to persist data
const { When } = CopySecrets;

/**
 * ---------------------------------------------------------------------------------------------------
 *                                   Mutate Action (Namespace)                                   *
 * ---------------------------------------------------------------------------------------------------
 *
 * This action removes the label `remove-me` when a Namespace is created.
 * Note we don't need to specify the namespace here, because we've already specified
 * it in the Capability definition above.
 */
When(a.Secret)
  .IsCreated()
  .InNamespace("minio")
  .WithName("minio-creds-secret")
  .Watch(async sec => {
    Log.info("minio-creds secret was created. Copying to cti-updater");

    try {
      await K8s(kind.Secret).Apply({
        metadata: {
          name: "minio-creds-secret",
          namespace: "cti-updater",
        },
        data: sec.data,
      });
    } catch (error) {
      Log.error(error, "Failed to copy minio-creds to cti-updater");
    }
  });
```

However, I think this pattern is common enough to warrant special functionality in the UDS Operator to support it.

### Proposed Solution

I propose two new custom resources and support in the UDS Operator to enable copying of secrets between namespaces.

1. A new key `secrets` in the `Package` CR to support a convenient, high-level syntax for copying of secrets in support of UDS Packages that need to reference secrets in other namespaces for their proper operation.

Basic syntax could look something like this:

```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata: ...
spec:
  sso: ...
  monitor: ...
  network: ...
  secrets:
    - name: minio-creds
      description: Copy MinIO credentials for my app to use for bucket creation
      ...
      from:
        namespace: minio
        name: minio-creds-secret
      to:
        namespace: app-that-uses-minio
        name: minio-creds-secret
```

With time, more sophisticated pattern matching, transformation and application logic can be added. For example, a field `required` that indicates whether UDS Package installation should succeed if the secret to be copied is not present.

2. The recognition of a new `Secret` CR in the uds.dev namespace, to support more general use of the UDS Operator to perform manipulation of secrets at runtime.

One possible schema:

```yaml
apiVersion: uds.dev/v1alpha1
kind: Secret
metadata: ...
spec:
  copy:
    from:
      namespace: minio
      name: minio-creds-secret
    to:
      namespace: app-that-uses-minio 
      name: minio-creds-secret
```

With time, new types of UDS Secret beyond the `copy` constructor could be created to allow for things like run-time generation of TLS certificates, passwords, or other types of material.

For example:

```yaml
apiVersion: uds.dev/v1alpha1
kind: Secret
metadata: ...
spec:
  tls:
    namespace: nginx
    name: nginx-creds
    cert_name: nginx.crt
    key_name: nginx.key
    algo: ed25519
    ...
---
apiVersion: uds.dev/v1alpha1
kind: Secret
metadata: ...
spec:
  password:
    namespace: apache
    name: admin_creds
    length: 24
    char_pattern: "[a-z][A-Z][0-9]"
    ...
```

This is merely an example of how a Secret object can be used more generally to support many other common types of run-time manipulation and generation. 

An alternate design might be separate CRs such as `uds.dev/CopySecret` or `uds.dev/GenerateSecret`  each with their own schema and behavior.

docandrew:
## Description

- Adds workaround for Zarf issue https://github.com/zarf-dev/zarf/issues/2713
- Adds secret copy functionality for issue #741

## Related Issue

Fixes #741

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/operator/index.ts
+++ b/src/pepr/operator/index.ts
@@ -83,9 +83,7 @@ When(UDSPackage)
   .WithName("keycloak")
   .Watch(() => {
     // todo: wait for keycloak and authservice to be running?
-    log.info(
-      "Identity and Authorization layer deployed, operator configured to handle SSO.",
-    );
+    log.info("Identity and Authorization layer deployed, operator configured to handle SSO.");
     UDSConfig.isIdentityDeployed = true;
   });
 When(UDSPackage)
@@ -93,9 +91,7 @@ When(UDSPackage)
   .InNamespace("keycloak")
   .WithName("keycloak")
   .Watch(() => {
-    log.info(
-      "Identity and Authorization layer removed, operator will NOT handle SSO.",
-    );
+    log.info("Identity and Authorization layer removed, operator will NOT handle SSO.");
     UDSConfig.isIdentityDeployed = false;
   });
 
--- a/src/pepr/operator/index.ts
+++ b/src/pepr/operator/index.ts
@@ -27,6 +27,9 @@ import { purgeAuthserviceClients } from "./controllers/keycloak/authservice/auth
 import { exemptValidator } from "./crd/validators/exempt-validator";
 import { packageReconciler } from "./reconcilers/package-reconciler";
 
+// Secret imports
+import { copySecret, labelCopySecret, validateSecret } from "./secrets";
+
 // Export the operator capability for registration in the root pepr.ts
 export { operator } from "./common";
 
@@ -80,7 +83,9 @@ When(UDSPackage)
   .WithName("keycloak")
   .Watch(() => {
     // todo: wait for keycloak and authservice to be running?
-    log.info("Identity and Authorization layer deployed, operator configured to handle SSO.");
+    log.info(
+      "Identity and Authorization layer deployed, operator configured to handle SSO.",
+    );
     UDSConfig.isIdentityDeployed = true;
   });
 When(UDSPackage)
@@ -88,6 +93,15 @@ When(UDSPackage)
   .InNamespace("keycloak")
   .WithName("keycloak")
   .Watch(() => {
-    log.info("Identity and Authorization layer removed, operator will NOT handle SSO.");
+    log.info(
+      "Identity and Authorization layer removed, operator will NOT handle SSO.",
+    );
     UDSConfig.isIdentityDeployed = false;
   });
+
+// Watch for secrets w/ the UDS secret label and copy as necessary
+When(a.Secret)
+  .IsCreatedOrUpdated()
+  .WithLabel(labelCopySecret, "true")
+  .Mutate(request => copySecret(request))
+  .Validate(request => validateSecret(request));
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -26,10 +26,16 @@ tasks:
       - description: "Create the dev cluster"
         task: setup:create-k3d-cluster
 
-      # Note: This currently is broken until https://github.com/zarf-dev/zarf/issues/2713 is resolved
-      # As a workaround you can edit the `src/istio/values/upstream-values.yaml` file to change ###ZARF_REGISTRY### to docker.io before running
+      # Workaround for https://github.com/zarf-dev/zarf/issues/2713
+      - description: "Modify Istio values w/ upstream registry"
+        cmd: "uds zarf tools yq -i '.global.proxy_init.image |= sub(\"###ZARF_REGISTRY###\", \"docker.io\") | .global.proxy.image |= sub(\"###ZARF_REGISTRY###\", \"docker.io\")' src/istio/values/upstream-values.yaml"
+
       - description: "Deploy the Istio source package with Zarf Dev"
         cmd: "uds zarf dev deploy src/istio --flavor upstream --no-progress"
+
+      # Undo previous registry workaround
+      - description: "Restore Istio registry values"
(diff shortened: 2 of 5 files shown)


bburky:
This appears to be unprivileged and effectively allows cluster-wide read of any Secrets if you have RBAC permissions of read/write to a Secret in a single namespace. Also, because the name is customizable, this is a bypass of `resourceNames` RBAC too.

It's reasonable for an app to have runtime read/write secrets in it's own namespace (possibly restricted by `resourceNames`) , but this would allow reading any secret in the cluster.

There's a few different ways you could address this, but I'd suggest driving it via a CR instead of the Secret resources alone. For example, editing Package resources is privileged (things shouldn't be giving away RBAC to this resource kind), which drives most of this operator and solves the permissions issue.

This is isn't a comment on the feature itself, I have no opinion on it.

tasks.yaml
```diff
@@ -26,11 +26,17 @@ tasks:
...
+      - description: "Modify Istio values w/ upstream registry"
+        cmd: "uds zarf tools yq -i '.global.proxy_init.image |= sub(\"###ZARF_REGISTRY###\", \"docker.io\") | .global.proxy.image |= sub(\"###ZARF_REGISTRY###\", \"docker.io\")' src/istio/values/upstream-values.yaml"
+
       - description: "Deploy the Istio source package with Zarf Dev"
         cmd: "uds zarf dev deploy src/istio --flavor upstream --no-progress"
 
+      # Undo previous registry workaround
+      - description: "Restore Istio registry values"
+        cmd: "uds zarf tools yq -i '.global.proxy_init.image |= sub(\"docker.io\", \"###ZARF_REGISTRY###\") | .global.proxy.image |= sub(\"docker.io\", \"###ZARF_REGISTRY###\")' src/istio/values/upstream-values.yaml"
+
```
mjnagel, line 39: Zarf 0.42.1 included [this fix](https://github.com/zarf-dev/zarf/pull/3064)/feature to allow us to pass in `--registry-url` during the `dev deploy` (set to `docker.io`). Once uds-cli consumes that zarf version this should be switched to use that flag instead of yq.

src/pepr/operator/index.ts
```diff
@@ -91,3 +94,10 @@ When(UDSPackage)
     log.info("Identity and Authorization layer removed, operator will NOT handle SSO.");
     UDSConfig.isIdentityDeployed = false;
   });
+
+// Watch for secrets w/ the UDS secret label and copy as necessary
+When(a.Secret)
+  .IsCreatedOrUpdated()
+  .WithLabel(labelCopySecret, "true")
+  .Mutate(request => copySecret(request))
+  .Validate(request => validateSecret(request));
```
mjnagel, line 103: Would be curious to get your thoughts on this design approach of mutating the "destination" secret. Initially I thought this was a mistake because I didn't read far enough in the PR 😅.

My inclination would be that it makes more sense to `Watch` (or `Reconcile`) secrets that have a label like `uds.dev/copyMe` (i.e. the "source" secret). That way you can guarantee you know when the secrets update at the source and ensure they are kept up to date in the destination. This current pattern would only watch changes/creations on the destination side, meaning you might miss changes if the source updates I think? There may be a few more design implementations to think through here, but overall I think it would make a bit more sense? (a few of those design decisions: do you need an `overwrite: true/false` annotation, do you support mutliple destinations and how, etc)

I think this might also resolve most of @bburky's security concerns since the only secret being read would be one that has been explicitly labelled to be copied. The only caveat would be on that `overwrite` piece - we would want to be careful implementing this to not overprivilege that label to write to any secret without some safeguards (maybe we only allow overwriting if pepr is already controlling the secret, use ownership metadata, etc).

The one obvious downside I could see to this is that you would need to have access to label/annotate the source secret which might be an issue with some upstream charts/operators? If that is a major sticking point we could move to a custom resource as @bburky recommended. If we want to go that route I think I'd lean slightly towards a new custom resource rather than using the `Package` CR.

mjnagel (changes requested):
Sorry for the delayed review on this - happy to have further convo on my comments. If you want to think through and propose other options here it might be good to write up a design doc for the team + security to review before we dive into another implementation.

mjnagel:
Going to close this PR for now as its grown stale and has conflicts. I think we should revisit this with a technical design doc to hash out and agree on the overall implementation before we proceed further with implementation.

@docandrew I'll leave the branch around so we don't lose the code here, happy to sync up on the need and starting on a design as well.

PR #953, opened 2024-10-23
feat: add finalizer for UDS Package CRs

### LINKED ISSUE #523: Leverage `finalizer` for `Package` deletions

mjnagel:
## Description

This PR moves our `onDelete` code into a finalizer, using Pepr's `Finalize` functionality. Due to an [open Pepr feature request](https://github.com/defenseunicorns/pepr/issues/1316) this does not update the Package CR status during deletion to avoid triggering the finalization multiple times. In order to provide visibility into deletion status and failure this adds eventing around the specific cleanup operations as well as specific failure events in catch blocks.

This PR does also add a new status to the Package CR of `Removing`, which was stubbed out for a future state where we may be able to use it (pending addition of the Pepr feature to not delete the finalizer).

3 additional related QoL changes:
- await removal of SSO tokens from the store
- modify log message around istio injection to be more accurate based on state
- ensure `addlicense` is run on generated CRD files post-generate

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/523

## Type of change

## Checklist before merging


src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -107,3 +110,51 @@ export async function packageReconciler(pkg: UDSPackage) {
...
+      type: "Normal",
+    });
+    // Cleanup the namespace
+    await cleanupNamespace(pkg);
+  } catch (e) {
+    await writeEvent(pkg, {
+      message: `Restoration of istio injection status failed, ${e.message}`,
+      reason: "RemovalFailed",
+    });
+  }
```
mjnagel: Obviously catching these and just tossing an event is non-ideal but in the current state of `Finalize` we don't have any options to fail the function here without deleting the `finalizer` from the CR ([ref](https://github.com/defenseunicorns/pepr/blob/63aebabb135099d523fe08dc220fab84027e1dc5/src/lib/watch-processor.ts#L113-L115)). This is also still a slight improvement over the previous state since we are able to leave some breadcrumb about the failure, whereas previously the Package CR instantly deleted with no events indicating progress or failure.

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -84,7 +84,7 @@ export async function purgeSSOClients(pkg: UDSPackage, newClients: string[] = []
     const token = Store.getItem(storeKey);
     if (token) {
       await apiCall({ clientId: ref }, "DELETE", token);
-      Store.removeItem(storeKey);
```
rjferguson21, line 87: Why this change? We are already awaited the DELETE to keycloak.  

I think waiting for the remove only makes sense if the potential update (if someone deleted and created the same client?) would result in some sort of conflict?

rjferguson21: This is a bit of a nit: because I think it doesn't matter but I think its ultimately a question of if this works as expected?

```
Store.setItem("foo". "bar")
Store.removeItem("foo")
```

mjnagel: If the token still existed in the store on a subsequent creation we would get it ([ref](https://github.com/defenseunicorns/uds-core/blob/finalizer-add/src/pepr/operator/controllers/keycloak/client-sync.ts#L135)) and try to use it ([ref](https://github.com/defenseunicorns/uds-core/blob/finalizer-add/src/pepr/operator/controllers/keycloak/client-sync.ts#L139)) on the first attempt, and then log an error + retry without the token on a second try. It would mostly just be a bit of clutter in the logs if we didn't wait for the token deletion OR best case we are gambling on the timing between token actually being removed and "new" client being requested.

Imo the whole point of the finalizer and this cleanup is to guarantee that everything connected to the Package is truly gone when it is deleted from the cluster. I'm not sure why we would do that with some pieces but not everything, but I might just be missing your point here with the token?

rjferguson21: After chatting about this you are correct.  I had forgotten about the store being used as an indication of whether or not the client existed.  

As a follow up to our conversation about if `removeItem` optimistically removed items from the local cache.
```
Store.removeItem("foo")
Store.getItem("foo")
```
It seems like you are right there too.  It does not, which means things go like this: 
```
# waits up to 5 seconds to flush/write this removal to kubernetes
Store.removeItem("foo")

# gets item from in memory cache
# in memory cache is updated exclusively from a 5 second debounced watch 
Store.getItem("foo")
```

rjferguson21:
Going to try to figure out exactly why but FYI it seems like if I Create and then Delete many (5 or 6) packages at once the authservice secret does not get cleaned up properly.  I see lingering chains for Packages that no longer exist. 

I checked back on `main` to make sure it was not a regression introduced  by [the authservice batch update PR](https://github.com/defenseunicorns/uds-core/pull/735) and it seemed to work as expected there.

mjnagel:
I was able to reproduce the same issue identified ^, digging into it.

rjferguson21:
Discussed offline with Micah but wanted to follow up here.  Part of this change included moving from [`Store.removeItem` to `Store.removeItemAndWait`](https://github.com/defenseunicorns/uds-core/pull/953/files#diff-d14f3bad79474ad3285f4b7d8802bd449f23dff4e4ac7746d2836b1d2a414defR87) when cleaning up SSO clients.  

This was causing resource contention for the in memory copy of the authservice configuration in circumstances where many packages are deleted simultaneously due to the fact that the underlying Pepr store implementation would resolve the removeItemAndWait calls for every package at the same time. 

I introduced a basic resource lock to avoid the contention on the in the memory update.

noahpb (approved):

PR #979, opened 2024-11-04
chore: regroup renovate support dependencies

mjnagel:
## Description

Regroups a few things to ensure support-deps grabs the right stuff:
- glob up all task files at the root level, with the exception of the create task (reordered pepr to ensure this one remains separate)
- includes `src/test` which contains test manifests/images
- includes `src/**/tasks.yaml` which sometimes contain playwright images or other support deps ([example](https://github.com/defenseunicorns/uds-core/blob/main/src/grafana/tasks.yaml#L42-L43))

## Related Issue

N/A

## Type of change

## Checklist before merging

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -73,14 +73,14 @@
       "commitMessageTopic": "grafana"
     },
     {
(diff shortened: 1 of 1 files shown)


chance-coleman (approved):
lgtm, i added something similar to the loki-e2e PR but this is more better so i'll fix that conflict after this is merged.

PR #982, opened 2024-11-05
feat: set Istio gateway TLS from Kubernetes secret

### LINKED ISSUE #976: Allow Istio Ingress TLS Certificate from Secret
Allow a method to configure Istio to use a TLS certificate from a Kubernetes secret, such as one created by cert-manager. This would allow for delivery engineers to automate the renewal of TLS certificates.

Currently, if TLS is enabled, it comes from the [gateway-tls secret](https://github.com/defenseunicorns/uds-core/blob/main/src/istio/chart/templates/tls-cert.yaml). I would like a way to override this configuration and provide my own secret to use instead.

JeffResc:
## Description
Adds the ability to set Istio gateway TLS from Kubernetes secret

## Related Issue

Fixes #976 

## Type of change

## Checklist before merging


docs/reference/configuration/ingress.md
```diff
@@ -93,3 +93,32 @@ variables:
 :::note
 If you are using Private PKI or self-signed certificates for your tenant certificates it is necessary to additionally configure `UDS_CA_CERT` with additional [trusted certificate authorities](https://uds.defenseunicorns.com/reference/configuration/uds-operator/#trusted-certificate-authority).
 :::
+
+#### Configuring TLS from a Secret
+
+As an alternative to specifying individual certificate, key, and CA certificate values, you can set `tls.credentialName` in the gateway configuration. This field specifies the name of a Kubernetes secret containing the TLS certificate, key, and optional CA certificate for the gateway. When `tls.credentialName` is set, it will override `tls.cert`, `tls.key`, and `tls.cacert` values, simplifying the configuration by allowing a direct reference to a Kubernetes TLS secret.
```
mjnagel: Small nits on the docs:
- might be good to explicitly call out the required fields in the existing secret if using that, or alternatively link to Istio docs on this (i.e. the section from [here](https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings))
- might also be good to explicitly call out the namespaces that these secrets should be in (iirc they need to be in the same namespace as the gateway)

src/istio/chart/templates/tls-cert.yaml
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 {{- $tls := .Values.tls }}
-{{ if $tls.cert }}
+{{ if and $tls.cert (not .Values.tls.credentialName) }}
```
mjnagel:
```suggestion
{{ if and $tls.cert (not $tls.credentialName) }}
```

Definitely a nit, just a suggested change for consistency with values references.

mjnagel (commented):
LGTM overall - two small comments.

mjnagel (approved):

PR #984, opened 2024-11-06
feat: allow workloads to see and process external IP addresses

Racer159:
## Description

This is a pair PR to the fix for GitLab to be able to process IP addresses directly so that Rack Attack can be configured and will work correctly:

https://github.com/defenseunicorns/uds-package-gitlab/pull/227

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging

### DIFF
--- a/src/istio/values/gateway-values.yaml
+++ b/src/istio/values/gateway-values.yaml
@@ -1,2 +1,5 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
 service:
   externalTrafficPolicy: Local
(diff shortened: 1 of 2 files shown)


Racer159:
For more context on this change: https://github.com/istio/istio/issues/7328#issuecomment-435752699

mjnagel (changes requested):
After reviewing upstream docs, etc we're leaning towards closing this for now and not adjusting the defaults for uds-core. There are definite downsides to this change with the loadbalancing behavior, but it is unclear how much of a downside these are at this point. Given that this is only a real "need"/request for GitLab we would like to defer this change for the time being.

This would remain as a configurable value at the bundle layer of course and we do think it's still worth exploring/monitoring the effects of this or looking into other options for rate limiting. Also definitely could revisit this if/when we identify additional places where having the accurate client IP would be useful.

PR #987, opened 2024-11-07
fix!: remove remoteCidr exception block

### LINKED ISSUE #950: `remoteCidr` generated policy should only include `except` if in the range
https://github.com/defenseunicorns/uds-core/blob/2f63db2f26cb30c056f376b1823f758cd403aefe/src/pepr/operator/controllers/network/generators/remoteCidr.ts

Our current remoteCidr policies will unconditionally add the `except` block with the cloud metadata IP. We should either:
1. Conditional add it: based on whether it is included in the `cidr` we put in the policy, this would require determining if the CIDR includes it.
2. Never include it: Just drop that entirely, simpler but would allow someone to set this to 0.0.0.0/0 or another similar range that includes metadata.

Historically I have encountered issues when `except` IPs/CIDRs are not contained in the range. I haven't identified any upstream docs or issues related to this but could search for those if required.

chance-coleman:
## Description
BREAKING CHANGE:Remove the generated exception block from the remoteCidr generation. This change means that a cidr containing the META_IP could be set. 

Conversation with @mjnagel resulted in this implementation but open to other suggestions if we want to try and lock down allowed IP ranges with either generation logic, validations, or other.

## Related Issue

Fixes #950

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/operator/controllers/network/generate.spec.ts
+++ b/src/pepr/operator/controllers/network/generate.spec.ts
@@ -134,7 +134,6 @@ describe("network policy generate with remoteCidr", () => {
             {
               ipBlock: {
                 cidr: "192.168.0.0/16",
-                except: ["169.254.169.254/32"], // Include the except field here
               },
             },
           ],
@@ -162,7 +161,6 @@ describe("network policy generate with remoteCidr", () => {
             {
               ipBlock: {
                 cidr: "10.0.0.0/8",
-                except: ["169.254.169.254/32"], // Include the except field here
               },
             },
           ],
--- a/src/pepr/operator/controllers/network/generators/remoteCidr.ts
+++ b/src/pepr/operator/controllers/network/generators/remoteCidr.ts
@@ -4,14 +4,12 @@
  */
 
 import { V1NetworkPolicyPeer } from "@kubernetes/client-node";
-import { META_IP } from "./cloudMetadata";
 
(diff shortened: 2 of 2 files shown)


catsby:
Are we confident this is a non-breaking change?

chance-coleman:
@catsby that is a good call out. IMO this is technically a breaking change because I'm removing the guardrails protecting the exposure of the META_IP.

mjnagel:
@catsby @UnicornChance I think my slight leaning is that this does not seem like a breaking change. In this case the API surface area (`Package` CR spec) hasn't changed and we aren't adding any additional unexpected restrictions.

Definitely worth calling out that this has a potential change in security protections, but there are two possible scenarios:
1. `remoteCidr` was used with a range that did not include the metadata IP: In this case the network policy would've been broken (the bug issue this PR is fixing).
2. `remoteCidr` was used with a range that did include the metadata IP: With the change in this PR functionality will shift and now allow metadata access, but I'd argue that is more in line with my expectations when I allow a specific CIDR so still feels in line with a "bug fix" rather than a breaking change?

Open to other interpretations here - since we are pre-1.x it won't affect the versioning significantly but would more clearly call it out in release notes if we marked it as breaking.

mjnagel (approved):
LGTM, as discussed I think this is the best path forward since it seemed like a potential unexpected side effect to block this IP when someone has requested access to an explicit CIDR.

PR #995, opened 2024-11-08
fix: sets `fail-fast` to `false` for matrix workflows

noahpb:
## Description
We've noticed some behavior on the nightly ci pipelines where one failed parallel job was causing other in-progress jobs in the same workflow run to non gracefully terminate as the [cancellation timeout](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/canceling-a-workflow#:~:text=After%20the%205%20minutes%20cancellation%20timeout%20period%2C%20the%20server%20will%20force%20terminate%20all%20jobs%20and%20steps%20that%20don%27t%20finish%20running%20or%20fail%20to%20complete%20the%20cancellation%20process.) would be met. This led to inconsistent behavior with subsequent workflow runs involving terraform where pre-existing state files were locked. Workflows are continually failing until a new commit is made, which forces the workflow to generate a new state key. This would circumvent the issue, but consequentially leave behind orphaned resources. 

The intent of this PR is to ensure that all failed jobs gracefully exit and do not impact the status of other jobs that are running in the same workflow. This will add additional time to workflow runs, but it will always ensure that resources are properly cleaned up and that processes gracefully terminate before the pipeline completing.
...

## Related Issue
Example workflow run:
https://github.com/defenseunicorns/uds-core/actions/runs/11747339383/job/32731009925?pr=989#step:9:63

## Type of change

## Checklist before merging

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -15,7 +15,7 @@ on:
 jobs:
   publish-uds-core:
     strategy:
-      fail-fast: false    
+      fail-fast: false
       matrix:
         flavor: [upstream, registry1, unicorn]
     runs-on: "uds-ubuntu-big-boy-8-core"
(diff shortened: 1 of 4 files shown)


chance-coleman (approved):
I think the time trade off is worth the cleanup

Issue #951, opened 2024-10-23
Allow custom network policies for Grafana, Prometheus, and AlertManager


nfoucha:
### Is your feature request related to a problem? Please describe.
When deploying a uds package that requires connectivity to/from the other components in the monitoring stack (e.g. Thanos) there currently isn't a way to supply custom network policies for these components.  This blocks required communication channels and prevents the packaged application from working correctly.

### Describe the solution you'd like

- **Given** a package that requires network connectivity to/from monitoring stack components
- **When** an override for uds-core includes custom network policies for the relevant components
- **Then** the custom network policies are deployed in the appropriate namespaces and managed as part of the uds-core deployment

### Describe alternatives you've considered
Patching the kubernetes cluster to add/modify/remove network policies after the deployment is complete.

### Additional context
In the specific example use-case, Thanos acts as long-term storage for the Prometheus stack components so it needs to scrape prometheus (which is possible currently with the overly permissive network policies).  Thanos also has it's own set of alerts and a rules engine that integrates with AlertManager so it needs bi-directional connectivity.  Lastly, Thanos can act as a (Prometheus) datasource in Grafana, so Grafana needs an egress policy to communicate with Thanos.

mjnagel:
I think custom netpols could make sense here but also would be good to try and find some standard things we could directly support to make the config even easier. Here's a few ideas:
- Grafana
  - "Extra datasources" values to enable connection to in-cluster things, CIDRs, or anywhere (this came up recently with something like a cloudwatch datasource
  - Maybe a specific toggle for thanos if we want to directly support that one
- Prometheus
  - Thanos toggle to enable a netpol with the right values for access to "uds-package-thanos"
  - Maybe some custom policy access
- Alertmanager
  - Definitely need a customized egress policy that allows egress to send alerts somewhere
  - Potentially a specific egress policy for mattermost?
  - Custom ingress policy to support "additional alert sources"

TLDR is I would much rather try to identify specific usecases that we can turn into a simple toggle - like "I have thanos" - versus someone needing to specify a big block of YAML and hope they don't make it overly permissive.

PR #997, opened 2024-11-09
feat: add ability to add custom netpols for prometheus-stack package

### LINKED ISSUE #996: Allow Custom Netpols for Prometheus Stack Package
Currently, alertmanager can't reach outside the `monitoring` namespace with the default netpols deployed with core.  This is a problem, because it is typical to setup alertmanager with some sort of notification system (i.e. slack, email, etc).  Currently there is no way to route this traffic out of alertmanager.  And since each destination is going to be different for each customer, it makes sense to allow them to configure this themselves.  Recommend we add a custom value that can be overriden and allows netpols for unanticipated scenarios

- **Given** a deploy of uds-core
- **When** a user specifies an override value to add a custom netpol to the prometheus-stack package to allow alertmanager traffic to their end destination
- **Then** alertmanager is able to properly send alerts to their notification system

- Deploy a custom package with a single netpol manifest in it (feels overkill to create a package for this single use case that could  be override-able with a value)
- Set a blanket allow alertmanager egress to everywhere (probably not the best for security reasons)

### Additional context
Originally mentioned this in a comment [here](https://github.com/defenseunicorns/uds-core/issues/967#issuecomment-2453220670), but feel that this is a different problem and merits it's own issue

joelmccoy:
## Description

Adds a value to allow custom netpols for the prometheus-stack package.  Following a similar pattern that we use for [vector](https://github.com/defenseunicorns/uds-core/blob/v0.31.2/src/vector/chart/values.yaml#L4-L19).  Additionally added some documentation on how to override and add custom netpols to prometheus stack and vector.

## Related Issue

Fixes #996 

## Type of change

## Checklist before merging


noahpb:
Based on the discussion [here](https://github.com/defenseunicorns/uds-core/issues/951), I'm not sure if this is the preferred approach that we'd like to take for creating additional Network Policies - we have some more design to do here. Given that we currently don't have a solution, I think this is a reasonable approach that can work in the interim. cc @mjnagel and @UnicornChance.

joelmccoy:
@noahpb Good callout. Missed that issue.  It sounds like the preferred approach would be to minimize the customizability of these netpols.  The main use case I am trying to support is to allow alertmanager to egress to some place.  This place is going to be different for each customer/environment.  Would it be preferred to have a value that is more specific to where alertmanager can egress to?

A simple value like `alertmanagerEgress: []` which provides a list of ip ranges / netpol targets that it can go to?

mjnagel:
@noahpb / @joelmccoy yeah my main thoughts with that mention on the other issue are that we should try to capture specific use cases where we can. I don't think we necessarily should lock down all customizability, but where there are known needs to support it would be great to explicitly have a rule for those.

A specific value for alertmanager egress might make sense in this case, with conditionals for inCluster, cidr, or fallback for anywhere (in the future ideal state we could support specific hosts, i.e. `slack.com` or something). I also know that linked issue called out a few other needs for Thanos support and other alert sources so I could also see an argument to allow any custom policies.

Two other comments on this:
- It might be good to try and document this somewhere that surfaces on the docs site, like a "networking configuration" page that gives an example + shows the value to override.
- I don't like the `custom` value name personally. I know this is a pattern emerging with SWF and other apps but without looking at the template/having the tribal knowledge of what that does the name `custom` means nothing. I think we should be a lot more explicit with the name, i.e. `customNetworking` or [`additionalNetworkAllow`](https://github.com/defenseunicorns/uds-core/blob/c0b1dbf66aea906ea7badb5bc90c07f6cc42cf50/src/vector/chart/templates/uds-package.yaml#L35-L49) which more directly connect to the field in the Package CR.

joelmccoy:
Good feedback and clarity.  Do we have a preference on how flexible we want this too be?  For the alertmanager use case, I am thinking of two routes we could take at this point:

1. Add a field like `customNetworking` or `additionalNetworkAllow` and then give access to the end user to configure the `network` properties directly?
or..
2. Do we take a less flexible approach and add a field called something like `alertmanagerEgressAllow` which is only allowed to customize where alertmanager can send alerts.

Happy to document example overrides in the docs for whatever route we prefer

mjnagel:
@joelmccoy I think option 1 for now. This provides the flexibility that people need and supports unknown end user needs. I think if/when we identify specific possible network targets we should add conditional templates for those specific pieces, but will likely still want this block to support other unknown needs.

joelmccoy:
Based on feedback added in an `additionalNetworkAllow` value for the prometheus stack.  This is the same pattern currently used for `vector`.  Additionally added some documentation to demonstrate how to add additional network allowances for both Prometheus Stack and Vector.

noahpb (dismissed):
LGTM - Thanks for adding a doc! :)

src/prometheus-stack/chart/templates/uds-package.yaml
```diff
@@ -65,3 +65,18 @@ spec:
...
+        {{- if not .remoteGenerated }}
+        remoteNamespace: {{ .remoteNamespace }}
+        remoteSelector:
+          {{ .remoteSelector | toYaml | nindent 10 }}
+        port: {{ .port }}
+        {{- else }}
+        remoteGenerated: {{ .remoteGenerated }}
+        {{- end }}
+        description: {{ .description }}
+      {{- end }}
```
mjnagel: Wondering if we might want to support `remoteCidr` here as well? Recognize we don't have this one on the Vector netpol either - might just be easier to allow anything here and `toYaml` it so we don't have to maintain when new options pop up here (in the future we might have something like `remoteHost`). Thoughts?

joelmccoy: good callout.  I like the idea of that.  instead of templating this out, lets just let them dump the yaml from this value.  should future proof for future fields (and also account for remoteCidr).  I can get this change pushed up later

joelmccoy: fixed this in latest commit

mjnagel (approved):

mjnagel:
Will rebase and merge once https://github.com/defenseunicorns/uds-core/pull/1020 merges.

PR #1030, opened 2024-11-20
feat: allow TLS to be disabled for gateways

willswire:
## Description

When deploying UDS Core Base to EKS, we can use annotations to populate certificates from Amazon’s AWS Certificate Manager (ACM) on the ELBs provisioned by Istio (see [here](https://faun.pub/managing-tls-keys-and-certs-in-istio-using-amazons-acm-8ff9a0b99033) and [here](https://github.com/istio/istio/issues/6566#issuecomment-427136888)).

Unfortunately, this does not work if an entry in the Gateway's `servers` spec contains `tls` -- likely a result of the conflicting certs provided for `*.uds.dev`. 

By exposing the option of enabling/disable TLS via a variable, we can opt to disable it, and therefore enable annotation-based cert management.

## Type of change

## Checklist before merging


src/istio/chart/templates/gateway.yaml
```diff
@@ -27,6 +27,7 @@ spec:
         {{- range $server.hosts | default (list "*") }}
         - "{{ . }}.{{ $domain }}"
         {{- end }}
+      {{- if $.Values.tls.enabled }}
```
mjnagel, line 30: Definitely see the utility here but would be very hesitant to ship this as-is without some safeguards to ensure that TLS is being enforced at a different layer (i.e. using those AWS LB annotations). I don't have any immediate great ideas here but there might be some ways to approach this using...
1. Zarf actions to check something `onDeploy.before`?
2. Helm jobs to check something pre-install/pre-upgrade?
3. Pepr Policy to block this gateway config except in x/y/z scenario?

The Pepr Policy is probably the most "uds friendly" approach, followed maybe by the zarf action. Regardless of approach we should also document this configuration/how it should be used in our ingress doc I think.

willswire: After working through some Pepr Policy stuff at first, it seems like the more straightforward way to implement some safeguards is adding a Helm job/chart hook that checks for the annotations prior to installing the gateway object. I've added this in my latest commit.

willswire (commented):
Implemented TLS configuration validation via a Helm chart hook.

src/istio/chart/templates/tls-validation-job.yaml
```diff
@@ -0,0 +1,95 @@
...
+    metadata:
+      name: "{{ .Release.Name }}-tls-validation"
+      labels:
+        {{- include "uds-istio-config.labels" . | nindent 8 }}
+    spec:
+      serviceAccountName: "{{ .Release.Name }}-tls-validation-sa"
+      restartPolicy: Never
+      containers:
+        - name: tls-validation
+          image: bitnami/kubectl:1.31.3
```
mjnagel: We'll need to make this image configurable per-flavor so that we properly use chainguard or registry1 where required.

While not required, it would also be ideal if we could use an image already in our stack so that we aren't adding extra images for something that is conditional and only used in a very specific AWS setup. I think we pull in kubectl already for Velero and possibly some other components so we could make sure the images are aligned if we continue down this route.

src/istio/chart/templates/tls-validation-job.yaml
```diff
@@ -0,0 +1,95 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
```
mjnagel: Can we make this all conditional on `not tls.enabled`? I think the check is only applicable in that case, so we might as well remove the complexity if TLS is enabled, rather than having that check in the job.

src/istio/chart/templates/tls-validation-job.yaml
```diff
@@ -0,0 +1,95 @@
...
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: "{{ .Release.Name }}-tls-validation-sa"
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/hook": pre-install, pre-upgrade
+    "helm.sh/hook-weight": "-6"
+    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
```
mjnagel: Open to discussion but I'd lean towards `hook-succeeded,before-hook-creation` here so that the job output is available on failure. Otherwise I think the job output would just disappear and make it hard to see what happened I think?

src/istio/chart/templates/tls-validation-job.yaml
```diff
@@ -0,0 +1,95 @@
+# Copyright 2024 Defense Unicorns
```
mjnagel: Broad comment on the helm hook approach...I'm wondering what the UX looks like on failure here? Zarf would probably just show a generic helm timeout failure which isn't ideal imo. If we moved this to a zarf action or pepr policy we'd get a more clear error in the zarf/uds output I think.

src/istio/chart/templates/gateway.yaml
mjnagel: We don't have to go the pepr policy route if it seems too challenging, but just as a note for reference - we generate all the types used by Pepr using [this KFC command](https://github.com/defenseunicorns/kubernetes-fluent-client?tab=readme-ov-file#generating-typescript-definitions-from-crds). It should be pretty easy to get the necessary gateway type added if desired. For example, the prometheus CRDs got added with this:
```console
npx kubernetes-fluent-client crd https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.76.1/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml src/pepr/operator/crd/generated/prometheus
```

mjnagel (changes requested):

src/istio/chart/templates/gateway.yaml
willswire: I decided to go back and work the Pepr Policy solution instead. Much cleaner and better UX!

Screenshot 2024-11-22 at 12 24 56 PM

src/istio/chart/templates/tls-validation-job.yaml
willswire: N/A as we're going Pepr route

mjnagel:
@willswire just to check on this one - I think we determined that we would close this out due to some of the complexity/issues with non-tls traffic in the loop here?

willswire:
> @willswire just to check on this one - I think we determined that we would close this out due to some of the complexity/issues with non-tls traffic in the loop here?

Yep! We're good to close this one out.

PR #1032, opened 2024-11-20
feat: istio native sidecars

### LINKED ISSUE #536: Switch to Istio native sidecars

Currently we use istio's sidecars across all pods in the mesh. Due to existing istio issues we have also implemented a Pepr watch action to handle [properly terminating sidecars for jobs](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/istio/index.ts). If we switch to ["native" sidecars](https://istio.io/latest/blog/2023/native-sidecars/) we will no longer need this action and may have some other benefits with regards to init containers accessing the network (common with things that validate dependencies that may be in cluster).

Istio should be deployed with native sidecars by default (`ENABLE_NATIVE_SIDECARS` env on pilot). When upgrading this should be a seamless rollout - unsure if we will need to cycle all pods or something similar here (may be worth evaluating [this issue](https://github.com/defenseunicorns/uds-core/issues/16) first).

Definition of done:
- Native sidecars are enabled by default in uds core
- Sidecar termination watch is removed from pepr code
- On an upgrade all pods are properly migrated to native sidecars
- Documentation may need an update to ensure our k8s version pre-req is captured (this would require 1.29, i.e. non-EOL versions)

We could continue to use non-native sidecars + our pepr capability.

### Additional context

This functionality in kubernetes is available in 1.28 behind a feature flag, and 1.29 by default. We should be fairly safe to enable this but should document this requirement well (maybe in our k8s/distro support page). In theory someone could use overrides to switch to non-native sidecars but they would be missing our pepr capability unless we maintain that behind a flag as well.

chance-coleman:
## Description

Implement istio native sidecars until we can move over completely to ambient.

## Type of change

## Checklist before merging


src/pepr/policies/security.ts
```diff
@@ -141,22 +141,28 @@ When(a.Pod)
...
     }
-    // Check if running as root by checking if runAsNonRoot is false or runAsUser is 0
-    const isRoot = (ctx: Partial<V1SecurityContext>) => {
+
+    // Helper function to check if a container requires root privileges
+    const isRoot = (ctx: Partial<V1SecurityContext>, containerName: string) => {
+      // Skip the root check for `istio-init`
+      if (containerName === "istio-init") {
+        return false;
+      }
```
mjnagel: So this does work as is, but there's some [existing code](https://github.com/defenseunicorns/uds-core/blob/istio-native-sidecars/src/pepr/policies/common.ts#L68) that was intended to filter out istio init containers that is now unused. The only problem with that logic is [this line](https://github.com/defenseunicorns/uds-core/blob/istio-native-sidecars/src/pepr/policies/common.ts#L79) because proxies are `initContainers` rather than `containers` when we switch to native sidecar. If we wanted to make the change there and revert the other changes to this file that would work.

I think we have in the past identified that the way we filter out istio containers is overly permissive, so from that perspective I do like the approach you took here. If we did want to go this route I think we'd want to make it a bit "safer" and do a few extra checks to ensure that the container is actually Istio and not just something named `istio-init` (similar to what the current filter logic does to ensure it has an istio annotation + an `istio-proxy` container + the expected command args). Short term it might be easier to just update the existing filter?

src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -13,6 +13,7 @@ const log = setupLogger(Component.OPERATOR_ISTIO);
 
 const injectionLabel = "istio-injection";
 const injectionAnnotation = "uds.dev/original-istio-injection";
+const nativeIstioAnnotation = "uds.dev/native-istio-migrated";
```
mjnagel: nit: don't love this annotation key 😆 - would `uds.dev/native-istio-sidecars` work instead?

src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -29,40 +30,61 @@ export async function enableInjection(pkg: UDSPackage) {
...
-          name: pkg.metadata.namespace,
-          labels,
-          annotations,
-        },
-      },
-      { force: true },
+    log.info(
+      `Updating package-specific annotation for ${pkg.metadata.name} in namespace ${pkg.metadata.namespace}.`,
     );
+    shouldRestartPods = true; // Pods need restarting due to package change
```
mjnagel: I don't think this case requires a pod restart? If you had two package CRs in the same namespace we would add a package annotation for the second one, but istio injection (and native sidecars) would already be implemented before it gets applied so we wouldn't need to restart for just the pkgKey annotation. (I don't think we should set `shouldRestartPods` to `false` here either though, just not touch it)

mjnagel: You could test this out and see the problem by applying a bogus extra package CR to any namespace like this:
```
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: test
  namespace: monitoring
```
You'll see the monitoring pods cycle when they shouldn't need to.

src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -135,12 +157,23 @@ async function killPods(ns: string, enableInjection: boolean) {
...
+
+    if (enableInjection) {
+      if (foundSidecar) {
+        log.info(
+          `Marking Pod ${ns}/${pod.metadata?.name} for deletion to reconcile Istio sidecars.`,
+        );
+      } else {
+        log.info(`Marking Pod ${ns}/${pod.metadata?.name} for deletion to inject Istio sidecars.`);
+      }
     }
```
mjnagel: I think this might make more sense here:
```ts
    // note that this is `initContainers` now because that is how native sidecar works
    const foundSidecar = pod.spec?.initContainers?.find(c => c.name === "istio-proxy");

    // Keep the original check/skip if sidecar exists, but because of the change ^ this skips only if a native sidecar exists
    // If enabling injection, ignore pods that already have the istio sidecar
    if (enableInjection && foundSidecar) {
      log.debug(`Ignoring Pod ${ns}/${pod.metadata?.name}, already has sidecar`);
      continue;
```

I might be missing something here, but I think this feels a bit more robust and would prevent potential for deleting a pod that already has a sidecar?

chance-coleman: offline conversation, agreed this is a better path forward

src/pepr/policies/security.ts
```diff
@@ -399,10 +399,18 @@ When(a.Pod)
...
     const authorized = ["NET_BIND_SERVICE"];
+    const istioInitAuthorized = ["NET_ADMIN", "NET_RAW"];
 
-    const violations = securityContextContainers(request).filter(
-      c => c.ctx?.capabilities?.add && !c.ctx?.capabilities.add.includes(authorized[0]),
-    );
+    const violations = securityContextContainers(request).filter(c => {
+      const containerCapabilities = c.ctx?.capabilities?.add || [];
+      // Separate handling for `istio-init` container
+      if (c.name === "istio-init") {
```
noahpb: Same thing here as @mjnagel's previous comment - should use `isIstioInitContainer` here.

chance-coleman: good call, thanks

src/pepr/policies/security.ts
chance-coleman: went with updating the existing filter and reverting changes to the security.ts file.

src/pepr/policies/common.ts
```diff
@@ -70,19 +70,25 @@ export function isIstioInitContainer(
...
+
+  // Check for an Istio proxy in initContainers
+  const hasInitContainerSidecar = request.Raw.spec?.initContainers?.some(
+    c => c.name === "istio-proxy" && c.args?.includes("proxy"),
+  );
+
+  // Exit if no Istio proxy is found in either container or initContainer
+  if (!hasContainerSidecar && !hasInitContainerSidecar) {
     return false;
   }
```
mjnagel: Might be missing the why here - do we need to account for both scenarios? Or could we just have the single "native sidecar" check since we will be switching entirely to native?

If we do want both here let's make sure we line up the checks if we can (I think the init one is missing the port check).

chance-coleman: Decided here to remove the container checks and forego any verbosity since to avoid maintaining the container sidecar code.

src/pepr/policies/common.ts
```diff
@@ -70,19 +70,17 @@ export function isIstioInitContainer(
...
 
-  // Check for what looks like an istio sidecar
-  const possibleSidecar = request.Raw.spec?.containers?.find(
-    c =>
-      c.name === "istio-proxy" &&
-      c.ports?.find(p => p.name === "http-envoy-prom") &&
-      c.args?.includes("proxy"),
+  // Check for an Istio proxy in initContainers
+  const hasInitContainerSidecar = request.Raw.spec?.initContainers?.some(
+    c => c.name === "istio-proxy" && c.args?.includes("proxy"),
```
noahpb: We should consider adding an additional check to verify that the init container image (`c.image`) contains `istio/proxyv2`. Theoretically a user could masquerade a malicious container image as an istio init container (see kyverno security best practices for service mesh [ref](https://kyverno.io/blog/2024/02/04/securing-services-meshes-easier-with-kyverno)).

mjnagel: Not opposed to that, but we would need to make sure we account for flavors that might name the image slightly differently (cgr names the image `istio-proxy` or something similar iirc). As you noted it would have to be a `contains` check rather than a `equals` check since other parts of the ref like the registry, etc aren't as guaranteed. It ultimately would just provide one extra hoop to jump through though since an attacker could name their image whatever and use the right combo of heuristics to bypass our checks. Real security here would require extremely strict heuristics + image signature validation, but that's not super viable short term.

chance-coleman: offline conversation, deciding to kick this down the road in favor of image signature validation that either core or cli (maybe?) will implement.

src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -29,40 +30,60 @@ export async function enableInjection(pkg: UDSPackage) {
...
+  } else {
+    log.debug(`No namespace updates needed for ${pkg.metadata.namespace}.`);
+  }
+
+  // Restart pods if required
+  if (shouldRestartPods) {
+    log.debug(`Restarting pods in ${pkg.metadata.namespace} due to configuration changes.`);
+    await killPods(pkg.metadata.namespace, true);
   }
 }
```
mjnagel, line 80: I can't comment on the actual line, but right below here we have a `cleanupNamespace` that gets run on removal of a Package CR. We should make sure we delete our `nativeIstioAnnotation` in that function, around the same place we delete the injectionAnnotation (`delete annotations[injectionAnnotation];`).

src/pepr/operator/controllers/istio/injection.ts
```diff
@@ -29,40 +30,60 @@ export async function enableInjection(pkg: UDSPackage) {
...
-  if (!annotations[pkgKey] || originalInjectionLabel !== "enabled") {
-    // Ensure Istio injection is enabled
+  let shouldRestartPods = false;
+
+  // Ensure native Istio migration annotation is present
+  if (!hasMigrationAnnotation) {
+    annotations[nativeIstioAnnotation] = "true";
+    log.info(`Adding native Istio migration annotation to namespace ${pkg.metadata.namespace}.`);
+    shouldRestartPods = true; // Pods need restarting for migration
+  }
```
mjnagel: Comment open for discussion - after seeing the performance of this rollout and cycling all pods in a large cluster I'd advocate that we "do nothing" here and just allow eventual consistency with native sidecars (i.e. don't immediately cycle pods).

chance-coleman: Offline conversation with team, we will move forward without rolling / cycling pods.

mjnagel (approved):
LGTM - validated the upgrade path, install, and several aspects of the logic around removal of Package CRs. As noted this will not immediately cycle all pods to native sidecars but we don't see any issue with that.

noahpb (approved):

PR #1037, opened 2024-11-21
fix: keycloak upgrade wait

chance-coleman:
## Description

keycloak upgrades on major versions seem to be a bit finicky. Adding a wait to kubectl scale down so that the scale down finishes, before the update happens.

## Type of change

## Checklist before merging


mjnagel (approved):
Validated locally with an upgrade from 0.30.0, this behaved as anticipated.

PR #1038, opened 2024-11-21
feat: allow additional network rules for grafana and neuvector

### LINKED ISSUE #951: Allow custom network policies for Grafana, Prometheus, and AlertManager

mjnagel:
## Description

Following the pattern on Vector and Prometheus this adds support for custom network policies via overrides, as well as the examples for this.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/951

Note that we may want to continue to track usage of these fields and as suggested [here](https://github.com/defenseunicorns/uds-core/issues/951#issuecomment-2432728395) add some specific templated policies for easy toggle on/off if we see them being adopted more widely.

## Type of change

## Checklist before merging

### DIFF
--- a/docs/reference/configuration/uds-networking-configuration.md
+++ b/docs/reference/configuration/uds-networking-configuration.md
@@ -103,7 +103,7 @@ Reference the [spec for allow](https://uds.defenseunicorns.com/reference/configu
 
 ### NeuVector
 
-It may be desired to connect Grafana to additional datasources in or outside of the cluster. To facilitate this, you can provide a bundle override as follows:
+It may be desired send alerts from NeuVector to locations in or outside of the cluster. To facilitate this, you can provide a bundle override as follows:
 
 ```yaml
 packages:
--- a/docs/reference/configuration/uds-networking-configuration.md
+++ b/docs/reference/configuration/uds-networking-configuration.md
@@ -33,9 +33,9 @@ packages:
                   port: 443
 ```
 
-The example above allows Alertmanager to send alerts to any external destination. Alternatively, you could use the remoteNamespace key to specify another namespace within the Kubernetes cluster.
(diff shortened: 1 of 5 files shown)


docs/reference/configuration/uds-networking-configuration.md
```diff
@@ -70,4 +70,60 @@ packages:
...
+                  description: "Thanos Query"
+```
+
+The example above allows Grafana to query a remote Thanos instance in your cluster.
+
+Reference the [spec for allow](https://uds.defenseunicorns.com/reference/configuration/custom-resources/packages-v1alpha1-cr/#allow) for all available fields.
+
+### NeuVector
+
+It may be desired to connect Grafana to additional datasources in or outside of the cluster. To facilitate this, you can provide a bundle override as follows:
```
noahpb:
```suggestion
It may be desired to connect NeuVector to additional datasources in or outside of the cluster. To facilitate this, you can provide a bundle override as follows:
```

mjnagel: Good catch - I'm going to modify the wording here a bit more and will fix this.

noahpb (approved):

PR #1098, opened 2024-12-06
chore(docs): replace promtail reference with vector in prerequisites

joelmccoy:
## Description

There was a lingering reference to promtail in the website docs on the prerequisites page.  Replaced promtail with vector.

## Type of change

## Checklist before merging


joelmccoy:
If there are new instructions for vector, I'm happy to add them in.  I just thought I would get this PR up to open the discussion.

mjnagel:
@joelmccoy I _think_ that Vector would have this same set of pre-reqs, since ultimately all of these are just tied to the tooling being able to open and tail all the logs files it needs too. Might be safe to just swap the section title in this case.

joelmccoy:
ack - swapped out promtail for vector

mjnagel (approved):

PR #1102, opened 2024-12-09
feat: configurable authentication flows

chance-coleman:
## Description

See [uds-identity-config PR](https://github.com/defenseunicorns/uds-identity-config/pull/282).

The real change is adding the secretRef for getting env's into the identity config init container.

Some whitespace cleanup in statefulset and some new values overrides for the realm init.

## Type of change

## Checklist before merging


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -28,7 +28,7 @@ spec:
         {{- end }}
        {{- if not .Values.devMode }}
       annotations:
-        postgres-hash: {{ include (print $.Template.BasePath "/secret-postgresql.yaml") . | sha256sum }}        
+        postgres-hash: {{ include (print $.Template.BasePath "/secret-postgresql.yaml") . | sha256sum }}
```
mjnagel, line 31: The space change here made me think about this - might be good to add a checksum for the realm env secret now since the theme envs could be hot reloaded.

chance-coleman: I was wondering about how to trigger that earlier. Appreciate the suggestion

mjnagel (dismissed):
LGTM

src/keycloak/chart/templates/secret-kc-realm.yaml
```diff
@@ -16,4 +16,13 @@ data:
...
+  {{- end }}
+
+  SOCIAL_LOGIN_ENABLED: {{ .Values.realmInitEnv.SOCIAL_AUTH_ENABLED | toString | b64enc }}
+  X509_LOGIN_ENABLED: {{ .Values.realmInitEnv.X509_AUTH_ENABLED | toString | b64enc }}
+  USERNAME_PASSWORD_AUTH_ENABLED: {{ .Values.realmInitEnv.USERNAME_PASSWORD_AUTH_ENABLED | toString | b64enc }}
+  REGISTER_BUTTON_ENABLED: {{ or .Values.realmInitEnv.USERNAME_PASSWORD_AUTH_ENABLED .Values.realmInitEnv.X509_AUTH_ENABLED | toString | b64enc }}
+  DENY_USERNAME_PASSWORD_ENABLED: {{ ternary "DISABLED" "REQUIRED" (.Values.realmInitEnv.USERNAME_PASSWORD_AUTH_ENABLED) | b64enc }}
+  RESET_CREDENTIAL_FLOW_ENABLED: {{ ternary "REQUIRED" "DISABLED" (.Values.realmInitEnv.USERNAME_PASSWORD_AUTH_ENABLED) | b64enc }}
+  REGISTRATION_FORM_ENABLED: {{ ternary "REQUIRED" "DISABLED" (or .Values.realmInitEnv.USERNAME_PASSWORD_AUTH_ENABLED .Values.realmInitEnv.X509_AUTH_ENABLED) | b64enc }}
+  OTP_ENABLED: {{ .Values.realmInitEnv.USERNAME_PASSWORD_AUTH_ENABLED | toString | b64enc }}
```
mjnagel: Related to https://github.com/defenseunicorns/uds-identity-config/pull/282/files#r1884096378

I think previously we allowed independent control over disabling OTP. This changes makes it so that OTP is always required when user/pass is enabled I think? Would it make sense to adjust this conditional to allow someone to disable OTP still, even if they only have user/pass?

ex:
```suggestion
  OTP_ENABLED: {{ (and .Values.realmInitEnv.OTP_ENABLED .Values.realmInitEnv.USERNAME_PASSWORD_AUTH_ENABLED) | toString | b64enc }}
```

src/keycloak/chart/values.yaml
```diff
@@ -27,17 +27,21 @@ realm: uds
...
-#   GOOGLE_IDP_ADMIN_GROUP: ""
-#   GOOGLE_IDP_AUDITOR_GROUP: ""
-#   PASSWORD_POLICY: "hashAlgorithm(pbkdf2-sha256) and forceExpiredPasswordChange(90) and specialChars(2) and lowerCase(0) and upperCase(0) and passwordHistory(5) and length(12) and notUsername(undefined) and digits(0)"
-#   EMAIL_VERIFICATION_ENABLED: true
-#   OTP_ENABLED: true
-#   TERMS_AND_CONDITIONS_ENABLED: true
-#   REALM_X509_OCSP_FAIL_OPEN: true
+  USERNAME_PASSWORD_AUTH_ENABLED: true
+  X509_AUTH_ENABLED: true
+  SOCIAL_AUTH_ENABLED: true
```
mjnagel: Just realized as well, since these values are under `realmInitEnv` they get added as part of the range in the secret also. So you end up with `USERNAME_PASSWORD_AUTH_ENABLED` and `REALM_USERNAME_PASSWORD_AUTH_ENABLED` both present in the secret. Would it make sense to move these "special" 3 values under a separate key like `realmAuthFlows`?

src/keycloak/chart/templates/secret-kc-realm.yaml
chance-coleman: agreed! Will add.

src/keycloak/chart/values.yaml
chance-coleman: per offline conversation, considered using an omit in the range loop but that add some unnecessary maintenance burden. Moving forward with a new key that is specific for auth flow toggles.

src/keycloak/chart/values.yaml
```diff
@@ -27,17 +27,24 @@ realm: uds
...
+  # TERMS_AND_CONDITIONS_ENABLED: true
+  # X509_OCSP_FAIL_OPEN: true
+  # DISABLE_REGISTRATION_FIELDS: false
+
+# UDS Identity Config Authentication Flows Environment Variables. More info here: https://github.com/defenseunicorns/uds-identity-config/blob/main/docs/CUSTOMIZE.md#templated-realm-values
+realmAuthFlows:
+  USERNAME_PASSWORD_AUTH_ENABLED: false
+  X509_AUTH_ENABLED: true
+  SOCIAL_AUTH_ENABLED: true
+  OTP_ENABLED: false
```
mjnagel: Shouldn't these all be true by default still to align with our previous defaults?

chance-coleman: yes...

chance-coleman: updated, thanks

mjnagel (approved):

PR #1114, opened 2024-12-12
chore: allow separate configuration of admin domain name

mjnagel:
## Description

Allows more flexibility in deployment by exposing a new zarf var for `ADMIN_DOMAIN`.

Note that an entirely different domain will require this change in Keycloak: https://github.com/defenseunicorns/uds-identity-config/pull/284

## Related Issue

N/A

## Type of change

## Steps to Validate

Testing can be done by exporting `UDS_ADMIN_DOMAIN=uds.dev`. With these changes you should be able to see all virtualservices and gateways use `uds.dev`. Applications will not immediately be accessible due to uds-k3d's nginx config routing them to the wrong gateway (plus the wrong certs used for the admin interfaces). You can also test with a different domain entirely but this requires similar setup.

## Checklist before merging


mjnagel:
This is ready for review, but noting that a completely different domain will require the new 0.9.0 identity-config release.

noahpb (approved):

Issue #970, opened 2024-10-30
prometheus is missing container metrics from certain nodes


noahpb:
### Environment
Device and OS: darwin arm64
App version: v0.29.1-unicorn
Kubernetes distro being used: k3d with two nodes

### Steps to reproduce
1. Create a k3d cluster with additional nodes
```
$ kubectl get node
NAME               STATUS   ROLES                  AGE   VERSION
k3d-agent1-0       Ready    <none>                 23m   v1.30.4+k3s1
k3d-uds-server-0   Ready    control-plane,master   25m   v1.30.4+k3s1
```
2. Deploy uds-core with monitoring

### Expected result
Container metrics such as CPU and Memory utilization should be queryable

### Actual Result
Prometheus only returns metrics from pods that are scheduled on control plane nodes

### Visual Proof (screenshots, videos, text, etc)
Metrics returned for `container_cpu_usage_seconds`
![image](https://github.com/user-attachments/assets/dbbca229-ebc4-4c23-82cb-740a3613b3e8)

No metrics returned when filtering out control plane node:
![image](https://github.com/user-attachments/assets/902a21b4-7a59-47d3-82c3-43f3bff1295d)

### Severity/Priority
Moderate

### Additional Context
Removing all `NetworkPolicies` in the `monitoring` namespace allows Prometheus to pick up metrics from the missing nodes.

joelmccoy:
Internal related issue: https://github.com/defenseunicorns/uds-infrastructure/issues/573

noahpb:
Thanks to @rjferguson21's suggestion, we've been able to confirm that the `allow-prometheus-stack-egress-metrics-scraping` NetworkPolicy generated by the operator needs to be adjusted. The `remoteNamespace: ""` specification is not permissive enough to allow egress traffic to the prometheus-node-exporter daemonset pods. Manually adjusting the egress specification of the NetworkPolicy to the CIDR range of the nodes worked in my local testing.

mjnagel:
Would suggest to resolve this we build an `AllNodes` generated target. We should be able to build that list of IPs using a watch on the nodes with Pepr, similar to our `KubeAPI` target. This would also be helpful for metrics-server which has an `Anywhere` rule with a todo comment to switch that to an all nodes target.

Code links for current kubeapi logic:
[...]

PR #1119, opened 2024-12-12
fix: add generated target for all node IPs

catsby:
## Description

Adds a new generator / target called `KubeNodes` that contains the internal IP addresses of nodes in the cluster. 

**NOTE:** ~I have no idea (yet) wher the `docs/reference/` file changes came from.~ They appear to be missing on `main`.

## Related Issue

Relates to #970 . `Steps to Validate` include steps to verify 970 gets fixed.

## Type of change

## Steps to Validate

<details>

### Setup and verify behavior of the target

Create a k3d cluster named `uds` (we use names later for adding nodes):

```bash
k3d cluster create uds
```

Deploy slim-dev:

```bash
uds run slim-dev
```

Create and deploy monitoring layer:

```bash
uds run -f ./tasks/create.yaml single-layer-callable --set LAYER=monitoring

uds run -f ./tasks/deploy.yaml single-layer-callable --set LAYER=monitoring
```

Create and deploy metrics-server layer:

```bash
uds run -f ./tasks/create.yaml single-layer-callable --set LAYER=metrics-server

uds run -f ./tasks/deploy.yaml single-layer-callable --set LAYER=metrics-server
```

Inspect the network policy for scraping of kube nodes:

```bash
kubectl describe networkpolicy allow-prometheus-stack-egress-metrics-scraping-of-kube-nodes -n monitoring
```

The `spec:` part is the relevant part, and should contain the IPs of the nodes:

```bash
Spec:
  PodSelector:     app.kubernetes.io/name=prometheus
  Not affecting ingress traffic
  Allowing egress traffic:
    To Port: <any> (traffic allowed to all ports)
    To:
      IPBlock:
        CIDR: 172.28.0.2/32
        Except:
  Policy Types: Egress

```

Add a node:

```bash
k3d node create extra1 --cluster uds --wait --memory 500M
```

Verify the internal IP of the new node:

```bash
kubectl get nodes -o custom-columns="NAME:.metadata.name,INTERNAL-IP:.status.addresses[?(@.type=='InternalIP')].address"
```

Re-get the netpol to verify the new ip is in the `spec:` block:

```bash
kubectl describe networkpolicy allow-prometheus-stack-egress-metrics-scraping-of-kube-nodes -n monitorin

```

Should now be something like this:

```bash
Spec:
  PodSelector:     app.kubernetes.io/name=prometheus
  Not affecting ingress traffic
  Allowing egress traffic:
    To Port: <any> (traffic allowed to all ports)
    To:
      IPBlock:
        CIDR: 172.28.0.2/32
        Except:
    To:
      IPBlock:
        CIDR: 172.28.0.4/32
        Except:
  Policy Types: Egress
```

### Verify Prometheus can read things

Connect directly to prometheus:

```bash
kubectl port-forward -n monitoring svc/kube-prometheus-stack-prometheus 9090:9090
```

Visit http://localhost:9090/ 

Execute this expression to see all node/cpu data:

```bash
node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate
```

To see just info from the `extra1` node:

```bash
node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate{node=~"^k3d-extra.*"}
```

Add a new node:

```bash
k3d node create extra2 --cluster uds --wait --memory 500M
```

Verify the netpol updates:

```bash
kubectl describe networkpolicy allow-prometheus-stack-egress-metrics-scraping-of-kube-nodes -n monitorin
```

Re-execute the Prometheus query from above. It make take a few minutes for `extra2` to show up though. Not sure why.

Delete a node and verify the spec updates again:

```bash
kubectl delete node k3d-extra1-0 && k3d node delete k3d-extra1-0
```

Re-reading the netpol should should the removal of that IP
</details>

## Checklist before merging


docs/reference/configuration/custom-resources/exemptions-v1alpha1-cr.md
```diff
@@ -0,0 +1,76 @@
+---
```
mjnagel: Not introduced by you, but these docs are supposed to be under [this folder](https://github.com/defenseunicorns/uds-core/tree/970-all-nodes/docs/reference/configuration/custom%20resources) (with a space instead of a `-`). It looks like we relocated them at one point but forgot to update the output dir in [this script](https://github.com/defenseunicorns/uds-core/blob/970-all-nodes/src/pepr/docs-gen/main.ts#L10). If you could also include that small fix in this PR that would be great - otherwise we can fix in a oneliner elsewhere, but we'll want to make sure the generated docs are not duplicated here.

I wasn't totally tracking that change but I believe we switched to a space due to how the docs site uses the folder name for generating sidebar nav.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,146 @@
...
+// Maintain a set of all node internal IPs
+const nodeSet = new Set<string>();
+
+/**
+ * Initialize the node targets by fetching the current nodes in the cluster
+ * and populating the nodeSet with their Internal IPs.
+ */
+export async function initAllNodesTarget() {
+  try {
+    const nodes = await fetchKubernetesNodes();
```
mjnagel: It's a bit of a safety net for us but would you be able to update this to use the `retryWithDelay` function ([example here](https://github.com/defenseunicorns/uds-core/blob/970-all-nodes/src/pepr/operator/controllers/network/generators/kubeAPI.ts#L30)). It provides a bit of safety around potential k8s api instability and gives Pepr a few chances to get the IPs.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,146 @@
...
+    log.error("error fetching node IPs:", err);
+  }
+}
+
+/**
+ * Returns the egress CIDRs of all known nodes as network policy peers.
+ * If none are known, defaults to 0.0.0.0/0 and logs a warning.
+ */
+export function nodeCIDRs(): V1NetworkPolicyPeer[] {
+  const policies = buildNodePolicies([...nodeSet]);
```
mjnagel, line 55: Bit of a nit, but is there a reason we shouldn't just store the `nodePeers` as our set in the first place? I'm comparing for consistency against the KubeAPI target where the `V1NetworkPolicyPeer` list is what we have [cached in memory](https://github.com/defenseunicorns/uds-core/blob/970-all-nodes/src/pepr/operator/controllers/network/generators/kubeAPI.ts#L19), rather than a string set.

I think it would clean up our code a bit since we have lines similar to this in several places:
```
  const policies = buildNodePolicies([...nodeSet]);
  await updateKubeNodesNetworkPolicies(policies);
```
Which could probably be reduced to `await updateKubeNodesNetworkPolicies(nodePeers)`. We could still make that a set to keep the benefits of our logic around add/delete and just add an `Array.from(nodePeers)` into the `updateKubeNodesNetworkPolicies` function or the spread operator to get it into the type we need for the netpols.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,146 @@
...
+  if (policies.length > 0) return policies;
+
+  log.warn("Unable to get API server CIDR, defaulting to 0.0.0.0/0");
```
mjnagel:
```suggestion
  log.warn("Unable to get Node CIDRs, defaulting to 0.0.0.0/0");
```
(feel free to modify, but this should be nodes rather than api server)

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,146 @@
...
+ */
+export async function updateKubeNodesFromCreateUpdate(node: kind.Node) {
+  const isReady = node.status?.conditions?.some(
+    (condition: V1NodeCondition) => condition.type === "Ready" && condition.status === "True",
+  );
+
+  if (isReady) {
+    const ip = getNodeInternalIP(node);
+    if (ip) nodeSet.add(ip);
+  }
```
mjnagel: I think there's an edge case we might be missing here where a node has an `Update` that results in it being not-ready, but we have it in our set and need to remove it?

Maybe that's unnecessary (and could cause some churn if there's a broken node) but I think if we want to strictly track Ready nodes it might make sense to add?

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * Update all NetworkPolicies labeled with uds/generated=KubeNodes to
+ * reflect the given node CIDRs.
+ */
+export async function updateKubeNodesNetworkPolicies(newNodes: V1NetworkPolicyPeer[]) {
+  const netPols = await K8s(kind.NetworkPolicy)
+    .WithLabel("uds/generated", RemoteGenerated.KubeNodes)
+    .Get();
```
mjnagel, line 99: Mostly just a comment because this line was extremely relevant and problematic in the past - the label we are selecting on gets added to our policies [here](https://github.com/defenseunicorns/uds-core/blob/970-all-nodes/src/pepr/operator/controllers/network/generate.ts#L101-L104). We could update that comment line to indicate that it's also use for the `KubeNodes` policies.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,146 @@
...
+      log.warn(
+        `KubeNodes NetworkPolicy ${netPol.metadata?.namespace}/${netPol.metadata?.name} is missing spec.`,
+      );
+      continue;
+    }
+
+    if (netPol.spec.egress) {
+      netPol.spec.egress[0] = netPol.spec.egress[0] || { to: [] };
+      netPol.spec.egress[0].to = newNodes;
+    }
```
mjnagel: Let's make sure we account for ingress here as well, this caught us recently for the KubeAPI policies so you should be able to copy the logic from [there](https://github.com/defenseunicorns/uds-core/blob/970-all-nodes/src/pepr/operator/controllers/network/generators/kubeAPI.ts#L173-L194).

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,146 @@
...
+    );
+
+    try {
+      await K8s(kind.NetworkPolicy).Apply(netPol, { force: true });
+    } catch (err) {
+      let message = err.data?.message || "Unknown error while applying KubeNode network policies";
+      if (UDSConfig.kubeApiCidr) {
+        message +=
+          ", ensure that the KUBEAPI_CIDR override configured for the operator is correct.";
+      }
```
mjnagel: This error message addition is irrelevant in the context of the `KubeNodes` policies, although it does bring up a potential question around this - we may want to add a manual "override" similar to the `UDSConfig.kubeApiCidr` override to allow people to opt out of auto-updates. The usage/explanation of that is documented [here](https://uds.defenseunicorns.com/reference/configuration/uds-networking-configuration/). It doesn't require too many code changes to support if we wanted to add it (should be able to copy most of what was done for KubeAPI).

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,146 @@
+/**
+ * Copyright 2024 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
```
mjnagel, line 4: General comment on this added file - would be great to add jest testing around these functions. We encountered a number of issues with the kubeAPI policy and added testing to cover those functions as a result, would be great to do similar here. You should be able to model after what was done for KubeAPI with Jest mocks to avoid any k8s calls, and most of that should be relatively easy to generate as well.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
mjnagel: (If we do end up adding this manual config override let's make sure we also add to that same doc)

docs/reference/configuration/custom-resources/exemptions-v1alpha1-cr.md
catsby: Removed in https://github.com/defenseunicorns/uds-core/pull/1123, and merged in to this branch to pick up the update.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
catsby: 👍 updated in aa4f8808

src/pepr/operator/controllers/network/generators/kubeNodes.ts
catsby: It made more sense to me (a human) to track the simple IP addresses as `string`'s in the set than to convert it to an `Object` and track those. We'd need to convert the string to the `V1NetworkPolicyPeer` type then add/remove it from the set. I assume Typescript/Javascript sets would support that, but as a human it felt easier to reason about. 

The duplication around the pair of function calls is real though so I refactored it some in [7bea9ca8](https://github.com/defenseunicorns/uds-core/pull/1119/commits/7bea9ca8a5fa6fc2ab673f57e1635ca5525bad91) to address that. `updateKubeNodesNetworkPolicies()` now just works off of the global(?) `nodeSet`.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
catsby: 🤔 I didn't consider that. It might cause churn but I added a `delete()` call to the set if the node is **not** ready. We can adjust it later if needed. 

Thinking though, the target is `KubeNodes`, and not guaranteed to be _ready_ nodes I suppose? Even if right now we're trying to only include the ready ones.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
catsby: 👍 added in 024b5874

src/pepr/operator/controllers/network/generators/kubeNodes.ts
catsby: added in 9b4bc342

src/pepr/operator/controllers/network/generators/kubeNodes.ts
mjnagel: Yeah, especially in this case of metrics I suppose there could be an argument that a slightly unhealthy node is the exact type of node that you'd want to get metrics from (if they were reachable). Now I'm wondering if we should just ignore health altogether...worst case a pod has an extra egress permission to a node that is unhealthy, or partially unhealthy 🤔

src/pepr/operator/controllers/network/generators/kubeNodes.ts
catsby: Support for this was added in 15e6235d , I believe

src/pepr/operator/controllers/network/generators/kubeNodes.ts
catsby: Tests were added in 6921d23b

catsby:
Rebased with main, and I think I've addressed all the feedback

src/pepr/uds-operator-config/values.yaml
```diff
@@ -10,6 +10,7 @@ operator:
   UDS_LOG_LEVEL: "###ZARF_VAR_UDS_LOG_LEVEL###"
   AUTHSERVICE_REDIS_URI: "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
   KUBEAPI_CIDR: ""
+  KUBENODE_CIDRS: ""
```
mjnagel, line 13: Let's make sure we document this with an example in [this doc](https://github.com/defenseunicorns/uds-core/blob/970-all-nodes/docs/reference/configuration/uds-networking-configuration.md), probably a `KubeNodes` section.

src/pepr/operator/index.ts
```diff
@@ -83,3 +92,13 @@ When(UDSPackage)
     log.info("Identity and Authorization layer removed, operator will NOT handle SSO.");
     UDSConfig.isIdentityDeployed = false;
   });
+
+// Watch for changes to the API server EndpointSlice and update the API server CIDR
+if (!UDSConfig.kubeNodeCidrs) {
+  When(a.Node).IsCreatedOrUpdated().Reconcile(updateKubeNodesFromCreateUpdate);
+}
+
+// Watch for changes to the API server EndpointSlice and update the API server CIDR
```
mjnagel: These comments don't match the functionality here (should be `Watch for changes to nodes ...`).

src/pepr/operator/index.ts
```diff
@@ -35,6 +42,8 @@ if (process.env.PEPR_WATCH_MODE === "true" || process.env.PEPR_MODE === "dev") {
   void initAPIServerCIDR();
 }
 
+void initAllNodesTarget();
```
mjnagel: We should wrap this in the same conditional as the one above:
```suggestion
// Pre-populate the Node CIDR list since we are not persisting it
// Note ignore any errors since the watch will still be running hereafter
if (process.env.PEPR_WATCH_MODE === "true" || process.env.PEPR_MODE === "dev") {
  void initAllNodesTarget();
}
```

This ensures that this init only runs on the watcher pod (1/3 of the pepr pods) since it's only relevant there. The admission pods don't control network policies so they don't need this code to run.

src/pepr/operator/index.ts
mjnagel:
```suggestion
// Watch for changes to the Nodes and update the Node CIDR list
if (!UDSConfig.kubeNodeCidrs) {
  When(a.Node).IsCreatedOrUpdated().Reconcile(updateKubeNodesFromCreateUpdate);
}

// Watch for Node deletions and update the Node CIDR list
```

mjnagel (commented):
A few follow-on comments here, the code logic looks good at this point though (comments are on comments/docs). I'll deploy and validate this resolves the prometheus error, wanted to post these comments first though.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
mjnagel: Would be curious on @UnicornChance / @noahpb opinions here...should we consider node health as part of the netpol construction? I very slightly lean towards not considering the health because pods may still want to be able to hit the node even if it's reporting unhealthy?

noahpb: I agree with @mjnagel  - a node might not appear as `Ready` when it has `DiskPressure` or `MemoryPressure`, but workloads could still be running on it. I think we should only cut off network access to the nodes once we're certain it's no longer needed.

catsby: I updated the behavior to only remove nodes if they're being deleted in 1d4f0af8

src/pepr/uds-operator-config/values.yaml
catsby: I added a section on `KubeNodes` in 1d4f0af8

docs/reference/configuration/uds-networking-configuration.md
```diff
@@ -25,6 +25,25 @@ This configuration directs the operator to use the specified CIDR range (`172.0.
...
+packages:
+  - name: uds-core
+    repository: ghcr.io/defenseunicorns/packages/uds/core
+    ref: x.x.x
+    overrides:
+      uds-operator-config:
+        uds-operator-config:
+          values:
+            - path: operator.KUBENODE_CIDRS
+              value: "172.28.0.2/32,172.28.0.3/32,172.28.0./32"
```
mjnagel: missing the last octet on the example here
```suggestion
              value: "172.28.0.2/32,172.28.0.3/32,172.28.0.4/32"
```

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,174 @@
...
+      // Remove managed fields to prevent server-side apply errors
+      netPol.metadata.managedFields = undefined;
+    }
+
+    log.debug(
+      `Updating KubeNodes NetworkPolicy ${netPol.metadata?.namespace}/${netPol.metadata?.name} with new CIDRs.`,
+    );
+
+    try {
+      await K8s(kind.NetworkPolicy).Apply(netPol, { force: true });
```
mjnagel: https://github.com/defenseunicorns/uds-core/blob/970-all-nodes/src/pepr/operator/controllers/network/generators/kubeAPI.ts#L173-L197

Would be good to compare with the existing node list before we make the apply to make sure we aren't unnecessarily calling k8s apply here.

catsby: Changed in 9b7a61c3

mjnagel (dismissed):

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,177 @@
...
+    const nodes = await retryWithDelay(fetchKubernetesNodes, log);
+    nodeSet.clear();
+
+    for (const node of nodes.items) {
+      const ip = getNodeInternalIP(node);
+      if (ip) nodeSet.add(ip);
+    }
+  } catch (err) {
+    log.error("error fetching node IPs:", err);
+  }
```
mjnagel, line 45: Slight modification to this logic - we would want to call `updateKubeNodesNetworkPolicies` after a fresh startup. The most clear place this might matter is if someone switched from using the `UDSConfig.kubeNodeCidrs` to the auto-genned version or vice-versa.
```suggestion
  if (UDSConfig.kubeNodeCidrs) {
    const nodeCidrs = UDSConfig.kubeNodeCidrs.split(",");
    for (const nodeCidr of nodeCidrs) {
      nodeSet.add(nodeCidr);
    }
    await updateKubeNodesNetworkPolicies();
    return;
  }

  try {
    const nodes = await retryWithDelay(fetchKubernetesNodes, log);
    nodeSet.clear();

    for (const node of nodes.items) {
      const ip = getNodeInternalIP(node);
      if (ip) nodeSet.add(ip);
    }
    await updateKubeNodesNetworkPolicies();
  } catch (err) {
    log.error("error fetching node IPs:", err);
  }
```

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,177 @@
...
+ * and populating the nodeSet with their Internal IPs.
+ */
+export async function initAllNodesTarget() {
+  // if a list of CIDRs is defined, use those
+  if (UDSConfig.kubeNodeCidrs) {
+    const nodeCidrs = UDSConfig.kubeNodeCidrs.split(",");
+    for (const nodeCidr of nodeCidrs) {
+      nodeSet.add(nodeCidr);
+    }
+    return;
```
mjnagel: One addition here, we should call an update to the netpols here in case the config value was just added or changed:
```suggestion
    }
    await updateKubeNodesNetworkPolicies();
    return;
```

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  try {
+    const nodes = await retryWithDelay(fetchKubernetesNodes, log);
+    nodeSet.clear();
+
+    for (const node of nodes.items) {
+      const ip = getNodeInternalIP(node);
+      if (ip) nodeSet.add(ip);
+    }
```
mjnagel, line 43: Similar suggestion here - assuming the fetch doesn't error we should update netpols as well.
```suggestion
    }
    await updateKubeNodesNetworkPolicies();
```

mjnagel (approved):

PR #1127, opened 2024-12-16
feat: experimental opt-in classification banner

### LINKED ISSUE #1079: Allow configuring a classification banner on all exposed user interfaces
As a user of UDS Core I have a requirement to mark all my user interfaces with a [classification banner](https://www.astrouxds.com/components/classification-markings/). I would like UDS Core to provide this banner when I request to expose my UI (i.e. via the `expose` field in the Package CR).

This should be configurable for different classification levels at deploy time, and ideally applied at the gateway level to ensure all services are covered.

mjnagel:
## Description

This PR adds an experimental (opt-in) classification banner provided via an envoyfilter. This filter injects html/script code to add a top (and optionally bottom) banner with the desired classification level and the expected colors. This has primarily been validated against Keycloak and Grafana, other UIs may not display perfectly (part of why this is noted as experimental).

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1079

## Type of change

## Steps to Validate

1. Modify the slim-dev bundle to add the below overrides:
```yaml
    overrides:
      istio-controlplane:
        uds-global-istio-config:
          values:
            - path: classificationBanner.text
              value: "UNCLASSIFIED" # Possible values: UNCLASSIFIED, CUI, CONFIDENTIAL, SECRET, TOP SECRET, TOP SECRET//SCI, UNKNOWN
            - path: classificationBanner.addFooter
              value: true
            - path: classificationBanner.enabledHosts
              value:
                - keycloak.admin.{{ .Values.domain }}
                - sso.{{ .Values.domain }}
```
2. Deploy slim-dev: `uds run slim-dev --set flavor=unicorn`
3. Validate that the banner with the expected classification level appears on the Keycloak admin and tenant interfaces and does not overlap with any content.

## Checklist before merging


docs/reference/configuration/uds-optional-features.md
chance-coleman, line 1: Looks good overall, wondering if its worth documenting that the defined classification level determines the color as well? Or directly linking in the doc to the envoy filter used so they could see that.

mjnagel: Yeah, will make those changes.

chance-coleman (approved):
lgtm

Issue #1027, opened 2024-11-20
Make least-privilege exemptions for ztunnel and install-cni pods


mjnagel:
https://github.com/defenseunicorns/uds-core/pull/699 currently has catch-all exemptions for the ztunnel and istio-cni pods. These exemptions should be slimmed down to just what is necessary for running these applications. All exemptions should have adequate descriptions to provide the "why".

sgettys:
This is part of the [ambient PR](https://github.com/defenseunicorns/uds-core/pull/1142) the minimum required exemptions for ztunnel and CNI are part of the ambient update to the istio package

PR #1142, opened 2024-12-20
feat: add ztunnel cni for ambient

sgettys:
## Description

This PR enables Istio Ambient for all of UDS Core. This includes Waypoint for L7 routing. It's still in draft until the FIPS discussion around ztunnel images and until the [CNI ipset fix](https://github.com/istio/istio/pull/53862) is released

## Related Issue

Fixes #

Relates to #

## Type of change

## Steps to Validate

## Checklist before merging


src/istio/common/chart/templates/exemptions.yaml
```diff
@@ -0,0 +1,29 @@
...
+        - RestrictCapabilities
+        - RestrictHostPathWrite
+        - RestrictHostPorts
+        - RestrictProcMount
+        - RestrictSeccomp
+      matcher:
+        namespace: istio-system
+        name: "^.*"
+      title: "istio exemptions"
+      description: "Istio does some privileged things"
```
mjnagel: This will still need some slimming down to specific pods/policies (https://github.com/defenseunicorns/uds-core/issues/1027).

src/istio/common/chart/templates/gateway-api.yaml
```diff
@@ -0,0 +1,10345 @@
+# Copyright 2024 The Kubernetes Authors.
```
mjnagel, line 1: Could we install these CRDs as a `manifest` in the `zarf.yaml` with a remote reference instead of copying them in here? May make it easier to maintain across upgrades.

src/istio/common/chart/templates/exemptions.yaml
sgettys: Yeah that is what I was going to dig into today

src/istio/common/chart/templates/gateway-api.yaml
sgettys: Good idea, what will happen if the CRs are already installed in the cluster though? How can we deal with that? Or just assume that it won't be?

mjnagel: Good q - I think in general it wouldn't be a problem to "re-apply" the CRDs via zarf manifest, but it could cause issues with helm ownership metadata on the CRDs. We could use zarf actions to add metadata (labels/annotations) to the existing CRDs if they exist, or just ignore it and assume they won't already be present on the cluster. Since the gateway CRDs are not a standard part of k8s clusters I think we might be okay to just assume that they aren't already present in the cluster?

sgettys: That was my first thought as well but I wonder if they might come in to some later flavor of k8s as a built in. I don't think we'll run into any clusters that are using these in the wild though so we should be safe to assume they don't exist

src/neuvector/chart/templates/peerauthentication/neuvector-controller-pa.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     matchLabels:
       app: neuvector-controller-pod
   mtls:
-    mode: STRICT
+    mode: PERMISSIVE
```
mjnagel, line 15: I missed this change earlier - what is the behavior with this set to STRICT? The way this behaves it would allow PERMISSIVE on all ports for the controller, I'd hope we could instead just set some ports to PERMISSIVE such as the two below here.

sgettys: I'll test this out and this will be a separate PR when we enable ambient. Going to revert for this PR

mjnagel:
Closing - this can be used for reference but we plan to attack ambient at a more granular level with smaller PRs.

PR #1152, opened 2025-01-06
docs(fix): changed name of policy so policy names were not duplicated

JoeHCQ1:
## Description

Minor fix to docs, fixed duplicated name.

## Type of change

## Checklist before merging


mjnagel (approved):

noahpb:
This appears to be a dup of #973

mjnagel:
Good catch @noahpb - once rebased there were no changes here.

JoeHCQ1:
Ah, I thought I'd merged it too, but the diff was still there when I pushed and compared. Thanks! I'm making sure I don't have anything of value left over as a client effort wraps up.

PR #1175, opened 2025-01-14
feat: optional istio cni ztunnel component

### LINKED ISSUE #1027: Make least-privilege exemptions for ztunnel and install-cni pods

### LINKED ISSUE #1033: Add ztunnel/cni components to Istio
Related to https://github.com/defenseunicorns/uds-core/issues/681

In order to incrementally work towards ambient mesh we should add the ztunnel and cni components to our istio setup. Ideally these should all be on v1.24.0 (when ambient was GA-ed). Definition of done here:
- Ztunnel and install-cni components included in default istio setup
- Automatic detection or documentation on any distro specific configuration for the CNI setup ([example](https://istio.io/latest/docs/ambient/install/platform-prerequisites/#k3d)) - if distro specifics are k3d exclusively this could be handled with bundle overrides?
- Full functionality of these two components in all 3 flavors (and FIPS images for registry1/unicorn flavors)

The scope of this issue does not require any workloads to be switched to ambient, or any support of ambient in the operator - but if those changes are necessary/helpful to validate ztunnel functionality they can be worked here.
```[tasklist]
### Tasks
- [ ] add optional ztunnel and cni components to istio chart
- [ ] make cni setup confirgurable
```

sgettys:
## Description
* optional Istio cni and ztunnel component
* required exemptions for CNI and ztunnel to operate
* CNI config via zarf variables

## Related Issue

Fixes #1033
Fixes #1027 

## Type of change

## Steps to Validate

## Checklist before merging


src/istio/ambient/zarf.yaml
```diff
@@ -0,0 +1,76 @@
...
+    actions:
+      onDeploy:
+        before:
+          - description: "Ensure CNI_CONF_DIR is set"
+            cmd: |
+              if [ \"${ZARF_VAR_CNI_CONF_DIR}\" == \"\" ]; then
+                if ./zarf tools kubectl version -o json 2>/dev/null | ./zarf tools yq '.serverVersion.gitVersion' 2>/dev/null | grep -q "k3s"; then
+                  if ./zarf tools kubectl get nodes -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null | grep -q "k3d"; then
+                    echo "/var/lib/rancher/k3s/agent/etc/cni/net.d"
+                  fi
```
mjnagel: Should this check for k3d here? Based on [the upstream profiles](https://github.com/istio/istio/tree/master/manifests/helm-profiles) I think we'd want `/var/lib/rancher/k3s/agent/etc/cni/net.d` for both k3s and k3d, so might not need the extra node check here?

sgettys: Oh yeah whoops I thought I took that out. Good catch!

sgettys:
Tested on AWS EKS, K3d, and K3s. Ambient component installs without issue

README.md
```diff
@@ -114,7 +113,14 @@ uds run test-single-layer --set LAYER=identity-authorization
 
 Note you can specify the `--set FLAVOR=registry1` flag to test using Iron Bank images instead of the upstream images.
 
+## Istio Ambient
+
+UDS Core now supports deploying Istio in Ambient mode. This feature is still in alpha and is not recommended for production use. To deploy Istio in Ambient mode deploy the optional `istio-ambient` component in the UDS Core Base layer. This will deploy the Istio CNI and ztunnel components but does not yet configure `Core` components to utilize Ambient mesh. This feature is under active development and will be updated in future releases.
```
mjnagel: Should this section be part of a [configuration doc](https://github.com/defenseunicorns/uds-core/tree/feat/istio-cni-ztunnel/docs/reference/configuration) or somewhere else? I don't have a super strong opinion but not sure we need to advertise/note this on the top level readme.

bundles/k3d-standard/uds-bundle.yaml
```diff
@@ -40,6 +40,7 @@ packages:
     ref: 0.34.1
     # x-release-please-end
     optionalComponents:
+      - istio-ambient
```
mjnagel, line 43: Think we chatted on this previously but can't remember where we landed - do we want to add ambient components to the [nightly bundles (eks, rke2, aks)](https://github.com/defenseunicorns/uds-core/tree/feat/istio-cni-ztunnel/.github/bundles) to further validate their healthy deployment?

README.md
mjnagel: Maybe even the change to `docs/reference/UDS Core/prerequisites.md` is sufficient to note ambient without adding anything here?

docs/reference/UDS Core/prerequisites.md
```diff
@@ -66,10 +66,29 @@ done
...
+cniConfDir: /var/lib/rancher/k3s/agent/etc/cni/net.d
+cniBinDir: /opt/cni/bin/
 
-NeuVector historically has functioned best when the host is using cgroup v2. Cgroup v2 is enabled by default on many modern Linux distributions, but you may need to enable it depending on your operating system. Enabling this tends to be OS specific, so you will need to evaluate this for your specific hosts. 
+# All other clusters
+cniConfDir: /etc/cni/net.d
+cniBinDir: /opt/cni/bin/
+```
+
+These values can be overwritten when installing core by setting the `cniConfDir` and `cniBinDir` values in the `istio-ambient` component.
```
mjnagel, line 87: nit: might be nice to provide an example of what an override would look like here. Especially since this should be set via zarf var (or in uds-config file) rather than a bundle override (since our action will look for that zarf var).

src/istio/common/zarf.yaml
```diff
@@ -41,3 +41,7 @@ components:
           - description: "Ensure istio-injection is enabled for Pepr"
             mute: true
             cmd: "./zarf tools kubectl label namespace pepr-system istio-injection=enabled --overwrite"
+          - description: "Cycle Pepr to ensure enrollment in Istio"
+            cmd: |
+              ./zarf tools kubectl rollout restart -n pepr-system deploy/pepr-uds-core-watcher
+              ./zarf tools kubectl rollout restart -n pepr-system deploy/pepr-uds-core
```
mjnagel: Definitely a nit but could we make these commands a bit "less aggressive"? Essentially add a check to see if a sidecar is already present, and only cycle the pod if there isn't one? I'm thinking in particular of upgrade scenarios where Pepr is already deployed and this would just add some lag time and unnecessary cycling of the pods. In particular it would also result in all `Package` CRs being re-reconciled since we currently do that on Pepr startup (optimizations could be made there as well for the future, but for now we do that to re-baseline everything).

src/istio/zarf.yaml
```diff
@@ -53,11 +97,29 @@ components:
...
+      path: ambient
+    charts:
+      - name: cni
+        valuesFiles:
+          - "values/unicorn/cni.yaml"
+      - name: ztunnel
+        valuesFiles:
+          - "values/unicorn/ztunnel.yaml"
+    images:
+      # Chainguard's install-cni-fips is not working right now, issue submitted
```
mjnagel: This comment might have been lingering from previous issues, seems like the image is working now?
```suggestion
```

src/istio/ambient/chart/templates/exemptions.yaml
```diff
@@ -0,0 +1,39 @@
...
+ exemptions:
+   - policies:
+       - DisallowPrivileged # CNI plugin requires NET_ADMIN, NET_RAW, SYS_PTRACE, SYS_ADMIN, and DAC_OVERRIDE capabilities
+       - RequireNonRootUser # CNI plugin requires root
+       - RestrictVolumeTypes # CNI plugin uses 'hostPath' volume type
+       - DisallowHostNamespaces # CNI plugin requires host namespace access for networking
+       - DropAllCapabilities # CNI plugin requires NET_ADMIN, NET_RAW, SYS_PTRACE, SYS_ADMIN, and DAC_OVERRIDE capabilities
+       - RestrictCapabilities # CNI plugin requires NET_ADMIN, NET_RAW, SYS_PTRACE, SYS_ADMIN, and DAC_OVERRIDE capabilities
+       - RestrictHostPathWrite # CNI plugin requires access to write to CNI
+       - RestrictHostPorts # CNI plugin requires access to host ports
```
mjnagel: Reviewing the pod security context I'm not sure all of these are required?
```yaml
    securityContext:
      appArmorProfile:
        type: Unconfined
      capabilities:
        add:
        - NET_ADMIN
        - NET_RAW
        - SYS_PTRACE
        - SYS_ADMIN
        - DAC_OVERRIDE
        drop:
        - ALL
      privileged: false
      runAsGroup: 0
      runAsNonRoot: false
      runAsUser: 0
```

In particular I was able to make these changes:
```suggestion
       - RequireNonRootUser # CNI plugin requires root
       - RestrictVolumeTypes # CNI plugin uses 'hostPath' volume type
       - DisallowHostNamespaces # CNI plugin requires host namespace access for networking
       - RestrictCapabilities # CNI plugin requires NET_ADMIN, NET_RAW, SYS_PTRACE, SYS_ADMIN, and DAC_OVERRIDE capabilities
       - RestrictHostPathWrite # CNI plugin requires access to write to CNI
       - RestrictHostPorts # CNI plugin requires access to host ports
```
since privileged is explicitly set as `false` and we do actually drop all capabilities (although we also add some).

src/istio/ambient/chart/templates/exemptions.yaml
```diff
@@ -0,0 +1,39 @@
...
+     title: "Istio CNI exemptions"
+     description: "Exemptions necessary for Istio CNI to manage network configurations and ensure proper traffic routing"
+   - policies:
+       - DisallowPrivileged # ztunnel requires NET_ADMIN, NET_RAW, and SYS_ADMIN capabilities
+       - RequireNonRootUser # ztunnel requires root
+       - RestrictVolumeTypes # ztunnel uses 'hostPath' volume type
+       - DisallowHostNamespaces # ztunnel requires host namespace access for networking
+       - DropAllCapabilities # ztunnel requires NET_ADMIN, NET_RAW, SYS_PTRACE, SYS_ADMIN, and DAC_OVERRIDE capabilities
+       - RestrictCapabilities # ztunnel requires NET_ADMIN, NET_RAW, SYS_PTRACE, SYS_ADMIN, and DAC_OVERRIDE capabilities
+       - RestrictHostPathWrite # ztunnel requires access to write to CNI
```
mjnagel: Similar comment as above, I didn't need all of these. For ref the securityContext:
```yaml
    securityContext:
      allowPrivilegeEscalation: true
      capabilities:
        add:
        - NET_ADMIN
        - SYS_ADMIN
        - NET_RAW
        drop:
        - ALL
      privileged: false
      readOnlyRootFilesystem: true
      runAsGroup: 1337
      runAsNonRoot: false
      runAsUser: 0
```
I was able to make the following change:
```suggestion
       - DisallowPrivileged # ztunnel runs with allowPrivilegeEscalation
       - RequireNonRootUser # ztunnel requires root
       - RestrictVolumeTypes # ztunnel uses 'hostPath' volume type
       - RestrictCapabilities # ztunnel requires NET_ADMIN, NET_RAW, and SYS_ADMIN capabilities
       - RestrictHostPathWrite # ztunnel requires access to write to CNI
```
and the pod still came up.

bundles/k3d-standard/uds-bundle.yaml
mjnagel: In particular I'm curious on the RKE2 install and if our auto-detection logic could/should be updated to handle RKE2 (I have suspicions based on the shared lineage of k3s).

mjnagel (commented):
In my testing everything seemed to work well on k3d. Given that this is optional I'm not overly concerned about anything here but have a handful of comments.

README.md
sgettys: I'm fine with that! This will come out of here anyway once it's fully switched over

src/istio/ambient/zarf.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - name: CNI_CONF_DIR
+    description: "CNI configuration directory"
+    default: ""
+  - name: CNI_BIN_DIR
+    description: "CNI binary directory"
+    default: ""
+
+components:
+  - name: istio-ambient
+    required: true
```
mjnagel: I think it still works as expected but for consistency sake and peace of mind, maybe this should be false?
```suggestion
    required: false
```

bundles/k3d-standard/uds-bundle.yaml
sgettys: I tested it locally on RKE2 and worked as expected

docs/reference/UDS Core/prerequisites.md
sgettys: Might need an example of what this looks like, the way I'm setting via uds-config.yaml is not working

src/istio/common/zarf.yaml
sgettys: Good call!

docs/reference/UDS Core/prerequisites.md
mjnagel: Within your `uds-config.yaml` file it should look something like this:
```yaml
variables:
  # name of the package in the bundle
  core:
    # zarf var name
    CNI_CONF_DIR: /path/to/cni/conf
    CNI_BIN_DIR: /path/to/cni/bin
```

Even though they are zarf vars, they still get set inside the `variables` section like an `override` would be.

sgettys: Hmm I tried that and it didn't do what I was expecting

src/istio/ambient/zarf.yaml
```diff
@@ -0,0 +1,74 @@
...
+                    echo "/bin/"
+                  else
+                    echo "/var/lib/rancher/k3s/data/cni"
+                  fi
+                else
+                  echo "/opt/cni/bin"
+                fi
+              fi
+            setVariables:
+              - name: CNI_BIN_DIR
```
mjnagel:
```suggestion
          - description: "Ensure CNI_CONF_DIR is set"
            cmd: |
              if [ "${ZARF_VAR_CNI_CONF_DIR}" = "" ]; then
                if ./zarf tools kubectl version -o json 2>/dev/null | ./zarf tools yq '.serverVersion.gitVersion' 2>/dev/null | grep -q "k3s"; then
                  echo "/var/lib/rancher/k3s/agent/etc/cni/net.d"
                else
                  echo "/etc/cni/net.d"
                fi
              else
                echo "${ZARF_VAR_CNI_CONF_DIR}"
              fi
            setVariables:
              - name: CNI_CONF_DIR
          - description: "Ensure CNI_BIN_DIR is set"
            cmd: |
              if [ "${ZARF_VAR_CNI_BIN_DIR}" = "" ]; then
                if ./zarf tools kubectl version -o json 2>/dev/null | ./zarf tools yq '.serverVersion.gitVersion' 2>/dev/null | grep -q "k3s"; then
                  if ./zarf tools kubectl get nodes -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null | grep -q "k3d"; then
                    echo "/bin/"
                  else
                    echo "/var/lib/rancher/k3s/data/cni"
                  fi
                else
                  echo "/opt/cni/bin"
                fi
              else
                echo "${ZARF_VAR_CNI_BIN_DIR}"
              fi
            setVariables:
              - name: CNI_BIN_DIR
```

Discussed a bit outside gh but these actions were missing the `else` clauses to echo out the zarf var (to set the same zarf var). I also nuked the escape syntax around the quotes because it didn't seem necessary but might've missed something there.

src/istio/zarf.yaml
```diff
@@ -39,11 +56,38 @@ components:
...
+          - "values/upstream/ztunnel.yaml"
+    images:
+      # Registry1 ztunnel image is not working right now
+      # https://repo1.dso.mil/dsop/tetrate/istio/1.24/ztunnel/-/issues/9
+      #- registry1.dso.mil/ironbank/tetrate/istio/ztunnel:1.24.2-tetratefips0
+      - docker.io/istio/ztunnel:1.24.2-distroless
+      # Registry1s install-cni-fips is not working right now
+      # https://repo1.dso.mil/dsop/tetrate/istio/1.24/install-cni/-/issues/11
+      #- registry1.dso.mil/ironbank/opensource/istio/install-cni:1.24.2
+      - docker.io/istio/install-cni:1.24.2-distroless
```
mjnagel, line 90: Might be good to make sure we have an issue tracking these todos to switch over - esp since the original issues got closed out (not sure all the problems are solved though!)

docs/reference/UDS Core/prerequisites.md
```diff
@@ -66,10 +66,44 @@ done
...
+
+To set these values add the following to the `uds-config.yaml` file:
 
-NeuVector historically has functioned best when the host is using cgroup v2. Cgroup v2 is enabled by default on many modern Linux distributions, but you may need to enable it depending on your operating system. Enabling this tends to be OS specific, so you will need to evaluate this for your specific hosts. 
+```yaml
+variables:
+  core:
+    cniConfDir: /etc/cni/net.d
+    cniBinDir: /opt/cni/bin
```
mjnagel: Needs more snake_case

sgettys: Now with 2x as much snake_case

mjnagel (approved):

PR #1186, opened 2025-01-15
feat: add logic to handle updates to operator config

### LINKED ISSUE #1130: Authservice config (ca cert/redis) only read at install
The [UDS Operator config](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/uds-operator-config/values.yaml) includes two config values for authservice, CA Cert and Redis. This config secret is provided to Pepr pods as env vars but changes to values are not properly handled. This particular issue is focused on the two authservice values.

In order to properly handle modifications to this configuration:
1. Pepr must update the [UDSConfig](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/config.ts#L27) when the secret changes. This could be done with a watch on the secret or a checksum on the pepr pods to cycle them when the secret updates.
2. Code around the authservice config must handle adding/changing the CA cert and redis configuration. Currently these values are only added as part of [buildInitialSecret](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/authservice/config.ts#L80) meaning this only works when the `authservice-uds` secret does not exist. We should ensure that when the config is changed we also update the secret and cycle authservice.

Definition of done: Slim-dev can be deployed without configuration for redis/ca-cert. After adding either or both of these parameters Pepr properly updates the `authservice-uds` secret and cycles the Authservice pod to read the new config.

Note: There are other "upgrade" that we do not account for in the config like changing domain. These should not be part of this issue but are good to track as known limitations as well.

mjnagel:
## Description

This PR adds logic to handle updates to the operator config, specifically managed via the operator secret. In particular this adds:
- A watch on the operator config secret
- A function to manage changes to that secret
- Logic to specifically handle additions/changes/deletions of the authservice config (redis and ca cert)
- Logic to handle updates to the CIDR ranges for nodes/kubeapi
- Stubbed out logic to handle updates to domain and admin domain (this would be a larger effort to fully implement so leaving a TODO on this)

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1130

## Type of change

## Steps to Validate

<details>

```console
# Deploy slim-dev with unicorn flavor (or other flavor)

# Validate an update to redis URI takes effect
## Edit the operator secret
kubectl edit secret -n pepr-system uds-operator-config
## Add `cmVkaXM6Ly90ZXN0` as the value for `AUTHSERVICE_REDIS_URI` (this is base64 encoded `redis://test`)
## Validate that Pepr watcher logs indicate it cycled authservice
kubectl logs -n pepr-system deploy/pepr-uds-core-watcher | grep "Updating Authservice secret"
## Validate that Authservice cycled and is trying to use our fake redis url
kubectl logs -n authservice deploy/authservice --all-pods # You should see `redis-url="redis://test"`
## Edit the operator secret
kubectl edit secret -n pepr-system uds-operator-config
## Remove the `AUTHSERVICE_REDIS_URI` value (change to empty string)
## Validate that Pepr watcher logs indicate it cycled authservice
kubectl logs -n pepr-system deploy/pepr-uds-core-watcher | grep "Updating Authservice secret"
## Validate that Authservice cycled and is no longer using redis
kubectl logs -n authservice deploy/authservice --all-pods # It shouldn't have redis listed

# You could redo these same steps with a CA Cert value, just make sure that the value you use is "double" base64 encoded since we expect it to be base64 encoded, and then the secret requires it to be base64 encoded as well

# Validate updates to the CIDR values
## Deploy monitoring layer to add some test resources
uds run test:single-layer --set LAYER=monitoring --set FLAVOR=unicorn
## Edit the operator secret
kubectl edit secret -n pepr-system uds-operator-config
## Set both `KUBEAPI_CIDR` and `KUBENODE_CIDRS` to `MTkyLjE2OC4wLjEvMzI=` (this is `192.168.0.1/32` base64 encoded)
## Validate that Pepr watcher logs indicate it updated network policies
kubectl logs -n pepr-system deploy/pepr-uds-core-watcher | grep "Updating KubeNodes"
kubectl logs -n pepr-system deploy/pepr-uds-core-watcher | grep "Updating KubeAPI"
## Validate the network policies actually have our new IP (`192.168.0.1/32`)
kubectl get networkpolicy -n monitoring allow-prometheus-stack-egress-kube-prometheus-stack-admission-create-kubeapi -o yaml
kubectl get networkpolicy -n monitoring allow-prometheus-stack-egress-metrics-scraping-of-kube-nodes -o yaml
```

</details>

## Checklist before merging


src/pepr/operator/controllers/config/config.ts
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+    UDSConfig.caCert = decodedConfigData.UDS_CA_CERT;
+    UDSConfig.authserviceRedisUri = decodedConfigData.AUTHSERVICE_REDIS_URI;
+
+    // Account for undefined or placeholder values (dev mode)
+    if (
+      !UDSConfig.authserviceRedisUri ||
+      UDSConfig.authserviceRedisUri === "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
+    ) {
+      UDSConfig.caCert = "";
```
noahpb: Should this be `UDSConfig.authserviceRedisUri`?

mjnagel: Good catch, fixed!

mjnagel:
@noahpb identified an issue where empty strings weren't being handled properly for the redis uri and ca cert, made some modifications and added tests to make sure those scenarios are properly handled. Should be g2g again for review.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -0,0 +1,111 @@
...
+  ) {
+    UDSConfig.domain = decodedConfigData.UDS_DOMAIN;
+    UDSConfig.adminDomain = decodedConfigData.UDS_ADMIN_DOMAIN;
+    if (!UDSConfig.domain || UDSConfig.domain === "###ZARF_VAR_DOMAIN###") {
+      UDSConfig.domain = "uds.dev";
+    }
+    if (!UDSConfig.adminDomain || UDSConfig.adminDomain === "###ZARF_VAR_ADMIN_DOMAIN###") {
+      UDSConfig.adminDomain = `admin.${UDSConfig.domain}`;
+    }
+    // todo: Add logic to handle domain changes and update across virtualservices, authservice config, etc
```
sgettys, line 104: Create an issue for this todo?

sgettys (approved):

src/pepr/operator/controllers/config/config.ts
mjnagel: https://github.com/defenseunicorns/uds-core/issues/1220

noahpb (approved):

PR #1201, opened 2025-01-16
fix: loki schema config update

### LINKED ISSUE #782: Update to TSDB by default for Loki (and latest schema version)
# Description

TSDB has been the recommended store for Loki for several releases now. UDS Core should update to use it by default in its [schemaConfig](https://github.com/defenseunicorns/uds-core/blob/main/src/loki/values/values.yaml#L42-L50). We should also upgrade to the latest version of the schema at this time this issue is picked up (v13 as of this writing).

In order to "cut-over" between schemas, Loki requires that dates be set to determine when to switch over (see [here](https://grafana.com/docs/loki/latest/operations/storage/schema/#changing-the-schema)). To effectively set this "by default" in core, we will need a mechanism to capture the user's upgrade date and switch the schema in the future (2 days). A few options have been explored for this:
- Use helm functions + secret lookups to manage schema change dates
- Use Pepr mutations to modify the loki config and set schema change dates
- Use a combination of the two above approaches - have pepr manage a configmap with dates + use helm lookups to query those dates

# Definition of Done

- TSDB is used as the default store for Loki in UDS Core
- Upgrading properly handles the switch from BoltDB to TSDB

# Additional Context

https://grafana.com/docs/loki/latest/operations/storage/tsdb/

https://lokidex.com/posts/tsdb/

chance-coleman:
## Description
TSDB has been the recommended store for Loki for several releases now. This set the default schemaConfig to TSDB and managed upgrading current clusters to TSDB. 

Includes unit tests for the necessary logic utilities.

## Related Issue

Fixes #782 

## Type of change

## Steps to Validate
- deploying uds core should result in a loki secret ( name: Loki, namespace: Loki ) that contains the config.yaml for configuring loki. That secret should have a schema config entry for TSDB that should start being used 2 days in the future.
- upgrading an existing cluster that has not been upgraded should upgrade that secret to use the TSDB two days in the future
- upgrading an existing cluster that already has the TSDB configured will not trigger the update again ( checks for annotation )

## Checklist before merging


chance-coleman:
closing in favor of https://github.com/defenseunicorns/uds-core/pull/1224

PR #1216, opened 2025-01-22
chore: add docs for layer selection

Racer159:
## Description

Adds some documentation for UDS Layer Selection and Pre-req packages.

## Related Issue

Fixes #N/A

## Type of change

## Checklist before merging


chance-coleman:
1. Can you upload the drawio file as well to the diagrams folder for future updates?
2. @mjnagel if we want to keep some consistency with the other diagrams that we just reworked, probably should update the color scheme. Also fine doing this myself once the drawio file is uploaded.

mjnagel:
@UnicornChance yeah agreed - we should try and keep color schemes aligned going forward.

@Racer159 maybe broader question on the diagram - is there a reason it needs to even be a diagram? I don't have a strong objection, but given that it's essentially a colored table (no diagram elements where layout/flow need to be consider) then would we be better served just making it in markdown?

Racer159:
> @UnicornChance yeah agreed - we should try and keep color schemes aligned going forward.
> 
> @Racer159 maybe broader question on the diagram - is there a reason it needs to even be a diagram? I don't have a strong objection, but given that it's essentially a colored table (no diagram elements where layout/flow need to be consider) then would we be better served just making it in markdown?

I can definitely make it a table - it just started as a diagram initially since I thought it might be more - I'll go ahead and massage it into a table

chance-coleman (dismissed):
thanks for making this a markdown table

docs/reference/UDS Core/functional-layers.md
```diff
@@ -46,3 +46,37 @@ packages:
...
     ref: 0.29.0-upstream
 ```
+
+## Layer Selection
+
+Layer selection will always be deployment-specific but below are guidelines for what layers to consider for your deployment.  The *required* layer guidance follows the [Big Bang Conformant Stack](https://repo1.dso.mil/big-bang/product/bbtoc/-/blob/master/policy/conformance.md?ref_type=heads) though if you are not bound by that document you may need to tweak them for your needs - UDS Core Base is the only *technical* layer required to install most UDS Packages.
+
+| Layer | Selection Criteria |
+|-------|--------------------|
+| **UDS Add-ons*** |
```
mjnagel: TBD on naming here, I think add-ons makes sense but not sure if it might conflict with other things we are calling add-ons? Cc @austenbryan

docs/reference/UDS Core/functional-layers.md
```diff
@@ -46,3 +46,37 @@ packages:
     repository: ghcr.io/defenseunicorns/packages/uds/core-backup-restore
     ref: 0.29.0-upstream
 ```
+
+## Layer Selection
+
+Layer selection will always be deployment-specific but below are guidelines for what layers to consider for your deployment.  The *required* layer guidance follows the [Big Bang Conformant Stack](https://repo1.dso.mil/big-bang/product/bbtoc/-/blob/master/policy/conformance.md?ref_type=heads) though if you are not bound by that document you may need to tweak them for your needs - UDS Core Base is the only *technical* layer required to install most UDS Packages.
```
mjnagel: On the last line that's mostly accurate, not sure if it's worth calling out that some layers require identity (monitoring and runtime-security) and logging provides no UI without monitoring (maybe not necessary to say, still an API provided?).

Racer159: Made some changes to notes that the monitoring and runtime security layers need IDAM and reworded logging / monitoring to make it more clear that one was "backend" and the other "frontend"

Racer159:
we should also try to define "UDS Edge" in here as well if we can.

docs/reference/UDS Core/functional-layers.md
```diff
@@ -46,3 +46,41 @@ packages:
...
+
+| Layer | Selection Criteria |
+|-------|--------------------|
+| **UDS Add-ons*** |
+| UDS Runtime      | Provides a common operating picture for a Kubernetes cluster and UDS deployments <br/> *(install if you would like to have an easy-to-use window into your cluster/deployments)* |
+| UDS Security Hub | Tracks CVE / vulnerability postures across UDS deployments <br/> *(install if you would like to more easily track vulnerabilities to meet compliance requirements)* |
```
mjnagel: Should this be removed? To my knowledge security hub isn't actually deployable today and I don't think there are plans to make it deployable with the future of a Registry UI.

Racer159: Yeah I need to update other things (since runtime is also UDS UI now)

docs/reference/UDS Core/functional-layers.md
```diff
@@ -46,3 +46,41 @@ packages:
     repository: ghcr.io/defenseunicorns/packages/uds/core-backup-restore
     ref: 0.29.0-upstream
 ```
+
+## Layer Selection
+
+Layer selection will always be deployment-specific but below are guidelines for what layers to consider for your deployment.  The layers marked with a four pointed star (🟄) are those needed to follow the [Big Bang Conformant Stack](https://repo1.dso.mil/big-bang/product/bbtoc/-/blob/master/policy/conformance.md?ref_type=heads) though if you are not bound by that document, UDS Core Base is the only *technical* layer required to install most UDS Packages.
```
mjnagel: Can we pick a different symbol here (for the four pointed star)? On GitHub and on the docs site this symbol shows as 🟄 (included screenshot in case this is machine/font specific).

Screenshot 2025-03-21 at 1 50 18 PM

docs/reference/UDS Core/functional-layers.md
```diff
@@ -46,3 +46,41 @@ packages:
...
+
+| Layer | Selection Criteria |
+|-------|--------------------|
+| **UDS Add-ons*** |
+| UDS UI           | Provides a common operating picture for a Kubernetes cluster and UDS deployments <br/> *(install if you would like to have an easy-to-use window into your cluster/deployments)* |
+| UDS Registry     | Provides a storage location for UDS components and mission applications <br/> *(install if you would like to be able to easily store and view the software available in your environment)* |
+| UDS Remote Agent | Allows for more advanced remote cluster management / deployment <br/> *(install if you would like to manage UDS deployments from more advanced clients than UDS CLI)* |
```
mjnagel: My leaning is to include this and the prerequisites _after_ the core layers, maybe even as separate tables - but open to other opinions.

docs/reference/UDS Core/functional-layers.md
```diff
@@ -46,3 +46,41 @@ packages:
...
+| **UDS Core Pre-Requisites**‡ |
+| UDS Package Minio Operator | Provides storage for Loki (Logging) and Velero (Backup and Restore) <br/> *(install after core base but before logging/backup and restore if selected)* |
+| UDS Package MetalLB        | Provides a simple LoadBalancer implementation <br/> *(install after Zarf init and before UDS Core Base)* |
+
+:::note
+*UDS Add-ons are not part of the open-source platform but are also not required to maintain / operate a UDS deployment.  They provide additional functionality to streamline the deployment, monitoring, and management of the deployment for the given organization.
+:::
+
+:::note
+†The Monitoring and Runtime Security layers provide user login and therefore require the Identity and Authorization layer
```
mjnagel:
```suggestion
†The Monitoring and Runtime Security layers provide user login and therefore require the Identity and Authorization layer.
```

docs/reference/UDS Core/functional-layers.md
mjnagel: One additional reason to split these into separate tables - the 4 back to back note blocks could be split up and placed after each separate table. Then we wouldn't need any symbols (*,†,††,‡) except for the two core notes - and could potentially use one of those unused symbols to denote BBCS requirements.

My suggestion would be ordering as prereqs, then core layers, then addons - each as separate tables with their individual notes underneath.

docs/reference/UDS Core/functional-layers.md
```diff
@@ -46,3 +46,41 @@ packages:
...
+| Backup and Restore         | Allows volumes and k8s objects to be backed up and restored <br/> *(install if deployment provides critical data or must maintain state)* |
+| Identity and Authorization | Provides authentication and authorization functionality <br/>*(install if deployment requires an auth mechanism (i.e. direct user login))*  |
+| Logging🟄                   | Provides backend log storage and log shipping capabilities <br/> *(install if the deployment requires log aggregation and shipping)* |
+| Metrics Server††🟄          | Provides metrics collection capabilities (req of UDS Runtime) <br/> *(install if the cluster does not provide its own metrics server)* |
+| Base🟄                      | Provides the base for all other functional layers <br/> *(required for all "UDS" deployments and all other functional layers)* |
+| **UDS Core Pre-Requisites**‡ |
+| UDS Package Minio Operator | Provides storage for Loki (Logging) and Velero (Backup and Restore) <br/> *(install after core base but before logging/backup and restore if selected)* |
```
mjnagel: 2 nits:
- might leave off the `loki` and `velero` names since we don't refer to those in the above layers (or at least invert so the layer name is first and app is in parenthesis)
- would we want to note/encourage someone to use minio for backup storage? I'm not super familiar with what this might look like on-prem/on-edge, would it be normal to backup to minio and then backup minio itself somewhere else?

mjnagel (approved):

Issue #1212, opened 2025-01-20
Keycloak deployments gets OOM Killed when using Kernel 6.12


slaskawi:
Upstream indicates the Keycloak heap workaround is fixed, so the temporary workaround from uds-core issue 1212 can be removed. This refers to the `JAVA_OPTS_KC_HEAP` environment variable and is expected to be a simple revert.

This is part of the bundles and can be deferred for now.

Relevant upstream context: [keycloak#36609](<https://github.com/keycloak/keycloak/issues/36609>).

mjnagel:
This should be resolved/able to be worked around at this point by using the documented override (see second code block under the [Keycloak section here](https://uds.defenseunicorns.com/reference/configuration/resource-configuration-and-ha/#keycloak)). For the dev/demo bundles this override is present by default.

@slaskawi would it be good to keep this issue open to track the upstream fix and removing this workaround at some point? Or maybe open a follow-on issue for that?

PR #1218, opened 2025-01-23
fix: add Keycloak workaround for Kernels 6.12+

slaskawi:
## Description

This Pull Request introduces a new `env` setting for Keycloak that enables Mission Heroes to specify additional Environment Variables. This setting can be used for Kernel 6.12+ workaround.

## More context

Linux Kernel 6.12 changed the way the missing CGroups v1 are reported. This caused any Java applications to interpret container limits incorrectly (more on this in #1212). 

As clarified in Cloud Native Slack with the Keycloak Team, the Red Hat JVM Team will backport all required fixes into lower Java versions. However, the timeline for this remains unknown and it is very likely that we may need this workaround at some point.

Once a patched JVM is out there and used in Keycloak container images, we will revert this change.

## Related Issue

Relates to #1212

## Type of change

## Steps to Validate

The easiest way to test this feature is to add this part of the override to the `k3d-slim-dev/uds-bundle.yaml`:

```yaml
            - path: env
              value:
                - name: JAVA_OPTS_KC_HEAP
                  value: "-XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G"
```

Then, call `uds run slim-dev`. Once Keycloak is up and running, check its Pod and the `JAVA_OPTS_KC_HEAP` Environment Variable should be there.

This Pull Request has been checked against regressions in [this](https://github.com/defenseunicorns/uds-core/actions/runs/13198583779/job/36845332819?pr=1218) Github Actions run. This patch contained hardcoded `JAVA_OPTS_KC_HEAP` set to `-XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G` and it didn't cause any test failures. 

## Checklist before merging


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -174,6 +174,10 @@ spec:
               value: "-Dcom.redhat.fips=true"
             {{- end }}
           {{- end }}
+          {{- if or .Values.workarounds.kernel6_12 }}
```
eddiezane: Please add a note that this value must match the value in the helm chart.

Here as well https://github.com/defenseunicorns/uds-core/blob/098ef3df502bbebf3ec3c57d6f3fede825877540/src/keycloak/chart/values.yaml#L130-L137

eddiezane (commented):
Thanks for adding this!

I manually tested this and confirmed it worked.

docs/reference/configuration/resource-configuration-and-ha.md
```diff
@@ -144,6 +144,8 @@ packages:
               path: postgresql.password
 ```
 
+When running Keycloak on new Kernels 6.12+, it is also necessary to specify the `KEYCLOAK_KERNEL_6_12_MODE` setting and set it `true`. This is a temporary workaround for [this](https://github.com/keycloak/keycloak/issues/36609) Keycloak issue. Once addressed, this setting will be removed.
```
mjnagel: We may want to adjust this note to provide an example of the bundle `overrides` block that would be necessary. For the dev/demo bundles this value is setup but for an end user making their own bundle they would need to set up overrides to match (with either values or variables) in order to have this set properly.

src/keycloak/chart/templates/statefulset.yaml
mjnagel: The templating could get particularly nasty but I think we could potentially "convert" the `resources.limits.memory` value to the format/syntax expected by the Java arg and dodge the need for the end user to set both. Helm template could look something like this (I haven't tested and not sure if this accurately captures _all_ possible k8s options for passing in memory limits):
```
{{- define "convertMemory" -}}
{{- $memory := . | quote -}}
{{- if hasSuffix "Gi" $memory -}}
    {{- printf "%sG" (trimSuffix "Gi" $memory) }}
{{- else if hasSuffix "Mi" $memory -}}
    {{- printf "%sM" (trimSuffix "Mi" $memory) }}
{{- else -}}
    {{- $memory }}  # Assume raw bytes / already correct (i.e. already using G, M, K)
{{- end -}}
{{- end -}}
```

(usage would be `-XX:MaxRAM={{ include "convertMemory" .Values.resources.limits.memory }}`)

If we can't do that or it seems too clunky we will want to just expose another value `workarounds.kernel6_12.memoryLimit` (and maybe switch `workarounds.kernel6_12` to `workarounds.kernel6_12.enabled`) and then document that value and note that it must be in line with the resource limit to work properly. Given that this is a (hopefully) temporary workaround, I think that would be okay to require as an additional value, just defaulting to 1G the same as the resource limit.

mjnagel (commented):
Appreciate the work on this! Just have one small comment on the docs note + a suggestion/thought on @eddiezane's comment.

src/keycloak/chart/templates/statefulset.yaml
slaskawi: I did a little bit of research and I think we need this function to accommodate a few more scenarios:

1. If we set the memory limits to `1G`, this essentially means we'll configure CGroups to allow `1,000 MB` max allocation. Passing `-XX:MaxRAM=1G` will set it to `1,024 MB` ([link to the GC configuration header file](https://github.com/openjdk/jdk/blob/master/src/hotspot/share/gc/shared/gc_globals.hpp#L275), [link to the GC configuration on x86](https://github.com/openjdk/jdk/blob/master/src/hotspot/cpu/x86/c1_globals_x86.hpp#L55), [link to calculations in utilities](https://github.com/openjdk/jdk/blob/master/src/hotspot/share/utilities/globalDefinitions.hpp#L307C1-L309C39)). This might cause overcommiting memory and essentially going OOM. The situation gets worse with larger memory values. 

2. The Quantity in Kubernetes can also use the exponent notation (see [the code](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L58)), for example `129e6`. 

I think I'll just error on the second scenario as it's very unlikely someone will use this for specifying memory. But the first one must be addressed.

mjnagel: Thanks for digging into this - definitely also keep in mind that we could add a separate value for the java memory limit here if the templating is getting tough to keep this in sync automatically.

docs/reference/configuration/resource-configuration-and-ha.md
slaskawi: Added a meaningful example.

slaskawi:
@mjnagel @eddiezane Thanks for the review!

I decided to change the approach and introduce a way to introduce non-standard environment variables in the Keycloak Helm Template - it's called `env`. With the new setting, you can leverage the Bundle Overrides and pass the `JAVA_OPTS_KC_HEAP`. I find this approach much more flexible and we could potentially leverage it for doing non-standard things. 

In the documentation, I described how to properly use this setting to bypass the Kernel 6.12+ issue.

src/keycloak/chart/README.md
```diff
@@ -44,3 +44,9 @@ We would have to truncate the chart's fullname to six characters because pods ge
 Using a StatefulSet allows us to truncate to 20 characters leaving room for up to 99 replicas, which is much better.
 Additionally, we get stable values for `jboss.node.name` which can be advantageous for cluster discovery.
 The headless service that governs the StatefulSet is used for DNS discovery via DNS_PING.
+
+## Workarounds
+
+### Kernel 6.12+
+
+When running Keycloak on new Kernels 6.12+, it is also necessary to specify the `KEYCLOAK_KERNEL_6_12_MODE` setting and set it `true`. This is a temporary workaround for [this](https://github.com/keycloak/keycloak/issues/36609) Keycloak issue. Once addressed, this setting will be removed.s
```
mjnagel: Given the change to just allow for ENV overrides we should update this note to indicate the new value for env and how to set java opts for the issue in 6.12+.

slaskawi: Thanks for spotting this! I forgot to revert this file.

mjnagel (dismissed):
LGTM at this point! Last question - should we maybe add the override of `env` by default to the dev/demo bundles? I think since we don't expose resources as configurable to the end user in those bundles we could just hardcode the 1G override you had initially. That would allow @eddiezane and others to deploy those bundles without needing to think about the values.

slaskawi:
> LGTM at this point! Last question - should we maybe add the override of env by default to the dev/demo bundles? I think since we don't expose resources as configurable to the end user in those bundles we could just hardcode the 1G override you had initially. That would allow @eddiezane and others to deploy those bundles without needing to think about the values.

Done! Thanks for suggesting this!

mjnagel (approved):

PR #1221, opened 2025-01-23
docs: add documentation on metrics/dashboards for apps

mjnagel:
## Description

Adds/modifies a docs page detailing how metrics can be collected for an additional application (outside core) and how to add on additional dashboards.

This doc is partially being added in response to a support thread requesting info on how to group dashboards within uds core when adding new ones on. Related to that grouping setup, I also added support for annotations on the loki dashboard configmap.

I also moved some of the information around our monitoring setup and why we made the choices we did into a new dev doc for this purpose. While it still has valuable information, it's not as important for the end user to know/understand unless they wanted to dig into why we chose our approach (in which case the dev docs are a good place for them to be looking).

## Related Issue

N/A

## Type of change

## Steps to Validate

If you want to test the grouping, that can be done by making a new bundle (or modifying the standard bundle locally) and confirming that all the core dashboards end up grouped together.

## Checklist before merging

### DIFF
--- a/docs/reference/configuration/uds-monitoring-metrics.md
+++ b/docs/reference/configuration/uds-monitoring-metrics.md
@@ -2,23 +2,11 @@
 title: Monitoring and Metrics
 ---
 
-UDS Core leverages Pepr to handle setup of Prometheus scraping metrics endpoints, with the particular configuration necessary to work in a STRICT mTLS (Istio) environment. We handle this via a default scrapeClass in prometheus to add the istio certs. When a monitor needs to be exempt from that tlsConfig a mutation is performed to leverage a plain scrape class without istio certs.
+UDS Core deploys Prometheus and Grafana to provide metrics collection and dashboarding. Out of the box all applications in UDS Core will have their metrics collected by Prometheus, with some default dashboards present in Grafana for viewing this data. This document primarily focuses on the integrations and options provided for extending this to monitor any additional applications you would like to deploy.
 
-## TLS Configuration Setup
+## Capturing Metrics
 
-Generally it is beneficial to use service and pod monitor resources from existing helm charts where possible as these may have more advanced configuration and options. The UDS monitoring setup ensures that all monitoring resources use a default [`scrapeClass`](https://github.com/prometheus-operator/prometheus-operator/blob/v0.75.1/Documentation/api.md#monitoring.coreos.com/v1.ScrapeClass) configured in Prometheus to handle the necessary `tlsConfig` setup for metrics to work in STRICT Istio mTLS environments (the `scheme` is also mutated to `https` on individual monitor endpoints, see [this doc](https://istio.io/latest/docs/ops/integrations/prometheus/#tls-settings) for details). This setup is the default configuration but individual monitors can opt out of this config in 3 different ways:
-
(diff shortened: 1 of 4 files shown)


docs/reference/configuration/uds-monitoring-metrics.md
```diff
@@ -58,21 +46,72 @@ spec:
...
+
+## Adding Dashboards
 
-This spec intentionally does not support all options available with a `PodMonitor` or `ServiceMonitor`. While we may add additional fields in the future, we do not want to simply rebuild these specs since we are handling the complexities of Istio mTLS metrics. The current subset of spec options is based on the common needs seen in most environments.
+Grafana within UDS Core is configured with [a sidecar](https://github.com/grafana/helm-charts/blob/6eecb003569dc41a494d21893b8ecb3e8a9741a0/charts/grafana/values.yaml#L926-L928) that will watch for new dashboards added via configmaps or secrets and load them into Grafana dynamically. In order to have your dashboard added the configmap or secret must be labelled with `grafana_dashboard: "1"`, which is used by the sidecar to watch and collect new dashboards.
 
-## Notes on Alternative Approaches
+Your configmap/secret must have a data key named `<dashboard_file_name>.json`, with a multi-line string of the dashboard json as the value. See the below example for a basic dashboard created this way:
```
sgettys: We can use .Files.Get to populate this too right? Makes managing dashboards easier

docs/reference/configuration/uds-monitoring-metrics.md
```diff
@@ -58,21 +46,72 @@ spec:
...
+            # This value adds an annotation to the loki dashboards to specify that they should be grouped under a `uds-core` folder
+            - path: dashboardAnnotations
+              value:
+                grafana_folder: "uds-core"
+```
 
-Another alternative approach would be to use a pod selector and port only. We would then generate both a service and servicemonitor, giving us full control of the port names and selectors. This seems like a viable path, but does add an extra resource for us to generate and manage. There could be unknown side effects of generating services that could clash with other services (particularly with istio endpoints). This would otherwise be a relative straightforward approach and is worth evaluating again if we want to simplify the spec later on.
+:::note
+If using this configuration, any dashboards without a `grafana_folder` annotation will still be loaded in Grafana, but will not be grouped (they will appear at the top level outside of any folders). Also note that new dashboards in UDS Core may also need to be overridden to add the folder annotation, this example represents the current set of dashboards deployed by default.
```
sgettys, line 136: Maybe an explicit call out that dashboards that are managed out of UDS Core can also be grouped into separate folders with the dashboard_folder annotation. Update the example above?

sgettys (dismissed):

docs/reference/configuration/uds-monitoring-metrics.md
mjnagel: Yeah that could be a good addition to the example here with a link to helm docs, will make that change.

docs/reference/configuration/uds-monitoring-metrics.md
mjnagel: Updated, great callout.

docs/reference/configuration/uds-monitoring-metrics.md
mjnagel: Updated!

sgettys (approved):
LGTM!

PR #1223, opened 2025-01-24
fix: add healthz port to neuvector services

mjnagel:
## Description

Issue reported in support channel. After review it seems like Istio protocol detection was causing the problem and dropping some traffic. This adds proper protocol detection/listeners on all the neuvector services and adds a validation to ensure jobs complete in CI.

## Related Issue

N/A

## Type of change

## Steps to Validate

Can be tested by running an upgrade from 0.34.0 -> this branch.

## Checklist before merging

### DIFF
--- a/src/neuvector/tasks.yaml
+++ b/src/neuvector/tasks.yaml
@@ -41,6 +41,11 @@ tasks:
             protocol: https
             address: neuvector.admin.uds.dev
             code: 200
+      - description: Wait for NeuVector Job Completion
+        cmd: |
+          if kubectl get jobs -n neuvector --no-headers 2>/dev/null | grep -q .; then
+            kubectl wait --for=condition=complete job --all -n neuvector --timeout=5m
+          fi
 
   - name: e2e-test
     actions:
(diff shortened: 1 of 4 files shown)


src/neuvector/tasks.yaml
```diff
@@ -41,6 +41,11 @@ tasks:
             protocol: https
             address: neuvector.admin.uds.dev
             code: 200
+      - description: Wait for NeuVector Job Completion
+        cmd: |
+          if kubectl get jobs -n neuvector --no-headers 2>/dev/null | grep -q .; then
```
sgettys: ./zarf tools kubectl?

sgettys (dismissed):

src/neuvector/tasks.yaml
mjnagel: ty yep.

chance-coleman (approved):

src/neuvector/tasks.yaml
mjnagel: Updated to `./uds zarf tools kubectl` based on usage in other tasks currently. Not a huge issue either way since this is just validations and not in the `zarf.yaml` but this is more portable 🚀

sgettys (approved):

PR #1224, opened 2025-01-24
feat: loki schema config management

### LINKED ISSUE #782: Update to TSDB by default for Loki (and latest schema version)

chance-coleman:
## Description

Feature to manage the loki schemaConfig date and version transitions with helm only changes. 

Walk through of what the helm is doing:

- Secret Retrieval: Fetches the loki secret from the loki namespace.
- Default Date Setup: Sets a default date that is 48 hours in the future.
- Initial Result Setting: Initializes the result variable with the default date.
- Check for Secret and Config Data: Verifies the secret and config.yaml file exist.
- Decode and Parse Config Data: Converts the base64-encoded config.yaml from the secret into a readable YAML format stored in a variable.
- Iterate Through Configs: Loops through each configuration in the schema_config.configs list.
- Conditionally Update Result: If a configuration with store set to "tsdb" and version set to "v13" is found, updates the result to the from date of that configuration and stops the loop.
- Output the Result: Outputs the date found or the default date if no matching configuration is found.

## Related Issue

Fixes #782 

## Type of change

## Steps to Validate
- `uds run dev-setup`
- `npx pepr deploy`
- At this point there is a few different configurations to test with the loki behavior the following are what i focused on in my testing:
  - new install of uds-core logging 
    - `uds run dev-deploy --set LAYER=logging`
    - this will test that a new install will set the tsdb "from" date 2 days into the future
  - upgrade from valid tsdb ( meaning the current loki is configured to use tsdb )
    - modify the values.yaml for loki to deploy with tsdb as a static "from" date like "2023-01-12"
    - deploy `uds run dev-deploy --set LAYER=logging`
    - revert that change to the values.yaml
    - deploy again `uds run dev-deploy --set LAYER=logging`
    - This should result in the tsdb "from" being left alone and still equal to what was originally set
  - upgrade from loki that wasnt configured with tsdb
    - modify the values.yaml to exclude (comment out or delete) the tsdb schemaConfig entry
    - deploy `uds run dev-deploy --set LAYER=logging`
    - revert the changes to the vlaues.yaml
    - deploy `uds run dev-deploy --set LAYER=logging`
    - this should result in the tsdb schemaConfig being created with a date 2 days into the future

## Checklist before merging


src/loki/values/values.yaml
```diff
@@ -46,6 +46,25 @@ loki:
         index:
           prefix: loki_index_
           period: 24h
+      - from: "{{- $secret := lookup \"v1\" \"Secret\" \"loki\" \"loki\" -}}
+          {{- $defaultDate := now | dateModify \"+48h\" | date \"2006-01-02\" -}}
+          {{- $result := $defaultDate -}}
+          {{- if and $secret (index $secret.data \"config.yaml\") -}}
```
mjnagel: Definite nit, but wondering if we should set the date to `now` (maybe with dateModify minus some amount of time) if the secret doesn't exist? Right now for a fresh install we'd be setting boltdb / old schema for the first two days when we could in theory just start right on the latest schema and tsdb.

chance-coleman: set the date to 24 hours previous to the creation/update of loki.

noahpb (dismissed):
Validated all scenarios locally, lgtm! nice work

sgettys:
Are there any upgrade scenarios we need to run for this? Also should we do a docs update on the what/why for doing this and how to override?

src/loki/values/values.yaml
mjnagel: Currently the `defaultDate` is used if (1) the secret doesn't exist or (2) the new tsdb store isn't being used.

I think these are actually 2 different scenarios that we need to be careful with:
1. If the secret doesn't exist then we can consider this a net-new install (not an upgrade). In this case, we can set the date to something in the past like we have here.
2. If the secret does exist, but the `tsdb` store isn't in the config, this is an upgrade where we need to set the date to the future.

So I think we basically need 2 different default dates? First one should be `-24h` (or maybe `-48h` bc timezones?). This should be used if and only if the secret doesn't exist. If the secret exists our default date should be in the future, `+48h`. And then if we find an existing `tsdb`, `v13` date we use that instead of the defaults.

chance-coleman: updated for this, as well as added a doc explaining the store type and version management

docs/reference/configuration/loki-storage.md
```diff
@@ -0,0 +1,69 @@
...
+            {{- end -}}
+          {{- end -}}
+          {{- $result -}}"
+        store: tsdb
+        object_store: "{{ .Values.loki.storage.type }}"
+        schema: v13
+        index:
+          prefix: loki_index_
+          period: 24h
+```
```
mjnagel: This should probably be documented as an `override` for a bundle since that is the main way an end user would be interacting with this. I think this is the right syntax but probably good to validate:
```yaml
  - name: core
    repository: ghcr.io/defenseunicorns/packages/uds/core
    ref: x.x.x
    overrides:
      loki:
        loki:
          values:
            # Setting this value will override the default auto-lookup behavior
            - path: loki.schemaConfig.configs
              value:
                # Self-manage configs here, making sure to include any previous dates you used
                - from: 2022-01-11
                  store: boltdb-shipper
                  object_store: "{{ .Values.loki.storage.type }}"
                  schema: v12
                  index:
                    prefix: loki_index_
                    period: 24h
                # Example: Explicitly set a date for TSDB and the v13 schema, making sure this is in the future
                - from: 2025-03-25
                  store: tsdb
                  object_store: "{{ .Values.loki.storage.type }}"
                  schema: v13
                  index:
                    prefix: loki_index_
                    period: 24h
```

docs/reference/configuration/loki-storage.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Forcing an Upgrade with Immediate Schema Change
+If you want to transition immediately to a new storage type without waiting for the automatic 48-hour logic, modify the Helm chart values by explicitly setting the schema configuration in `values.yaml`.
```
mjnagel: This might also be helpful to have an example override for. Would also maybe link to the upstream docs on changing the schema and specifically call out the nuances around making sure the date is in the future (and specifically future with timezones): https://grafana.com/docs/loki/latest/operations/storage/schema/#changing-the-schema

src/loki/values/values.yaml
mjnagel: Haven't run through any deploys/upgrades but I think the logic looks good now so going to resolve this thread.

docs/reference/configuration/loki-storage.md
```diff
@@ -0,0 +1,73 @@
...
+                  object_store: "{{ .Values.loki.storage.type }}"
+                  schema: v13
+                  index:
+                    prefix: loki_index_
+                    period: 24h
+```
+
+It is important to ensure that the from date is set in the future, accounting for time zones, to avoid potential indexing issues. Additionally, schema configurations must be listed in sequential order, with the latest configuration appearing last to ensure proper data indexing.
+
+For more details on changing the schema, refer to the official Loki documentation: [Changing the Schema](https://grafana.com/docs/loki/latest/operations/storage/schema/#changing-the-schema)..
```
mjnagel:
```suggestion
For more details on changing the schema, refer to the official Loki documentation: [Changing the Schema](https://grafana.com/docs/loki/latest/operations/storage/schema/#changing-the-schema)
```

docs/reference/configuration/loki-storage.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Managing Upgrades
+When upgrading Loki with an existing deployment, follow these best practices to ensure a seamless transition:
+
+1. Check Existing Configuration
+   - Retrieve the existing Loki `Secret` using:
+     ```sh
+     kubectl get secret loki -n loki -o yaml
+     ```
+   - Check if `config.yaml` exists and whether `schema_config.configs` contains an entry for the intended storage type.
```
mjnagel, line 33: I think we feel pretty good about the upgrade logic and it relies on "proven" helm tech, but maybe we should note "save off a copy of the `schema_config.configs` list" to use with overrides if something goes wrong? Essentially a backup/restore type thing - should only be relevant when you're making a real change (upgrade).

mjnagel (dismissed):
Tested 3 scenarios a few times:
- Fresh install from this branch, tsdb date set to past as expected
- Upgrade from last release -> this branch, tsdb date set to future as expected
- Upgrade from a "past deploy" of tsdb (manually changed secret to an older date for tsdb), tsdb date maintained across upgrade

mjnagel (approved):

Issue #1220, opened 2025-01-23
Allow updating domain and admin domain seamlessly on upgrade


mjnagel:
### Is your feature request related to a problem? Please describe.

Currently UDS Core does not (easily) support updating the domain on an existing deployment. While anywhere the domain is directly used in helm values will typically get updated, there are a number of operator-controlled-resources that need to be adjusted when the domain changes. In particular the following need to happen but do not currently:
- All virtualservices and serviceentries need to update to the new domain
- ~Keycloak's trusted host policy needs to be updated to the new domain~ not required with client credentials
- All SSO clients need to be updated with the right domain (for redirect uris, etc)
- Related to ^ Authservice configuration must be updated with the right domain
- Potential pod cycling to ensure all the updated config is read (unsure if this is required)

### Describe the solution you'd like

When I update my domain, all deployments properly read in the new domain and Pepr updates any necessary configuration so that everything works!

### Describe alternatives you've considered

Document a manual process for how to shift to a new domain - this could be partially necessary with the current state of Keycloak management, but ideally we would manage this all for the end user.

### Additional context

This was a TODO from https://github.com/defenseunicorns/uds-core/pull/1186

mjnagel:
With the switch to Keycloak client credentials management this should be more feasible.

mjnagel:
Also to note as a workaround this can be manually accomplished in a few basic steps:
- Deploy your updated config with the new domain
- Cycle the Pepr watcher pod - this will force a re-reconciliation of all Package CRs
- (per-application) validate application is exposed on the new domain as expected and functionality like SSO is working - in some cases you may need to cycle individual applications to reload the domain/sso config

PR #1233, opened 2025-01-27
feat!: create ClusterConfig CRD

### LINKED ISSUE #1077: Create ClusterConfig CRD to configure cluster/cloud parameters

This came out of CTO office hours. I was chatting about this strange EKS edge-case where we do not receive an event when the Kubernetes `EndpointSlice` and `Service` . Also, I know that @mjnagel is already working on a derivative of this.

`ClusterConfig` as a way to parameter things like CIDR Range, Domain, ect. This should allow us to hedge against strange edge-cases that are present in specific environments. This would be a more robust solution that is highly configurable. 

- **Given** I want to have a very specific cluster config for a given cloud environment
- **When** I create an Instance of a `ClusterConfig` 
- **Then** it will program specific configurations for edge cases that exist in given clouds.

TristanHoladay:
## Description
This PR:
* adds a ClusterConfig CRD
* moves the CRD generation into a manifest component
* updates uds-config watcher / logic to use the ClusterConfig

## Related Issue

- resolves #1077 

## Design Doc
https://www.notion.so/ClusterConfig-v1-183e512f24fc80879e7cce7731617039

## Type of change

## Steps to Validate

Modify slim-dev to add some overrides for the cluster config, for example:
```
      uds-operator-config:
        uds-operator-config:
          values:
            - path: cluster.policy.allowAllNsExemptions
              value: "true"
            - path: operator.KUBEAPI_CIDR
              value: "172.0.0.0/8"
            - path: operator.KUBENODE_CIDRS
              value: "172.0.0.0/8,192.0.0.0/8"
```

Deploy slim-dev and validate all items behave as expected (for example netpol/authpol have the expected CIDRs, exemptions are allowed in all namespaces).

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: create ClusterConfig CRD (#1233)

BREAKING CHANGE: The following bundle overrides are no longer valid, and you will need to use the corresponding `cluster` values instead, or where applicable use the Zarf variable. If currently using the Zarf variables to set these configs there is no change. If using helm/bundle overrides, review the table below: <br><table><thead><tr><th>Removed Value</th><th>Replacement</th></tr></thead><tbody><tr><td><code>operator.UDS_DOMAIN</code></td><td><code>cluster.expose.domain</code> or Zarf variable <code>DOMAIN</code></td></tr><tr><td><code>operator.UDS_ADMIN_DOMAIN</code></td><td><code>cluster.expose.adminDomain</code> or Zarf variable <code>ADMIN_DOMAIN</code></td></tr><tr><td><code>operator.UDS_CA_CERT</code></td><td><code>cluster.expose.caCert</code> or Zarf variable <code>CA_CERT</code></td></tr><tr><td><code>operator.UDS_ALLOW_ALL_NS_EXEMPTIONS</code></td><td><code>cluster.policy.allowAllNsExemptions</code> or Zarf variable <code>ALLOW_ALL_NS_EXEMPTIONS</code></td></tr><tr><td><code>operator.UDS_LOG_LEVEL</code></td><td>Zarf variable <code>UDS_LOG_LEVEL</code> <em>(no bundle/Helm override available)</em></td></tr></tbody></table>
END_COMMIT_OVERRIDE


src/pepr/operator/controllers/config/config.ts
```diff
@@ -3,60 +3,97 @@
...
     if (UDSConfig.caCert) {
-      try {
-        atob(UDSConfig.caCert);
-      } catch (e) {
-        log.error(
-          "Invalid CA Cert provided in uds-operator-config secret, falling back to no CA Cert",
+      if (!isBase64(UDSConfig.caCert)) {
+        configLog.error(
+          "Invalid CA Cert provided in uds-operator-config ClusterConfig, falling back to no CA Cert",
         );
```
mjnagel: This might be a good one to add an event on the cluster config CR since we fallback to something and they might get unexpected behavior if they just made a typo or something.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -69,43 +106,126 @@ export async function updateUDSConfig(config: kind.Secret) {
...
+}
 
-  log.info("Updated UDS Config based on uds-operator-config secret changes");
+export async function loadUDSConfig() {
+  // Run in Admission and Watcher pods
+  if (process.env.PEPR_WATCH_MODE || process.env.PEPR_MODE === "dev") {
+    const cfgList = await K8s(ClusterConfig).InNamespace("pepr-system").Get();
+    const cfgSecretList = await K8s(kind.Secret).InNamespace("pepr-system").Get();
+
+    const cfgSecret = cfgSecretList.items.find(s => s.metadata?.name === "uds-operator-config");
```
mjnagel: I might be missing intent here but `.Get()` supports getting a secret by name, i.e. this could just be one line:
```suggestion
    const cfgSecret = await K8s(kind.Secret).InNamespace("pepr-system").Get("uds-operator-config");
```

This also could be used for the cluster config, but I see that you're also using the full get to error if multiple cluster configs are found so maybe that one makes sense as is?

src/pepr/operator/crd/register.ts
```diff
@@ -82,4 +83,39 @@ export async function registerCRDs() {
         process.exit(1);
       });
   }
+
+  // Register the Exemption CRD if we're in "admission" or dev mode (Exemptions are watched by the admission controllers)
```
mjnagel: Comment needs updating to reflect the right CRD, but also probably should update the conditional to run regardless of `PEPR_WATCH_MODE`? (although this isn't actually in the execution code anymore)

src/pepr/operator/crd/sources/cluster-config/v1alpha1.ts
```diff
@@ -0,0 +1,94 @@
...
+                    "cidr range for your Kubernetes control plane nodes. This is a manual override that can be used instead of relying on Pepr to automatically watch and update the values",
+                },
+                kubenodeCIDRS: {
+                  type: "array",
+                  description:
+                    "cidr(s) for all Kubernetes nodes (not just control plane). Similar reason to above,annual override instead of relying on watch",
+                  items: {
+                    type: "string",
+                    pattern:
+                      "^(([0-9]{1,3}\\.){3}[0-9]{1,3}\\/[0-9]+)|(([a-fA-F0-9:]+:+)+[a-fA-F0-9:]+(/[0-9]+)?)$", // Matches IPv4 and IPv6 CIDR
```
mjnagel: Assuming this regex works, can we use the same on the `kubeapiCIDR`?

mjnagel: It might also be nice to have tests against this regex in our code somewhere - I think @UnicornChance briefly looked at this for the `remoteCidr` stuff and we ended up just leaving the validation to k8s because we couldn't find a succinct and solid approach to validating both ipv4 and ipv6.

src/pepr/operator/crd/sources/cluster-config/v1alpha1.ts
```diff
@@ -0,0 +1,94 @@
...
+              type: "object",
+              properties: {
+                domain: {
+                  type: "string",
+                  description: "Domain all cluster services will be exposed on",
+                },
+                adminDomain: {
+                  type: "string",
+                  description:
+                    "Domain all cluster services on the admin gateawy will be exposed on",
```
mjnagel:
```suggestion
                    "Domain all cluster services on the admin gateway will be exposed on",
```

src/pepr/operator/crd/sources/cluster-config/v1alpha1.ts
```diff
@@ -0,0 +1,94 @@
...
+                },
+                caCert: {
+                  type: "string",
+                  description:
+                    "The trusted CA that signed your domain certificates if using Private PKI ",
```
mjnagel:
```suggestion
                    "The trusted CA that signed your domain certificates if using Private PKI",
```

src/pepr/tasks.yaml
```diff
@@ -15,6 +16,10 @@ tasks:
 
       - cmd: "npx kubernetes-fluent-client crd exemptions.uds.dev src/pepr/operator/crd/generated"
 
+      - cmd: "npx kubernetes-fluent-client crd clusterconfig.uds.dev src/pepr/operator/crd/generated"
```
mjnagel: If you merge in `main` you will also see some new schema generation tasks in here. I don't know that we need to publish the cluster config schema given that it's not really end user facing in an editor the way package/exemption CRs are...but we may want to for consistency sake?

src/pepr/uds-cluster-crds/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+apiVersion: v2
+name: uds-cluster-crds
+description: UDS Core cluster level configurations
```
mjnagel: Description is probably more accurate as:
```suggestion
description: UDS Core CRDs
```
Since the "configurations" are really the CRs, not the CRDs?

src/pepr/uds-cluster-crds/templates/clusterconfig.uds.dev.yaml
```diff
@@ -0,0 +1,86 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    meta.helm.sh/release-name: zarf-1480d00ca85972047319d79d2ea7253325be2d4c
+    meta.helm.sh/release-namespace: default
+  labels:
+    app.kubernetes.io/managed-by: Helm
```
mjnagel: I'm not sure these should be here? Especially since we explicitly set them in the action, might be candidates to delete in the `yq` part of the task?

src/pepr/tasks.yaml
```diff
@@ -28,17 +33,29 @@ tasks:
...
 
+  - name: gen-manifests
+    description: "Generate CRD manifests"
+    actions:
+      - cmd: |
+          CRD_LIST="exemptions.uds.dev packages.uds.dev clusterconfig.uds.dev"
+          for CRD in $CRD_LIST; do
+             FILE="src/pepr/uds-cluster-crds/templates/$CRD.yaml"
+             kubectl get crd $CRD -o yaml > $FILE
+             yq -i 'del(.metadata.creationTimestamp,.metadata.generation,.metadata.resourceVersion,.metadata.uid,.status)' $FILE
```
mjnagel: These two commands should be `uds zarf tools` just to make sure this works well on dev machines where kubectl or yq aren't present/on the path.

src/pepr/uds-operator-config/values.yaml
```diff
@@ -2,18 +2,35 @@
...
+  expose:
+    # Domain configuration (admin defaults to `admin.UDS_DOMAIN`)
+    UDS_DOMAIN: "###ZARF_VAR_DOMAIN###"
+    UDS_ADMIN_DOMAIN: "###ZARF_VAR_ADMIN_DOMAIN###"
+    UDS_CA_CERT: "###ZARF_VAR_CA_CERT###"
+  policy:
+    UDS_ALLOW_ALL_NS_EXEMPTIONS: "###ZARF_VAR_ALLOW_ALL_NS_EXEMPTIONS###"
+  networking:
+    KUBEAPI_CIDR: ""
+    KUBENODE_CIDRS: []
```
mjnagel: imo all of these can follow better values naming standards. As long as we are maintaining parity on the zarf vars, these are net new keys so we can name them better. For example:
```suggestion
cluster:
  attributes:
    clusterName: "###ZARF_VAR_CLUSTER_NAME###"
    clusterTags: "###ZARF_VAR_CLUSTER_TAGS###"
  expose:
    # Domain configuration (admin defaults to `admin.<domain>`)
    domain: "###ZARF_VAR_DOMAIN###"
    adminDomain: "###ZARF_VAR_ADMIN_DOMAIN###"
    caCert: "###ZARF_VAR_CA_CERT###"
  policy:
    allowAllNsExemptions: "###ZARF_VAR_ALLOW_ALL_NS_EXEMPTIONS###"
  networking:
    kubeApiCidrs: ""
    kubeNodeCidrs: []
```

For context the other keys were all caps + `_` to match with env standards since they were directly mounted as env by pepr. Since we're no longer doing that it would be good to align with helm norms of camelcase here.

src/pepr/uds-operator-config/values.yaml
```diff
@@ -2,18 +2,35 @@
...
   PEPR_LAST_SEEN_LIMIT_SECONDS: "300"
   # Allow Pepr to re-list resources more frequently to avoid missing resources
   PEPR_RELIST_INTERVAL_SECONDS: "600"
   # Configure Pepr reconcile strategy to have separate queues for faster reconciliation
   PEPR_RECONCILE_STRATEGY: "kindNsName"
+
+cluster:
+  attributes:
+    UDS_CLUSTER_NAME: "###ZARF_VAR_CLUSTER_NAME###"
+    UDS_CLUSTER_TAGS: "###ZARF_VAR_CLUSTER_TAGS###"
```
mjnagel: For cluster name and tags imo we don't need to expose a zarf var for these, unless there's a specific UI use case I'm missing here - they could just be set as overrides?

Specific to tags, shouldn't that be a list rather than a string (maybe this is string because of zarf var right now though)? It could also be useful to have a values schema for this chart now that we're adding more complexity to it.

src/pepr/uds-operator-config/templates/clusterconfig.yaml
```diff
@@ -0,0 +1,33 @@
...
+{{- end }}
+  networking:
+    kubeapiCIDR: {{ .Values.cluster.networking.KUBEAPI_CIDR | default .Values.operator.KUBEAPI_CIDR | quote }}
+    kubenodeCIDRS:
+{{- if .Values.cluster.networking.KUBENODE_CIDRS }}
+{{ toYaml .Values.cluster.networking.KUBENODE_CIDRS | indent 6 }}
+{{- else if .Values.operator.KUBENODE_CIDRS }}
+{{- range $index, $cidr := .Values.operator.KUBENODE_CIDRS }}
+      - {{ $cidr | quote }}
+{{- end }}
```
mjnagel: I'll probably have to test this out locally so putting a placeholder comment for now but this templating looks off to me. I'd expect that we need to split on `,` here and range over that ([example of this](https://stackoverflow.com/questions/73127947/helm-iterate-over-string-with-comma-separated-values)). By default I think range here will either glob things together or it might range over each character?

src/pepr/uds-operator-config/values.yaml
mjnagel: On the tag comment I think that would also help clean up the helm template for tags if switched to a proper list.

src/pepr/zarf.yaml
```diff
@@ -32,6 +32,28 @@ variables:
...
+            mute: true
+            cmd: |
+              # Label the specific CRDs
+              ./zarf tools kubectl label $(./zarf tools kubectl get crds | awk '/exemptions.uds.dev|packages.uds.dev/ {print "crd/" $1}') "app.kubernetes.io/managed-by=Helm" --overwrite || true
+
+              # Annotate with release name
+              ./zarf tools kubectl annotate $(./zarf tools kubectl get crds | awk '/exemptions.uds.dev|packages.uds.dev/ {print "crd/" $1}') "meta.helm.sh/release-name=uds-cluster-crds" --overwrite || true
+
+              # Annotate with release namespace
+              ./zarf tools kubectl annotate $(./zarf tools kubectl get crds | awk '/exemptions.uds.dev|packages.uds.dev/ {print "crd/" $1}') "meta.helm.sh/release-namespace=pepr-system" --overwrite || true
```
mjnagel: This feels overly complex, we should be able to just directly label/annotate two CRDs in a single kubectl call without needing a get/awk in there?
```suggestion
              # Label the specific CRDs
              ./zarf tools kubectl label crd exemptions.uds.dev packages.uds.dev "app.kubernetes.io/managed-by=Helm" --overwrite || true
              # Annotate with release name
              ./zarf tools kubectl annotate crd exemptions.uds.dev packages.uds.dev "meta.helm.sh/release-name=uds-cluster-crds" --overwrite || true
              # Annotate with release namespace
              ./zarf tools kubectl annotate crd exemptions.uds.dev packages.uds.dev "meta.helm.sh/release-namespace=pepr-system" --overwrite || true
```

I think the resulting command ends up being the same unless someone had only one of the CRs on their cluster - but the net result should be the same always due to the `|| true` fallback.

src/pepr/uds-cluster-crds/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.1.0
```
mjnagel, line 21: Placeholder note mostly for myself - I've encountered issues in the past when charts are updated without a version bump. I _think_ that zarf will upgrade/apply regardless but we may want to validate to ensure that we have the ability to upgrade CRDs easily here without versioning this chart. If not we may want to automate the version bump or ensure there's clear dev notes on how to.

mjnagel (commented):
Still need to go through the config functions more and test out a few deploy scenarios but wanted to get this initial round of feedback out. The approach overall looks sound and aligns with the design so no overall concerns - just will want to validate there's no timing or upgrade issues with the new CRD approaches.

src/pepr/operator/controllers/config/config.ts
TristanHoladay: mmm. ok i wondered if i could do that but the intellisense wasn't indicating `.Get()` accepted params. It's difficult to say whether the # check for the ClucterConfig is valuable here or not. For cluster setups, probably not. On pod cycling or upgrades it makes more sense. I like making the calls the same though .... :shrug:

src/pepr/operator/crd/register.ts
TristanHoladay: nice catch

src/pepr/operator/crd/sources/cluster-config/v1alpha1.ts
TristanHoladay: yeah looks like that one isn't working and a refactored version doesn't either. i'll yank it out.

src/pepr/tasks.yaml
TristanHoladay: i could see it as useful if a user can't figure out why the values they're feeding the chart are not working, but i agree that sense users don't need to create ClusterConfig CRs it might be confusing. I say we keep it separate for now and publish the schema later if desired.

src/pepr/operator/controllers/config/config.ts
TristanHoladay: i see that we can write an Event with `writeEvent()`, but i don't see anything that's writing any events to CRs. There is the `handleFailure()` that we could open up to more than just Package CRs but that is writing to the `status`. Are you saying we should update the `ClusterConfig` CR status and write an Event or actually write an event on to the CR?

mjnagel: Yeah the [`writeEvent`](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/reconcilers/index.ts#L89) function is the way I'd expect us to be doing this. The event ends up attached to a specific CR through the [`involvedObject`](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/reconcilers/index.ts#L101-L107), which will make the event show when describing the CR (so not directly "writing to" the CR, but they get attached that way). This function currently only allows passing in the `UDSPackage` type, but I think it can be easily updated to allow our ClusterConfig as well. [Here's an example](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/reconcilers/package-reconciler.ts#L129-L133) that also overrides the reason/type.

TristanHoladay: cool, i wondered if there was an indirect attachment somehow happening. I'll try to add this today.

TristanHoladay: as note in sync chat, this check was actually redundant now that we're validating on updates.

src/pepr/operator/controllers/config/config.spec.ts
```diff
@@ -55,21 +45,152 @@ jest.mock("../../../logger", () => {
...
+    data: {
+      AUTHSERVICE_REDIS_URI: btoa("mock-redis-uri"),
+    },
+  } as kind.Secret;
+
+  return {
+    K8s: jest
+      .fn()
+      // valid ClusterConfig and config secret
+      .mockReturnValueOnce({
```
mjnagel: Maybe a nit but I don't really like this approach to mocking as it relies on the test ordering (i.e. using `mockReturnValueOnce` to mock it differently per test).

I think we could refactor the mocks here a bit. First we'd want to setup some const jest functions outside of the test code:
```ts
const mockClusterConfGet = jest.fn();
const mockSecretGet = jest.fn();
```
Then in the current pepr mock (where this comment is), setup the mocks to mock the implementati0on and return these functions:
```ts
    (K8s as jest.Mock).mockImplementation(resourceKind => {
      if (resourceKind === ClusterConfig) {
        return { InNamespace: jest.fn().mockReturnValue({ Get: mockClusterConfGet }) };
      }
      if (resourceKind === kind.Secret) {
        return { InNamespace: jest.fn().mockReturnValue({ Get: mockSecretGet }) };
      }
```
Then per test we can mock the return value:
```ts
    mockClusterConfGet.mockReturnValue({
      // Specific return value for the test
    });
    mockSecretGet.mockReturnValue({
      // Specific return value for the test
    });
```

We have something very similar to this in the istio namespace testing for reference: https://github.com/defenseunicorns/uds-core/blob/9b6f6812042aa5345efc69c7242922c0014ae867/src/pepr/operator/controllers/istio/namespace.spec.ts#L29-L68

src/pepr/operator/controllers/config/config.ts
```diff
@@ -3,64 +3,91 @@
...
+  allowAllNSExemptions: false,
+  kubeApiCidr: "",
+  kubeNodeCidrs: [],
+  isIdentityDeployed: false,
+  isAmbientDeployed: false,
+};
 
-export async function updateUDSConfig(config: kind.Secret) {
-  log.info("Updating UDS Config from uds-operator-config secret change");
+export const configLog = setupLogger(Component.CONFIG);
```
mjnagel: Why did this change from `setupLogger(Component.OPERATOR_CONFIG)`?

src/pepr/operator/controllers/config/config.ts
```diff
@@ -3,64 +3,91 @@
...
     if (
-      !UDSConfig.authserviceRedisUri ||
-      UDSConfig.authserviceRedisUri === "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
+      decodedCfgData.AUTHSERVICE_REDIS_URI &&
+      decodedCfgData.AUTHSERVICE_REDIS_URI !== "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
     ) {
+      UDSConfig.authserviceRedisUri = decodedCfgData.AUTHSERVICE_REDIS_URI;
+    } else {
       UDSConfig.authserviceRedisUri = "";
     }
```
mjnagel: The ordering of these checks can result in `reconcileAuthservice` getting called unnecessarily. If we check inequality before filtering out the "empty string" cases then undefined/"###ZARF_VAR_AUTHSERVICE_REDIS_URI###" will result in calling this function even if they "evaluate" to `""` in the end.

May be good to add test cases for this as well (I did this locally to validate my assumption).

src/pepr/operator/controllers/config/config.ts
```diff
@@ -69,43 +96,128 @@ export async function updateUDSConfig(config: kind.Secret) {
...
+    const cfgSecret = await K8s(kind.Secret).InNamespace("pepr-system").Get("uds-operator-config");
+
+    if (cfgList.items.length === 0) {
+      throw new Error("No ClusterConfig found");
+    }
+
+    // Important check in case additional configs are created in window between uds-operator-config chart and pepr module chart
+    if (cfgList.items.length > 1) {
+      throw new Error(
+        `ClusterConfig Processing: only one ClusterConfig is allowed -- found: ${cfgList.items.length}`,
```
mjnagel: nit: should this log out the names as well?

src/pepr/operator/controllers/keycloak/authservice/config.ts
```diff
@@ -30,20 +31,27 @@ let debounceTimer: NodeJS.Timeout | null = null;
...
+
+export function initializeOperatorConfig() {
+  operatorConfig = {
+    namespace: "authservice",
+    secretName: "authservice-uds",
+    baseDomain: `https://sso.${UDSConfig.domain}`,
+    realm: "uds",
+  };
 
+  log.info(operatorConfig, `Operator config initialized`);
```
mjnagel:
```suggestion
  log.info(operatorConfig, `Authservice operator config initialized`);
```

src/pepr/operator/crd/register.ts
```diff
@@ -6,12 +6,50 @@
...
+import { v1alpha1 as clusterConfig } from "./sources/cluster-config/v1alpha1";
 import { v1alpha1 as exemption } from "./sources/exemption/v1alpha1";
 import { v1alpha1 as pkg } from "./sources/package/v1alpha1";
 
 // configure subproject logger
 const log = setupLogger(Component.OPERATOR_CRD);
 
+export async function registerClusterConfig() {
+  // Register the ClusterConfig CRD
+  if (process.env.PEPR_WATCH_MODE || process.env.PEPR_MODE === "dev") {
```
mjnagel:
```suggestion
  if (process.env.PEPR_WATCH_MODE === "true" || process.env.PEPR_MODE === "dev") {
```

I think? Unless I'm missing a reason this is applied from all pods.

src/pepr/operator/crd/sources/cluster-config/v1alpha1.ts
```diff
@@ -0,0 +1,92 @@
...
+                  description:
+                    "cidr range for your Kubernetes control plane nodes. This is a manual override that can be used instead of relying on Pepr to automatically watch and update the values",
+                },
+                kubenodeCIDRS: {
+                  type: "array",
+                  description:
+                    "cidr(s) for all Kubernetes nodes (not just control plane). Similar reason to above,annual override instead of relying on watch",
+                  items: {
+                    type: "string",
+                  },
```
mjnagel:
```suggestion
                kubeApiCIDR: {
                  type: "string",
                  description:
                    "CIDR range for your Kubernetes control plane nodes. This is a manual override that can be used instead of relying on Pepr to automatically watch and update the values",
                },
                kubeNodeCIDRs: {
                  type: "array",
                  description:
                    "CIDR(s) for all Kubernetes nodes (not just control plane). Similar reason to above,annual override instead of relying on watch",
                  items: {
                    type: "string",
                  },
```
Will likely need to regenerate after making these changes, and maybe update some code refs:
1. CIDR is an acronym so should be capitalized everywhere (in the description).
2. CIDRs is just the plural, so the s should be lowercase.
3. To keep with camelCase-ing things `Api` and `Node` should have capital letters in the property keys.

src/pepr/operator/crd/validators/clusterconfig-validator.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import { PeprValidateRequest } from "pepr";
+import { isBase64 } from "../../controllers/utils";
+import { ClusterConfig } from "../generated/clusterconfig-v1alpha1";
+
+export async function validateCfgUpdate(req: PeprValidateRequest<ClusterConfig>) {
+  try {
+    validateCfg(req.Raw);
+  } catch (e) {
+    return req.Deny(`Failed to validate UDSConfig update: ${e}`);
```
mjnagel: This message reads funny to me? Imo we have "successfully validated" the UDS Config, but it failed the validation?
```suggestion
    return req.Deny(`ClusterConfig failed validation: ${e}`);
```

But potentially that is weird too, since it would end up being `ClusterConfig failed validation: ClusterConfig Validation: namespace or name is invalid`. May need to rework the messaging here?

src/pepr/operator/crd/register.ts
mjnagel: Revisiting this after reviewing the rest, since this is only called by the task maybe it's not a big issue?

mjnagel (commented):
Doing some more validation of upgrade scenarios and backwards compatibility, but wanted to add a few smaller code comments first.

src/pepr/values.yaml
```diff
@@ -4,12 +4,12 @@
 watcher:
   serviceMonitor:
     enabled: ###ZARF_VAR_PEPR_SERVICE_MONITORS###
-  envFrom:
-    - secretRef:
-        name: uds-operator-config
+  env:
+    - name: UDS_LOG_LEVEL
+      value: ###ZARF_VAR_UDS_LOG_LEVEL###
```
mjnagel, line 9: I think this loses two values/overrides we still want:
- PEPR_RECONCILE_STRATEGY
- PEPR_KEYCLOAK_CLIENT_STRATEGY

I'm not seeing another place/way those are loaded? We could hardcode `PEPR_RECONCILE_STRATEGY` if we needed/wanted to. `PEPR_KEYCLOAK_CLIENT_STRATEGY` is a new one, but should be overridable by the end user. Not sure if it needs to be part of the clusterconfig because ideally it's temporary but curious on @slaskawi opinion?

Related, maybe we discussed this but can/should we clean up the existing secret? Leaving all the existing config values in there could be confusing if that's no longer the source of truth (domains, cacert, etc)?

src/pepr/uds-operator-config/templates/clusterconfig.yaml
```diff
@@ -0,0 +1,31 @@
...
+spec:
+  attributes:
+    clusterName: {{ .Values.cluster.attributes.clusterName | quote }}
+    tags:
+{{ toYaml .Values.cluster.attributes.clusterTags | nindent 6 }}
+  networking:
+    kubeapiCIDR: {{ .Values.cluster.networking.kubeApiCidr | default .Values.operator.kubeApiCidr | quote }}
+    kubenodeCIDRS:
+{{- if .Values.cluster.networking.kubeNodeCidrs }}
+{{ toYaml .Values.cluster.networking.kubeNodeCidrs | indent 6 }}
```
mjnagel, line 20:
```suggestion
{{- toYaml .Values.cluster.networking.kubeNodeCidrs | indent 6 }}
```

src/pepr/uds-operator-config/templates/clusterconfig.yaml
```diff
@@ -0,0 +1,31 @@
...
+{{- else if .Values.operator.KUBENODE_CIDRS }}
+{{- range ( split "," .Values.operator.KUBENODE_CIDRS ) }}
+      - {{ . | quote }}
+{{- end }}
+{{- end }}
```
mjnagel:
```suggestion
    tags:
      {{- toYaml .Values.cluster.attributes.clusterTags | nindent 6 }}
  networking:
    kubeapiCIDR: {{ .Values.cluster.networking.kubeApiCidr | default .Values.operator.KUBEAPI_CIDR | quote }}
    kubenodeCIDRS:
    {{- if .Values.cluster.networking.kubeNodeCidrs }}
      {{- toYaml .Values.cluster.networking.kubeNodeCidrs | nindent 6 }}
    {{- else if .Values.operator.KUBENODE_CIDRS }}
      {{- range ( split "," .Values.operator.KUBENODE_CIDRS ) }}
      - {{ . | quote }}
      {{- end }}
    {{- end }}
```
Small readability improvements for the template. Also fixed a bad fallback value in the default for kubeapi.

src/pepr/uds-cluster-crds/Chart.yaml
mjnagel: Validated locally that CRD upgrades get applied without needing to bump this.

src/pepr/values.yaml
slaskawi: Yes, I agree - there should be an easy way to override these. Right now they are defined here:

https://github.com/defenseunicorns/uds-core/blob/05def890049a4cb33e19634f93f08bed008cb09f/src/pepr/uds-operator-config/values.yaml#L20-L21

Just for the record, once the Client Credentials proves to run without any problems, the Dynamic Client Registration will be deprecated and potentially removed.

src/pepr/operator/crd/validators/clusterconfig-validator.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+  return req.Approve();
+}
+
+export function validateCfg(cfg: ClusterConfig) {
+  if (cfg.metadata?.namespace !== "pepr-system" || cfg.metadata?.name !== "uds-cluster-config") {
+    throw new Error("ClusterConfig Validation: namespace or name is invalid");
```
slaskawi: nit: Might be worth adding an information about the supported values.

src/pepr/operator/crd/generated/clusterconfig-v1alpha1.ts
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * cidr range for your Kubernetes control plane nodes. This is a manual override that can be
+   * used instead of relying on Pepr to automatically watch and update the values
+   */
+  kubeapiCIDR?: string;
+  /**
+   * cidr(s) for all Kubernetes nodes (not just control plane). Similar reason to above,annual
+   * override instead of relying on watch
+   */
+  kubenodeCIDRS?: string[];
```
slaskawi: Do we care about configuring other CIDRs like Services?

src/pepr/operator/controllers/utils.ts
```diff
@@ -132,3 +132,13 @@ export async function retryWithDelay<T>(
...
 }
+
+/**
+ * Node.js friendly base64 validator.
+ *
+ * @param {string} str - string to validate as base64
+ * @returns {boolean} - The result of the validation.
+ */
+export function isBase64(str: string) {
+  return Buffer.from(str, "base64").toString("base64") === str;
```
slaskawi: Does this throw errors?

slaskawi (commented):
Added a few questions and comments inline but I wonder if this change is backwards compatible. In other words, if I upgrade my cluster and for some reason I won't have a `ClusterConfig` CR - would UDS boot up successfully?

src/pepr/operator/crd/generated/clusterconfig-v1alpha1.ts
mjnagel: I think for now this is just to keep track of the 2 cidrs / cidr lists that we need for various network policies (in particular for kubeapi and kube node access). In theory could add kube services / kube pods as well but might defer that to when we have a usecase for tracking/using those values?

src/pepr/operator/controllers/config/config.ts
TristanHoladay: i think i did this early on when trying to separate out the config from the `operator` keyword. thinking of it now, i guess it does make sense to leave as the operator config because the operator chart is still the owner / deployment mechanism / controller for this.

mjnagel: I think we also generally try to line the components up with their place in the code, so since this is under `operator/` it's helpful to still have that reference in there.

src/pepr/values.yaml
mjnagel: `PEPR_KEYCLOAK_CLIENT_STRATEGY` can be removed now I think, since we deprecated the options for that.

src/pepr/operator/controllers/config/config.ts
TristanHoladay: image

src/pepr/operator/crd/validators/clusterconfig-validator.ts
TristanHoladay: changed to "Validation Failed: ClusterConfig: ..."

src/pepr/operator/controllers/utils.ts
TristanHoladay: yeah, i think it can. I'm not sure if we want to pass those specific errors on, so for now i threw a try-catch around it and just return false on error.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -51,18 +51,19 @@ export async function updateCfgSecrets(cfg: kind.Secret) {
...
-      decodedCfgData.AUTHSERVICE_REDIS_URI !== "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
-    ) {
-      UDSConfig.authserviceRedisUri = decodedCfgData.AUTHSERVICE_REDIS_URI;
-    } else {
-      UDSConfig.authserviceRedisUri = "";
-    }
+  // no data key then return
+  if (!Object.keys(decodedCfgData).includes("AUTHSERVICE_REDIS_URI")) {
+    return;
+  }
```
mjnagel: I think we discussed this one and it should be close to impossible to hit this case, but if it's undefined should we still treat that as "not set" and set to `""`? The place where this would matter is if someone set a redis uri before, then upgraded to a state where the key wasn't defined. Right now we would leave their redis uri, but I'm thinking we should instead remove it?

src/pepr/operator/controllers/config/config.ts
```diff
@@ -75,18 +76,18 @@ export async function updateCfgSecrets(cfg: kind.Secret) {
   }
 }
 
-export async function updateCfg(cfg: ClusterConfig) {
-  configLog.info("Updating UDS Config from uds-operator-config ClusterConfig change");
+async function handleCAUpdate(expose: ConfigExpose) {
+  if (!Object.keys(expose).includes("caCert")) {
+    return;
+  }
```
mjnagel: Same consideration here, if someone removes their caCert later we would want to make sure it gets removed from the authservice config (treat this as `""`).

mjnagel (commented):
LGTM overall, majority of comments resolved since last review. 2 small things here with undefined values.

src/pepr/operator/controllers/config/config.ts
TristanHoladay: yeah as we talked about it's really unlikely we'll ever hit this, but what you're saying makes sense.

src/pepr/operator/crd/sources/cluster-config/v1alpha1.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export const v1alpha1: V1CustomResourceDefinitionVersion = {
+  name: "v1alpha1",
+  served: true,
+  storage: true,
+  schema: {
+    openAPIV3Schema: {
+      type: "object",
+      properties: {
+        spec: {
```
mjnagel, line 25: I think we can do the name validation of the CRD in the schema here:
```ts
        metadata: {
          type: "object",
          properties: {
            name: {
              type: "string",
              enum: ["uds-cluster-config"],
            },
          },
        } as V1JSONSchemaProps,
```

src/pepr/operator/crd/register.ts
```diff
@@ -6,12 +6,48 @@
...
+      {
+        apiVersion: "apiextensions.k8s.io/v1",
+        kind: "CustomResourceDefinition",
+        metadata: {
+          name: "clusterconfig.uds.dev",
+        },
+        spec: {
+          group: "uds.dev",
+          versions: [clusterConfig],
+          scope: "Namespaced",
```
mjnagel: I was thinking this over this morning and reviewing the validation/apply logic.... I think it makes more sense for this to be `Cluster` scoped. Since we only intend to have a single cluster config, and it's not _necessarily_ tied to a specific namespace I'd vote for use switching this to `Cluster` scope. That should reduce some of the validation we need to do as well.

docs/reference/configuration/uds-cluster-configuration.md
```diff
@@ -0,0 +1,88 @@
...
+    adminDomain: "###ZARF_VAR_ADMIN_DOMAIN###"
+    caCert: "###ZARF_VAR_CA_CERT###"
+  policy:
+    allowAllNsExemptions: "###ZARF_VAR_ALLOW_ALL_NS_EXEMPTIONS###"
+  networking:
+    kubeApiCidr: ""
+    kubeNodeCidrs: []
+```
+
+> Note
```
chance-coleman, line 45:
```suggestion
:::note
Many of the values under `operator` are deprecated and will be removed. Please use `cluster` going forward.
:::
```

nit: Syntax for making a note in the docs.

docs/reference/configuration/uds-cluster-configuration.md
```diff
@@ -0,0 +1,88 @@
...
+> Many of the values under `operator` are deprecated and will be removed. Please use `cluster` going forward.
```
chance-coleman:
```suggestion
:::note
Many of the values under `operator` are deprecated and will be removed. Please use `cluster` going forward.
:::
```

docs/reference/configuration/uds-cluster-configuration.md
```diff
@@ -0,0 +1,88 @@
...
+  KUBENODE_CIDRS: ""
+  ### END DEPRECATED ###
+  AUTHSERVICE_REDIS_URI: "###ZARF_VAR_AUTHSERVICE_REDIS_URI###"
+  # Allow Pepr watch to be configurable to react to dropped connections faster
+  PEPR_LAST_SEEN_LIMIT_SECONDS: "300"
+  # Allow Pepr to re-list resources more frequently to avoid missing resources
+  PEPR_RELIST_INTERVAL_SECONDS: "600"
+  # Configure Pepr reconcile strategy to have separate queues for faster reconciliation
+  PEPR_RECONCILE_STRATEGY: "kindNsName"
+
```
mjnagel: I might lean towards only documenting the non-deprecated values here. The rest would be in the chart still if someone is using them, but that way we lead users toward the newer approach. If you do make this change, may be able to drop the note block below as well.

docs/reference/configuration/uds-cluster-configuration.md
```diff
@@ -0,0 +1,88 @@
...
+    ref: x.x.x
+    overrides:
+      uds-operator-config:
+        uds-operator-config:
+          values:
+            - path: cluster.networking.kubeNodeCidrs
+              value: |
+                172.28.0.2/32
+                172.28.0.3/32
+                172.28.0.4/32
```
mjnagel: Since this is an array shouldn't it be:
```suggestion
            - path: cluster.networking.kubeNodeCidrs
              value:
                - 172.28.0.2/32
                - 172.28.0.3/32
                - 172.28.0.4/32
```

Or:
```suggestion
            - path: cluster.networking.kubeNodeCidrs
              value: ["172.28.0.2/32","172.28.0.3/32","172.28.0.4/32"]
```

docs/reference/configuration/uds-networking-configuration.md
```diff
@@ -40,8 +40,11 @@ packages:
       uds-operator-config:
         uds-operator-config:
           values:
-            - path: operator.KUBENODE_CIDRS
-              value: "172.28.0.2/32,172.28.0.3/32,172.28.0.4/32"
+            - path: cluster.networking.kubeNodeCidrs
+              value: |
+                172.28.0.2/32
+                172.28.0.3/32
+                172.28.0.4/32
```
mjnagel: Same comment as above, I think this should be formatted as an array (right now this is a multi-line string).

src/pepr/uds-operator-config/values.schema.json
```diff
@@ -0,0 +1,103 @@
...
+        "UDS_CA_CERT": { "type": "string", "description": "Deprecated: CA certificate." },
+        "UDS_ALLOW_ALL_NS_EXEMPTIONS": {
+          "type": "string",
+          "description": "Deprecated: Allow all namespace exemptions."
+        },
+        "UDS_LOG_LEVEL": {
+          "type": "string",
+          "enum": ["debug", "info", "warn", "error", "fatal"],
+          "description": "Deprecated: Log level configuration."
+        },
```
mjnagel: I think this is missing trace (it is a valid log level and worked previously). Maybe not the most important because this is set directly with a zarf var on the env now (see other comment as well).

src/pepr/uds-operator-config/values.yaml
```diff
@@ -2,18 +2,35 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 operator:
+  ### DEPRECATED: This section is deprecated and will be removed in a future release. ###
   # Domain configuration (admin defaults to `admin.UDS_DOMAIN`)
   UDS_DOMAIN: "###ZARF_VAR_DOMAIN###"
   UDS_ADMIN_DOMAIN: "###ZARF_VAR_ADMIN_DOMAIN###"
   UDS_CA_CERT: "###ZARF_VAR_CA_CERT###"
   UDS_ALLOW_ALL_NS_EXEMPTIONS: "###ZARF_VAR_ALLOW_ALL_NS_EXEMPTIONS###"
   UDS_LOG_LEVEL: "###ZARF_VAR_UDS_LOG_LEVEL###"
```
mjnagel: I think this value (`UDS_LOG_LEVEL`) is just not used anymore? Since we don't set envFrom this secret anymore this just gets thrown into a secret but not used.

src/pepr/uds-operator-config/values.schema.json
```diff
@@ -0,0 +1,103 @@
...
+        "UDS_DOMAIN": { "type": "string", "description": "Deprecated: Domain configuration." },
+        "UDS_ADMIN_DOMAIN": {
+          "type": "string",
+          "description": "Deprecated: Admin domain configuration."
+        },
+        "UDS_CA_CERT": { "type": "string", "description": "Deprecated: CA certificate." },
+        "UDS_ALLOW_ALL_NS_EXEMPTIONS": {
+          "type": "string",
+          "description": "Deprecated: Allow all namespace exemptions."
+        },
```
mjnagel: This caused an upgrade failure for me. In my bundle override I was passing in:
```yaml
    overrides:
      uds-operator-config:
        uds-operator-config:
          values:
            - path: operator.UDS_ALLOW_ALL_NS_EXEMPTIONS
              value: "true"
```
Deploying the same overrides on this branch I get this error:
```console
     ERROR:  failed to deploy bundle: unable to deploy component "uds-operator-config": unable to install chart after 3 attempts: All attempts fail:
             #1: values don't meet the specifications of the schema(s) in the following chart(s):
             uds-operator-config:
             - operator.UDS_ALLOW_ALL_NS_EXEMPTIONS: Invalid type. Expected: string, given: boolean
```

Seems like somehow I'm ending up with a boolean here, maybe due to some CLI stuff but I think we might want to account for this in our schema and templating...

src/pepr/uds-operator-config/templates/clusterconfig.yaml
```diff
@@ -0,0 +1,31 @@
...
+  labels:
+    {{- include "uds-operator-config.labels" . | nindent 4 }}
+spec:
+  attributes:
+    clusterName: {{ .Values.cluster.attributes.clusterName | quote }}
+    tags:
+{{ toYaml .Values.cluster.attributes.clusterTags | nindent 6 }}
+  networking:
+    kubeApiCIDR: {{ .Values.cluster.networking.kubeApiCIDR | default .Values.operator.KUBEAPI_CIDR| quote }}
+    kubeNodeCIDRS:
```
mjnagel: This needs an update, lowercase `s`?:
```
I0709 17:06:59.720915   91671 warnings.go:110] "Warning: unknown field \"spec.networking.kubeNodeCIDRS\""
```

```suggestion
    kubeNodeCIDRs:
```

src/pepr/uds-operator-config/templates/clusterconfig.yaml
```diff
@@ -0,0 +1,31 @@
...
+{{- if .Values.cluster.networking.kubeNodeCIDRS }}
+{{ toYaml .Values.cluster.networking.kubeNodeCIDRS | nindent 6 }}
```
mjnagel: Value should also have the lowercase `s` (same with the schema).

src/pepr/operator/controllers/config/config.ts
```diff
@@ -69,43 +97,127 @@ export async function updateUDSConfig(config: kind.Secret) {
...
+    kubeApiCidr: cfg.spec?.networking?.kubeApiCIDR || "",
+
+    // Static CIDRs to use for KubeNodes instead of k8s watch. Comma separated list of CIDRs.
+    kubeNodeCidrs: cfg.spec?.networking?.kubeNodeCIDRs || [],
+
+    // Track if UDS Core identity-authorization layer is deployed
+    isIdentityDeployed: false,
+  };
+
+  configLog.info(UDSConfig, "Loaded UDS Config");
```
mjnagel: This will log out the `authserviceRedisUri` which may contain a user/pass. We should make sure we don't leak secrets here.

mjnagel (commented):
I'm still debugging but I performed an upgrade with some overrides and things did not go great 😅. I had `operator.KUBEAPI_CIDR` overridden, and in my resulting `clusterconfig` CR I see my override. I also see it in the logs:
```
{"level":30,"time":1752095246910,"pid":18,"hostname":"pepr-uds-core-watcher-94859bb84-zfsr9","component":"operator.config","domain":"mjnagel.dev","adminDomain":"admin.mjnagel.dev","caCert":"","allowAllNSExemptions":false,"authserviceRedisUri":"","kubeApiCidr":"172.0.0.0/8","kubeNodeCidrs":[],"isIdentityDeployed":false,"msg":"Loaded UDS Config"}
```

However...I don't see `Static CIDR`... from `initAPIServerCIDR` - I suspect there is a timing issue based on where those init functions are called, compared to when the config is loaded? We may need to move those into the main `pepr.ts` flow to sequence them properly 🤔

src/pepr/uds-operator-config/values.schema.json
TristanHoladay: hmmm, interesting. i did a similar test but i think i set it via `cluster.` and it worked just fine.

src/pepr/operator/controllers/config/config.ts
TristanHoladay: thanks for catching that. I created a function instead of in-lining the fix in case we want to expand this later and to make it explicit.

image

TristanHoladay: just thinking, though.... do we want to show the `****` if the Uri is `" "` or show empty string in that case?

src/pepr/operator/controllers/config/config.ts
```diff
@@ -3,109 +3,276 @@
...
+    let cfg: ClusterConfig = {};
+    let cfgSecret: kind.Secret = {};
 
-  log.info("Updated UDS Config based on uds-operator-config secret changes");
+    try {
+      cfg = await K8s(ClusterConfig).Get("uds-cluster-config");
+      cfgSecret = await K8s(kind.Secret).InNamespace("pepr-system").Get("uds-operator-config");
+    } catch (e) {
+      configLog.warn(e, "Error while fetching cluster config and operator config secret");
+    }
```
chance-coleman: Should this log be a bit more descriptive? seems like it would be useful to know more.

```suggestion
    } catch (error) {
        const err = error as Error;
        configLog.warn({
            error: err.message,
            stack: process.env.NODE_ENV === 'development' ? err.stack : undefined,
            operation: 'fetchConfig'
        }, "Failed to load cluster configuration");
        throw new Error(`Configuration load failed: ${err.message}`);
    }
```

src/pepr/operator/controllers/config/config.ts
```diff
@@ -3,109 +3,276 @@
...
+        ? parseInt(process.env.PEPR_RESYNC_DELAY_SECONDS, 10)
+        : 5,
+      lastSeenLimitSeconds: process.env.PEPR_LAST_SEEN_LIMIT_SECONDS
+        ? parseInt(process.env.PEPR_LAST_SEEN_LIMIT_SECONDS, 10)
+        : 300,
+      relistIntervalSec: process.env.PEPR_RELIST_INTERVAL_SECONDS
+        ? parseInt(process.env.PEPR_RELIST_INTERVAL_SECONDS, 10)
+        : 1800,
+    };
+    const watcher = K8s(ClusterConfig).Watch(async (cfg, phase) => {
```
chance-coleman: Add types to callback:

```suggestion
    const watcher = K8s(ClusterConfig).Watch(async (cfg: ClusterConfig, phase: WatchPhase) => {
```

src/pepr/operator/controllers/config/config.ts
```diff
@@ -3,109 +3,276 @@
...
+      resyncDelaySec: process.env.PEPR_RESYNC_DELAY_SECONDS
+        ? parseInt(process.env.PEPR_RESYNC_DELAY_SECONDS, 10)
+        : 5,
+      lastSeenLimitSeconds: process.env.PEPR_LAST_SEEN_LIMIT_SECONDS
+        ? parseInt(process.env.PEPR_LAST_SEEN_LIMIT_SECONDS, 10)
+        : 300,
+      relistIntervalSec: process.env.PEPR_RELIST_INTERVAL_SECONDS
+        ? parseInt(process.env.PEPR_RELIST_INTERVAL_SECONDS, 10)
+        : 1800,
+    };
```
chance-coleman: Mostly preference: using `process.env...` everywhere adds a lot of extra fluff when we could define top level variables to use everywhere. something like this:
```
const DEFAULT_CONFIG = {
  WATCH: {
    RESYNC_FAILURE_MAX: 5,
    RESYNC_DELAY_SEC: 5,
    LAST_SEEN_LIMIT_SEC: 300,
    RELIST_INTERVAL_SEC: 1800
  },
  PLACEHOLDERS: {
    ZARF_VAR_DOMAIN: '###ZARF_VAR_DOMAIN###',
    ZARF_VAR_CA_CERT: '###ZARF_VAR_CA_CERT###',
    ZARF_VAR_AUTHSERVICE_REDIS_URI: '###ZARF_VAR_AUTHSERVICE_REDIS_URI###'
  }
};
```

src/pepr/operator/controllers/config/config.ts
```diff
@@ -3,109 +3,276 @@
...
+      const authserviceUpdate: AuthServiceEvent = {
+        name: "global-config-update",
+        action: Action.UpdateGlobalConfig,
+        // Base64 decode the CA cert before passing to the update function
+        trustedCA: atob(UDSConfig.caCert),
+        redisUri: UDSConfig.authserviceRedisUri,
+      };
+      configLog.debug("Updating Authservice secret based on change to Redis URI");
+      await reconcileAuthservice(authserviceUpdate);
     }
```
chance-coleman: I believe this reconcile Authservice functionality is used at least twice, should dry this up.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -3,109 +3,276 @@
...
+    } else {
       UDSConfig.domain = "uds.dev";
     }
-    if (!UDSConfig.adminDomain || UDSConfig.adminDomain === "###ZARF_VAR_ADMIN_DOMAIN###") {
+    if (expose.adminDomain && expose.adminDomain !== "###ZARF_VAR_ADMIN_DOMAIN###") {
+      UDSConfig.adminDomain = expose.adminDomain;
+    } else {
       UDSConfig.adminDomain = `admin.${UDSConfig.domain}`;
     }
     // todo: Add logic to handle domain changes and update across virtualservices, authservice config, etc
```
chance-coleman, line 163: is this todo comment relevant?

src/pepr/operator/crd/validators/clusterconfig-validator.spec.ts
```diff
@@ -0,0 +1,87 @@
+/**
```
chance-coleman, line 1: Both this test file and the exempt-validator.spec.ts file will be excluded from the vitest coverage because of the vitest exclude block in the config: [link](https://github.com/defenseunicorns/uds-core/blob/main/vitest.config.js#L17)

We probably want to udpate that config to look like this:
```
import { defineConfig } from 'vitest/config';

const isCi = process.env.CI_COVERAGE_ALL === 'true';

export default defineConfig({
  test: {
    globals: true,
    environment: 'node',
    globalSetup: ['./vitest.setup.js'],
    include: ['**/*.spec.ts'],
    coverage: {
      provider: 'v8',
      reporter: ['text', 'html'],
      all: isCi,
      include: ['src/pepr/**'],
      exclude: ['**/docs-gen/**', '**/crd/generated/**', '**/crd/sources/**'],
    },
  },
});
```

We can do this in a follow on PR as well, just something I had to do in testing this code.

chance-coleman (commented):
Looking good! Few small observations.

src/pepr/operator/controllers/config/config.ts
mjnagel: It's still a todo here and not one intended to be resolved by this PR. It's being tracked in https://github.com/defenseunicorns/uds-core/issues/1220 though.

src/pepr/operator/crd/validators/clusterconfig-validator.spec.ts
mjnagel: I'll go ahead and update here just so that we don't forget.

src/pepr/operator/controllers/config/config.ts
mjnagel: Open to changing a bit, although I think it's worth providing some context here:
- this func is only called on startup of pepr
- we don't want to immediately throw errors from this, since the `uds-operator-config` secret might not exist (and that's okay)
- we DO throw errors to crash the process if the clusterconfig doesn't exist
- this should only fail if (1) the clusterconfig/secret don't exist or (2) kubeapi issues
- the log includes the error (`e`) so the full error is present in the log message already
- in general that error will essentially be the http response from the k8s api so we'd have to parse a bit differently looking at `e.data?.message` for example

I'm going to separate this out/clean this up a bit to be more clear what is happening, but I think if we just log the full error object we're probably okay here?

src/pepr/operator/controllers/config/config.ts
mjnagel: This is a direct copy/paste of the upstream Pepr watch: https://github.com/defenseunicorns/pepr/blob/a451070517adefdde49da50f47914181c8ac796f/src/lib/processors/watch-processor.ts#L59-L73

Essentially it just makes sure that we are running our watches in the admission pods the same way as Pepr runs its watches in the watcher pods. I think I'd still prefer to read all these envs for that reason (this could be modified by a downstream user if they need to tweak parameters). However definitely agree on copy/paste so I pulled this out into the utils file where it can be reused across the 3 watches we are running.

src/pepr/operator/controllers/config/config.ts
mjnagel: Pulling into a function.

src/pepr/operator/controllers/config/config.ts
chance-coleman: reviewed IRL looks good to me

src/pepr/operator/controllers/config/config.ts
```diff
@@ -3,109 +3,270 @@
...
+    } catch (e) {
+      configLog.error(e);
+      throw e;
+    }
+  }
+}
+
+// Helper function for redacting sensitive values from the UDS Config
+function redactConfig() {
+  const authserviceRedisUri = UDSConfig.authserviceRedisUri ? "****" : "";
```
chance-coleman, line 217: Would it make sense to make this a configuration based loop that controls which values are redacted? Then we could add and remove those values anytime we or the user needs. I don't think this needs to be captured in this PR, can be a follow on enhancement.

mjnagel: I can update if we want to. With the current config the `authserviceRedisUri` is the only sensitive value (since it could have a user/pass). Might vote for extending/changing this if/when we have another sensitive value?

chance-coleman (commented):
i deployed this branch with slim-dev + e2e-tests and the clusterconfig resource looked like this: 
```
apiVersion: uds.dev/v1alpha1
kind: ClusterConfig
metadata:
  annotations:
    meta.helm.sh/release-name: uds-operator-config
    meta.helm.sh/release-namespace: pepr-system
  creationTimestamp: "2025-07-21T13:32:28Z"
  generation: 1
  labels:
    app.kubernetes.io/instance: uds-operator-config
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: uds-operator-config
    helm.sh/chart: uds-operator-config-0.1.0
  name: uds-cluster-config
  resourceVersion: "985"
  uid: 4d2ecf09-4c6b-48f7-b056-d51ab3f7d2db
spec:
  attributes:
    clusterName: ""
    tags: []
  expose:
    adminDomain: ""
    caCert: ""
    domain: uds.dev
  networking:
    kubeApiCIDR: ""
    kubeNodeCIDRs: []
  policy:
    allowAllNsExemptions: false
```
Not the biggest fan of empty fields. They are slightly confusing and made we wonder if i misconfigured something (even though i didn't configure anything). Not sure what to do with this information, could be follow up task to cleanup, also okay leaving it if there's a good reason. To be clear i'm mostly concerned with empty values.

src/pepr/operator/controllers/config/config.ts
chance-coleman: works for me! i was just looking forward and thinking that we could make use of this clusterconfig for a lot of things and potentially one day that includes more secure fields.

chance-coleman (approved):
tested locally, LGTM. thanks for taking this on

PR #1235, opened 2025-01-27
chore: add docs for wiring in pre-reqs for S3/MetalLB

Racer159:
## Description

Adds some docs for S3 pre-reqs and mentions the MetalLB package
## Related Issue

Fixes #N/A

## Type of change

## Steps to Validate

## Checklist before merging


chance-coleman (dismissed):
LGTM, thanks!

docs/reference/UDS Core/prerequisites.md
```diff
@@ -136,3 +140,180 @@ Metrics server is provided as an optional component in UDS Core and can be enabl
...
+                  - uds-loki-chunks
+                  - uds-loki-ruler
+                  - uds-loki-admin
+                copyPassword:
+                  enabled: true
+
+- name: core-logging
+  ...
+  overrides:
+    loki:
```
mjnagel, line 241: We should also add the overrides to lock down the egress traffic to strictly what is necessary here (`uds-loki-config` chart has some [values for storage egress](https://github.com/defenseunicorns/uds-core/blob/add-docs-s3-lb-prerequisites/src/loki/chart/values.yaml) that when not specified default to egress anywhere).

docs/reference/UDS Core/prerequisites.md
```diff
@@ -136,3 +140,180 @@ Metrics server is provided as an optional component in UDS Core and can be enabl
...
+                  - uds-velero
+                copyPassword:
+                  enabled: true
+                  secretIDKey: AWS_ACCESS_KEY_ID
+                  secretPasswordKey: AWS_SECRET_ACCESS_KEY
+
+- name: core-backup-restore
+  ...
+  overrides:
+    velero:
```
mjnagel, line 357: Same comment here on the egress policy overrides, see [the values here](https://github.com/defenseunicorns/uds-core/blob/add-docs-s3-lb-prerequisites/src/loki/chart/values.yaml).

docs/reference/UDS Core/prerequisites.md
```diff
@@ -136,3 +140,180 @@ Metrics server is provided as an optional component in UDS Core and can be enabl
...
+                bucket: "uds-velero"
+                config:
+                  region: ""
+                  s3ForcePathStyle: true
+                  s3Url: "http://uds-minio-hl.minio.svc.cluster.local:9000/"
+```
+
+:::caution
+If you are using the in-cluster Minio Operator UDS Package for backups you must ensure that the volumes that back that storage are themselves backed up!  Cluster or deployment issues may result in a loss of Minio as well as the application you intend to back up.
+:::
```
mjnagel: nit, but I wonder if there's a better place to put this caution note given that it applies to both Loki and Velero (and is currently "nested" under the Velero section). We could just copy/paste it to Loki but that doesn't feel right 😅

docs/reference/UDS Core/prerequisites.md
```diff
@@ -136,3 +140,180 @@ Metrics server is provided as an optional component in UDS Core and can be enabl
...
     - metrics-server
 ```
+
+#### Loki
+
+The Loki deployment is (by default) backed by an object storage provider for log retention.  For cloud environments you can wire this into the environment's storage provider with the following overrides:
+
+```yaml
+loki:
+  values:
```
mjnagel: nit: this confused me initially (especially when comparing against the minio example) - can we add the boilerplate of `overrides`, package, component, etc?

docs/reference/UDS Core/prerequisites.md
```diff
@@ -136,3 +140,180 @@ Metrics server is provided as an optional component in UDS Core and can be enabl
...
+                  key: secretKey
+```
+
+#### Velero
+
+The Velero deployment is (by default) backed by an object storage provider for backup retention.  For cloud environments you can wire this into the environment's storage provider with the following overrides:
+
+```yaml
+velero:
+  values:
```
mjnagel: Same comment as loki on adding the boilerplate to the example.

mjnagel (commented):
A few minor comments, appreciate you taking the time to test this out for us! One other note we _may_ want to add, while the examples are `values`, for a prod bundle the end user would likely want `variables` for connecting to cloud storage at least.

docs/reference/UDS Core/prerequisites.md
slaskawi: I echo what @mjnagel said. It would be great to align these two examples.

docs/reference/UDS Core/prerequisites.md
slaskawi: Also, we we have any other recommendations for backups? For example, in the [UDS docs](https://uds.defenseunicorns.com/reference/uds-core/prerequisites/#default-storage-class), we wrote:

> For instance, EBS volumes have a minimum size of 1Gi, which could lead to unexpected behavior, especially during Velero’s CSI backup and restore process. These constraints may also necessitate adjustments to default PVC sizes, such as Keycloak’s PVCs, which default to 512Mi in devMode.

Perhaps we could align these bits of the documentation somehow?

Racer159: added a common caution under that

mjnagel (approved):

PR #1238, opened 2025-01-29
fix: address AKS ci flakiness

noahpb:
## Description
Creates 3 worker nodes for the AKS CI User nodepool. Fixes current setup where zero nodes are being added to the pool. Also switches to using [API Server VNET](https://learn.microsoft.com/en-us/azure/aks/api-server-vnet-integration) integration to address dropped connection issues with the API server during UDS Core install. 

## Type of change

## Steps to Validate
- N/A
## Checklist before merging


chance-coleman (dismissed):

mjnagel (approved):
LGTM based on CI results and the write up 👍

PR #1250, opened 2025-02-03
fix: added network restriction tests

### LINKED ISSUE #1207: Add CI testing of network restrictions

Currently the operator generates network policies based on the Package CR spec (`network.allow`). These are not currently validated in CI (e2e) to confirm that they are properly restricting ingress/egress.

Add testing of network restrictions to e2e test suite covering:
- ingress restictions
- egress restrictions
- allowed and denied networking

### Additional context

As we move to istio ambient we will be [changing the mechanism for network traffic restrictions](https://github.com/defenseunicorns/uds-core/issues/1028) and would like to have confidence that the new method of restricting traffic is working 1:1 with the current network policies.

chance-coleman:
## Description

Currently the operator generates network policies based on the Package CR spec (network.allow). This adds tests that can be used to validate the networks locally and in CI.

### Test Categories & Cases

1. Default Deny Behavior
    - External traffic (demo-8080.uds.dev) is blocked by default.
    - Internal traffic (curl-2 in curl-test-1) is blocked by default.
    - Egress to Google (www.google.com) is blocked by default.
    - Traffic to an explicitly blocked port (9999) is denied.
1. Wide Open Ingress & Egress
    - Ingress & Egress policies allow pod-to-pod communication.
    - Traffic to an unallowed port (8081) is blocked.
    - Traffic to an undefined port (9999) is blocked.
    - Wide open Egress allows external traffic (Google) to succeed.
1.  Egress to Anywhere
    - Egress Anywhere policy allows unrestricted outbound traffic.
    - External requests (demo-8080.uds.dev) succeed.
    - Requests to Google succeed.
1. RemoteNamespace Policy Validation
    - Ingress & Egress policies between different namespaces allow communication.
    - Egress to Google is still blocked due to restricted policy.
    - Traffic to a blocked port (9999) is denied.
1. Kube API Restrictions
    - Kubernetes API calls fail with 401 Unauthorized.
    - External traffic (Google) is blocked.
    - Internal traffic to blocked port (9999) is denied.
1. RemoteCIDR Policy Validation
    - RemoteCIDR-based policies allow inbound connections.
    - External traffic (Google) is allowed due to unrestricted RemoteCIDR policy.

## Related Issue

Fixes #1207 

## Type of change

## Steps to Validate
- `uds run slim-dev`
- `uds run -f src/test/tasks.yaml create-deploy-test-resources`
- `uds run -f src/test/tasks.yaml e2e-test`

## Checklist before merging


test/jest/network.spec.ts
```diff
@@ -0,0 +1,411 @@
...
+  namespace: string,
+  name: string,
+  patchPayload: object[],
+  waitTime = 3000,
+): Promise<void> {
+  const patchJson = JSON.stringify(patchPayload);
+  const command = `kubectl patch package ${name} -n ${namespace} --type=json -p='${patchJson}'`;
+  await execAsync(command);
+  await new Promise(resolve => setTimeout(resolve, waitTime)); // Allow changes to propagate
+}
```
mjnagel: We should probably use the `K8s` (kubernetes fluent client) to apply/patch here. Not a 1:1 example, but we apply netpols this way in our [operator code](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/policies.ts#L153).

src/test/tasks.yaml
```diff
@@ -146,9 +146,95 @@ tasks:
...
+
+      - description: "Run KubeAPI Network Restrictions tests"
+        cmd: |
+          npx jest test/jest/network.spec.ts -t "KubeAPI Network Restrictions"
+        dir: test/jest
+
+  - name: reapply-testing-apps
+    actions:
+      - description: Delete the Test Curl App and Test Tenant App
+        cmd: kubectl delete -f src/test/app-curl.yaml && kubectl delete -f src/test/app-tenant.yaml
```
mjnagel: On the delete here you can probably use `K8s` / KFC to delete pods (not 1:1 example but we have a generic delete call [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/utils.ts#L72)).

test/jest/network.spec.ts
mjnagel: Might also be able to use `beforeEach` to do an `Apply` of the original Package CR before each test and save needing to delete/apply that.

src/test/tasks.yaml
mjnagel: Alternatively might be able to:
1. Delete the Package CRs by name
2. Apply the manifests
3. Delete the pods by label
4. Wait for pods to be ready

sgettys:
Can you add another test for Explicit Port Traffic Restrictions?
* test-tenant-app has ingress rule for 8080 from curl-test
* curl-test tries to connect to 8081

this would catch the netpols leak that is caused by ztunnel that the auth pols will fix

test/jest/network.spec.ts
chance-coleman: following up on offline conversation here, KFC won't work because i'm editing the package here, which isn't a valid K8s kind. Potentially could do some wrapper code to make it work but not sure it's worth it.

src/test/tasks.yaml
chance-coleman: this helps speed up the tests but not enough in my opinion. Originally these tests would take 8 minutes to run, majority of which was the deletion and apply of resources, with these proposed changes it takes about 4 minutes

src/test/tasks.yaml
```diff
@@ -3,6 +3,22 @@
...
+      - description: "Run Test Apps E2E tests"
+        task: create-deploy-test-resources
+      - description: "Run Network Tests"
+        cmd: |
+          npx jest test/jest/network.spec.ts
+        dir: test/jest
+      - description: "Remove Test Apps resources"
+        task: remove-test-resources
+
+  - name: create-deploy-test-resources
```
mjnagel: nit: changing this to just `create-deploy` as the name might clean it up a bit since any calls to this task would already have the context of this being for the test task file (for example where we call it right now it feels a bit redundant as `test-resources:create-deploy-test-resources`). If you agree here probably also the same for `remove-test-resources`?

test/jest/network.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+// Execute commands inside a pod
+async function execInPod(namespace: string, podName: string, containerName: string, command: string[]) {
+  const cmd = `kubectl exec -n ${namespace} ${podName} -c ${containerName} -- ${command.join(" ")}`;
+  try {
+    const { stdout } = await execAsync(cmd);
+    return { stdout: stdout.trim(), stderr: "" };
+  } catch (error) {
+    return { stdout: "000", stderr: "Curl failed" };
+  }
+}
```
mjnagel: I know we discussed the issues with using KFC for anything to do with the Package CRs, but I think these two uses of `kubectl` could be migrated to KFC/the k8s javascript client to avoid the need for any host execution of CLI commands?
- Get Pods example: https://github.com/defenseunicorns/uds-core/blob/v0.32.0/src/pepr/operator/controllers/istio/injection.ts#L127 (you should be able to use `WithLabel` here to do an equivalent call)
- Exec into pod example: https://github.com/defenseunicorns/uds-core/blob/v0.32.0/src/pepr/istio/index.ts#L75-L101 (note that this uses client-node because KFC doesn't support exec)

This might make the exec one a bit more complex but I think it might be worth trying to do to buy us the type safety and just avoid shelling out to `kubectl`.

test/jest/network.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+// Curl Commands
+const CURL_EXTERNAL = [
+  "curl",
+  "-s",
+  "-o",
+  "/dev/null",
+  "-w",
+  "%{http_code}",
+  "https://demo-8080.uds.dev",
+];
```
mjnagel: nit on the name of this one - technically curls to `*.uds.dev` will stay (routing wise) inside of the cluster, and the netpols would just need to allow egress to the gateway in cluster. Based on how you're using this in the tests I think it's still valid, just wanted to make that clarification. Maybe could just rename `CURL_GATEWAY`?

test/jest/network.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+  });
+
+  test.concurrent("RemoteCidr Restrictions", async () => {
+    // Validate successful request when using RemoteCidr
+    const success_response = await execInPod("test-admin-app", testAdminApp, "curl", INTERNAL_CURL_COMMAND_7);
+    expect(success_response.stdout).toBe("200");
+
+    // Validate successful request to Google because of wide open remoteCidr
+    const denied_google_response = await execInPod("test-admin-app", testAdminApp, "curl", GOOGLE_CURL);
+    expect(denied_google_response.stdout).toBe("200");
```
mjnagel: nit: name seems wrong here? should be a successful google response?

test/jest/network.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    const denied_google_response = await execInPod("curl-test-6", curlPodName8, "curl-8", GOOGLE_CURL);
+    expect(denied_google_response.stdout).toBe("000");
+
+    // Default Deny for Blocked Port
+    const blocked_port_curl = getCurlCommand("curl-2", "curl-test-1", 9999);
+    const denied_port_response = await execInPod("curl-test-6", curlPodName8, "curl-8", blocked_port_curl);
+    expect(denied_port_response.stdout).toBe("503");
+  });
+
+  test.concurrent("RemoteCidr Restrictions", async () => {
```
mjnagel, line 260: Might not be tracing things right here but I think the pod we're using for the curls here is actually a [`remoteGenerated: Anywhere`](https://github.com/defenseunicorns/uds-core/blob/network-testing/src/test/app-admin.yaml#L41)? Is that intentional/should this test name be different?

test/jest/network.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      "%{http_code}",
+      "-k",
+      "-H",
+      "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)",
+      `https://kubernetes.default.svc.cluster.local/api/v1/namespaces`,
+    ];
+
+    // Validate unauthorized response when using Kube API ( because of missing authorization token )
+    const success_unauthorized_response = await execInPod("curl-test-6", curlPodName8, "curl-8", kubeApi_curl);
+    expect(success_unauthorized_response.stdout).toContain("401");
```
mjnagel: So there's a few issues here that cause the 401:
- The api endpoint we're trying to hit isn't allowed with our default svc account. You can however hit just the base API regardless of role: `https://kubernetes.default.svc.cluster.local/api`
- The `cat` command here is not getting evaluated in the right place (i.e. it's being evaluated on the host rather than in the pod). If you change the exec command to run inside a basic shell and wrap everything in single quotes this will work (execInPod cmd should end with `-- sh -c '${command.join(" ")}'`), but that might change if you want to try the client-node exec approach.

With those two changes locally I got this to respond with a 200 which feels better than checking for a 401 😄

src/test/app-curl.yaml
```diff
@@ -0,0 +1,479 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    uds: curl-testing-namespace
+  name: curl-test-1
```
mjnagel: I struggled with the naming when trying to trace through these tests. Two different approaches that would help I think:
1. Number all the namespaces + letter all the deployments/pods. For example `curl-test-1` namespace would have two deployments/packages `curl-1-a` and `curl-1-b`. I think this would help because I started getting confused further along by the pod names/numbers not lining up with the namespace names/numbers.
2. Maybe better but not sure if this is clean: Name everything by the test functionality. For example `curl-test-1` namespace would be `curl-deny-all` (pods could probably just be numbered here), `curl-test-2` namespace would be `curl-ingress-port` or similar, etc.

It's probably harder to go fix all the places for option 2 but I think it would make the tests and manifests a lot more readable - I'm stepping through the tests and looking at the manifests on the side to make sure I understand desired/expected behavior with each, if they had descriptive names it might be easier to validate.

test/jest/network.spec.ts
mjnagel: If we do stick with the `kubectl` commands (I don't feel as strongly on exec because of the complexity) let's use `uds zarf tools kubectl` since we generally don't assume anything beyond `uds` for running our tasks and tests.

test/jest/network.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+      "-o",
+      "/dev/null",
+      "-w",
+      "%{http_code}",
+      "http://curl-4.curl-test-2.svc.cluster.local:8081",
+    ];
+
+    // Deny request when port is not allowed on ingress
+    const denied_incorrect_port_response = await execInPod("test-admin-app", testAdminApp, "curl", CURL_INTERNAL_8081);
+    expect(denied_incorrect_port_response.stdout).toBe("503");
```
mjnagel, line 180: Just want to call out here that with authorization policies this specific scenario (potentially other tests) will end up returning a 403 instead of a 503 I believe. I'm okay updating at that point if we want, or maybe we just check that it's a code in the range of 400-599 ([ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)).

mjnagel (commented):
Happy to walk through any of these comments synchronously. From what I can tell the tests look like they provide good coverage so mostly comments on the execution and a few things I think we can do to make things cleaner.

test/jest/network.spec.ts
chance-coleman: Addressed this specific test and we can revisit other tests with authpols

src/test/app-curl.yaml
chance-coleman: went with a prefix of `curl-ns` and `curl-pkg` followed by test type.

test/jest/network.spec.ts
chance-coleman: was able to get both of these kubectl commands moved over to KFC and client-node

mjnagel (dismissed):

test/jest/network.spec.ts
```diff
@@ -199,6 +199,28 @@ describe("Network Policy Validation", () => {
...
+    ];
+
+    // Validate that request is successful when using Ingress Gateway Bypass
+    const success_response = await execInPod("test-admin-app", testAdminApp, "curl", authserivce_curl);
+    expect(success_response.stdout).toBe("503");
+
+    // Validate that request is successful when using Ingress Gateway Bypass
+    const success_response2 = await execInPod("test-admin-app", testAdminApp, "curl", authserivce_curl_header);
+    expect(success_response2.stdout).toBe("503");
+  });
```
mjnagel: naming/comments these should be "blocked" / unsuccessful rather than success?

chance-coleman: shoot, yeah good call

mjnagel (approved):

PR #1259, opened 2025-02-05
chore(ci): add e2e tests for cloud distros

### LINKED ISSUE #729: Run validation tasks against nightly deployments

noahpb:
## Description
Adapts our nightly pipelines (RKE2, EKS and AKS) to run our e2e tests. This PR generally takes our e2e tests that we currently use for k3d and expands them to run against additional k8s distributions and architectures. Notable changes are:
- Adds public load balancers for RKE2 to make Ingress Gateways accessible from github workflows
- Adds a fix for metrics scraping of kube-apiserver components for STIG'd RKE2
- Adds coredns setup tasks for each distro to allow connectivity to `*.uds.dev` from apps within the cluster
- Adds a new task to `./tasks/test.yaml` called `uds-core-non-k3d` for triggering e2e tests against various distros. This task includes an optional input called `EXCLUDED_PACKAGES` that allows callers to skip uneccessary tests, such as `metrics-server` for some distros.
- Allows configuration of the `architecture` for the `./src/test/package`
- Adds a new task to `./tasks/utils.yaml` called `setup-hosts` that configures the local `/etc/hosts` file to point to `*.uds.dev` and `*.admin.uds.dev` domains that are hosted on the cluster that e2e tests are running against.
- Adds timeouts to the curl commands in `./test/jest/network.spec.ts` to produce more meaningful error messages when timeouts are occurring during connectivity tests (before these would run indefinitely and the jest timeout would ultimately be reached).
- In `./test/jest/network.spec.ts`, changes explicit `expect().toBe("503")` responses to `expect().not.toBe("200")` to account for different behavior of cluster CNIs when rejecting traffic due to `NetworkPolicy` deny rules.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/729

## Type of change

## Steps to Validate

## Checklist before merging


src/test/tasks.yaml
```diff
@@ -20,9 +23,13 @@ tasks:
 
   - name: create-deploy
     description: Test app used for UDS Core validation
+    inputs:
+      architecture:
+        description: "System architecture that the test-apps package should be built for."
+
```
mjnagel: Default here should be `${UDS_ARCH}` I think to make sure that our github runners use an autodetected ARCH.

tasks/test.yaml
```diff
@@ -73,6 +80,11 @@ tasks:
 
   - name: e2e-tests
     description: "E2E Test all packages"
+    inputs:
+      architecture:
+        description: "System architecture that the test-apps package should be built for."
+        required: true
+        default: "arm64"
```
mjnagel: https://github.com/defenseunicorns/uds-core/pull/1259/files#r1971899902

.github/workflows/test-eks.yaml
```diff
@@ -92,12 +92,19 @@ jobs:
...
       - name: Deploy Core Bundle
         env:
           UDS_CONFIG: .github/bundles/eks/uds-config.yaml
         run: uds deploy .github/bundles/eks/uds-bundle-uds-core-eks-nightly-*.tar.zst --confirm
         timeout-minutes: 30
 
+      - name: Test UDS Core
+        run: uds run -f tasks/test.yaml uds-core-non-k3d --set EXCLUDED_PACKAGES="metrics-server"
+        continue-on-error: true
+
```
noahpb, line 107: Noting the `continue-on-error: true` here. EKS VPC CNI [permits](https://github.com/defenseunicorns/uds-core/actions/runs/13862860013/job/38795298976?pr=1259#step:15:708) outbound traffic to the internet for the network tests, despite not having a NetworkPolicy to explicitly allow this. Ignoring failures for now until we decide how to proceed.

src/istio/tasks.yaml
```diff
@@ -6,7 +6,7 @@ tasks:
     inputs:
       validate_passthrough:
         description: Whether to validate the passthrough gateway
-        default: "true"
+        default: "false"
```
chance-coleman: Can provide me some context to why this change was needed?

tasks/iac.yaml
```diff
@@ -118,18 +120,19 @@ tasks:
...
+            if [ $(uds zarf tools kubectl get po,job -A --no-headers=true | egrep -v 'helm-install|Running|Complete' | wc -l) -gt 0 ]; then
               echo "Waiting for cluster components to be ready...";
               sleep 5;
             else
               echo "Cluster is ready"
               break
             fi
           done
-          uds zarf tools kubectl apply -f ./metallb.yaml
+          #uds zarf tools kubectl apply -f ./metallb.yaml
```
chance-coleman: can we just remove this line?

tasks/utils.yaml
```diff
@@ -64,3 +64,158 @@ tasks:
...
+                  rewrite stop {
+                    name regex (.*\.admin\.uds\.dev) admin-ingressgateway.istio-admin-gateway.svc.cluster.local answer auto
+                    }
+                  rewrite stop {
+                  name regex (.*\.uds\.dev) tenant-ingressgateway.istio-tenant-gateway.svc.cluster.local answer auto
+                    }
+              }
+          kind: ConfigMap
+          metadata:
+            name: rke2-coredns-rke2-coredns
```
chance-coleman: could we remove the duplicity and simplify this name?

src/istio/tasks.yaml
noahpb: I set this to false because the nightly CI tests (rke2, eks, aks) currently don't deploy the passthrough gateway, and because it was removed by default. I now realize though that the k3d tests rely on this check being present as the k3d-standard bundle still deploys this component. I can add this back in and start deploying the passthrough gateway for the nightly tests. Thoughts?

tasks/iac.yaml
noahpb: Yes, good call

tasks/utils.yaml
noahpb: Sort of? I tried to configure the default coredns settings for rke2, but it's baked into the install scripts that are run during the AMI build process. I tried manually modifying the helm chart for coredns to add an additional volume mount for a `coredns-custom` configmap like seen on other distros, but it was tricky and unreliable. I figured the safest option would be to modify the default configmap for coredns. I do agree though that having duplicate files with similar config is not desirable. I will explore templating these.

.github/test-infra/aws/rke2/scripts/user_data.sh
```diff
@@ -42,32 +40,39 @@ spec:
...
+  name: aws-load-balancer-controller
   namespace: kube-system
 spec:
-  chart: longhorn
-  repo: https://charts.longhorn.io
-  version: 1.7.1
-  targetNamespace: kube-system 
+  chart: aws-load-balancer-controller
+  repo: https://aws.github.io/eks-charts
+  version: 1.11.0
```
mjnagel: Is this (and longhorn below) captured by renovate currently? I'm not seeing any regex matches that would pick this up as-is, but we should be able to use one of the `#` comment matches right before the version.

tasks/iac.yaml
```diff
@@ -118,18 +120,19 @@ tasks:
               break
             fi
           done
-
           # wait for cluster components
           while true; do
-            if [ $(uds zarf tools kubectl get po,job -A --no-headers=true | egrep -v 'Running|Complete' | wc -l) -gt 0 ]; then
+            if [ $(uds zarf tools kubectl get po,job -A --no-headers=true | egrep -v 'helm-install|Running|Complete' | wc -l) -gt 0 ]; then
```
mjnagel: Probably missing the obvious - what was this change needed for? Iirc this originally was used to make sure all pods/jobs were either completed or running before moving on.

tasks/utils.yaml
```diff
@@ -64,3 +64,158 @@ tasks:
...
+  - name: rke2-allow-prom-kube-dns
+    actions:
+      - description: Create NetworkPolicy to allow Prometheus to scrape kube-dns
+        cmd: |
+          uds zarf tools kubectl apply -f - <<EOF
+          apiVersion: networking.k8s.io/v1
+          kind: NetworkPolicy
+          metadata:
+            name: allow-prometheus-to-kube-dns
+            namespace: kube-system
```
mjnagel, line 96: Don't have to do this now, but still worth evaluating if this should be a conditional netpol in core?

test/jest/network.spec.ts
```diff
@@ -143,12 +147,13 @@ describe("Network Policy Validation", () => {
 
   test.concurrent("Denied Requests by Default and Incorrect Ports and Labels", async () => {
     // Default Deny when no Ingress or Egress defined or Exposed Endpoints
+    // The HTTP response code could either be 000 or 503, depending on the K8s distro
     const denied_external_response = await execInPod("curl-ns-deny-all", curlPodName1, "curl-pkg-deny-all-1", CURL_GATEWAY);
-    expect(denied_external_response.stdout).toBe("000");
-
+    expect(denied_external_response.stdout).not.toBe("200");
```
mjnagel: My only concern with this is that it might allow some things to slip through. Looking at the [HTTP Codes that are possible](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status), we really would want to assert that the response code is not 100-399 I think ("information"/"success"/"redirect"). Or alternatively assert 400+ or 000?

src/istio/tasks.yaml
mjnagel: Could we default this to a variable, with that variable defaulted to `true` and then pass in `false` for the e2e tests? Imo we probably don't need to add the passthrough for nightly, unless it makes this testing easier.

.github/workflows/test-eks.yaml
mjnagel: I think this makes sense for now to keep moving - definitely good to track this down and identify if there are any solutions out there though. Otherwise this might be a limitation for us to document with Core/EKS.

.github/test-infra/aws/rke2/scripts/user_data.sh
noahpb: I'm not sure how I'd be able to start capturing these without adding a `customManager` to renovate to look for these charts. I tried experimenting locally with the `# renovate: ...` comments but have not had any luck.

tasks/iac.yaml
noahpb: This change was added to filter out any of the pods that are spawned by rancher's helm controller. In some cases, a helm release may succeed but the `helm-install` pod dedicated to rolling out the helm release can end up in a failed state.

.github/test-infra/aws/rke2/scripts/user_data.sh
noahpb: Wait, hold on. I might be able to get this with all of the `regex` custom managers we have in uds-common.

tasks/iac.yaml
mjnagel: We actually encountered an error that is semi-related over the weekend:
- MetalLB CRDs don't exist: https://github.com/defenseunicorns/uds-core/actions/runs/14013744264/job/39236727846#step:12:39
- Helm install of MetalLB failed/restarting: https://github.com/defenseunicorns/uds-core/actions/runs/14013744264/job/39236727846#step:14:75

While this PR does remove the metallb install I think, I'm worried we might hit similar issues with this check. We probably need one check for `job` resources to make sure they complete, and a separate check for pods to make sure they are running or completed? Do you have more insight into what these "false failures" are? Otherwise I'm worried we'll hit some race conditions here because this will hide real failures.

.github/test-infra/aws/rke2/scripts/user_data.sh
mjnagel: My hope was that we could use [this match](https://github.com/defenseunicorns/uds-common/blob/main/config/renovate.json5#L133-L147) for this, we'd just need to make sure we set all the config appropriately:
- [helm datasource](https://docs.renovatebot.com/modules/datasource/helm/)
- [helm versioning](https://docs.renovatebot.com/modules/versioning/helm/)
- registryUrl: set to the repo I think
- depName: set to the chart name

You could also cross-reference this other matcher that we use for helm charts in zarf files: https://github.com/defenseunicorns/uds-common/blob/main/config/renovate.json5#L206-L223

tasks/iac.yaml
noahpb: Good point, thanks for calling this out! I updated the `rke2-cluster-ready` task to wait for all 4 cluster nodes to provision and wait for all `helm-install` pods to be in a `Completed` state before continuing.

.github/test-infra/aws/rke2/scripts/user_data.sh
noahpb: Thanks for the pointer! I had to move the `HelmChart` resources out of the `.sh` script and into a new `.yaml` file in order for the first regex manager you linked in the previous comment to be able to pick up these `HelmCharts`. I successfully validated this works as expected locally.

test/jest/network.spec.ts
noahpb: Good point. I'll update these to check that it's either less than 100 or greater than 399.

noahpb: Updated tests in https://github.com/defenseunicorns/uds-core/pull/1259/commits/161d3726b9cfaffee6f041a2965e4ae76902f272

tasks/iac.yaml
```diff
@@ -109,27 +111,32 @@ tasks:
     actions:
       - task: rke2-nodes-ready
       - cmd: |
-          # wait for at least 3 nodes
+          # wait for nodes to be ready
           while true; do
-            if [ $(uds zarf tools kubectl get nodes -o jsonpath='{range .items[*]}{.status.conditions[-1].type}={.status.conditions[-1].status}{"\n"}{end}' | egrep -i '^ready.*true' | wc -l) -lt 3 ]; then
-              echo "Waiting for at least 3 nodes to be ready...";
+            if [ $(uds zarf tools kubectl get nodes -o jsonpath='{range .items[*]}{.status.conditions[-1].type}={.status.conditions[-1].status}{"\n"}{end}' | egrep -i '^ready.*true' | wc -l) -lt 4 ]; then
```
mjnagel: Was this an artifact of some local testing? Looking at the tf vars I think we would have 3 nodes - https://github.com/defenseunicorns/uds-core/blob/feat/e2e-test-nightly/.github/test-infra/aws/rke2/variables.tf#L38-L48 - which may be why the RKE2 tasks are failing in the latest run?

test/jest/network.spec.ts
```diff
@@ -254,7 +278,7 @@ describe("Network Policy Validation", () => {
     // Default Deny for Blocked Port
     const blocked_port_curl = getCurlCommand("curl-pkg-deny-all-2", "curl-ns-deny-all", 9999);
     const denied_port_response = await execInPod("curl-ns-kube-api", curlPodName8, "curl-pkg-kube-api", blocked_port_curl);
-    expect(denied_port_response.stdout).toBe("503");
+    expect(denied_port_response.stdout).not.toBe("200");
```
mjnagel:
```suggestion
    expect(isResponseError(denied_port_response)).toBe(true);
```
I think?

.github/workflows/test-eks.yaml
mjnagel: @noahpb on this one did we identify if overrides to the eksctl config would work here?

noahpb: I think you're referring to the enable network policy override? Last we checked, it's not [supported](https://redirect.github.com/eksctl-io/eksctl/issues/7770) in eksctl. Perhaps we can enable this when we switch to tf modules.

tasks/iac.yaml
noahpb: So there's 3 control nodes and one worker node (the extra control node is the bootstrap node). I'll look into why it started failing.

.github/workflows/test-eks.yaml
noahpb: I am seeing ways to [workaround](https://redirect.github.com/2i2c-org/infrastructure/issues/4660) this in eksctl that we could explore.

mjnagel: Yeah going off of that issue in eksctl I had thought we could install the vpc-cni separately as an addon to enable netpols potentially (seems like maybe the same as what you linked in that workaround). Let's leave that as a follow-on though maybe?

.github/test-infra/aws/rke2/iam.tf
```diff
@@ -77,6 +77,20 @@ data "aws_iam_policy_document" "aws_ccm" {
   }
 }
 
+data "local_file" "helm_template" {
+  filename = "./scripts/helmchart-template.yaml"
+}
+
+data "http" "aws-lb-controller-iam" {
+  url = "https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/${split(":", yamldecode(local.aws_load_balancer_controller_template[4]).image)[1]}/docs/install/iam_policy_us-gov.json"
```
mjnagel: This (and the supporting yaml/split logic) feel incredibly clunky and error-prone compared to just renovate-ing this directly? There is already a matcher in our common renovate config for this type of string, the only necessary change is [adding support for TF file matching](https://github.com/defenseunicorns/uds-common/pull/464). Seems like this would be a simpler approach?

tasks/iac.yaml
mjnagel:
```suggestion
            if [ $(uds zarf tools kubectl get po,job -A --no-headers=true | egrep -v 'Running|Complete' | wc -l) -gt 0 ]; then
```

If I'm following correctly, helm-installs should already be complete before this, so we don't need to filter them out?

tasks/test.yaml
```diff
@@ -53,11 +58,21 @@ tasks:
...
     actions:
       - cmd: |
-          for package in src/*; do
-            uds run -f ${package}/tasks.yaml validate --no-progress
+          for package in $(ls src); do
+            if [ ! $(echo ${EXCLUDED_PACKAGES} | grep ${package}) ]; then
+              if [ ${package} = "istio" ]; then
+                uds run -f src/${package}/tasks.yaml validate --no-progress --with validate_passthrough=${{ .inputs.validate_passthrough }}
+              fi
+             uds run -f src/${package}/tasks.yaml validate --no-progress
```
mjnagel: Should this be an if/else? Otherwise I think we're running the istio validation twice?

.github/test-infra/aws/rke2/scripts/user_data.sh
```diff
@@ -2,73 +2,37 @@
...
+./yq -i 'select(documentIndex < 4)' helmchart-template.yaml
+envsubst < helmchart-template.yaml > /var/lib/rancher/rke2/server/manifests/00-helmcharts.yaml
+HELM_LATEST=$(curl -L --silent --show-error --fail "https://get.helm.sh/helm-latest-version" 2>&1 || true)
+curl https://get.helm.sh/helm-$HELM_LATEST-linux-amd64.tar.gz --output helm.tar.gz
+tar -xvf ./helm.tar.gz && rm -rf ./helm.tar.gz
+chmod +x ./linux-amd64/helm
+./linux-amd64/helm repo add longhorn https://charts.longhorn.io
+./linux-amd64/helm repo update
+./linux-amd64/helm template longhorn longhorn/longhorn --set defaultSettings.deletingConfirmationFlag=true --set longhornUI.replicas=0 --set namespaceOverride=kube-system --no-hooks > /var/lib/rancher/rke2/server/manifests/01-longhorn.yaml
+rm -rf ./linux-amd64
```
mjnagel: Could we add a comment on why we're installing helm and templating this out to apply later? I know there are some limitations with Rancher's HelmChart CR but that was the approach we used before so it would be good to at least keep track of why we're doing this instead since I think the longhorn install seemed okay previously? One risk to note with this as well - it's not pinned to any specific version for helm or for longhorn.

mjnagel (commented):
Pending resolution of the CI issue this looks good for the most part. 2 questions/comments on the execution of the helm chart/iam role.

test/jest/network.spec.ts
```diff
@@ -143,21 +164,22 @@ describe("Network Policy Validation", () => {
 
   test.concurrent("Denied Requests by Default and Incorrect Ports and Labels", async () => {
     // Default Deny when no Ingress or Egress defined or Exposed Endpoints
-    const denied_external_response = await execInPod("curl-ns-deny-all-1", curlPodName1, "curl-pkg-deny-all-1", CURL_GATEWAY);
-    expect(denied_external_response.stdout).toBe("000");
+    // The HTTP response code could either be 000 or 503, depending on the K8s distro
+    const denied_external_response = await execInPod("curl-ns-deny-all", curlPodName1, "curl-pkg-deny-all-1", CURL_GATEWAY);
```
mjnagel:
```suggestion
    const denied_external_response = await execInPod("curl-ns-deny-all-1", curlPodName1, "curl-pkg-deny-all-1", CURL_GATEWAY);
```

test/jest/network.spec.ts
```diff
@@ -143,21 +164,22 @@ describe("Network Policy Validation", () => {
...
+    expect(isResponseError(denied_external_response)).toBe(true);
 
     // Default deny when no Ingress or Egress for internal curl command
-    const denied_internal_response = await execInPod("curl-ns-deny-all-1", curlPodName1, "curl-pkg-deny-all-1", INTERNAL_CURL_COMMAND_1);
-    expect(denied_internal_response.stdout).toBe("503");
+    const denied_internal_response = await execInPod("curl-ns-deny-all", curlPodName1, "curl-pkg-deny-all-1", INTERNAL_CURL_COMMAND_1);
```
mjnagel:
```suggestion
    const denied_internal_response = await execInPod("curl-ns-deny-all-1", curlPodName1, "curl-pkg-deny-all-1", INTERNAL_CURL_COMMAND_1);
```

test/jest/network.spec.ts
```diff
@@ -143,21 +164,22 @@ describe("Network Policy Validation", () => {
...
     // Default Deny for Google Curl when no Egress defined
     const denied_google_response = await execInPod("curl-ns-deny-all-1", curlPodName1, "curl-pkg-deny-all-1", GOOGLE_CURL);
     expect(denied_google_response.stdout).toBe("000");
 
     // Default Deny for Blocked Port
-    const blocked_port_curl = getCurlCommand("curl-pkg-deny-all-2", "curl-ns-deny-all-2", 9999);
-    const denied_port_response = await execInPod("curl-ns-deny-all-1", curlPodName1, "curl-pkg-deny-all-1", blocked_port_curl);
-    expect(denied_port_response.stdout).toBe("503");
+    const blocked_port_curl = getCurlCommand("curl-pkg-deny-all-2", "curl-ns-deny-all", 9999);
+    const denied_port_response = await execInPod("curl-ns-deny-all", curlPodName1, "curl-pkg-deny-all-1", blocked_port_curl);
```
mjnagel:
```suggestion
    const blocked_port_curl = getCurlCommand("curl-pkg-deny-all-2", "curl-ns-deny-all-2", 9999);
    const denied_port_response = await execInPod("curl-ns-deny-all-1", curlPodName1, "curl-pkg-deny-all-1", blocked_port_curl);
```

mjnagel (commented):
I think these should fix the CI error? Might've missed one+ though.

mjnagel (approved):

PR #1268, opened 2025-02-11
feat: introduced a new option CREATE_OPTIONS and skip SBOMs in tests

### LINKED ISSUE #1168: Skip sbom for dev/test tasks

Currently we use the same creation tasks for both dev/test and release processes:
- Release Creation: [Create tasks](https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/publish.yaml#L64-L74)
- Test Creation: [Test tasks](https://github.com/defenseunicorns/uds-core/blob/main/.github/workflows/test.yaml#L65-L90) which call the [same creation tasks](https://github.com/defenseunicorns/uds-core/blob/main/tasks/test.yaml#L83)

Our creation tasks (for example [standard package](https://github.com/defenseunicorns/uds-core/blob/main/tasks/create.yaml#L25)) do not specify any special options for `--skip-sbom` so that SBOMs are always generated. For dev/test this is not necessary and could add a minute or two due to the amount of images included in our core package.

### Possible Fixes

UDS Common includes [an input in their create task](https://github.com/defenseunicorns/uds-common/blob/main/tasks/create.yaml#L11-L13) that allows for passing in additional options to the zarf args. We could use this to pass in `--skip-sbom` when we are certain we are running a dev/test creation.

### Additional context

We need to be careful to make sure we still include SBOMs for release CI.

slaskawi:
## Description

This Pull Request introduces a new option `CREATE_OPTIONS` that being passed
into the `create:single-layer` task. This option can be used to pass for
example `--skip-sbom` command and speed up development process. Here's an
example:

```bash
uds run create:standard-package --log-level debug --set CREATE_OPTIONS="--skip-sbom"
```

Skipping SBOMs has also been added to all tests and dependent tasks. In other
words - in all places where we are certain that we won't need them.

## Related Issue

Fixes #1168

## Type of change

## Steps to Validate

Comparing CI results from a Unicorn flavor from this Pull Request and a baseline, we shortened the CI tests from 22m 11s  to 17m 56s. That's roughly 4 mins per each job.

## Checklist before merging

### DIFF
--- a/tasks/create.yaml
+++ b/tasks/create.yaml
@@ -15,11 +15,16 @@ variables:
 
   - name: LAYER
 
+  - name: CREATE_OPTIONS
+    description: "Additional options passed in when creating Zarf packages. For example: --skip-sbom"
+    default: ""
+
 tasks:
   - name: standard-package
     description: "Create the UDS Core Zarf Package"
     inputs:
       create_options:
+        default: ${CREATE_OPTIONS}
         description: "Additional options passed in when creating Zarf packages. For example: --skip-sbom"
     actions:
       - task: single-layer
@@ -57,7 +62,7 @@ tasks:
         with:
           path: packages/checkpoint-dev
           config: ./zarf-config.yaml
-          options: "--skip-sbom"
+          options: "${CREATE_OPTIONS} --skip-sbom"
           architecture: ${ZARF_ARCHITECTURE}
 
   # This task is a wrapper to support --set LAYER=identity-authorization
@@ -73,6 +78,7 @@ tasks:
         default: base
         description: The UDS Core layer to build
       create_options:
+        default: ${CREATE_OPTIONS}
         description: "Additional options passed in when creating Zarf packages. For example: --skip-sbom"
     actions:
       - task: pepr-build
--- a/tasks/create.yaml
+++ b/tasks/create.yaml
@@ -15,14 +15,22 @@ variables:
 
   - name: LAYER
 
+  - name: CREATE_OPTIONS
+    description: "Additional options passed in when creating Zarf packages. For example: --skip-sbom"
+    default: ""
+
 tasks:
   - name: standard-package
     description: "Create the UDS Core Zarf Package"
+    inputs:
+      create_options:
+        default: ${CREATE_OPTIONS}
+        description: "Additional options passed in when creating Zarf packages. For example: --skip-sbom"
     actions:
-      - task: pepr-build
-
-      - description: "Create the UDS Core Standard Zarf Package"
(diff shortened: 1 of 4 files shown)


tasks/test.yaml
```diff
@@ -26,6 +26,7 @@ tasks:
       - task: create:single-layer
         with:
           layer: ${LAYER}
+          create_options: "--skip-sbom"
```
mjnagel: I think this is problematic for our single-layer publishing. Tracing through the tasks we call in the [core layers publish job](https://github.com/defenseunicorns/uds-core/blob/1168-skip-sbom/.github/workflows/publish.yaml#L101), [`tasks/publish.yaml single-layer`](https://github.com/defenseunicorns/uds-core/blob/1168-skip-sbom/.github/workflows/publish.yaml#L132) actually [calls this task](https://github.com/defenseunicorns/uds-core/blob/1168-skip-sbom/tasks/publish.yaml#L75) if I'm not mistaken, which would skip sboms.

We could modify that workflow to call sub-tasks necessary there instead, which might line up better with how we are doing the standard package publish? Just will want to make sure we are very careful to not break those steps since we don't have any real validation/testing of the publish until we hit release time.

tasks/create.yaml
```diff
@@ -50,7 +58,12 @@ tasks:
     description: "Create the K3d + UDS Core Checkpoint Zarf Package"
     actions:
       - description: "Create the UDS Core Checkpoint Zarf Package"
-        cmd: "uds zarf package create packages/checkpoint-dev --confirm --no-progress --skip-sbom"
+        task: common:package
+        with:
+          path: packages/checkpoint-dev
+          config: ./zarf-config.yaml
+          options: "${CREATE_OPTIONS} --skip-sbom"
```
mjnagel, line 65: I think this makes sense for checkpoint - @Racer159 concur? I don't think there's any need/usecase for the sbom on the checkpoint package (and I'd wonder if it even provides any value given the composition of that package).

Racer159: yeah there is no reason for sboms there - you should _never_ run checkpoint in prod since it reuses keys

tasks/test.yaml
slaskawi: You're eagle-eyed @mjnagel ! I refactored this piece and followed your advise there. 

I also agree with you, it's better structured now as relying on `test:single-layer` for building process makes publishing depend on side effects of tests. This could be quite fragile.

mjnagel (approved):

mjnagel:
Ran through the publish workflow to validate:
- all still respect architecture as expected
- all still respect flavor as expected
- all still create/save SBOMs as expected
- all tasks pass still 😄 

Encountered no issues so this should be g2g - thanks @slaskawi !

PR #1288, opened 2025-02-15
feat: kstatus for Pepr

### LINKED ISSUE #991: Implement .status.conditions on `package.uds.dev` to allow for evaluation with kstatus
Since Zarf v0.42.0 deployments automatically wait for all resources in the deployed chart to be ready using [kstatus](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md). This allows users to avoid the `.actions.wait.cluster` syntax and have their packages automatically wait for resources that Helm won't wait for such as CRDs. However, not all CRDs are implemented in such a way that allows kstatus to tell if they are reconciled. `packages.uds.dev` custom resources are immediately considered reconciled upon creation, because the `.status.conditions` field is not implemented. The Package CRD uses `.status.phase` instead, however according to the Kubernetes [API-convention docs](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties), `.status.phase` is deprecated.

The only condition required for kstatus to work is the [Ready condition](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md#the-ready-condition). The UDS operator should create `packages.uds.dev` objects with an entry in the `.status.conditions` map with status set to "False", then update the status field to "True" when the resource is ready. 
```yaml
status:
   conditions:
    - type: Ready 
      status: "False" # Update to "True" when ready
```

This will allow owners of UDS packages to delete the common wait entry that looks like the yaml below. Additionally, by adopting the community standard other tools, such as Lula, will have an easier time integrating when waits are required.
```yaml
      - description: Validate nginx Deployment
        maxTotalSeconds: 300
        wait:
          cluster:
            kind: Deployment
            name: nginx
            namespace: nginx
            condition: Available
```

slaskawi:
## Description

This Pull Request introduces [kstatus](https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/README.md) to the Pepr Operator. At the moment it supports setting only one condition:

```yaml
 status:
   phase: Ready
   conditions:
    - lastTransitionTime: "2025-02-17T09:33:00.748Z"
      message: The package is ready for use.
      reason: ReconciliationComplete
      status: "True"
      type: Ready
```

## Related Issue

Fixes #991

## Type of change

## Steps to Validate

Run UDS Core with this patch on and observe `kubectl get packages -o yaml --all-namespaces`. 

## Checklist before merging


src/pepr/policies/storage.spec.ts
```diff
@@ -8,7 +8,7 @@ import { K8s, kind } from "pepr";
 
 const failIfReached = () => expect(true).toBe(false);
 
-describe("storage policies", () => {
+describe.skip("storage policies", () => {
```
slaskawi: To be fixed in https://github.com/defenseunicorns/uds-core/issues/1290

src/pepr/policies/security.spec.ts
```diff
@@ -8,7 +8,7 @@ import { K8s, kind } from "pepr";
 
 const failIfReached = () => expect(true).toBe(false);
 
-describe("security policies", () => {
+describe.skip("security policies", () => {
```
slaskawi: To be fixed in https://github.com/defenseunicorns/uds-core/issues/1290

src/pepr/policies/network.spec.ts
```diff
@@ -8,7 +8,7 @@ import { K8s, kind } from "pepr";
 
 const failIfReached = () => expect(true).toBe(false);
 
-describe("network policies", () => {
+describe.skip("network policies", () => {
```
slaskawi: To be fixed in https://github.com/defenseunicorns/uds-core/issues/1290

schemas/package-v1alpha1.schema.json
```diff
@@ -954,7 +954,7 @@
       ],
       "title": "ProtocolMapper"
     },
-    "Status": {
+    "StatusObject": {
```
slaskawi, line 957: A note for reviewers: this conversation happened automatically by kubernetes-fluent-client. This is due to duplicated `status` name - one of which is an Object and one is String.

.github/workflows/test.yaml
```diff
@@ -89,6 +89,13 @@ jobs:
         if: ${{ inputs.package == 'all' && inputs.test_type == 'upgrade' }}
         run: uds run test-uds-core-upgrade --set FLAVOR=${{ inputs.flavor }} --no-progress
 
+      # This test requires a running k3d cluster. It needs to go the last.
+      - name: Test Pepr
+        if: ${{ (inputs.package == 'all' || inputs.package == 'base') && inputs.test_type == 'install' }}
+        run: |
+          npm ci
+          uds -f ./src/pepr/tasks.yaml run validate
+
```
mjnagel: I might not be following but these tests should already be getting run.

Install:
- [`test:uds-core-e2e`](https://github.com/defenseunicorns/uds-core/blob/991-status_conditions_pepr/tasks/test.yaml#L111) calls
- [`test:uds-core`](https://github.com/defenseunicorns/uds-core/blob/991-status_conditions_pepr/tasks/test.yaml#L101) calls
- [`test:validate-packages`](https://github.com/defenseunicorns/uds-core/blob/991-status_conditions_pepr/tasks/test.yaml#L54) which loops through the `src` files and [calls `validate`](https://github.com/defenseunicorns/uds-core/blob/991-status_conditions_pepr/tasks/test.yaml#L60) from each task file

Upgrade:
- [`test:uds-core-upgrade`](https://github.com/defenseunicorns/uds-core/blob/991-status_conditions_pepr/tasks/test.yaml#L127) calls
- [`test:validate-packages`](https://github.com/defenseunicorns/uds-core/blob/991-status_conditions_pepr/tasks/test.yaml#L54) (loops through src files as mentioned)

On the upgrade the `validate` task only gets called _after_ the upgrade but otherwise this should already be running?

mjnagel (commented):
This looks pretty great otherwise - maintains our old custom status as well as adding the kstatus conditions! I'll validate a deploy locally as well just to confirm.

We don't have to do this in this PR but I'd be curious if could drop the `onDeploy.after` actions that we have in several of our `src/` zarf packages. As mentioned in the original issue Zarf should honor the kstatus of these CRDs now so that we can delete those 😄

src/pepr/policies/storage.spec.ts
mjnagel: May be semi-related to [this thread](https://github.com/defenseunicorns/uds-core/pull/1288#discussion_r1959987097) - curious on why these tests were failing since they should be run in CI. I believe our policy tests might require a running cluster and active pepr deployment, so wondering if timing of running them was causing an issue here?

slaskawi: That was a mistake on my side. Fixed.

.github/workflows/test.yaml
slaskawi: Thanks a million for explanation. Fixed.

slaskawi:
@mjnagel Absolutely! The test results look very promising: https://github.com/defenseunicorns/uds-core/pull/1307

src/pepr/tasks.yaml
```diff
@@ -67,7 +67,7 @@ tasks:
               # Extract the version (e.g., v1alpha1)
               version=$(kubectl get crd $CRD -o json | jq -r '.status.storedVersions[0]')
               # Typescript format markdown
-              npx ts-node src/pepr/docs-gen/main.ts "$CRD-crd.json" "$version"
+              npx ts-node --project src/pepr/tsconfig-docs-gen.json src/pepr/docs-gen/main.ts "$CRD-crd.json" "$version"
```
chance-coleman, line 70: I'm probably just slow here, but what is this change doing with the empty json file?

slaskawi: The empty Project file forces `ts-node` to load the defaults. The defaults do not include the `exclusions` we have here: https://github.com/defenseunicorns/uds-core/blob/4d0b9d0497e76793ff7de18a4ab7fe2bdfbfc747/tsconfig.json#L24-L30

For some reason, the docs-gen requires transitively "test/jest" and its dependencies. So with the exclusion it just fails. 

I tried to override these settings in place with `ts-node` or somehow instruct it to ignore these dependencies (as they are needed for tests) but failed to so do. Overriding the entire Project configuration was the easiest way I could find.

chance-coleman: ahh i see, okay. i had a different approach to bring in the `@types/node` in a different PR. But i think you're approach is cleaner and simpler.

src/authservice/common/zarf.yaml
```diff
@@ -15,14 +15,3 @@ components:
         localPath: ../chart
         version: 1.0.1
         namespace: authservice
-    actions:
-      onDeploy:
-        after:
-          - description: Validate Authservice Package
-            maxTotalSeconds: 300
```
mjnagel, line 22: One thing I noticed in testing this - the zarf behavior default seems to be waiting for 15m on resources (both the helm waits on "standard" resources and the kstatus waits). As this diff shows, previously we had a single wait set for 5 minutes on the Package CR only.

I haven't yet identified if we can override the default 15m wait but it might be worth evaluating whether we would want to. The default is probably fine for now - just means a longer wait time for failure potentially?

mjnagel (approved):
This looks great overall @slaskawi - validated locally that it behaves as expected for both working and "broken" packages. Put one thread on the diff for consideration but I'm good with this PR as-is.

src/authservice/common/zarf.yaml
mjnagel: Just checked with zarf folks - sounds like the timeout is a CLI option so end users could override this when deploying. Probably can ignore this comment for now!

slaskawi: Yes, I agree. It seems the on failure cases we'll be waiting a bit longer. Let's see if this will require any adjustments in the future.

Issue #1309, opened 2025-02-20
k3d-core-demo:0.36.0 failing with a java error in keycloak


devonjones:
### Environment
Device and OS: Macbook Pro, Apple M4 Max, 64 gig, OSX 15.3.1
App version: 0.36.0
Kubernetes distro being used: k3d version v5.8.3
    k3s version v1.31.5-k3s1
    docker desktop v4.38.0
Other:

### Steps to reproduce
1. uds deploy k3d-core-demo:0.36.0 --confirm

### Expected result
uds runs

### Actual Result
keycloak dies with a java problem

$ kubectl get pods -A

NAMESPACE                   NAME                                         READY   STATUS    RESTARTS      AGE
istio-admin-gateway         admin-ingressgateway-6b9766b975-h275g        1/1     Running   0             83s
istio-passthrough-gateway   passthrough-ingressgateway-cb6f9b8df-r8znt   1/1     Running   0             63s
istio-system                istio-cni-node-98prh                         1/1     Running   0             87s
istio-system                istiod-9bf475cd-h2swq                        1/1     Running   0             94s
istio-system                ztunnel-822g5                                1/1     Running   0             85s
istio-tenant-gateway        tenant-ingressgateway-6b7cfd9bf-jxgth        1/1     Running   0             66s
keycloak                    keycloak-0                                   1/2     Error     2 (25s ago)   33s
kube-system                 coredns-9889f7446-mfsjz                      1/1     Running   0             4m46s
metrics-server              metrics-server-78db54bff9-w2qnr              2/2     Running   0             56s
pepr-system                 pepr-uds-core-6896dc6948-lxv5p               2/2     Running   0             70s
pepr-system                 pepr-uds-core-6896dc6948-v86fx               2/2     Running   0             90s
pepr-system                 pepr-uds-core-watcher-85ff5df7d5-jqnfz       2/2     Running   0             85s
uds-dev-stack               ensure-machine-id-vt8n8                      1/1     Running   0             5m10s
uds-dev-stack               local-path-provisioner-6c786cf79d-kbt5z      1/1     Running   0             5m10s
uds-dev-stack               metallb-controller-dc88974b6-8pphm           1/1     Running   0             5m34s
uds-dev-stack               metallb-speaker-ldvrn                        4/4     Running   0             5m34s
uds-dev-stack               minio-5c6dd989b7-fdqn9                       1/1     Running   0             5m6s
[...]

PR #1314, opened 2025-02-24
feat: new bundle variable - KEYCLOAK_HEAP_OPTIONS

slaskawi:
## Description

This Pull Request introduces `KEYCLOAK_HEAP_OPTIONS` to control the Java Heap
configuration passed into Keycloak deployments.

## Related Issue

Relates to #1309

## Type of change

## Steps to Validate

Run the CI and check the Keyclaok Pod Spec

## Checklist before merging

### DIFF
--- a/bundles/k3d-slim-dev/uds-bundle.yaml
+++ b/bundles/k3d-slim-dev/uds-bundle.yaml
@@ -163,6 +163,14 @@ packages:
             - name: KEYCLOAK_DEVMODE
               description: "Enables Keycloak dev mode"
               path: devMode
+            # This is a workaround for Keycloak and Kernel 6.12+ memory issue. It will be removed once
+            # https://github.com/defenseunicorns/uds-core/issues/1212 is sorted
+            - name: KEYCLOAK_HEAP_OPTIONS
+              description: "Sets the JAVA_OPTS_KC_HEAP environment variable in Keycloak"
+              path: env
+              default:
+                - name: JAVA_OPTS_KC_HEAP
+                  value: "-XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G"
           values:
             - path: realmInitEnv
               value:
@@ -173,9 +181,3 @@ packages:
                 GOOGLE_IDP_CORE_ENTITY_ID: "https://sso.uds.dev/realms/uds"
                 GOOGLE_IDP_ADMIN_GROUP: "uds-core-dev-admin"
                 GOOGLE_IDP_AUDITOR_GROUP: "uds-core-dev-auditor"
-            # This is a workaround for Keycloak and Kernel 6.12+ memory issue. It will be removed once
-            # https://github.com/defenseunicorns/uds-core/issues/1212 is sorted
-            - path: env
-              value:
-                - name: JAVA_OPTS_KC_HEAP
-                  value: "-XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G"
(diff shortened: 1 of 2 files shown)


slaskawi:
Checked the CI logs and the Keycloak Pod configuration seems to be correct:

```
        - name: JAVA_OPTS_KC_HEAP
          value: -XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50
            -XX:MaxRAM=1G
```

bundles/k3d-slim-dev/uds-bundle.yaml
```diff
@@ -163,6 +163,14 @@ packages:
...
               description: "Enables Keycloak dev mode"
               path: devMode
+            # This is a workaround for Keycloak and Kernel 6.12+ memory issue. It will be removed once
+            # https://github.com/defenseunicorns/uds-core/issues/1212 is sorted
+            - name: KEYCLOAK_HEAP_OPTIONS
+              description: "Sets the JAVA_OPTS_KC_HEAP environment variable in Keycloak"
+              path: env
+              default:
+                - name: JAVA_OPTS_KC_HEAP
+                  value: "-XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G"
```
mjnagel: Could you provide an example of how an end user would override this? For the immediate issue we encountered with m4 devices it would be ideal if we had a `uds deploy` command where the end user could `--set` this variable to add on the extra arg.

I spent a short bit of time trying to override this but I couldn't find a syntax that worked (some made my shell mad, some made CLI mad). One other thing that felt a bit weird is that this is a var named `KEYCLOAK_HEAP_OPTIONS` but is actually overriding the `env` value. I came up with something that _works_ but open to other ideas here:
```yaml
          variables:
...
            - name: KEYCLOAK_HEAP_OPTIONS
              description: "Sets the JAVA_OPTS_KC_HEAP environment variable in Keycloak"
              path: env[0].value
          values:
...
            - path: env[0]
              value:
                name: JAVA_OPTS_KC_HEAP
                value: "-XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G"
```

It's a bit weird but it sets up the env + default under `values` and then exposes a variable to specifically target `env[0].value` (the value for the first env, `JAVA_OPTS_KC_HEAP`). This makes it more targeted to the actual java opts (rather than just env as a whole) and allows for overriding very easily: `uds deploy bundles/k3d-slim-dev/uds-bundle-k3d-core-slim-dev-arm64-0.36.2.tar.zst --set KEYCLOAK_HEAP_OPTIONS="-XX:MaxRAMPercentage=70"`.

Very open to other paths here but I think we should make sure that we have a clear/simple path to override this - and we may want to document the m4 issue/override example somewhere as well.

slaskawi: You're absolutely right. I adjusted the code to according to your suggestions.

mjnagel (approved):
LGTM - do we want to add a note anywhere to document the flag required for m4 macs?

mjnagel:
@slaskawi going to merge this as-is but might be good for us to document the m4 flag as a follow on PR 😄

bundles/k3d-slim-dev/uds-bundle.yaml
bhearn7: You guys rock! This worked for me: `uds deploy k3d-core-slim-dev:0.37.0 --set KEYCLOAK_HEAP_OPTIONS="-XX:UseSVE=0 -XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G" --confirm`

PR #1293, opened 2025-02-18
feat: sso doc restructure

slaskawi:
## Preview

https://github.com/defenseunicorns/uds-docs/pull/150

## Description

This Pull Request restructures SSO documentation. Highlights:
- Moved most of the examples into a separate page (`/reference/configuration/sso/`) with introduction

## Definition of done checklist (copied from the main ticket)

- [x] SSO examples pulled out of `docs/reference/configuration/uds-operator/package` (undecided on whether client attribute validation should stay or move)
- [x] New SSO section added in `docs/reference/configuration/uds-operator/package` with a link to a new document
- New document should contain:
  - [x] Existing examples (SSO secret templating, Authservice protection, Device Flow, Service account roles), with narrative added/updated as needed
  - [x] New examples (Group based authentication, ?) with narrative

Additional potential things to add here over time (likely follow on issues):
- [x] ~~Flow diagram of Pepr / Keycloak configuration~~
- [x] ~~Flow diagram of Pepr / Authservice configuration (and resource tree for authorization policies)~~
- [x] A diagram and a description of interactions between Pepr, AuthService and Keycloak
- [x] Additional IDP specific docs - https://github.com/defenseunicorns/uds-identity-config/issues/302

## Related Issue

Fixes #1282
Fixes https://github.com/defenseunicorns/uds-identity-config/issues/302

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Steps to Validate

Rebuild the uds-docs to see the result.

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md) followed

PR #1315, opened 2025-02-24
chore(docs): keycloak session timeout doc

chance-coleman:
## Description
There has been some confusion about how keycloak manages session timeouts. After some investigation by @mjnagel this document was created to address those knowledge gaps.

## Type of change

## Checklist before merging


docs/reference/configuration/keycloak-session-timeouts.md
```diff
@@ -0,0 +1,75 @@
...
+
+### What Happens?
+1. **At 5 minutes of inactivity** → **Client's token expires**
+   - The access token is now **invalid**, meaning it can no longer be used for authentication.
+   - The **realm session is still active** (since 1 hour hasn't passed).
+   - If Client makes an API request, it will get a **401 Unauthorized** error.
+
+2. **At 6 minutes**, the user **reactivates Client**.
+   - If the refresh token is still valid, Client can obtain a new access token.
+   - If the refresh token has expired, the user must **reauthenticate**.
```
mjnagel: I think at 6 minutes the refresh token would be expired no matter what, since that's tied to the client session timeout (going based on the testing with Authservice). But the user doesn't necessarily have to "reauthenticate" since their overall session is still valid? The client/application just needs to request a new client token?

This is where I get a bit confused on the behavior and how to word this - @slaskawi might have some thoughts here?

mjnagel (commented):
Overall LGTM - just reviewed https://github.com/defenseunicorns/uds-core/pull/1293 as well which might move where this doc lands.

docs/reference/configuration/keycloak-session-timeouts.md
chance-coleman: Ahh yeah good call, i missed that.

docs/reference/configuration/keycloak-session-timeouts.md
```diff
@@ -0,0 +1,75 @@
+---
+title: Keycloak Session Timeout
+---
+
+## Understanding Keycloak Session Idle Timeouts
+
+Keycloak has two session idle timeouts: the **realm session idle timeout** and the **client session idle timeout**. These settings control session expiration differently, and their interaction determines how long a user remains authenticated across different clients.
```
slaskawi: I think we might want to clarify here that we're talking about Realm > Realm Settings > Sessions configuration screen here. 

The Client Session settings can also be overridden per-Client in Client > Advanced > Advanced Settings:

image

docs/reference/configuration/keycloak-session-timeouts.md
```diff
@@ -0,0 +1,75 @@
...
+
+---
+
+## Scenario 1: Client Session Idle Timeout is Shorter than Realm Session Idle Timeout
+
+- **Realm Session Idle Timeout** = **1 hour**
```
slaskawi: I think it might be a good idea to use Keycloak terminology here (taken from [here](https://docs.redhat.com/en/documentation/red_hat_build_of_keycloak/22.0/html/server_administration_guide/managing_user_sessions#timeouts)):

- SSO Session Idle 
- SSO Session Max 
- Client Session Idle 
- Client Session Max 

This approach might help Mission Heroes find these options in Keycloak UI

docs/reference/configuration/keycloak-session-timeouts.md
```diff
@@ -0,0 +1,75 @@
...
+- **Client Session Idle Timeout** = **5 minutes**
+- **User signs into Client.**
+- **User is inactive for 6 minutes**, then reactivates Client.
```
slaskawi: If you're aiming to be fully correct, that needs to be 7 minutes. There's 2 min window there to ensure invalidation is propagated across the entire fleet of Keycloak Servers. See [this note](https://docs.redhat.com/en/documentation/red_hat_build_of_keycloak/22.0/html/server_administration_guide/managing_user_sessions#timeouts):

>  For idle timeouts, a two-minute window of time exists that the session is active. For example, when you have the timeout set to 30 minutes, it will be 32 minutes before the session expires.

docs/reference/configuration/keycloak-session-timeouts.md
```diff
@@ -0,0 +1,75 @@
...
+
+### What Happens?
+1. **At 5 minutes of inactivity** → **Client's token expires**
+   - The access token is now **invalid**, meaning it can no longer be used for authentication.
+   - The **realm session is still active** (since 1 hour hasn't passed).
+   - If Client makes an API request, it will get a **401 Unauthorized** error.
+
+2. **At 6 minutes**, the user **reactivates Client**.
+   - Since the **client session timeout** controls the refresh token expiration, the refresh token has also expired.
+   - However, since the **realm session is still active**, the client/application can initiate a new authentication request to obtain fresh tokens. This process does not prompt the user for credentials again as long as the realm session remains valid.
```
slaskawi: > This process does not prompt the user for credentials again as long as the realm session remains valid

This is probably true for Browser application as we have an HTTP Session and a Cookie at hand. What happens to Applications that use Bearer Tokens?

docs/reference/configuration/keycloak-session-timeouts.md
```diff
@@ -0,0 +1,75 @@
...
+
+---
+
+## Scenario 2: Realm Session Idle Timeout is Shorter than Client Session Idle Timeout
+
+- **Realm Session Idle Timeout** = **10 minutes**
+- **Client Session Idle Timeout** = **30 minutes**
```
slaskawi: In general, this example uses a non-recommended configuration. Here's a note from [Keycloak manual](https://docs.redhat.com/en/documentation/red_hat_build_of_keycloak/22.0/html/server_administration_guide/managing_user_sessions#timeouts):

> [...] This value should specify a shorter idle timeout than the SSO Session Idle [...]

If we decided to keep this example, we should probably add a warning note that this is purely for illustration purposes and shouldn't be configured that way in production.

docs/reference/configuration/keycloak-session-timeouts.md
```diff
@@ -0,0 +1,75 @@
...
+  2. It reaches its **max lifespan**.
+  3. The **realm session expires**.
+  4. The **user logs out**.
+
+---
+
+## Additional Considerations
+- **Client session idle timeout only affects token expiration**, not the realm session itself.
+- If multiple clients are accessed under the same session, different client timeouts can cause token expiration at different intervals.
+- A realm session timeout takes **precedence**, meaning once the realm session expires, all client sessions are forcibly logged out.
```
slaskawi: We might want to add that Realm takes priority only if at the Client level setting we use "Inherits from Realm settings". Otherwise, the Client settings takes priority.

docs/reference/configuration/keycloak-session-timeouts.md
```diff
@@ -0,0 +1,75 @@
...
+- **Offline sessions** allow longer session persistence beyond normal idle timeouts if configured properly.
+- To prevent unwanted logouts, applications can implement **silent authentication** via refresh tokens or use offline tokens when necessary.
```
slaskawi: I would probably remove this line. Using Offline Tokens has quite some security implications. If Clients validate Tokens themselves (by checking the signature), there's no way to sign an application out. We should probably have some internal conversation and vendor this risk in.

slaskawi (commented):
Great article that clarifies how session works! 

My biggest change request (mentioned inline) is to use Keycloak UI terminology of these things. Otherwise, it's very easy to confuse some of the settings.

docs/reference/configuration/keycloak-session-timeouts.md
chance-coleman: Yes agreed, this is to show the incorrect settings that we've observed a few people setting. Will add a warning that this is not a recommended configuration.

docs/reference/configuration/keycloak-session-timeouts.md
chance-coleman: Good point, i'm not sure if we have anyone utilizing bearer tokens. Added two new bullets for addressing those applications to cover our bases.

docs/reference/configuration/keycloak-session-timeouts.md
chance-coleman: Interesting, I never noticed this!

docs/reference/configuration/keycloak-session-timeouts.md
chance-coleman: Yeah, good call. kind of weird when the UI uses different terms than the documentation. Will add both here for more clarity.

docs/reference/configuration/keycloak-session-timeouts.md
chance-coleman: Added a couple lines outlining where the realm and client settings can be configured from within the UI.

slaskawi (dismissed):

mjnagel (approved):

PR #1331, opened 2025-03-03
feat: egress gw

### LINKED ISSUE #776: Egress gateway / external service entry configuration

Provide a way to allow egress traffic to a known `remoteHost` via the `Package` CR and operator. This should utilize egress gateways / service entries as described/approved in the egress gateway configuration.

Where possible in core we should expose configuration to utilize this for external dependencies like S3, RDS, etc (this could be done as a follow on).

meganwolf0:
## Description

Adds egress gateway and relevant Istio resources to UDS Core.

This PR contains:
* Addition of `egressgateway` component in the Istio zarf package (default is not included)
* Update to `dev-setup` task to optionally pass istio_components (e.g., `uds run dev-setup --with istio_components=istio-egress-gateway`)
* Package CR update - extension of `network.allow` parameters to support a defined `remoteHost` and `remoteProtocol` for workloads to send traffic
* Controller updates to create needed egress resources to route traffic to the egress gateway workload
* Tests
  * Unit alongside functions (largely egress.spec.ts)
  * E2E in `test/jest/network.spec.ts` - optional follow on to this is to possibly have an `test/jest/egress.spec.ts` that batch adds, updates, validates, and removes packages with egress
* Docs

Notes to reviewers
* The addition of shared resources that multiple Package CRs are updating seemed to be a fairly uncommon pattern in the uds-operator (the exception probably being the authservice secret) - the reconciliation of these values lives in an in-memory map. Handling the concurrent operation of both updating the map and applying the shared resources turned out to be a fairly complex operation. Due to limitations, batch processing of packages with egress will not correctly run in pepr dev mode, but will perform when executed in cluster.
* When packages are force deleted, the finalizer does not run and the shared resources will not be cleaned up (suggest follow-on issue to address)
* This is the design for sidecar only, assuming that at some point an ambient design will accompany or entirely replace this
* The grafana playwright test needed to be updated because for some reason with the addition of two namespaces, the grafana namespace fell out of the "visible" portion of the namespace dropdown that was being selected

## Related Issue

Fixes #776 

## Type of change

## Steps to Validate
* Create a new Package CR with the appropriate `network.allow` specification to allow egress to only a particular host. See [this file](https://gist.github.com/meganwolf0/6d02ecf09bcb3e6ba7aaae721918690b) for examples.
* Run `uds run dev-setup --with istio_components=istio-egress-gateway` - note limitation above on batch processing, run slim dev with the appropriate zarf modifications if attempting to batch reconcile a large amount of packages
* Apply packages and check for the following resources:
  * Scoped to package namespace: Service entry and sidecar; service entry per host and sidecar per workload
  * Scoped to istio-egress-gateway namespace: Virtual Service, Gateway, and ServiceEntry per unique host. Annotations linking back to parent package
* Delete or modify packages, ensure shared resources are updated appropriately
* (optional) Cycle watcher pod, ensure shared resources are re-reconciled correctly

## Checklist before merging


meganwolf0:
I guess I'm going to have to fix the unsigned commits 🙃

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -0,0 +1,256 @@
...
+      inMemoryPackageMap[pkgId] = hostResourceMap;
+    } else {
+      removeEgressResources(pkgId);
+    }
+  } else if (action == PackageAction.Remove) {
+    removeEgressResources(pkgId);
+  }
+
+  // Check if there is an istioEgressGatewayNamespace, if not don't proceed
+  log.debug("HEREHEREHERE");
```
chance-coleman: probably can remove this test log 😄

docs/reference/configuration/egress.md
```diff
@@ -0,0 +1,114 @@
+---
+title: Istio Egress
+---
+
+UDS Core leverages Istio to route dedicated egress out of the service mesh. This document provides an overview and examples of the Istio resources that UDS Core deploys to handle egress.
+
+>[!NOTE]
+>This is only usable for workloads implementing an Istio sidecar proxy
```
mjnagel: Is the intent of this "part of the istio mesh" or does this mean explicitly has to have a sidecar (does not work with ambient workloads). Just want to confirm, if the second we should likely explicitly call out `spec.network.serviceMesh.mode` here and note that `ambient` is not supported.

docs/reference/configuration/egress.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Configuring the Egress Workload
+
+The dedicated egress workload is an *optional* component of UDS Core. To enable it in the UDS Bundle, add it to the `optionalComponents` as follows:
```
mjnagel:
```suggestion
The dedicated egress gateway is an *optional* component of UDS Core. To enable it in the UDS Bundle, add it to the `optionalComponents` as follows:
```

docs/reference/configuration/egress.md
mjnagel: Can we also switch these note blocks to the format that the docs site will show (see [here](https://github.com/defenseunicorns/uds-core/blob/1e7ebce255256c3b3bb7a5407acdd43473fb918f/docs/reference/UDS%20Core/dns.md?plain=1#L14-L16) for an example).

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -62,9 +85,79 @@ export async function istioResources(pkg: UDSPackage, namespace: string) {
...
+      // Apply the ServiceEntry and force overwrite any existing resource
+      await K8s(IstioServiceEntry)
+        .InNamespace(namespace)
+        .Get(serviceEntry.metadata!.name!)
+        .catch(async err => {
+          if (err.status === 404) {
+            await K8s(IstioServiceEntry).Apply(serviceEntry, { force: true });
+          } else {
+            log.error(`Failed to reconcile Service Entry ${serviceEntry.metadata?.name}: ${err}`);
+          }
```
mjnagel: Should this be thrown so that it hits our handle failure function? Or is it okay to just log the error here? I suspect we'd want to throw this back to the reconciler, where it will hit retry logic, create events, etc.

(this comment would also apply to the sidecar error handling below)

src/pepr/operator/controllers/istio/gateway.ts
```diff
@@ -0,0 +1,111 @@
...
+      mode: IstioTLSMode.Passthrough,
+    },
+  };
+}
+
+// Check for other gateways that might conflict - gateways that are not added by the operator
+// Note: Users adding their own Istio resources will need to understand the possible conflicts with the spec. This is not an operation
+// blocked by K8s, but will be identified as invalid by Istio. The UDS operator will only manage/deconflict resources it creates or those
+// that follow the deterministic naming convention.
+export async function warnMatchingExistingGateways(host: string) {
```
mjnagel, line 83: I think my personal leaning is that we should fail reconciliation of the package CR if this is hit. Otherwise we are marking the package as Ready even though it is unlikely to be working as expected/intended. With this current approach, the only indication of an issue would be a debug log in the pepr watcher pod.

(this comment would also apply to the VS)

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -6,28 +6,39 @@
...
 /**
  * Creates a VirtualService and ServiceEntry for each exposed service in the package
+ * and creates or merges Istio resources for allowing egress traffic to external services
  *
  * @param pkg
  * @param namespace
  */
 export async function istioResources(pkg: UDSPackage, namespace: string) {
   const pkgName = pkg.metadata!.name!;
+  // const pkgId = getPackageId(pkg);
```
mjnagel: Can be cleaned up?

src/pepr/operator/controllers/istio/service-entry.ts
```diff
@@ -82,3 +86,110 @@ export function generateSEName(pkgName: string, expose: Expose) {
...
+      namespace,
+      labels: {
+        "uds/package": pkgName,
+        "uds/generation": generation,
+      },
+      // Use the CR as the owner ref for each ServiceEntry
+      ownerReferences: ownerRefs,
+    },
+    spec: {
+      // Append the UDS Domain to the host
```
mjnagel: Don't think this comment applies here?

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -65,33 +65,68 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+      policy.remoteProtocol &&
+      (policy.remoteGenerated ||
+        policy.remoteNamespace ||
+        policy.remoteSelector ||
+        policy.remoteCidr ||
+        !policy.remoteHost)
+    ) {
+      return req.Deny(
+        "remoteProtocol cannot be combined with remoteGenerated, remoteNamespace, remoteSelector, or remoteCidr and must have remoteHost",
       );
```
mjnagel: Should we also deny if `remoteHost` is used with `Ingress` since I don't think that is supported? Maybe also would be good to validate/deny wildcard subdomains since I think those were documented as not supported?

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -0,0 +1,256 @@
...
+        IstioServiceEntry,
+        log,
+      );
+    })
+    .catch(e => {
+      if (e.status == 404) {
+        log.debug(
+          `Namespace ${istioEgressGatewayNamespace} not found. Skipping shared egress resource reconciliation.`,
+        );
+        return; // Exit the routine
```
mjnagel: This also feels like a potential place to fail the reconciliation? I think this would indicate that the egress gateway is not deployed, so all the expected egress control wouldn't work as expected.

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -0,0 +1,256 @@
...
+      } else {
+        log.error(`Error getting namespace ${istioEgressGatewayNamespace}: ${e}`);
+        throw e;
+      }
```
mjnagel: I think this could be any sort of error from `applyEgressResources` or `purgeOrphans` even, is that true? May want to reword the error log accordingly.

src/pepr/operator/crd/index.ts
```diff
@@ -28,6 +29,7 @@ export {
 export {
   HTTP as IstioHTTP,
   HTTPRoute as IstioHTTPRoute,
+  Tl as IstioTL,
```
mjnagel: Even though the upstream is `Tl`, thoughts on exporting as `IstioTLS` for our own readability?

src/test/tasks.yaml
```diff
@@ -218,6 +220,179 @@ tasks:
...
+        wait:
+          cluster:
+            kind: Pod
+            name: app=curl
+            namespace: egress-gw-2
+            condition: Ready
+
+  - name: shared-egress-resources
+    description: Test reconciliation and clean-up of shared egress resources
+    actions:
```
mjnagel: Could this all be done in the [network test](https://github.com/defenseunicorns/uds-core/blob/main/test/jest/network.spec.ts) rather than in a shell script? We could do better typed assertions and patches there I think with KFC.

src/pepr/operator/controllers/istio/egress.ts
mjnagel: Just to reinforce this comment - I deployed slim-dev forgetting that the egress gateway was optional and was confused why everything appeared to work, but my pods couldn't connect to the hosts I allowed 😅

test/playwright/grafana.test.ts
```diff
@@ -46,7 +46,7 @@ test("validate namespace dashboard", async ({ page }) => {
     await page.goto(`/dashboards`);
     await page.click('text="Kubernetes / Compute Resources / Namespace (Pods)"');
     await page.getByTestId('data-testid Dashboard template variables Variable Value DropDown value link text authservice').click();
-    await page.getByRole('option', { name: 'grafana' }).click();
+    await page.getByRole('option', { name: 'authservice-test-app' }).click();
```
mjnagel, line 54: I think this may result in failure on the monitoring single-layer tests since the test apps are not deployed there (and these same tests are run for both). We should be able to conditionally set a different namespace here based on `fullCore` to cover each of the two tests (continuing to use `grafana` for `!fullCore`).

test/jest/network.spec.ts
```diff
@@ -266,4 +272,42 @@ describe("Network Policy Validation", () => {
...
+      `curl -s -o /dev/null -w "%{http_code}" https://httpbin.org/headers`
+    ];
+
+    // Validate successful tls request when using Egress Gateway for egress-gw-1
+    const success_response_tls = await execInPod("egress-gw-1", curlPodNameEgress1, "curl", egress_gateway_tls_curl);
+    expect(success_response_tls.stdout).toBe("200");
+
+    // Validate unsuccessful http request when using Egress Gateway for egress-gw-1
+    const not_success_response_http = await execInPod("egress-gw-1", curlPodNameEgress1, "curl", egress_gateway_http_curl);
+    expect(not_success_response_http.stdout).toBe("503");
```
mjnagel: Probably okay to use this for now, but we likely will need to switch this to something that checks for any error response code like [this](https://github.com/defenseunicorns/uds-core/pull/1259/files#diff-8715c128e04f8a56d28cbafd9d43d96dfd055cd1e2a47960f52d2917eedf0795R111-R127) in the future (necessary to support these tests on our IAC clusters). If that other PR lands first we will want to leverage that function.

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -92,14 +95,15 @@ export async function packageReconciler(pkg: UDSPackage) {
...
-    // Create the VirtualService and ServiceEntry for each exposed service
+    // Create the Istio Resources per the package configuration
     endpoints = await istioResources(pkg, namespace!);
 
     // Configure the ServiceMonitors
     const monitors: string[] = [];
     monitors.push(...(await podMonitor(pkg, namespace!)));
     monitors.push(...(await serviceMonitor(pkg, namespace!)));
 
+    // TODO: add status field for exposedHosts
```
mjnagel: Is this TODO still valid? I think this may have been dropped from the design?

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -219,6 +223,27 @@ export async function packageFinalizer(pkg: UDSPackage) {
...
+    }, log);
+  } catch (e) {
+    log.debug(
+      `Removal of shared egress resources during finalizer failed for ${pkg.metadata?.namespace}/${pkg.metadata?.name}: ${e.message}`,
+    );
+    await writeEvent(pkg, {
+      message: `Removal of shared egress resources failed: ${e.message}`,
+      reason: "RemovalFailed",
+    });
+  }
```
mjnagel: In general on finalizer failure we should `return false` (and update status) to make sure the finalizer is not removed if cleanup fails. Otherwise we won't have a great way to indicate to the user that deletion was unsuccessful and they may have manual cleanup required.

In this case it's a bit more complex since the end user probably doesn't know how to cleanup the shared resources so I'm not sure what to log out, but this would get "fixed" the next time pepr restarts...

mjnagel (commented):
Did some local validation and it appears to work as expected for the most part. I want to spend a bit more time looking at the share resources and testing some things there as well but wanted to get this first batch of feedback submitted.

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -0,0 +1,256 @@
...
+    hostPortsProtocol = {
+      host,
+      ports,
+      protocol,
+    };
+  }
+  return hostPortsProtocol;
+}
+
+function removeEgressResources(pkgId: string) {
```
chance-coleman: While this function is small and straightfoward it'd be good to have some test coverage here.

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -0,0 +1,256 @@
...
+  const protocol = allow.remoteProtocol ?? RemoteProtocol.TLS;
+
+  // reconcile ports
+  let ports = [];
+  if (allow.ports) {
+    ports = allow.ports;
+  } else if (allow.port) {
+    ports = [allow.port];
+  } else {
+    ports = [443];
```
chance-coleman, line 438: doesnt appear that we have any tests that test this usecase

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -0,0 +1,256 @@
...
+    await K8s(IstioServiceEntry)
+      .Apply(serviceEntry, { force: true })
+      .catch(async e => {
+        log.error(
+          `Failed to apply Service Entry ${serviceEntry.metadata?.name} of generation ${generation}`,
+        );
+        throw e;
+      });
+  }
+}
```
chance-coleman: largely this function is untested, probably a bit difficult with K8s things. Could either break it up to test non-apply things or mock the K8s return like we've done in other places.

src/pepr/operator/controllers/istio/gateway.ts
```diff
@@ -0,0 +1,111 @@
...
+            log.debug(
+              `Found existing Gateway ${gw.metadata?.name}/${gw.metadata?.namespace} with matching host. Istio will not behave properly with multiple Gateways using the same hosts.`,
+            );
+            break;
+          }
+        }
+      }
+    }
+  }
+}
```
chance-coleman: I think whatever is decided based on Micah's comment, should verify the functionality with some testing.

docs/reference/configuration/egress.md
```diff
@@ -0,0 +1,114 @@
...
+When a Package CR specifies the `network.allow` field with, at minimum, the `remoteHost` and `port` or `ports` parameters, the UDS Core operator will create the necessary Istio resources to allow traffic to egress from the mesh. These include the following:
+* An Istio ServiceEntry, in the package namespace, which is used to define the external service that the workload can access.
+* An Istio Sidecar, in the package namespace, which is used to enforce that only registered traffic can egress from the workload. This is only applied to the workload selected in the `network.allow`.
+* A shared Istio VirtualService, in the istio egress gateway namespace, which is used to route the traffic to the egress gateway.
+* A shared Istio Gateway, in the istio egress gateway namespace, which is used to expose the egress gateway to the outside world.
+* A shared Istio Service Entry, in the istio egress gateway namespace, to register the hosts and the ports for the egress gateway.
+
+## Limitations
+
+The configuration Package CRs in combination with the behavior of Istio should be understood when using egress. There are a few "gotchas" that might occur while using the egress configurations.
```
noahpb:
```suggestion
The configuration in Package CRs in combination with the behavior of Istio should be understood when using egress. There are a few "gotchas" that might occur while using the egress configurations.
```

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -6,28 +6,39 @@
...
+export const istioEgressGatewayNamespace = "istio-egress-gateway";
 
 /**
  * Creates a VirtualService and ServiceEntry for each exposed service in the package
+ * and creates or merges Istio resources for allowing egress traffic to external services
  *
  * @param pkg
  * @param namespace
  */
 export async function istioResources(pkg: UDSPackage, namespace: string) {
```
chance-coleman: We didn't have much testing here to begin with, but maybe this could be a good time to start? Also open to adding a follow on issue here since it was so untested before this PR.

src/pepr/operator/controllers/istio/virtual-service.ts
```diff
@@ -93,3 +109,130 @@ export function generateVSName(pkgName: string, expose: Expose) {
...
+        if (vsHost === host) {
+          log.debug(
+            `Found existing Virtual Service ${vs.metadata?.name}/${vs.metadata?.namespace} with matching host. Istio will not behave properly with multiple Virtual Services routing the same hosts.`,
+          );
+          break;
+        }
+      }
+    }
+  }
+}
```
chance-coleman: should probably add a test case for this scenario, also the note in the comment would be good to link to the egress.md doc limitations section

src/pepr/operator/controllers/istio/egress.ts
noahpb: To add on to this - should we back out of creating all egress related sources initially if `istioEgressGatewayNamespace` is not found? I did the same as @mjnagel above.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -65,33 +65,68 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+        policy.remoteHost ||
+        policy.remoteProtocol)
     ) {
       return req.Deny(
-        "remoteCidr cannot be combined with remoteGenerated, remoteNamespace, or remoteSelector",
+        "remoteCidr cannot be combined with remoteGenerated, remoteNamespace, remoteSelector, remoteHost, or remoteProtocol",
+      );
+    }
+
+    // If 'remoteHost' is set, it cannot be combined with 'remoteGenerated', 'remoteNamespace', 'remoteSelector', or 'remoteCidr'.
```
noahpb, line 120: There's no validation on a user specifying multiple `remoteHosts` in a given `allow` object (ie, literally specifying `remoteHosts` twice), but this should be okay for now. There are other places in our Package CR where we should be validating this. No action needed here, more of an FYI.

docs/reference/configuration/egress.md
meganwolf0: From what I can tell about the ambient implementation for egress, it really just requires the waypoint proxy - and I believe this architecture (with the virtual service and sidecar resources) would fail on a workload using ambient (tbh I haven't tested this yet, so might be worthwhile to do so before actually publishing this statement). The justification for this statement is based on this [gh thread](https://github.com/istio/istio/issues/52532) though.

I assume a follow on to this PR is to actually configure the proper resources in the case of ambient. However for this I'll just confirm this fails for an ambient workload (or no sidecar workload + no ambient - we do have some folks running this way currently) and add the suggestion you have above, if that works?

src/pepr/operator/controllers/istio/gateway.ts
meganwolf0: Ok that makes sense - should I move these calls to the end of the `applyEgressResources` routine so the resources are still created?

src/pepr/operator/controllers/istio/egress.ts
meganwolf0: yeah so I added this after I basically did the same thing that you and Noah mentioned above - deployed the default uds core without an egress gateway (because there was no namespace the package would never finalize when I tried to remove it) - I'm don't believe this error catch is OBE by making the updates suggested above - I suppose a user could manually remove the ns and a package still wouldn't finalize...

Anyway with that context, I do still feel like this catch is a nice to have, I could do like "unable to reconcile shared egress resources" or something, as the encapsulated logic both applies and removes old resources. Does that seem more appropriate?

docs/reference/configuration/egress.md
mjnagel: Yeah I think that works - just wanted to be extra explicit that this is only supported for sidecar mode and not supported for ambient (yet).

src/pepr/operator/controllers/istio/egress.ts
meganwolf0: Ok thinking about this one a bit, as this routine is used in both reconciling and finalizing the shared resources... I'm not sure how guarded you all want to be about users doing stuff... but I'm thinking if I do a ns existence check in `istioResources` and just don't run the egress methods without that passing, that handles the case brought up here. The case I mentioned below on finalizing (where it'll fail if the ns is removed ahead of package finalizing), then maybe I'd leave this alone to guard against that case?

Hopefully this makes sense, lmk if I need to clarify this point further or am missing some assumptions/behaviors

src/pepr/operator/controllers/istio/gateway.ts
mjnagel: I think it's actually better to have it up front (where it is) probably? That way we don't get "halfway" through egress resource creation and fail (leaving a bunch of resources but only "half" what is needed)?

src/test/tasks.yaml
meganwolf0: I was going to do this as a jest test at first, then I couldn't import the UDS Package type. Might have been missing a typescript-ism 🤷

src/pepr/operator/controllers/istio/egress.ts
mjnagel: Yeah I think updating the log line to what you stated (`unable to reconcile shared egress resources`) seems like a better, more generic fit here.

src/pepr/operator/controllers/istio/egress.ts
meganwolf0: Since this function modifies the local variable `inMemoryPackageMap`, I'm not really sure how to test this without modifying the function - I think I'd have to pass the map, modify and return it, then wherever it's called would have to enforce some kind of "lock" to ensure it's not being modified in parallel. Is there another way you see this happening that I might be missing? 

I believe this is indirectly tested in the e2e functionality, but if this does need some coverage I can modify the signature/add locks.

src/pepr/operator/crd/validators/package-validator.ts
meganwolf0: added the ambient check here as well

src/test/tasks.yaml
meganwolf0: sorry, going back through this - I think what I wanted to do was import the istio resources then validate those are created/destroyed. It seems like jest doesn't have access to the right packages in `src` to actually import those types here. So I just wrote those kubectl commands as a workaround for testing. But yeah if I'm missing something on how to do that, or if I simply need to redefine those types in the jest directory and import them and that is the preferred approach, then I can re-write this as such.

src/pepr/operator/controllers/istio/egress.ts
chance-coleman: yeah i see what you're saying there now. It's probably not worth the effort when this is covered in the E2E tests. The only alternative i can think of is `spy`-ing the debug log, but this is only partially testing the function.

src/test/tasks.yaml
meganwolf0: alrighty - I removed these manual command tests and added a decent bit of mocking tests in `ingress-resources.spec.ts` and `egress.spec.ts`. Tried to organize some of the shared concepts in `defaultTestMocks.ts`.

src/pepr/operator/reconcilers/package-reconciler.ts
meganwolf0: added an additional finalizer test for this, as I think that was missing previously

src/pepr/operator/controllers/istio/egress.ts
meganwolf0: should be added

src/pepr/operator/controllers/istio/egress.ts
meganwolf0: Added some mocking tests for the functions in this file, could use a sanity check on full coverage but did my best there.

src/pepr/operator/controllers/istio/istio-resources.ts
meganwolf0: I elected to not add tests for this function, largely because the scope of this PR is already pretty huge... also a lot of this is possibly OBE once ambient sweeps in... however I did end up breaking out the egress-specific portion into a separate function which I then did add testing for.

src/istio/values/base-istiod.yaml
```diff
@@ -15,6 +15,8 @@ meshConfig:
       envoyExtAuthzGrpc:
         service: "authservice.authservice.svc.cluster.local"
         port: "10003"
+  outboundTrafficPolicy:
+    mode: "ALLOW_ANY"
```
chance-coleman, line 19: i probably don't understand enough about this, but i believe this disables istios defaullt `REGISTRY_ONLY` posture. Is that safe? Could it we use a ServiceEntry + Sidecar to whitelist egress entries? @mjnagel gpt probably has a better understanding of this.

docs/reference/configuration/egress.md
```diff
@@ -0,0 +1,120 @@
...
+
+* Currently, egress will only work for workloads that are using the Istio sidecar proxy.
+
+* Specifying a port in a Package that is not exposed via the workload: This will be allowed with a warning from the operator, but the traffic will not be able to egress. An `istioctl analyze` will show an error such as: `Referenced host:port not found: "egressgateway.istio-egress-gateway.svc.cluster.local:9200"`
+
+* Specifying a remote host that is also used in other Gateways or VirtualServices: This will be allowed with a warning from the operator, but some unexpected behavior may occur. An `istioctl analyze` will show an error such as: `The VirtualServices ... define the same host ... which can lead to unexpected behavior` and `Conflict with gateways ...`
+
+* For all egresses defined within a single Package CR, all workloads that also have egress will have shared access to any host defined (is that true with the VS?)
+
+## Security Considarations
```
chance-coleman:
```suggestion
## Security Considerations
```

src/istio/values/base-istiod.yaml
mjnagel: `ALLOW_ANY` is actually the [default posture](https://istio.io/latest/docs/tasks/traffic-management/egress/egress-control/#envoy-passthrough-to-external-services:~:text=ALLOW_ANY%20is%20the%20default%20value%2C%20allowing%20you%20to%20start%20evaluating%20Istio%20quickly) so this doesn't change anything beyond explicitly setting it to that. In a previous thread (or maybe the design doc) I think we discussed this - defaulting to `REGISTRY_ONLY` would cause issues for anything allowing "egress anywhere" so the recommendation would be that end users apply `Sidecar` CRs to set this per namespace/workload where they want to lock it down further.

chance-coleman: ahh okay thanks for the context!

src/pepr/operator/controllers/istio/egress.ts
mjnagel: I think this is fine to move forward - still handles the reconciliation issue I was concerned with, but also won't fail finalizing for a case where there's nothing to clean up.

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -62,9 +89,108 @@ export async function istioResources(pkg: UDSPackage, namespace: string) {
...
+  // Error if egress gateway is NOT enabled in the cluster and we have egress resources
+  await K8s(kind.Namespace)
+    .Get(istioEgressGatewayNamespace)
+    .catch(e => {
+      if (hostResourceMap) {
+        log.error(
+          `Egress gateway is not enabled in the cluster. Please enable the egress gateway and retry.`,
+        );
+        throw e;
+      }
```
mjnagel: I might throw the `log.error` message here instead of the original error, at least conditionally if `e.status` is a 404.

Currently this throw causes a buggy behavior when you don't have the egress component deployed because our `handleFailure` function [sees the 404 and skips any status update/events](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/reconcilers/index.ts#L128-L132). More generally I find that the kubeapi responses aren't as useful as wrapping our own specific error message (like what you have logged here) when we know the specific failure.

chance-coleman (commented):
I'm happy with this as is, would like to get approvals from rest of team before merge though.

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -62,9 +89,112 @@ export async function istioResources(pkg: UDSPackage, namespace: string) {
...
+  await K8s(kind.Namespace)
+    .Get(istioEgressGatewayNamespace)
+    .catch(e => {
+      if (hostResourceMap) {
+        if (e.status == 404) {
+          const errText = `Egress gateway is not enabled in the cluster. Please enable the egress gateway and retry.`;
+          log.error(errText);
+          throw new Error(errText);
+        } else {
+          log.error(e, `Unable to reconcile shared egress resources.`);
```
mjnagel: slight nit, this is still an error caught when getting the namespace - so maybe something like this.

```suggestion
          log.error(e, `Unable to reconcile get the egress gateway namespace ${istioEgressGatewayNamespace}.`);
```

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -0,0 +1,268 @@
...
+        IstioVirtualService,
+        log,
+      );
+      await purgeOrphans(
+        generation.toString(),
+        istioEgressGatewayNamespace,
+        sharedEgressPkgId,
+        IstioServiceEntry,
+        log,
+      );
```
mjnagel: In some local testing with a large amount of packages (10x in my case) using 2 different host and a mix of ports/protocols...I think this generation + purge logic _can_ be pretty problematic. In my case I applied packages using 2 different hosts across http and tls, and I ended up with no gateways/virtualservices. In examining the logs the watcher purged both gateways/vs resources.

It seems like we might need locking and/or debounce around the shared resources (including the in-memory maps n such)? Locking would ensure that generation is only bumped once at a time, guaranteeing that we don't have concurrent operations editing/purging these. Debouncing would help so that on multiple package applies/pepr restart we aren't spamming the API server with updates. I can't really think of alternatives that would cover all the edge cases here, but open to other options as long as we can handle concurrency here well.

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -62,9 +89,115 @@ export async function istioResources(pkg: UDSPackage, namespace: string) {
...
+) {
+  // Error if egress gateway is NOT enabled in the cluster and we have egress resources
+  await K8s(kind.Namespace)
+    .Get(istioEgressGatewayNamespace)
+    .catch(e => {
+      if (hostResourceMap) {
+        if (e.status == 404) {
+          const errText = `Egress gateway is not enabled in the cluster. Please enable the egress gateway and retry.`;
+          log.error(errText);
+          throw new Error(errText);
```
mjnagel: Commenting here because it's the closest to relevant...

Curious on your thoughts, wondering if we should also fail reconciliation if a port is requested that isn't present in the egress gateway service? Just again leans more into surfacing information to the end user that something is misconfigured. As-is I can apply a package requesting port 1234 and it shows as ready/successful (and even creates the associated istio resources) but in practice wouldn't work because the egress gateway doesn't have that port configured.

mjnagel (commented):
Overall this is looking pretty good. My main concern at this point is with concurrency and race conditions, otherwise all the logic and resources seem sound.

src/pepr/operator/controllers/istio/egress.ts
meganwolf0: Tested this locally with this function updated - added both a lock and debounce as suggested. Hopefully that addresses the concurrency issue 🤞

meganwolf0:
Latest updates address failed batch package reconciliation, [here](https://gist.github.com/meganwolf0/6d02ecf09bcb3e6ba7aaae721918690b) is the test manifest. Expected that some packages fail, either due to invalid port egress or conflicting port/protocol match-ups.

chance-coleman (approved):
LGTM, thanks for sticking with this to get it over the finish line

PR #1334, opened 2025-03-04
feat: add status for removing / removalfailed

### LINKED ISSUE #963: Deleting Package before status is Ready does not properly clean up clients
### Environment
Device and OS:
App version:
Kubernetes distro being used:
Other:

### Steps to reproduce
1.  `kubectl apply -f package-a.yaml && sleep 1 && kubectl delete -f package-a.yaml`

### Expected result
Package should not have any Keycloak clients in cluster.

### Actual Result
Package leaves lingering Keycloak client

### Additional Context

### LINKED ISSUE #1159: Update Pepr finalizer code to reflect deletion status and failure
With the upstream Pepr changes in https://github.com/defenseunicorns/pepr/pull/1321 we are now able to handle our finalizer a bit better. The upstream changes allow us to `return false` from the finalizer function to skip removal of the finalizer from the resource. In particular this will allow us to:
- We can perform a state update to the `Package` CR to indicate that the CR is in a `Deleting` state
- Catch errors on deletion (i.e. keycloak client cleanup) and retry or put the CR into a `Failed Deletion` state?
- Handle issues like https://github.com/defenseunicorns/uds-core/issues/963 by waiting to run the finalizer until the Package CR is in a ready or Failed state?

Definition of done:
- When a package is deleted, it is updated with a status of `Deleting` while the finalizer runs
- When an error/failure is hit during finalizing, the Package is marked as `DeletionFailed` with an event noting the error and manual cleanup steps for the user (ex: delete client from keycloak manually)

mjnagel:
## Description

This PR utilizes the "new" ability in a Pepr finalizer to not remove the `finalizer`. This enables us to update the status while finalizing, and catch errors if cleanup does not work as expected. Changes:
- Skip finalizer if it's already running (based on status) 
- Skip finalizer if Package isn't ready/failed yet (for https://github.com/defenseunicorns/uds-core/issues/963)
- Patch `Removing` status on the CR
- Catch errors on finalization and patch `RemovalFailed` status and create a failure event
- Retry each cleanup/purge function using `retryWithDelay`

I also updated the diagram to support these changes, as well as adding test cases for the finalizer function. Diagram update can be previewed on the docs by using [this link](https://raw.githubusercontent.com/defenseunicorns/uds-core/c41964d426b8bb9780c26d41c631dbe6f50e854a/docs/.images/diagrams/uds-core-operator-uds-package.svg) on `docs/reference/configuration/UDS operator/package.md`, specific changes:
- Moved finalizer section to the right of reconciler
- Simplified flow of validator (to make more space in the diagram)
- Added new pieces of finalizer flow (failure, status patching, etc)

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/963

Fixes https://github.com/defenseunicorns/uds-core/issues/1159

## Type of change

## Steps to Validate

<details><summary>Testing Steps</summary>

Test setup:
```console
# Install slim-dev (unicorn flavor to avoid pull rate limiting)
# Create the test packages
zarf p create src/test --skip-sbom
# Deploy the test packages
zarf p deploy build/zarf-package-uds-core-test-apps-*.tar.zst --confirm
# Validate all package CRs go to Ready status
kubectl get pkg -A # should all show ready
```

Test that normal deletion works and makes events:
```console
# Delete a package CR
kubectl delete pkg -n test-tenant-app test-tenant-app
# Validate success and events
kubectl get pkg -n test-tenant-app # should show no resources
kubectl get events -n test-tenant-app | grep package # should show 3 removal events
```

Test that finalizer doesn't run until CR is ready:
```console
# This forces a re-reconcile of the package and then deletes immediately
# If you watch while this happens (k9s, etc) you should see it go to Pending before Removing
kubectl patch pkg httpbin-other -n authservice-test-app --subresource=status --type=json  -p='[{"op": "remove", "path": "/status"}]' && kubectl delete pkg httpbin-other -n authservice-test-app
# Validate that the watcher waited to finalize
kubectl logs -n pepr-system -l app=pepr-uds-core-watcher --tail=-1 | grep "Waiting"
kubectl get events -n authservice-test-app | grep package # should show 3 removal events
```

Test that finalizer places CR in RemovalFailed state on failed cleanup:
```console
# Deploy the test apps again (we need the sso client)
zarf p deploy build/zarf-package-uds-core-test-apps-*.tar.zst --confirm
# Edit the peprstore
kubectl edit peprstore -n pepr-system pepr-uds-core-store
# Delete the line with `uds-core-operator-v2-sso-client-uds-core-httpbin`, this is the client token and will make Pepr unable to cleanup the client
# Save the peprstore
# Delete the package CR
kubectl delete pkg httpbin-other -n authservice-test-app
# Make sure that status is marked as RemovalFailed (after ~15 seconds)
kubectl get pkg httpbin-other -n authservice-test-app
# Make sure events show up that client failed to be removed
kubectl describe pkg httpbin-other -n authservice-test-app
# Make sure that the SSO client removal was retried 4 times before final failure
kubectl logs -n pepr-system -l app=pepr-uds-core-watcher --tail=-1 | grep "cleanupSSOClients"
```

</details>

Also note the automated jest unit tests and validate those.

## Checklist before merging


src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -58,7 +58,13 @@ export async function keycloak(pkg: UDSPackage) {
     clients.set(client.clientId, client);
   }
 
-  await purgeSSOClients(pkg, [...clients.keys()]);
+  // Purge orphaned clients
+  try {
+    await purgeSSOClients(pkg, [...clients.keys()]);
+  } catch (e) {
+    log.error(e, `Failed to purge orphaned clients in for ${pkg.metadata!.name!}: ${e}`);
+  }
```
mjnagel, line 66: Just to note, this maintains the previous behavior where failures would just log out but we could also throw an error here or make an event. I made this change because `purgeSSOClients` will throw an error now if a token can't be found in the store.

src/pepr/operator/reconcilers/package-reconciler.spec.ts
chance-coleman, line 1: Do we have a test elsewhere that tests the retry functionality for 3 attempts? Going through your testing steps that seems to be a common use case for this functionality so could be a useful test ( if its not tested elsewhere ).

chance-coleman: Otherwise i can't find anything that needs updating.

mjnagel: Yeah this is a good callout - I'll add a test for this. I took an easy path of mocking out the entire retry function (and just having it call the function directly). With the delayed retries on all tests it was timing out on the ones where I expect failure. But I should be able to mock it more careful and have a test that throws an error the first time only.

mjnagel: On second thought - there are some existing tests of the retry function (see [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/utils.spec.ts)) that test a number of scenarios. I can add more wholistic testing here if we want, but I might lean towards keeping the tests here focused on this function/unit of code and trust those existing tests to handle the retry validation. Wdyt?

chance-coleman: Yeah definitely think those tests suffice!

mjnagel: Added some validation that `retryWithDelay` is getting called (or not called for our first few tests). I think this should cover it well - ensure the func is called and then the utils tests will cover that function behaving as expected.

chance-coleman (dismissed):
looks good to me (LGTM), ran through tests, everything is working as expected

chance-coleman (approved):

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -114,54 +115,108 @@ export async function packageReconciler(pkg: UDSPackage) {
...
-      message: `Removal of SSO / AuthService clients failed: ${e.message}`,
+      message: `Removal of SSO clients failed: ${e.message}. Clients must be manually removed from Keycloak.`,
       reason: "RemovalFailed",
+      type: "Warning",
     });
+    await updateStatus(pkg, { phase: Phase.RemovalFailed });
+    return false;
   }
 
+  // Indicate success - all other resources (network policies, virtual services, etc) are cleaned up through owner references
```
noahpb, line 218: <3 appreciate this comment as it answers a question that I had

noahpb (approved):

PR #1337, opened 2025-03-05
feat: document workaround for Keycloak and Apple M4 Macs

slaskawi:
## Description

This issue documents the workaround introduced in https://github.com/defenseunicorns/uds-core/pull/1314

## Related Issue

Relates to #1309

## Type of change

## Steps to Validate

N/A

## Checklist before merging


docs/reference/UDS Core/prerequisites.md
```diff
@@ -103,6 +103,14 @@ uds zarf package deploy uds-core --set CNI_CONF_DIR=/etc/cni/net.d --set CNI_BIN
 
 If you are using Cilium you will also need to make some additional configuration changes and add a cluster wide network policy to prevent Cilium's CNI from interfering with the Istio CNI plugin (part of the ambient stack). See the [upstream documentation](https://istio.io/latest/docs/ambient/install/platform-prerequisites/#cilium) for these required changes.
 
+#### Keycloak
+
+It has been reported that some version of Keycloak crash on Apple M4 Macbooks (the issue is tracked by [#1309](https://github.com/defenseunicorns/uds-core/issues/1309)). In order to apply a workaround, you have to override the `KEYCLOAK_HEAP_OPTIONS` variable and apply the `-XX:UseSVE=0 -XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G` value, for example:
```
chance-coleman:
```suggestion
It has been reported that some versions of Keycloak crash on Apple M4 Macbooks (the issue is tracked by [#1309](https://github.com/defenseunicorns/uds-core/issues/1309)). In order to apply a workaround, you have to override the `KEYCLOAK_HEAP_OPTIONS` variable and apply the `-XX:UseSVE=0 -XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G` value, for example:
```

docs/reference/UDS Core/prerequisites.md
```diff
@@ -103,6 +103,14 @@ uds zarf package deploy uds-core --set CNI_CONF_DIR=/etc/cni/net.d --set CNI_BIN
...
+
+```bash
+uds deploy k3d-core-slim-dev:0.37.0 --set KEYCLOAK_HEAP_OPTIONS="-XX:UseSVE=0 -XX:MaxRAMPercentage=70 -XX:MinRAMPercentage=70 -XX:InitialRAMPercentage=50 -XX:MaxRAM=1G" --confirm
+```
```
mjnagel: I think this is a good section for how to override on the dev/demo bundles. For end users making their own bundle they would need to setup an overrides section in their bundle. It might be good to give an example of that (keep it simple with `values` similar to what we did for the [Java memory example](https://uds.defenseunicorns.com/reference/configuration/resource-configuration-and-ha/#keycloak)). This might also be overkill since people generally don't make bundles when deploying on Macs (typically Mac deployments are dev/demo) - so open to not adding if we think the callout for dev/demo is sufficient. We should clarify that the variable `KEYCLOAK_HEAP_OPTIONS` is only available for the dev/demo bundles though 😄

slaskawi: In general, this section should go away mid April, once we have a Keycloak release that fixes both https://github.com/defenseunicorns/uds-core/issues/1309 and https://github.com/defenseunicorns/uds-core/issues/1212. 

Once that happens, we should probably stick with a generic way of passing environment variables in and suggest users to use `JAVA_OPTS_APPEND` (which has specifically been designed for this). The Heap workaround was a bit special and the problem happened only for containers and therefore using `JAVA_OPTS_KC_HEAP` was a better fit. But that was a very specific case. 

So let me add a more generic section describing the overrides, so that it can stay regardless to the workaround we describe here. Hope this sounds good to you!

mjnagel (approved):

Issue #1296, opened 2025-02-18
Use Client Credentials Grant for managing Keycloak Clients in UDS Operator


slaskawi:
## Summary

This ticket is the governing issue of redesigning how UDS Operator manages Keycloak Clients. 

At the time of writing, the UDS Operator uses the Dynamic Client Registration feature in Keycloak. This causes a number of issues and the goal of the new implementation is to switch to the Client Credentials Grant and Keycloak Admin APIs. 

The implementation will follow the [Managing Keycloak Clients from Pepr](https://www.notion.so/Managing-Keycloak-Clients-from-Pepr-1a5e512f24fc80c4bc35d13b9d7a1367) Technical Design

## Checklist

- [x] Handling of the Client Secret rotation
- [x] Figure out Access Rules
- [x] Update the logic in Pepr to leverage 3 states - old, new and auto
- [x] Enable Client Policy and Fine-grained Authorization, so that Pepr could only see its own Clients
- [x] Tests for Client Credentials grant in UDS Identity Config
- [x] Document the Realm upgrade (docs, a video and maybe an OpenTofu file)
- [x] Document Pepr logic for picking the implementation
- [x] Update the design with the latest findings

slaskawi:
Implemented.

Issue #354, opened 2024-04-18
Provide automated way of updating pods leveraging generated SSO secrets


mjnagel:
### Is your feature request related to a problem? Please describe.

Currently Grafana and NeuVector in uds-core mount secrets that are generated by pepr. If these secrets were to update due to (for example) a client id change or other field update, the pods mounting them would need to be cycled.

This can be (and as of https://github.com/defenseunicorns/uds-core/pull/351 is being) done using actions to annotate pods with a checksum of the secret. Then on updates if the secret has changed, the checksum will, resulting in a cycling of the pod.

While this does work, for neuvector specifically, anything on top of uds-core (or even Grafana) would have to build their own bespoke action to handle this.

### Describe the solution you'd like

This seems like a viable place to leverage pepr/operator intelligence to auto-cycle things for us (maybe handled the same checksum way, maybe just directly cycled). Two potential paths I can think of:
- Provide an optional field in the spec for a selector of pods that leverage the secret
- Automatically look up places the secret is used (env/volumes)

### Describe alternatives you've considered

- Leave this to the end user/actions + document?
- Something else with pepr/default actions?

Racer159:
This is something that SWF is interested in as well - not just for SSO secrets but out config secrets as well (e.g. the one hooking up Postgres or Object Storage) - if UDS Core had a way to watch secrets and run rollout restarts on Pods or checksum and annotate or another mechanism that would be :man_cook: :kiss: )

mjnagel:
> This is something that SWF is interested in as well - not just for SSO secrets but out config secrets as well (e.g. the one hooking up Postgres or Object Storage) - if UDS Core had a way to watch secrets and run rollout restarts on Pods or checksum and annotate or another mechanism that would be 👨‍🍳 💋 )

If the easiest path forward for SSO secrets is to do something more generic I'm good with that. First thoughts are we'd probably want/need to label the secrets with something like `uds.dev/auto-update-subscribers` (terrible name) and then have Pepr watch all create/update events on secrets with that label. We would then potentially need to lookup all "subscribers" to that secret and/or use a different label on them to keep them up to date.

If we do go down the "generic" route we should probably write up a design doc on this.

PR #1341, opened 2025-03-06
feat: use Client Credentials for managing Keycloak Clients

### LINKED ISSUE #1330: Redesign Client Management algorithm in Pepr
## Summary

This is a sub-task of https://github.com/defenseunicorns/uds-core/issues/1296

The goal of this ticket is redesign the algorithm for managing Clients used by Pepr

slaskawi:
## Description

This Pull Request introduces an alternative mode for managing Keycloak Clients by the UDS Operator.

The UDS Operator now discovers if Keycloak has been properly configure (and if it supports the configuration delivered by https://github.com/defenseunicorns/uds-identity-config/pull/358) and switches to Client Credentials if it does. In other case, the UDS Operator will keep using the Dynamic Client Registration feature. 

The behavior might be overridden by using the `PEPR_KEYCLOAK_CLIENT_STRATEGY` Environment Variable in the UDS Operator that can have 3 values:

* `dynamic_client_registration` - the Dynamic Client Registration mode
* `client_credentials` - using the Client Credentials mode
* `auto` - the Operator probes if a token obtained by the Client Credentials Grant looks good and picks proper mode according to the results of the check.

The documentation for this feature will be delivered here: https://github.com/defenseunicorns/uds-core/pull/1367

## Related Issue

* Fixes #1330
* Relates to #1296
* [Technical Design](https://www.notion.so/Managing-Keycloak-Clients-from-Pepr-1a5e512f24fc80c4bc35d13b9d7a1367)

## Type of change

## Steps to Validate

This functionality primarily relies on the regression testing. 

## Checklist before merging


src/keycloak/chart/templates/secret-client-secrets.yaml
```diff
@@ -0,0 +1,13 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "keycloak.fullname" . }}-client-secrets
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "keycloak.labels" . | nindent 4 }}
+type: Opaque
+data:
+  ignored-value: {{ "ignored-value" | b64enc }}
```
slaskawi: Remove this line.

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -97,9 +97,9 @@ spec:
             - name: redirect-welcome
               uri:
                 exact: /
-            - name: redirect-admin
-              uri:
-                prefix: /admin
+{{/*            - name: redirect-admin*/}}
```
slaskawi: Can I revert this piece?

src/keycloak/chart/values.yaml
```diff
@@ -278,4 +278,6 @@ autoscaling:
           value: 1
           periodSeconds: 300
 
-env: []
+env:
```
slaskawi: Revert this

src/pepr/operator/controllers/keycloak/client-secret-sync.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { K8s, kind } from "pepr";
+import { v4 as uuidv4 } from "uuid";
+import { Component, setupLogger } from "../../../logger";
+
+const KEYCLOAK_CLIENT_SECRET_KEY = "uds-operator";
+
+const log = setupLogger(Component.OPERATOR_CONFIG);
+
+export async function updateKeycloakClientsSecret(config: kind.Secret) {
```
slaskawi: Add docs

src/pepr/operator/controllers/keycloak/keycloak-client.ts
```diff
@@ -0,0 +1,409 @@
...
+
+    throwErrorIfNeeded(response);
+    await this.updateRegistrationTokenInStore(response.data);
+
+    return response.data;
+  }
+
+  async delete(
+    client: Partial<Client>,
+    previouslyObtainedRegistrationToken: string = "",
```
slaskawi: Does the `previouslyObtainedRegistrationToken` make sense here?

src/pepr/operator/controllers/keycloak/keycloak-client.ts
```diff
@@ -0,0 +1,409 @@
...
+  private readonly clientsAdminUrl: string;
+
+  constructor(baseUrl: string) {
+    this.clientsAdminUrl = baseUrl + "/admin/realms/uds/clients";
+    this.clientCredentialsUrl = baseUrl + "/realms/uds/protocol/openid-connect/token";
+  }
+
+  public async getAccessToken(previouslyObtainedToken: string = "") {
+    log.warn(`ClientCredentialsKeycloakClient: getting token`);
+    if (previouslyObtainedToken) {
```
slaskawi: Store token as a field

src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -37,22 +38,33 @@ spec:
...
         - source:
-            notNamespaces: 
+            notNamespaces:
               - "pepr-system"
+    - to:
+        - operation:
+            ports:
+              - "8080"
+            paths:
+              - "/admin/realms/{{ .Values.realm }}/clients"
```
slaskawi, line 56: Update the design: mention that we can't block clients because Keycloak doesn't have a method to get a Client by the Client ID. This happens through the collection endpoint.

src/pepr/operator/controllers/keycloak/keycloak-client.ts
```diff
@@ -0,0 +1,397 @@
...
+ * Dynamic Keycloak Client implementation that chooses between different Keycloak client implementations
+ */
+export class DynamicKeycloakClient implements KeycloakClient {
+  public static readonly ENV_KEYCLOAK_CLIENT_IMPLEMENTATION = "PEPR_KEYCLOAK_CLIENT_STRATEGY";
+  public static readonly ENV_KEYCLOAK_CLIENT_IMPLEMENTATION_DYNAMIC = "dynamic_client_registration";
+  public static readonly ENV_KEYCLOAK_CLIENT_IMPLEMENTATION_CLIENT_CREDENTIALS =
+    "client_credentials";
+  public static readonly ENV_KEYCLOAK_CLIENT_IMPLEMENTATION_AUTO = "auto";
+
+  private readonly dynaamicClientRegistrationKeycloakClient: DynamicClientRegistrationClient;
```
chance-coleman:
```suggestion
  private readonly dynamicClientRegistrationKeycloakClient: DynamicClientRegistrationClient;
```

typo

src/pepr/operator/controllers/keycloak/keycloak-client.ts
```diff
@@ -0,0 +1,397 @@
...
+    throwErrorIfNeeded(response);
+
+    const accessToken = response.data.access_token;
+    this.validateAccessToken(accessToken);
+    this.cachedToken = accessToken;
+    return this.cachedToken;
+  }
+
+  validateAccessToken(accessToken: string) {
+    /* eslint-disable @typescript-eslint/no-explicit-any */
```
chance-coleman: generally speaking this kinda frowned upon in Typescript since its bypassing the strong typed language requirements. It would probably be best to avoid this. I think for this use case you're going to want to define an interface for the JWT, something like :
```
interface KeycloakToken {
  exp: number;
  resource_access: {
    "realm-management": {
      roles: string[];
    };
    [key: string]: unknown;
  };
  [key: string]: unknown;
}
```
which can then be referenced as `const jwt = decodeJwt(accessToken) as KeycloakToken;`

src/pepr/operator/controllers/keycloak/keycloak-client.ts
```diff
@@ -0,0 +1,397 @@
...
+  async createOrUpdate(client: Partial<Client>): Promise<Client> {
+    const obtainedClient = await this.get(client);
+    if (obtainedClient) {
+      return this.update(client, obtainedClient);
+    }
+    return this.create(client);
+  }
+}
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
chance-coleman: As mentioned above, probably should avoid this eslint-disable. I think, not completely confident in this suggestion, you can use a generic type like this:
```
function throwErrorIfNeeded<T>(response: FetchResponse<T>): void {
  log.debug(`Checking response: ${JSON.stringify(response)}`);
  if (!response.ok) {
    const { data, status, statusText } = response;
    if (data) {
      throw new Error(
        `${JSON.stringify(status)}, ${JSON.stringify(statusText)}, ${JSON.stringify(data)}`
      );
    }
    throw new Error(`${JSON.stringify(status)}, ${JSON.stringify(statusText)}`);
  }
}
```

might lose some of the verbosity of the log though..

src/pepr/operator/controllers/keycloak/keycloak-client.ts
chance-coleman, line 1: I know you're coming from a Java background, so i can definitely see why and what you've done here.

With that said in TS/JS it's typically more functional based (idiomatic) and less class based. What you have definitely works and is one way to do this. However it comes at the cost of complexity, readability, and maintainability. With a functional approach we can avoid the boiler plate interface as well as different classes, instead there would be functions that are specific to the different client implementations. This would result in common code being pulled into shared functions and minimizing the duplicity, also improves test-ability, and with smaller functions its easier to refactor in the future. 

I'd like to get @mjnagel 's thoughts on this as well before you go down this rabbit hole.

src/pepr/operator/controllers/keycloak/keycloak-client.ts
```diff
@@ -0,0 +1,397 @@
...
+          log.info("Client Credentials Keycloak Client is available");
+          return this.clientCredentialsKeycloakClient;
+        } catch {
+          log.info(
+            "Client Credentials Keycloak Client is not properly configured, falling back to Dynamic Client Registration...",
+          );
+          return this.dynaamicClientRegistrationKeycloakClient;
+        }
+      default:
+        throw new Error(
```
chance-coleman: Im not sure im following how this Error will be handled elsewhere in the implementation, but i wonder if it would be better to fail gracefully with a `log.error` and default of `auto`?

src/pepr/operator/controllers/keycloak/keycloak-client.ts
mjnagel: While classes are definitely supported and work here, I tend to agree that the functional approach might be better to align with our current code "style". In general I'd expect this to look like 3-4 (directly called) common functions for create/update/delete (and maybe create/update). Then each of those might have common logic to detect the proper mode and call either dynamic or clientCred functions for create/update/delete. Potentially could also continue to pull out (or pull out more of) the common functionality for api requests (if/where it makes sense)?

src/pepr/operator/controllers/keycloak/keycloak-client.ts
```diff
@@ -0,0 +1,397 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
+import { Client } from "./types";
+import { fetch, K8s, kind } from "pepr";
+import { Store } from "../../common";
+import { retryWithDelay } from "../utils";
+import { Component, setupLogger } from "../../../logger";
+import { FetchResponse } from "kubernetes-fluent-client/src/fetch";
+import { decodeJwt } from "jose";
```
mjnagel: I haven't fully validated any of [these stackoverflow solutions](https://stackoverflow.com/questions/38552003/how-to-decode-jwt-token-in-javascript-without-using-a-library), but it would be good to remove this external dependency if we can and implement this JWT decoding on our own. Looking upstream at the [source](https://github.com/panva/jose/blob/08d1bd21eebf1fcd6a05c05e9e1dc469213b34bd/src/util/decode_jwt.ts#L31) for this I don't think it's overly complex to reimplement on our own (and the stackoverflow answers look similar).

There's in general no issue with us importing new libraries, but we should try to be critical and recreate/copy logic for simple things like this (<10 lines for some of those stackoverflow solves) to reduce our supply chain risks as much as possible.

src/pepr/operator/controllers/keycloak/client-secret-sync.ts
```diff
@@ -0,0 +1,25 @@
...
+  config.data = config.data || {};
+
+  // This might be a bug but it seems Zarf adds managedFields, which is prohibited in Secrets.
+  delete config.metadata?.managedFields;
+
+  if (!config.data[KEYCLOAK_CLIENT_SECRET_KEY]) {
+    log.info("Generating new Keycloak client secret");
+    config.data[KEYCLOAK_CLIENT_SECRET_KEY] = Buffer.from(uuidv4()).toString("base64");
+    await K8s(kind.Secret).Apply(config);
+  }
```
mjnagel: This will generate a new random secret when one doesn't exist. Is there anything following that generation that cycles Keycloak to make sure the new secret is mounted/loaded to the pod? Might just be missing something here, but it seems like the identity-config code assumes the secret exists (which I think makes sense) but there's no real guarantee around the timing of this secret being created/modified before Keycloak starts up and mounts it.

src/pepr/operator/index.ts
```diff
@@ -114,3 +115,10 @@ When(a.Secret)
   .InNamespace("pepr-system")
   .WithName("uds-operator-config")
   .Reconcile(updateUDSConfig);
+
+// Watch the Kubernetes Clients Secret
+When(a.Secret)
+  .IsCreatedOrUpdated()
+  .InNamespace("keycloak")
+  .WithName("keycloak-client-secrets")
+  .Reconcile(updateKeycloakClientsSecret);
```
mjnagel: This probably makes more sense as a mutation, rather than a watch/reconcile? If the function is entirely, only modifying the contents of the secret then it would be better to do this at mutate time. This ties into my other comment on timing/cycling of Keycloak - if you have the secret mutated before Keycloak mounts it you can dodge _some_ of those concerns.

src/keycloak/chart/templates/secret-client-secrets.yaml
mjnagel: Semi-related to the other two secret comments - that fact that this secret exists in the helm chart without any value for the `uds-operator` key means that each time the Keycloak chart is installed (whenever a helm install/upgrade happens) it will wipe out that key. Then the `updateKeycloakClientsSecret` function will generate a new one. Is that the expected/desired behavior? It does make for a handy automatic rotation, likely every week or two weeks at minimum (with new releases), but may also have new implications/considerations for cycling Keycloak and making sure Pepr waits to make requests until the new secret is loaded, etc.

src/pepr/operator/controllers/keycloak/client-secret-sync.ts
slaskawi: At the Keycloak boot time, we just need this secret to exist (as otherwise, Kubernetes won't schedule the Pod due to a missing Secret). The value is totally dynamic and we need it only when UDS Operator authenticates to Keycloak. On both sides (UDS Operator and Keycloak plugin) there's no caching whatsoever, and the value is read from the volume every time to ensure it's fresh and valid. 

In other words - this value is intended to be dynamic. You should be able to delete this key whenever you want and it will get instantly regenerated and used on both sides.

src/pepr/operator/controllers/keycloak/keycloak-client.ts
slaskawi: I've completely rewritten this code and this comment is probably not valid anymore.

src/pepr/operator/controllers/keycloak/keycloak-client.ts
slaskawi: After spending quite a while in the IDE, I couldn't figure it out and went a different route:

```typescript
export interface RestResponse {
  ok: boolean;
  status: number;
  statusText: string;
  data: unknown;
}

export async function throwErrorIfNeeded(response: RestResponse) {
  if (!response.ok) {
    const { status, statusText, data } = response;
    throw new Error(`${status}, ${statusText}, ${data ? JSON.stringify(data) : ""}`);
  }
}
```

It's slightly more complicated but seem to do the job

src/pepr/operator/controllers/keycloak/keycloak-client.ts
slaskawi: Agreed! 

Thanks to @mjnagel's example, I refactored the entire file. Please check it out again.

src/pepr/operator/controllers/keycloak/keycloak-client.ts
slaskawi: I refactored this piece very heavily and vendored your suggestion in.

src/pepr/operator/controllers/keycloak/keycloak-client.ts
slaskawi: Done! Thanks for suggestion.

src/pepr/operator/index.ts
slaskawi: I'll skip the timing/ordering issue here (as I think I described it in https://github.com/defenseunicorns/uds-core/pull/1341#discussion_r2007033576) and focus only on the general approach.

I tried switching to the `.Mutate` function and found it problematic. If UDS Operator malfunctions and when it's down I modified the secret and removed `uds-operator` key, it has never been recreated. With the Reconciler approach, my code is always called (in the worst case - when pepr get back to the operational state). Moreover, I think writing this as a Reconciler gives us more flexibility for the future - we may want to rotate it every X days or maybe even hours.

src/keycloak/chart/templates/secret-client-secrets.yaml
slaskawi: I think we'll have to make some compromise here. 

If we decided to let the UDS Operator create this secret, we'll end up in a weird situation where it will have to create the `keycloak` namespace to ensure the secret is ready when Keycloak StatefulSet is deployed. I found this reference across namespaces weird.

The alternative is let the Keycloak Helm Chart create it and the UDS Operator will just watch. The downside is like you said - reinstalling will cause the secret rotation. As I explained earlier (in https://github.com/defenseunicorns/uds-core/pull/1341#discussion_r2007033576), this shouldn't be a big deal and as a matter of fact, the upgrade tests seem to be green. 

Finally, even if we are very unlucky with the timing/ordering and the secret is for some reason empty, the retry mechanism from the UDS Operator should do the trick.

src/pepr/operator/controllers/keycloak/keycloak-client.ts
slaskawi: Done! Thanks for a great hint!

src/pepr/operator/controllers/keycloak/client-secret-sync.ts
mjnagel: Got it - I think that answers this specific concern, was overlooking that this being a volume on the pod guarantees its existence before the pod starts up.

src/keycloak/chart/templates/secret-client-secrets.yaml
mjnagel: EDIT: Striking out the first half of this comment as it's not accurate, see below replies. The volume will be eventually consistent with the secret contents. I believe the second half of this comment is still valid/helpful though.

~~The problem with this is that a secret updating in cluster is not the same as it updating in the pod (on the volume). The volume contents will only be "reloaded" on startup of the pod (there are tools that solve this like [reloader](https://github.com/stakater/Reloader) for example because it's a prominent issue - we have a similar issue with [SSO secrets](https://github.com/defenseunicorns/uds-core/issues/354)).~~

~~So with this code/the identity config code Pepr is pulling from the secret in the cluster each time (with the latest) whereas Keycloak/our plugin is pulling from the volume (which may or may not be the same, depending on when the secret changed and when Keycloak last cycled). Essentially we would need to update the in-cluster secret, then cycle Keycloak immediately after to be guaranteed that the plugin has the same secret as the operator. While Keycloak is cycling, especially with multiple replicas, we would probably hit errors trying to register clients, depending on which replica is hit and whether it has the updated secret I suspect? While the reconciliation retries might handle this, to some extent it's gambling that our retry count + delay is greater than the time it takes Keycloak to cycle.~~

I think controlling the secret with the operator might actually be more ideal here rather than including in the chart. We already have [a case where we do that today for authservice](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/authservice/config.ts#L45-L72) so it is something we've done before. We could also run this with an [`OnSchedule`](https://docs.pepr.dev/v0.46.3/user-guide/onschedule/) to cycle the secret daily/weekly (on some cadence), in addition to whenever it doesn't exist (for first startup and if the user wants to force a cycling of the secret).

This would handle some of the secret management and cycling, but wouldn't entirely handle the auth "problem" while the secret/volume is updating on Keycloak's side. I'd probably have to see the actual behavior here to understand how much failure this causes / how quick the secret/volume updates here. If the operator is the one cycling the secret we could maintain (in-memory) a cache of the "old secret" to try if the new one fails. On this comment - `this shouldn't be a big deal and as a matter of fact, the upgrade tests seem to be green` - my understanding is that this PR actually isn't running the new client credentials code at all because that doesn't exist in the identity-config build being used here? So I'm not sure the current CI upgrade tests are a good indication of how this will perform.

src/pepr/operator/index.ts
mjnagel: I typed too many words on the [other thread here](https://github.com/defenseunicorns/uds-core/pull/1341#discussion_r2007722289) which I think is the overarching issue here so going to resolve this thread to keep the discussion in one place there.

src/pepr/operator/controllers/keycloak/clients/dynamic-client.ts
```diff
@@ -0,0 +1,52 @@
...
+      log.warn("Cannot use Client Credentials, falling back to dynamic registration");
+      return "dynamic";
+    }
+  } else {
+    log.warn(
+      `Invalid ${process.env.PEPR_KEYCLOAK_CLIENT_STRATEGY} parameter value, falling back to dynamic registration`,
+    );
+    return "dynamic";
+  }
+}
```
mjnagel: Bit of a nit but it would be nice to use something "typed" here like an enum rather than just a string. We can keep it really simple, for example:
```ts
enum ClientStrategy {
  AUTO = "auto",
  CLIENT_CREDENTIALS = "client-credentials",
  DYNAMIC = "dynamic",
}
```

This just helps make sure we have better type checking/protection in the code, rather than relying on string matching. If you make this change, we'd want to use that enum everywhere (in the other functions when we pick strategy as well).

src/pepr/operator/controllers/keycloak/clients/dynamic-client.ts
```diff
@@ -0,0 +1,52 @@
+/**
```
mjnagel: I think naming this file with `dynamic` could be confusing since we have the "dynamic-client-registration" as well - it took me a bit to figure out where these functions were and why. While calling this the dynamic client is totally accurate I think we should pick a different name for this file and/or move this into the common file maybe?

chance-coleman: This definitely caught me off guard at first, +1 for putting this code in the common.ts file.

src/keycloak/chart/templates/secret-client-secrets.yaml
mjnagel: I don't know if we'd want to pull this in, but one way to avoid cycling keycloak would be to use something like the [k8s-sidecar](https://github.com/kiwigrid/k8s-sidecar) as a container in the sts. We already use this for Grafana so it's not a new dependency, but we might want to review the resource impact, etc against our own solution. In effect this would be a sidecar watching the secret for updates and changing a file on a volume immediately, removing the limitation of "you must cycle the pod for it to read the new secret".

noahpb: Is cycling Keycloak necessary? I originally thought cycling was necessary as well, but per the Kubernetes Documentation ([ref](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod)) and my own local testing, the secret contents will eventually get updated on the pod's filesystem (usually around ~1min or less in my tests). Or is cycling necessary to ensure that Keycloak only uses the updated/most recent value after `updateKeycloakClientsSecret` runs?

src/pepr/operator/README.md
```diff
@@ -155,6 +155,20 @@ spec:
           bearer_only: clientField(bearerOnly)
 ```
 
+### Controlling how UDS Operator interacts with Keycloak
+
+The UDS Operator can interact with Keycloak in two primary ways: using the Dynamic Client Registration or Client Credentials Grant. The method used is determined automatically or by specifying the environment variable `PEPR_KEYCLOAK_CLIENT_STRATEGY`.
+
+Dynamic Client Registration allows the UDS Operator to dynamically register new Clients in the Keycloak server. A successful registration flow results in the Registration Token to be stored in Pepr store, which can be later used for modifying and remocing the client.
```
noahpb:
```suggestion
Dynamic Client Registration allows the UDS Operator to dynamically register new Clients in the Keycloak server. A successful registration flow results in the Registration Token to be stored in Pepr store, which can be later used for modifying and removing the client.
```

src/keycloak/chart/templates/secret-client-secrets.yaml
mjnagel: That's a good callout - I think typically it is "required" because of the way applications function (i.e. they load config at startup, so the process has to be restarted to load the config on disk). In this case since our plugin code pulls the file each time, we might not need to restart the process/pod.

We probably should review/evaluate the impact of delayed updates here, and may still want to consider some of the suggestions in this thread such as:
- Operator managing the secret entirely (removes unnecessary changes of the secret)
- Caching the "old" secret when an update happens and trying that on failure (reducing failures/retries needed when the secret hasn't yet updated on disk)

bburky: Volume mounts will be automatically updated by Kubernetes (but environment variables cannot be updated this way).

It depends on the app if this is sufficient alone. Is this our custom Java code reading this? Just make sure to always read the file off disk before using it to make sure it's up to date (slightly inefficient, but should be fine). If you really want to you can monitor the file for changes with inotify I think, but it's really tricky with k8s.

bburky: You maybe could try adding `optional: true` to the volume mount too? Then let Pepr handle the creation of this Secret entirely? Let it be missing at Helm install time and Pod startup.

`optional: true` will allow the StatefulSet to start with the volume missing. I'm unsure though if it will automatically get added once it does appear. That's a different situation than the normal "update an existing Secret and update the active volume mounts" kind of thing

mjnagel: > You maybe could try adding optional: true to the volume mount too? Then let Pepr handle the creation of this Secret entirely? Let it be missing at Helm install time and Pod startup.

Since Pepr comes up before Keycloak we can create the secret at startup of Pepr (similar to the Authservice one) and then if something is off with timing the pod will only schedule/startup once that secret exists. I don't think we need `optional` here and the potential volume update impacts?

bburky: > Since Pepr comes up before Keycloak we can create the secret at startup of Pepr 


Ah, yes. Ok. I would avoid `optional: true` if you can. If Pepr is installed first (or concurrently), I think you're fine? Pepr may not have finished creating it when Keycloak is installed, but it once it creates it, it will allow the Pod to be scheduled.

src/pepr/operator/controllers/keycloak/clients/dynamic-client.ts
slaskawi: I renamed it to `keycloak-client.ts`

src/pepr/operator/controllers/keycloak/clients/dynamic-client.ts
slaskawi: Fixed

src/keycloak/chart/templates/secret-client-secrets.yaml
slaskawi: @mjnagel @bburky Here's the summary of this thread as I understand it:

1. Let's move the responsibility of creating the Client Secrets to the UDS Operator and implement it similarly to the AuthService Secret.
2. Keycloak shall use the `optional: true` flag to prevent blocking pods rollout
3. The suggestion is to add an `OnSchedule` and automatically rotate the secret. 

For the 1 and 2 - 100% agree. I'm a bit hesitant of point no 3 as we don't have this type of code in the UDS Operator yet. I'll add it as it has been requested but let's double check if we really need this @mjnagel ?

bburky: It's really just:
1. Let's move the responsibility of creating the Client Secrets to the UDS Operator and implement it similarly to the AuthService Secret.
   * Do not use `optional: true`. Keycloak will wait to start until pepr creates it 
2. Make sure the custom Java code re-reads the secret off disk every time you need it 
   * I think you may already be doing this. This avoids needing to monitor the file for changes. We don't use this secret too often, so I don't think we need to be any more efficient than this

bburky: Oh, sorry I missed the `OnSchedule` part, that is a @mjnagel question

src/pepr/operator/controllers/keycloak/client-secret-sync.ts
```diff
@@ -0,0 +1,31 @@
...
+const log = setupLogger(Component.OPERATOR_CONFIG);
+
+export async function updateKeycloakClientsSecret(
+  config: kind.Secret,
+  forceRotation: boolean = false,
+) {
+  config.data = config.data || {};
+
+  // This might be a bug but it seems Zarf adds managedFields, which is prohibited in Secrets.
+  delete config.metadata?.managedFields;
```
mjnagel, line 32: This is actually "normal" and your solve here should work, we've done this with some other updates like our netpols: https://github.com/defenseunicorns/uds-core/blob/c1703221b85d37451d60c226863a0b168e702e01/src/pepr/operator/controllers/network/generators/kubeNodes.ts#L128-L131

src/keycloak/chart/templates/secret-client-secrets.yaml
mjnagel: @slaskawi agree with what @bburky noted above. For the `OnSchedule` I'm okay leaving off for now if we want - the idea there was just that we would have a way for the secret to automatically cycle on some cadence, making sure that it's never older than a week or so. I'm not sure that it's strictly required, but it would seem to provide a better security posture around this "admin secret"? We haven't had a great use case for `OnSchedule` yet but this seems like one that would be good. We could make it conditional and/or expose the cadence as configuration too if we want to leave this to the end user risk posture.

One more thing I'd point out from this thread:
> Caching the "old" secret when an update happens and trying that on failure (reducing failures/retries needed when the secret hasn't yet updated on disk)

If we're not cycling the secret often this might not be important, but curious on your thoughts. If Pepr changes the secret (i.e. `forceRotation` true) then there is likely a short window where Keycloak hasn't updated yet (per @noahpb's testing maybe upwards of 1 minute). Perhaps we can cache the older value when we `forceRotation` and have it as a fallback in the code?

src/pepr/operator/controllers/keycloak/clients/client-credentials.ts
```diff
@@ -0,0 +1,143 @@
...
+    try {
+      const jwt = parseKeycloakToken(cachedToken);
+      if (jwt.exp && jwt.exp > Math.floor(Date.now() / 1000) + 5) return cachedToken;
+    } catch (e) {
+      log.error(e, "Failed to parse cached token");
+      cachedToken = null;
+    }
+  }
+
+  const secret = await K8s(kind.Secret).InNamespace(SECRET_NAMESPACE).Get(SECRET_NAME);
```
mjnagel, line 56: I'm not totally against this, but I wonder if it makes sense to just cache this secret? Since we have a watch on the secret, we can "refresh the cache" inside the `updateKeycloakClientsSecret` function and avoid needing the network call here?

slaskawi: I would prefer to leave it as is. This code shouldn't be very problematic as we already cache the token that has TTL set to 1 minute. My intuition is that UDS Operator will probably reconcile most of the SSO Clients during this time-frame if the cluster is in steady state. 

Let's leave it for now. I'd like to avoid making some silly mistake in caching this.

src/keycloak/chart/templates/secret-client-secrets.yaml
slaskawi: > Do not use optional: true. Keycloak will wait to start until pepr creates it

@bburky Removed!

slaskawi: We had a long chat about the `OnSchedule` piece with @mjnagel and @UnicornChance. We realized that we don't really know what is the desired cadence for rotating them. If we rotate not often enough, we won't check all the marks. If we rotate too often, we might cause disruptions for no good reason. 

We agreed to create a ticket to collect the secret rotation requirements and remove this piece of code from this Pull Request. 

I'll also add an information how to rotate the secret manually for the sake of completeness.

docs/.images/diagrams/uds-core-pepr-operator-flow.drawio
```diff
@@ -1,9 +1,9 @@
-<mxfile host="app.diagrams.net" agent="Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36" version="26.0.16">
```
mjnagel, line 1: Two comments on the diagram:
- Can we represent that it only uses one of the two client registration mechanisms depending on config? Right now it sort of looks like it calls both (which might be _partially_ accurate with auto mode 😄)
- Related to ^ I don't think we're using the peprstore is the keycloak admin api is used, so let's make sure that is only in the client registration "flow".

src/pepr/operator/controllers/keycloak/config.ts
```diff
@@ -0,0 +1,61 @@
...
+      const secret = await K8s(kind.Secret)
+        .InNamespace(KEYCLOAK_CLIENTS_SECRET_NAMESPACE)
+        .Get(KEYCLOAK_CLIENTS_SECRET_NAME);
+      await updateKeycloakClientsSecret(secret, false);
+    } catch (err) {
+      log.error(
+        err,
+        "Failed to rotate Keycloak Clients Secret. It might be an temporary error. Skipping",
+      );
+    }
```
mjnagel: I think this works but I'm wondering why we wouldn't just create the secret with the secret key when we first apply it? As-is this logic is:
- Apply an "empty" secret
- Get the empty secret
- Apply an update to the empty secret

Seems like we probably could just call `updateKeycloakClientsSecret` passing in an in-memory empty secret without applying it to the cluster first? It's a small change but it reduces the calls we're making here and simplifies this func imo.

src/pepr/operator/controllers/keycloak/clients/keycloak-client.ts
```diff
@@ -0,0 +1,59 @@
...
+  CLIENT_CREDENTIALS = "client_credentials",
+  DYNAMIC_CLIENT_REGISTRATION = "dynamic_client_registration",
+}
+
+export async function getStrategy() {
+  const strategy = process.env.PEPR_KEYCLOAK_CLIENT_STRATEGY || ClientStrategy.AUTO;
+  switch (strategy) {
+    case ClientStrategy.CLIENT_CREDENTIALS:
+      log.debug("Using Client Credentials strategy");
+      return "client_credentials";
```
mjnagel:
```suggestion
      return ClientStrategy.CLIENT_CREDENTIALS;
```

src/pepr/operator/controllers/keycloak/clients/keycloak-client.ts
```diff
@@ -0,0 +1,59 @@
...
+      log.warn(
+        `Invalid ${process.env.PEPR_KEYCLOAK_CLIENT_STRATEGY} parameter value, falling back to dynamic registration`,
+      );
+      return ClientStrategy.DYNAMIC_CLIENT_REGISTRATION;
+  }
+}
+
+export async function createOrUpdateClient(client: Partial<Client>) {
+  const strategy = await getStrategy();
+  if (strategy === ClientStrategy.CLIENT_CREDENTIALS.toString()) {
```
mjnagel:
```suggestion
  if (strategy === ClientStrategy.CLIENT_CREDENTIALS) {
```

I might just be missing something here, but we should be able to use the enum type directly in check, assuming `getStrategy` strictly returns the enum value (maybe related to other comment).

src/pepr/operator/controllers/keycloak/clients/keycloak-client.ts
```diff
@@ -0,0 +1,59 @@
...
+    return credentialsCreateOrUpdate(client);
+  }
+  return dynamicCreateOrUpdate(client);
+}
+
+export async function deleteClient(client: Partial<Client>) {
+  const strategy = await getStrategy();
+  if (strategy === "client_credentials") {
```
mjnagel:
```suggestion
  if (strategy === ClientStrategy.CLIENT_CREDENTIALS) {
```

Similar comment here.

src/pepr/uds-operator-config/values.yaml
```diff
@@ -17,3 +17,5 @@ operator:
   PEPR_RELIST_INTERVAL_SECONDS: "600"
   # Configure Pepr reconcile strategy to have separate queues for faster reconciliation
   PEPR_RECONCILE_STRATEGY: "kindNsName"
+  # Keycloak Client Mode. Possible values: "dynamic_client_registration", "client_credentials" and "auto"
+  PEPR_KEYCLOAK_CLIENT_STRATEGY: "auto"
```
mjnagel, line 21: Wondering if this is a good reason to add a chart values schema - would ensure people pass in the correct values here sooner. We don't have to add the schema for everything right now, but might be good to do it for this enum (could reference the [keycloak one](https://github.com/defenseunicorns/uds-core/blob/93a44e6d67b36f92c204d043484d996e877194ac/src/keycloak/chart/values.schema.json) as an example here).

src/pepr/operator/controllers/keycloak/clients/client-credentials.ts
mjnagel: In reviewing this again I'm OK with this - I had missed how the cachedToken above was being used and that it was being returned before this code executed. Fine to proceed with this as-is.

src/pepr/operator/controllers/keycloak/clients/client-credentials.ts
```diff
@@ -0,0 +1,143 @@
...
+  const token = await credentialsGetAccessToken();
+  const url = `${clientsAdminUrl}?clientId=${encodeURIComponent(client.clientId!)}`;
+  const response = await fetch<ClientWithId[]>(url, {
+    method: "GET",
+    headers: { "Content-Type": "application/json", Authorization: `Bearer ${token}` },
+  });
+  await throwErrorIfNeeded(response, () => {
+    resetCachedToken();
+  });
+  return response.data[0];
```
mjnagel, line 116: Is this data field the client id, full client object, something else? Might be good to comment on this code to take note of what we are returning (mostly wondering because with the dynamic response we return the full data, I assume this is different but it's hard to tell what this response looks like).

src/pepr/operator/controllers/keycloak/clients/client-credentials.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+export async function credentialsUpdate(client: Partial<Client>) {
+  log.info(`credentialsUpdate: updating client ${JSON.stringify(client)}`);
+  const token = await credentialsGetAccessToken();
+  const existing = await credentialsGet(client);
+  const url = `${clientsAdminUrl}/${encodeURIComponent(existing.id!)}`;
```
mjnagel: Do we need validation that `existing.id` is non-null? The `!` operator makes me worried this could hide errors if the `existing` response is not what we expect? (same comment for deletion)

src/pepr/operator/controllers/keycloak/clients/dynamic-client-registration.spec.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright 2025 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
+import { addRequiredAttributesToClient } from "./dynamic-client-registration";
+
+describe("addReuiredAttributesToClient", () => {
```
mjnagel:
```suggestion
describe("addRequiredAttributesToClient", () => {
```

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -138,31 +131,22 @@ async function syncClient(
...
+    client = await createOrUpdateClient(client);
   } catch (err) {
     const msg =
       `Failed to process Keycloak request for client '${client.clientId}', package ` +
       `${pkg.metadata?.namespace}/${pkg.metadata?.name}. Error: ${err.message}`;
 
     // Throw the error if this is the retry or was an initial client creation attempt
-    if (isRetry || !token) {
+    // if (isRetry || !token) {
+    if (isRetry) {
```
mjnagel: I think this change in logic did lose one aspect of this that we actually want to maintain - not using the token on a retry. This was intentionally added to handle cases where the PeprStore has a "stale token" after a client gets deleted. In these cases we want to make sure the retry tries to create the client rather than update. Without that logic we could get stuck (if using dynamic registration) when the PeprStore has an old token.

That's actually the main reason we had this retry logic here to my knowledge - which is why the second check here (`if (isRetry || !token)`) would not call the retry if we didn't have a token (i.e. we should only retry if we're retrying because we might have a stale token). I'm not sure if there's a different targeted retry approach we would want for the new client-credentials approach (probably just always retry because we may have an expired jwt?) - but for dynamic-registration I think we probably want to maintain this same approach. Regardless this may mean passing `isRetry` down to `createOrUpdateClient` unless there's a different way to handle it at this level.

slaskawi: I see! Thanks for explaining. I modified the logic to call the `dynamicCreate` on a retry code path.

As for the Client Credentials - yes, you're right. We don't anything else as we first obtain a Client from Keycloak and then decide what to do. In other words, the decision is made based on the Keycloak state - not Pepr Store state.

src/pepr/operator/controllers/keycloak/clients/client-credentials.ts
slaskawi: I don't think so. From the Keycloak side, this needs to be filled in as this is the database id (the primary key, which cannot be null in the database). If we ever see a null there, this needs to be a broader system failure (like Node.js misbehaving or some Linux sockets errors). 

I think we should be safe just to leave it as is.

src/pepr/operator/controllers/keycloak/clients/client-credentials.ts
slaskawi: This is one of the quirks of the Keycloak Admin APIs.

There's no Client GET REST endpoint that obtains a client based on `client_id` (the logical client name, like `uds-operator`). All Admin REST endpoints for client operator on the database Client ID, which is a UUID. The only interface that allows to obtain the Client using the `client_id` is the collection interface which returns you the client in a singular collection. I know - it's a terrible design but we'll be addressing this in v2 Admin API initiative. 

I'll add a comment explaining this.

src/pepr/uds-operator-config/values.yaml
slaskawi: I thought about this and I'm not convinced that the right thing to do. The ultimate goal is to get rid of the Dynamic Client registration and this switch as a consequence. We just want to keep it around for some time to ensure it works in all the cases.

src/pepr/operator/controllers/keycloak/config.ts
slaskawi: The reason I separated these two is the error handling. We allow the operator to continue working if adding the `uds-operator` key fails (we'll just retry) but we crash it when the Secret itself can't be created. 

But you're right - simplicity matters and I'll refactor the code to your guideline.

docs/.images/diagrams/uds-core-pepr-operator-flow.drawio
slaskawi: Comments addressed with one small caveat. Instead of using Pepr configuration variables, I just wrote "UDS Operator Configuration". Otherwise, we'd end up with a big decision block trying to accommodate `PEPR_KEYCLOAK_CLIENT_STRATEGY`, which is pretty long.

If you don't like, please let me know and I'll update it again.

mjnagel: This is perfect!

src/pepr/operator/controllers/keycloak/config.ts
```diff
@@ -0,0 +1,51 @@
...
+            namespace: KEYCLOAK_CLIENTS_SECRET_NAMESPACE,
+            name: KEYCLOAK_CLIENTS_SECRET_NAME,
+          },
+          type: "Opaque",
+        };
+        await updateKeycloakClientsSecret(secret, false);
+      } catch (err) {
+        log.error(
+          err,
+          "Failed to rotate Keycloak Clients Secret. It might be an temporary error. Skipping",
```
mjnagel:
```suggestion
          "Failed to create Keycloak Clients Secret",
```

src/pepr/uds-operator-config/values.yaml
mjnagel: Okay to skip for now in this case - I suspect most won't even override this.

src/pepr/operator/controllers/keycloak/clients/client-credentials.ts
mjnagel: Thanks this is perfect!

src/pepr/operator/controllers/keycloak/clients/client-credentials.ts
mjnagel: My initial thought was just safe guarding against a bad API response, but might be unnecessary since we're catching !ok responses in the other functions. Okay to leave this as-is for now if you deem it safe - I just tend to prefer the code guaranteeing it rather than relying on an API call to return what we expect. I'm not sure what the result would be here if we called just `${clientsAdminUrl}/` without the ID - if that would return an error and not have unexpected changes then should be fine to allow other error handling to handle this edge case.

My thought was a very lightweight check that guarantees in the typescript code that id is always populated:
```suggestion
  if (!existing || !existing.id) {
    throw new Error(`Failed to retrieve existing client, ${client.clientId}`);
  }
  const url = `${clientsAdminUrl}/${encodeURIComponent(existing.id)}`;
```

mjnagel (commented):
LGTM at this point - validated the code with the current auto mode (using dynamic) as well as with the identity-config changes and all works as expected. Feel free to address or resolve these final comments and I will approve.

src/pepr/operator/controllers/keycloak/clients/client-credentials.ts
```diff
@@ -0,0 +1,150 @@
...
+    resetCachedToken();
+  });
+  return credentialsGet(client);
+}
+
+export async function credentialsDelete(client: Partial<Client>) {
+  log.info(`credentialsDelete: deleting client ${JSON.stringify(client)}`);
+  const token = await credentialsGetAccessToken();
+  const existing = await credentialsGet(client);
+  const url = `${clientsAdminUrl}/${encodeURIComponent(existing.id!)}`;
```
mjnagel:
```suggestion
  if (!existing || !existing.id) {
    throw new Error(`Failed to retrieve existing client, ${client.clientId}`);
  }
  const url = `${clientsAdminUrl}/${encodeURIComponent(existing.id)}`;
```

Same change as for the update func above.

mjnagel (approved):

chance-coleman (approved):

PR #1343, opened 2025-03-06
feat: root domain templating

### LINKED ISSUE #1301: feat: Append DOMAIN to gateways
When deploying uds-core, gateways are created per ingress. There are different hosts added based on uds-core requirements (keycloak, sso etc). Something that has come up recently with a mission hero is that their UI needs to sit on the root domain (i.e. foo.example.com) which is not being set at the Gateway currently.

- **Given** the domain is set
- **When** gateway appends base domain to hosts array
- **Then** a user is able navigate to the base domain

N/A

### Additional context

Not sure if it's as easy as this to include it, but more so what should be done for SSL certificates if this was to occur.

```yaml
{{- $domain := tpl .Values.domain . }}
{{- if .Values.tls }}
apiVersion: networking.istio.io/v1beta1
kind: Gateway
metadata:
  name: {{ .Values.name }}-gateway
spec:
  selector:
    app: {{ .Values.name }}-ingressgateway
  servers:
    {{ range $name,$server := .Values.tls.servers }}
    - hosts:
        - "{{ $domain }}" # <--- add here
        {{- range $server.hosts | default (list "*") }}
        - "{{ . }}.{{ $domain }}"
        {{- end }}
      port:
        name: "http-{{ $name }}"
        number: 80
        protocol: HTTP
      tls:
        httpsRedirect: true
    - hosts:
        - "{{ $domain }}" # <--- add here
        {{- range $server.hosts | default (list "*") }}
        - "{{ . }}.{{ $domain }}"
        {{- end }}
      port:
        name: "https-{{ $name }}"
        number: 443
        protocol: HTTPS
      tls:
        mode: {{ $server.mode }}
        {{- if ne $server.mode "PASSTHROUGH" }}
        credentialName: {{ $.Values.tls.credentialName | default "gateway-tls" | quote }}
        # if supportTLSV1_2 is both defined and true, use TLSV1_2, otherwise use TLSV1_3
        minProtocolVersion: {{ if $.Values.tls.supportTLSV1_2 }}TLSV1_2{{ else }}TLSV1_3{{ end }}
        {{- end }}
    {{ end }}
{{ end }}
```

chance-coleman:
## Description
New feature to add root domain configuration to the istio package. This allows for applications to sit on the root domain and be configured with a virtual service to be accessed. Templating includes options for TLS mode, TLSversion, cert, key, and cacert. Also a credential override if not using the templated tls secret.

## Related Issue

Fixes #1301

## Type of change

## Steps to Validate
This is a bit complicated to test but here are some steps:
- Configure the values.yaml (certs for `uds.dev` and `uds.admin.dev`):
```yaml
enableRootDomain: true
rootDomainTLS:
  mode: SIMPLE                        # e.g., SIMPLE, MUTUAL, etc. default SIMPLE
  credentialName: ""                  # If set to a non-empty value, the chart will assume this secret already exists and will not auto-generate it using the provided cert data.
  supportTLSV1_2: true                # Set to false to enforce TLSV1_3 only
  cert: "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUM5akNDQWQ2Z0F3SUJBZ0lVS2xZTlg5TG9CcSt4QytOWEF4eHg0R3VDZ0RZd0RRWUpLb1pJaHZjTkFRRUwKQlFBd0VqRVFNQTRHQTFVRUF3d0hkV1J6TG1SbGRqQWVGdzB5TlRBek1EWXhORFUwTlRsYUZ3MHlOakF6TURZeApORFUwTlRsYU1CSXhFREFPQmdOVkJBTU1CM1ZrY3k1a1pYWXdnZ0VpTUEwR0NTcUdTSWIzRFFFQkFRVUFBNElCCkR3QXdnZ0VLQW9JQkFRQ3hMTHJ2ZWFHNDQ0NDFsbk8xaEVTSXUvM2UvOTRFVkxMZE5NSnpTNHJxOHl1Qi8wZGUKeVdaRFpISUJpb0RnbGFhaENUUUtUNFlFWTlkenBmNVFiWW9PUkZON1BoWEJUSTBacjlNUjRBQk9SeEpLbUREawpxelBWOGZPZW41a0dKaEFtWmllS0NuWk44a0FUZ29zQVVNUlh0Mk9laXE3Rmh0VUFQU0JaaGRueXJSOWZobGtsCnl0dGRxYXAxcFJZNmtzbStuWjVsbTdxc2RqMXdoQnVvN1lZcmtoYWlqUEF6WWFISkJCZ0VjWnVyMVN0a05BcjEKU2FmM1V5SmwxSkVMTG1ta2hiV0FBSDI2eVZKcGsyZTRkUVRrM3U3eWw3c1A2ZkhPSUoyeTdDbWZTM2xzV3hWMApBK0VUVC9DUVpBVHVXMHVYenowWFVjaDduL3EybnlyVHArREpBZ01CQUFHalJEQkNNQ0VHQTFVZEVRUWFNQmlDCkIzVmtjeTVrWlhhQ0RXRmtiV2x1TG5Wa2N5NWtaWFl3SFFZRFZSME9CQllFRk03dGdpdlovNC9LNzlwTzBUcnYKd2hvZUJaUm5NQTBHQ1NxR1NJYjNEUUVCQ3dVQUE0SUJBUUFYd1p5MkdTaksreVJ2SGpBd0J5UlRUUW9hZ1RKeQpIUG1zK2kxY0xRMEJJaWE1b2hsMmE2U1d2eHpVdzcxczNCK0kyaHo5aTQ3TFlaQ1hCYWpsUzZLeEZ6bXRneXNaCitBa0VWOWFNZk9kb1dLMVA3UHVEQWZtYXFSY21pQU1SYWd2NDZCRnYyL2ozNU9FUTZUQTNGZGNIYitrN0tQWjMKUmJwcFVTWk84RHJpN0dLMnJXeVZBMHBLK0FUdzBtYWJzS1FNYVVmNFZ1cEhxNWRON21VS2U1S2cxZmluS2ltZAppZEhBYWxuREw4c3c0UTYzbTBpWTd0T1h1eGIvRlFmM0c4Smx5RXF0SUZyTWFndkZ6UzhoVExxTlRpRVdSbTJpClkvYitZMGZyazRyQTR2RUowWGlMVEpnSzJlOTVCZU1IMWJlUWV1ZG9DY3JyRmNuYnpMTmJBalpSCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K"
  key: "LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCk1JSUV2UUlCQURBTkJna3Foa2lHOXcwQkFRRUZBQVNDQktjd2dnU2pBZ0VBQW9JQkFRQ3hMTHJ2ZWFHNDQ0NDEKbG5PMWhFU0l1LzNlLzk0RVZMTGROTUp6UzRycTh5dUIvMGRleVdaRFpISUJpb0RnbGFhaENUUUtUNFlFWTlkegpwZjVRYllvT1JGTjdQaFhCVEkwWnI5TVI0QUJPUnhKS21ERGtxelBWOGZPZW41a0dKaEFtWmllS0NuWk44a0FUCmdvc0FVTVJYdDJPZWlxN0ZodFVBUFNCWmhkbnlyUjlmaGxrbHl0dGRxYXAxcFJZNmtzbStuWjVsbTdxc2RqMXcKaEJ1bzdZWXJraGFpalBBellhSEpCQmdFY1p1cjFTdGtOQXIxU2FmM1V5SmwxSkVMTG1ta2hiV0FBSDI2eVZKcAprMmU0ZFFUazN1N3lsN3NQNmZIT0lKMnk3Q21mUzNsc1d4VjBBK0VUVC9DUVpBVHVXMHVYenowWFVjaDduL3EyCm55clRwK0RKQWdNQkFBRUNnZ0VBTlk5MEk1ajlqc3NiM285UkEzcXN6VGtua2haL280ZUhXdC9zT0xhSmxHVlcKcmNIaWJZRXM3UXFjZkdMR2V4NUNkVVEyK3piM0tMU1dHVndBK1lkanlOUlcrRmJZZE1zVGpNUmVablQxSXJUUgpqc05iWklWczhpeG1uWGxaNVVYMGMrZEY1TEVzK250VmE1QjNQTzB0cmVhK3N0cng0cnpULzNKN0tSVVJ3ZzZhCnM5c1R2T3dNU2RGUkxvWHU5bUE1RXlrWndTcjFuaVNkblp4OUlvNlJ1OFdRNEY5eXBuVVplQUd3VDBlWWd5QjcKcnczNmlzR3kzZC92SlJuNnZIcklDMk5rc2dqQURJYkpLUzdNNG5yc0s0WkRqYm5SZHlOVnBYYkc2UFdyMTJsTQp0bC9ZNndKeXhmaG5kSURYWERlYkVJL3Nxa2hjTTlqYitoaXV3aDd6MFFLQmdRRHBWbzMxZThRTDhvbGNBZ1loCnUvV0xDWlh3cjVFc2xXWkduWXhlSllYQjMwNE82T0M0SnlPTXd3dHJDOUlhMk9kK1IrMTc2VC8xb3Z4RFoxNHAKaWZNdFJnWEpQNTlQQXhVMGNpQUhoQUsrbWNUUXk1Q0JETXNUbmlXdUZIUDBab3g1WHpYRXN3U0JUck9DWExyTQpPSGVzZ25DVElCWlpYUFJ3WGlWWlozdE5id0tCZ1FEQ1ljdVZPeHlONkl3SVhpR0x4bXFabEM1dVhkc2ZoUkthCmJzWDdwdXJXbmJ0UUQzZVFUWGtvOWJPRUV6NFQvWFA0cmw2NkpMMllGQnY4MnVOY05KMEV4MkNtM3VHbGh1QWgKc0Rwa0t3ckQ2UGlJclZlT2p0RXNpR2pEZTVMNGgwczQ2UFZjTVBVUXlXQjAyNGhuSEZnRytFaWhRUWE3N0J1TAp3UzdQazh1SlJ3S0JnRFp2VzlUT0Y5RlZ0cGZCWFI3WGs1UHBHNUszMHAxZENTd21LdzMzb1BtMmw2WkF5OVFLClJXL2NQTGl2WEVlcEhIQklaVzNIMzRUMWpmWkhraDhNc0s5NkszVmtvMHl5Z21ybXlQUVg1dkFDUFhrY1ZFelcKNkRWZWFwbnU2WkkxcmtYT1lXUFBBNWtLL0RQSVlFZXVVTFR0QlVnOWJ4bFA1ektqdEFEa2RFNS9Bb0dBZktTSgpEdUFncHo1K3pMN1BjL2wvVVl1Yldsb0VzR2w4VEZyTHlWcDNLN2RvN1NWOTNhSlFGdHUyaVBKdDFCT3AwY1MrCktVNTE1SmJBZTV0QVAxZmRkN3NCVnYrRVljZmk0TGRJT2dML25EQS9iaVQ1Q0FpOFNIb1A4NG5CN0d1VTZLRTQKOUN6UWVEc3BCc1hlNlg5YlV0elNkZFJrcFF2NWZkd2FVREROU3ZFQ2dZRUEzT0hhV2k3eDF0c1ZLSWpjMEtGZwpxbmtjOTZBQlQ5dTR3V0t0bG9ZNXZFNDNWN2M1ZGM4QnZ0UGp1QUNucjZ2Q253YzZ3MjlhT3RzVUJPdVJNZ2QrCmlFRlpGcTN4bjVGK2I5MURLUzZDN2FLTjcvRzY3OUl6OFpZaGdVZGhuL1I4SjRPNkdXbGJEL2gvQThGQUJDOHIKZzlMTW9KUUxlc1VNQXVma1I5V25udGc9Ci0tLS0tRU5EIFBSSVZBVEUgS0VZLS0tLS0K"
  cacert: "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUM5akNDQWQ2Z0F3SUJBZ0lVS2xZTlg5TG9CcSt4QytOWEF4eHg0R3VDZ0RZd0RRWUpLb1pJaHZjTkFRRUwKQlFBd0VqRVFNQTRHQTFVRUF3d0hkV1J6TG1SbGRqQWVGdzB5TlRBek1EWXhORFUwTlRsYUZ3MHlOakF6TURZeApORFUwTlRsYU1CSXhFREFPQmdOVkJBTU1CM1ZrY3k1a1pYWXdnZ0VpTUEwR0NTcUdTSWIzRFFFQkFRVUFBNElCCkR3QXdnZ0VLQW9JQkFRQ3hMTHJ2ZWFHNDQ0NDFsbk8xaEVTSXUvM2UvOTRFVkxMZE5NSnpTNHJxOHl1Qi8wZGUKeVdaRFpISUJpb0RnbGFhaENUUUtUNFlFWTlkenBmNVFiWW9PUkZON1BoWEJUSTBacjlNUjRBQk9SeEpLbUREawpxelBWOGZPZW41a0dKaEFtWmllS0NuWk44a0FUZ29zQVVNUlh0Mk9laXE3Rmh0VUFQU0JaaGRueXJSOWZobGtsCnl0dGRxYXAxcFJZNmtzbStuWjVsbTdxc2RqMXdoQnVvN1lZcmtoYWlqUEF6WWFISkJCZ0VjWnVyMVN0a05BcjEKU2FmM1V5SmwxSkVMTG1ta2hiV0FBSDI2eVZKcGsyZTRkUVRrM3U3eWw3c1A2ZkhPSUoyeTdDbWZTM2xzV3hWMApBK0VUVC9DUVpBVHVXMHVYenowWFVjaDduL3EybnlyVHArREpBZ01CQUFHalJEQkNNQ0VHQTFVZEVRUWFNQmlDCkIzVmtjeTVrWlhhQ0RXRmtiV2x1TG5Wa2N5NWtaWFl3SFFZRFZSME9CQllFRk03dGdpdlovNC9LNzlwTzBUcnYKd2hvZUJaUm5NQTBHQ1NxR1NJYjNEUUVCQ3dVQUE0SUJBUUFYd1p5MkdTaksreVJ2SGpBd0J5UlRUUW9hZ1RKeQpIUG1zK2kxY0xRMEJJaWE1b2hsMmE2U1d2eHpVdzcxczNCK0kyaHo5aTQ3TFlaQ1hCYWpsUzZLeEZ6bXRneXNaCitBa0VWOWFNZk9kb1dLMVA3UHVEQWZtYXFSY21pQU1SYWd2NDZCRnYyL2ozNU9FUTZUQTNGZGNIYitrN0tQWjMKUmJwcFVTWk84RHJpN0dLMnJXeVZBMHBLK0FUdzBtYWJzS1FNYVVmNFZ1cEhxNWRON21VS2U1S2cxZmluS2ltZAppZEhBYWxuREw4c3c0UTYzbTBpWTd0T1h1eGIvRlFmM0c4Smx5RXF0SUZyTWFndkZ6UzhoVExxTlRpRVdSbTJpClkvYitZMGZyazRyQTR2RUowWGlMVEpnSzJlOTVCZU1IMWJlUWV1ZG9DY3JyRmNuYnpMTmJBalpSCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0K"

```
- modify /etc/hosts file to include the ingress-gateway external ip `<external-ip> uds.dev`
- deploy slim-dev `uds run slim-dev`
- add virtual service to app-tenant manifest
```yaml
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: test-tenant-app
  namespace: test-tenant-app
spec:
  hosts:
    - uds.dev
  gateways:
    - istio-tenant-gateway/tenant-gateway
  http:
    - match:
        - uri:
            prefix: /port8080
      name: route-8080
      rewrite:
        uri: /
      route:
        - destination:
            host: test-tenant-app
            port:
              number: 8080
    - match:
        - uri:
            prefix: /port8081
      name: route-8081
      rewrite:
        uri: /
      route:
        - destination:
            host: test-tenant-app
            port:
              number: 8081
```
- build and deploy test apps `uds zarf package create src/test --confirm --no-progress --skip-sbom && uds zarf package deploy build/zarf-package-uds-core-test-apps-*.zst --confirm --no-progress`
- curl the uds.dev url: `curl -vk https://uds.dev/port8080` and should receive a `Hello from port 8080`

## Checklist before merging


docs/reference/UDS Core/dns.md
```diff
@@ -6,13 +6,17 @@ UDS Core deploys two Gateways by default - a Tenant Gateway for end-user applica
 
 ### Domain Configuration
 
-Each Gateway requires a wildcard DNS entry corresponding with the chosen `DOMAIN` and `ADMIN_DOMAIN` [variables](https://github.com/defenseunicorns/uds-core/blob/f6b0b59060a14febd11b0cdc7480f853a57f8520/src/istio/zarf.yaml#L10-L16) (or `admin.<DOMAIN>` if not specifying a separate admin domain). When deploying UDS Core, you can expect two Gateways to be created that match the following domain names:
+Each Gateway requires a wildcard DNS entry corresponding with the chosen `DOMAIN` and `ADMIN_DOMAIN` [variables](https://github.com/defenseunicorns/uds-core/blob/f6b0b59060a14febd11b0cdc7480f853a57f8520/src/istio/zarf.yaml#L10-L16) (or `admin.<DOMAIN>` if not specifying a separate admin domain).When deploying UDS Core, you can expect two Gateways to be created that match the following domain names:
```
mjnagel:
```suggestion
Each Gateway requires a wildcard DNS entry corresponding with the chosen `DOMAIN` and `ADMIN_DOMAIN` [variables](https://github.com/defenseunicorns/uds-core/blob/f6b0b59060a14febd11b0cdc7480f853a57f8520/src/istio/zarf.yaml#L10-L16) (or `admin.<DOMAIN>` if not specifying a separate admin domain). When deploying UDS Core, you can expect two Gateways to be created that match the following domain names:
```

docs/reference/configuration/ingress.md
```diff
@@ -132,3 +132,29 @@ packages:
...
+### Apex (Root) Domain Configuration
+By default, the UDS Core Gateways are configured with wildcard hosts (for example, `*.uds.dev`), which match only subdomains (such as `demo.uds.dev` or `keycloak.admin.uds.dev`). The apex domain (i.e. `uds.dev`) is not covered by a wildcard. This is important if you need an application to be accessible at the root of your domain.
+
+To support this use case, UDS Core provides an optional configuration to enable a dedicated server block for the apex domain. When enabled, two additional server blocks are added to your Istio Gateway:
+- HTTP on port 80: Redirects traffic to HTTPS.
+- HTTPS on port 443: Terminates TLS using settings from the rootDomainTLS section.
+
+For example, if you want your application to be reachable at `https://uds.dev`, you would enable the apex domain by setting the appropriate flags in your values file:
+```yaml
+# Enable apex (root) domain configuration. When true, the Gateway creates dedicated server blocks
```
mjnagel: Can we restructure this example as a bundle override?

docs/reference/configuration/ingress.md
```diff
@@ -132,3 +132,29 @@ packages:
...
+rootDomainTLS:
+  mode: SIMPLE                        # TLS mode (e.g., SIMPLE, MUTUAL). Default is SIMPLE.
+  credentialName: ""                  # Specify a TLS secret name if pre-created. Set to "" to auto-create using the cert data.
+  supportTLSV1_2: true                # Set to true to support TLS 1.2, or false to enforce TLS 1.3 only.
+  cert: "BASE64_ENCODED_CERTIFICATE"  # Base64-encoded certificate data. For self-signed certs, cert and cacert are typically the same.
+  key: "BASE64_ENCODED_PRIVATE_KEY"   # Base64-encoded private key.
+  cacert: "BASE64_ENCODED_CERTIFICATE"  # Base64-encoded CA certificate (use the same as cert for self-signed).
+```
+:::note
+- If you provide a non-empty value for credentialName, UDS Core assumes that you have pre-created the Kubernetes secret and will not auto-generate it using the certificate data.
```
mjnagel, line 169: Is it worth mentioning here that someone could provide the other secret name (`gateway-tls`) if they are using a [SAN cert](https://www.digicert.com/faq/public-trust-and-certificates/what-is-a-multi-domain-san-certificate)?

src/istio/chart/templates/gateway.yaml
```diff
@@ -39,4 +39,27 @@ spec:
...
+        name: "https-root-domain"
+        number: 443
+        protocol: HTTPS
+      tls:
+        mode: {{ .Values.rootDomainTLS.mode | default "SIMPLE" }}
+        {{- if ne (.Values.rootDomainTLS.mode | default "SIMPLE") "PASSTHROUGH" }}
+        credentialName: {{ .Values.rootDomainTLS.credentialName | default "root-domain-tls" | quote }}
+        # if supportTLSV1_2 is both defined and true, use TLSV1_2, otherwise use TLSV1_3
+        minProtocolVersion: {{ if .Values.rootDomainTLS.supportTLSV1_2 }}TLSV1_2{{ else }}TLSV1_3{{ end }}
+        {{- end }}
```
mjnagel: I don't have a strong feeling here but I wonder if we should fallback to the `.Values.tls` values for most of these if not set (and then fallback to hardcoded values if neither are set), at least for mode/supportTLSV1_2 (not sure it makes as much sense for credential)?

docs/reference/configuration/ingress.md
```diff
@@ -132,3 +132,29 @@ packages:
             - path: tls.credentialName
               value: tenant-gateway-tls-secret # Reference to the Kubernetes secret for the tenant gateway's TLS certificate
 ```
+
+### Apex (Root) Domain Configuration
```
mjnagel: It might also be good in this section to add a note/example of a virtualservice to expose a service on the root domain?

docs/reference/configuration/ingress.md
chance-coleman: yeah good call out, will modify

docs/reference/configuration/ingress.md
chance-coleman: yeah good call out, will add

docs/reference/configuration/ingress.md
chance-coleman: yeah good call out, will add

src/istio/chart/templates/gateway.yaml
chance-coleman: yeah i think that is valid and probably safer that way. i pulled the logic into a variable for easier reference.

src/istio/chart/templates/gateway.yaml
```diff
@@ -2,6 +2,9 @@
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 {{- $domain := tpl .Values.domain . }}
+{{- $rootTLS := .Values.rootDomainTLS | default dict -}}
+{{- $rootMode := $rootTLS.mode | default ($.Values.tls.mode | default "SIMPLE") -}}
+{{- $rootSupportTLS := $rootTLS.supportTLSV1_2 | default ($.Values.tls.supportTLSV1_2 | default true) -}}
```
mjnagel: I think our default is typically 1.3 only, so `false` here?

docs/reference/configuration/ingress.md
```diff
@@ -132,3 +132,68 @@ packages:
...
+            - path: rootDomainTLS.credentialName
+              value: ""  # Leave blank to auto-create the secret using the provided cert data.
+            - path: rootDomainTLS.supportTLSV1_2
+              value: true
+            - path: rootDomainTLS.cert
+              value: "BASE64_ENCODED_CERTIFICATE"
+            - path: rootDomainTLS.key
+              value: "BASE64_ENCODED_PRIVATE_KEY"
+            - path: rootDomainTLS.cacert
+              value: "BASE64_ENCODED_CERTIFICATE"
```
mjnagel: nit: these would/should normally be variables rather than values (see other example above for the wildcard)

mjnagel (approved):

PR #1357, opened 2025-03-13
fix: renovate not checking test directory versions

chance-coleman:
## Description
Noticed that KFC was way out of date in out jest package.json. This was because the renovate only checked the top level of the test directory and not the sub directories.

## Type of change

## Checklist before merging


noahpb (approved):
Great catch!!

PR #1370, opened 2025-03-19
feat: new webauth and mfa flows

chance-coleman:
## Description
Related PR to the [uds-identity-config PR](https://github.com/defenseunicorns/uds-identity-config/pull/380). This allows for new templated values to be passed in to the identity config for configuring webauthn MFA as well as MFA for everywhere ( x509 included).

## Related Issue

Related: [uds-identity-config PR](https://github.com/defenseunicorns/uds-identity-config/pull/380)
Related: [Issue](https://github.com/defenseunicorns/uds-identity-config/issues/236)

## Type of change

## Steps to Validate
- not much to verify here other than deploying uds-core and making sure that new values are present in the keycloak realm secret. additional testing steps present in the identity-config PR.

## Checklist before merging


noahpb (approved):
LGTM 🚀

slaskawi (approved):

mjnagel (approved):
LGTM - I assume any documentation for these overrides will be included in either the identity-config updates or the follow on PR to pull the new identity-config version in.

src/keycloak/chart/values.yaml
```diff
@@ -51,6 +51,8 @@ realmAuthFlows:
   X509_AUTH_ENABLED: true
   SOCIAL_AUTH_ENABLED: true
   OTP_ENABLED: true
+  WEBAUTHN_ENABLED: false
+  X509_MFA_ENABLED: false
```
slaskawi, line 56: Do we want to put `WEBAUTHN_FLOW_ENABLED` here as well?

chance-coleman: Yeah good callout, the `WEBAUTHN_FLOW_ENABLED` is controlled by the `WEBAUTHN_ENABLED` key. The only reason its necessary is because it uses a different value `REQUIRED or DISABLED` vs `true of false`. So in this case we don't supply that.

```WEBAUTHN_FLOW_ENABLED: {{ ternary "REQUIRED" "DISABLED" (.Values.realmAuthFlows.WEBAUTHN_ENABLED) | b64enc }}```

PR #1372, opened 2025-03-21
feat!: only allow creation of one `UDSPackage` per namespace

### LINKED ISSUE #1352: Restrict usage of multiple `Package` CRs in a single namespace
This will be a breaking change in UDS Core.

Currently we have some support for creating multiple `Package` CRs in a single namespace, however this can cause confusion and unexpected results especially in relation to networking and service mesh configuration:
- Network Policies can be applied across the entire namespace, allowing workloads more ingress/egress than expected
- Service mesh configuration (mTLS posture, ambient mode) is typically applied across the entire namespace, which can result in unexpected configuration

To resolve this we should only allow a single `Package` CR per namespace. This could be done in either of two ways:
- Validator logic to deny packages if another package already exists: This would be the most direct approach but require a `K8s.Get` call in the validator
- Reconcile logic to fail reconciliation and create an Error status/event if other packages exist

We should also document the rationale for this and encourage best practices around separating applications into different namespaces.

noahpb:
## Description
This introduces a breaking change to the UDS Operator. 

As part of our ongoing work to introduce ambient mesh support to UDS Core, we have made the decision to restrict consumers to creating one UDS Package custom resource in any given namespace. This PR introduces this change to the operator.

Notable changes:
- Adds a new Pepr watch for `UDSPackage`
- Adds a `PackageStore` intended to serve as an in-memory map of all namespaces that have packages
- Adds a check to package validator that will block the creation of a UDSPackage if there is a preexisting UDSPackage in the target namespace

This change has limited backwards compatibility. Users with namespaces that have one or more Package CR may be prevented from making additional modifications to their existing Package CRs. Delete actions against these Package CRs will never be blocked, but updates/patches can be. Users are encouraged to elect a single Package CR to remain in the namespace and migrate configurations accordingly. 

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1352

## Type of change

## Steps to Validate
```bash
# check out this branch and then run:
$ uds run dev-setup

# deploy the operator
$ npx pepr deploy

# create a test package in a valid namespace:
$ kubectl create ns test
$ kubectl apply -f - <<EOF
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: foo
  labels: {}
  namespace: test
spec: {}
EOF
package.uds.dev/foo created

# create a new package in the same namespace:
$ kubectl apply -f - <<EOF
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: foo-1
  labels: {}
  namespace: test
spec: {}
EOF
Error from server: error when creating "STDIN": admission webhook "pepr-uds-core.pepr.dev" denied the request: A package with the name "foo" already exists in the namespace "test". Only one package can exist in a namespace.

# apply an update to the existing package
$ kubectl apply -f - <<EOF
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: foo
  namespace: test
spec: {}
EOF
package.uds.dev/foo configured
```
## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: only allow creation of one `UDSPackage` per namespace (#1372)

BREAKING CHANGE: `Package` CR validation will now prevent creating multiple `Package` CRs in the same namespace. Ensure that you only have a single `Package` CR per namespace before this upgrade, otherwise you may be unable to update them going forward. 
END_COMMIT_OVERRIDE


docs/reference/configuration/UDS operator/package.md
```diff
@@ -5,7 +5,12 @@ title: UDS Package
 ![UDS Operator Package Flowchart](https://github.com/defenseunicorns/uds-core/blob/main/docs/.images/diagrams/uds-core-operator-uds-package.svg?raw=true)
 
 ## Package
+:::note
+Only one UDS Package Custom Resource can exist in a namespace. This pattern was chosen by design to better enable workload isolation and to reduce complexity.
+:::
+Namespaces are common boundary for isolating workloads in multi-tenant clusters. When defining a UDS Package resource, consider the implications for all workloads that exist in the target namespace. If the UDS Package that you are defining has configurations that conflict each other or would be simplified by using a separate UDS Package definition, consider using a separate Kubernetes namespace. Read more about namespaces and mulitenancy [here](https://kubernetes.io/docs/concepts/security/multi-tenancy/).
```
chance-coleman:
```suggestion
Namespaces are a common boundary for isolating workloads in multi-tenant clusters. When defining a UDS Package resource, consider the implications for all workloads that exist in the target namespace. If the UDS Package that you are defining has configurations that conflict with each other or would be simplified by using a separate UDS Package definition, consider using a separate Kubernetes namespace. Read more about namespaces and mulitenancy [here](https://kubernetes.io/docs/concepts/security/multi-tenancy/).
```

src/pepr/operator/controllers/packages/package-store.spec.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * Copyright 2025 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
+//should add packages
```
chance-coleman: nit: these comments seem unnecessary. also line 14 below, the commented out packageNamespaceMap can be cleaned up.

noahpb: Ah, that was my pseudo code! Thanks for catching this.

src/pepr/operator/controllers/packages/package-store.ts
```diff
@@ -0,0 +1,138 @@
...
+ */
+function remove(pkg: UDSPackage, logger: boolean = true): void {
+  const namespace = pkg.metadata?.namespace || "";
+  const pkgToRemove = pkg.metadata?.name;
+  const items = packageNamespaceMap.get(namespace) || [];
+
+  if (items.length > 1) {
+    items.forEach((value, index) => {
+      if (value.metadata?.name === pkgToRemove) {
+        items.splice(index, 1);
```
chance-coleman: modifying the array returned from Map.get() directly with splice, which can lead to bugs if the same array reference is shared elsewhere. At the moment this probably isnt an issue, but for the sake of future proofing and maintainability. I think something like this would also work and avoid the direct manipulation:
```
const updatedItems = items.filter(pkg => pkg.metadata?.name !== pkgToRemove);
if (updatedItems.length > 0) {
  packageNamespaceMap.set(namespace, updatedItems);
} else {
  packageNamespaceMap.delete(namespace);
}
```

src/pepr/operator/controllers/packages/package-store.ts
```diff
@@ -0,0 +1,138 @@
...
+ * the package to be added is appended to the existing list of Packages. If not, a new key is created
+ * in packageNamespaceMap for that namespace, containing a list of one element.
+ */
+function add(pkg: UDSPackage, logger: boolean = true): void {
+  const namespace = pkg.metadata?.namespace || "";
+  // Get existing packages if any and merge into the one being added
+  const existingValue = packageNamespaceMap.get(namespace);
+  if (existingValue) {
+    existingValue.push(pkg);
+    packageNamespaceMap.set(namespace, existingValue);
```
chance-coleman: i think this is unnecessary because the existingValue used a get() its referencing the packageNamespaceMap already.

src/pepr/operator/controllers/packages/package-store.ts
```diff
@@ -0,0 +1,138 @@
...
+ * @param {boolean} [logger=true] - Optional flag to enable logging. Defaults to true.
+ *
+ * This function retrieves the namespace from the package metadata and adds the package
+ * to the packageNamespaceMap. If the namespace is not present, it defaults to an empty string.
+ * For backwards compatibility, the function checks if the namespace has an existing package. If it does,
+ * the package to be added is appended to the existing list of Packages. If not, a new key is created
+ * in packageNamespaceMap for that namespace, containing a list of one element.
+ */
+function add(pkg: UDSPackage, logger: boolean = true): void {
+  const namespace = pkg.metadata?.namespace || "";
```
chance-coleman: are there any concerns with namespace defaulting to an empty string? should we throw some kind of error/warn here?

src/pepr/operator/controllers/packages/package-store.ts
```diff
@@ -0,0 +1,138 @@
...
+let packageNamespaceMap: PackageNamespaceMap;
+
+/**
+ * Initializes the package namespace map.
+ *
+ * This function creates a new `Map` object and assigns it to the `packageNamespaceMap` variable.
+ * The `packageNamespaceMap` is used to store packages, using their namespace as the key.
+ */
+function init(): void {
+  packageNamespaceMap = new Map();
```
chance-coleman, line 26: This can be a fragile initialization method. If we keep this function then we should probably add checks for making sure its initialized. Alternatively, the typical TS approach would be an eager init like this:
```
let packageNamespaceMap: PackageNamespaceMap = new Map();
```

src/pepr/operator/controllers/packages/package-store.spec.ts
chance-coleman, line 1: A couple additional tests that might be beneficial, don't feel super strongly here.
- duplicate package with same name and how its handled
- empty namespace
- getPkgName() Behavior because it grabs on the first package in the namespace

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -21,10 +22,27 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
   const pkgName = pkg.metadata?.name ?? "_unknown_";
   const ns = pkg.metadata?.namespace ?? "_unknown_";
 
+  // Always permit packages that are being removed
+  const deletionTimestamp = pkg.metadata?.deletionTimestamp ?? null;
+  if (deletionTimestamp) {
+    return req.Approve();
+  }
```
mjnagel: I think our reconciler code would skip this and ensure we don't "reconcile" a package that is being deleted, but should this be part of the specific `PackageStore` check instead of a standalone check? Again not sure there's a way to exploit this, but if a package has a deletion timestamp this would mean we skip _ALL_ of the validation logic.

src/pepr/operator/reconcilers/index.ts
```diff
@@ -177,3 +179,18 @@ export function getReadinessConditions(ready: boolean = true) {
...
+    const watcher = K8s(UDSPackage).Watch(async (pkg, phase) => {
+      log.debug(`Processing package ${pkg.metadata?.name}, watch phase: ${phase}`);
+
+      processPackages(pkg, phase);
+    });
+    // This will run until the process is terminated or the watch is aborted
+    log.debug("Starting package watch...");
+    await watcher.start();
+  }
+}
```
mjnagel: I'm not sure the best place for this function (maybe `controllers/packages`?) but it seems out of place in this file in my opinion.

src/pepr/operator/controllers/packages/package-store.ts
noahpb: Good point, I'll add a check here to make sure it's defined and throw an error if not.

src/pepr/operator/crd/validators/package-validator.ts
noahpb: Good point. I'll remove this if statement and add an additional clause to the `PackageStore` check below.

noahpb: Updated - LMK if that is what you were looking to see @mjnagel.

src/pepr/operator/reconcilers/index.ts
noahpb: Moved to `src/pepr/operator/controllers/packages/packages.ts`

src/pepr/operator/controllers/packages/package-store.ts
noahpb: Discussed offline - thanks for the pointers here!

docs/reference/configuration/UDS operator/package.md
```diff
@@ -5,7 +5,12 @@ title: UDS Package
 ![UDS Operator Package Flowchart](https://github.com/defenseunicorns/uds-core/blob/main/docs/.images/diagrams/uds-core-operator-uds-package.svg?raw=true)
```
mjnagel, line 5: On the diagram changes can we:
- "Center align" the 3 gray flows (validate, reconcile, finalize) - right now the validate starts "lower" than reconcile, but continues below it
- Straighten out the arrow going from validate -> reconcile (may be fixed by ^)
- _Maybe_ shrink the arrows in between each check to make the validator box shorter - this one isn't a big deal but the shorter we can keep this diagram the better it will display on the docs site

I think we also would want/need to update the [source drawio file](https://github.com/defenseunicorns/uds-core/blob/feat/single-pkg-namespace/docs/.images/diagrams/uds-core-pepr-operator-flow.drawio)?

src/pepr/operator/crd/validators/package-validator.ts
mjnagel: LGTM!

src/pepr/operator/controllers/packages/packages.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @returns {void} This function does not return a value; it performs actions based on the input.
+ */
+
+// configure subproject logger
+const log = setupLogger(Component.OPERATOR_RECONCILERS);
+export async function startPackageWatch() {
+  PackageStore.init();
+  // only run in admission controller or dev mode
+  if (process.env.PEPR_WATCH_MODE === "false" || process.env.PEPR_MODE === "dev") {
+    const watcher = K8s(UDSPackage).Watch(async (pkg, phase) => {
```
mjnagel, line 45: I don't know if it's strictly required, but we may want to add the same [default watch config](https://github.com/defenseunicorns/uds-core/blob/d29d1b5abfe484747631b53b7475af83fa7946d0/src/pepr/policies/index.ts#L30-L43) here as we have in exemptions. That will just make sure that we get any upstream pepr changes to these values, and run this watch the same as a Pepr `.Watch` would work.

src/pepr/operator/controllers/packages/package-store.ts
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * A collection of functions related to watching UDSPackages
+ * Manages an in-memory map of UDSPackage resources
+ * Used in Pepr Validating Webhook Pods when vetting UDS Package resources for admission
+ */
+import { Component, setupLogger } from "../../../logger";
+import { UDSPackage } from "../../crd";
+const log = setupLogger(Component.OPERATOR_PACKAGES);
+
+export type PackageNamespaceMap = Map<string, UDSPackage[]>;
```
mjnagel: Probably just missing something, but why is this a list of `UDSPackage` resources if we only allow/want to allow a single package? Couldn't it just be a single Package, or even just the name of the package?

noahpb: Good question - I set it up this way to address an edge case where a consumer may upgrade to this version of the UDS Operator with more than one UDS Package in a namespace. In this event, if the user deletes a single Package CR from the namespace, `PackageStore.remove(namespace)` would delete that namespace from `packageNamespaceMap`. The next time the user creates a Package CR in that namespace, it would be approved by the package validator as `PackageStore.hasKey(ns)` would return false, despite there already being an existing Package CR(s) in the namespace.

noahpb: Acknowledging that this case is likely not to be encountered, but it felt to me that it would be a good thing to get ahead of.

>  Couldn't it just be a single Package, or even just the name of the package?

Definitely! I can slim it down to a list of Package names for now. I figured this could be expanded on in the future if we wished to fetch more details about packages from the `PackageStore`.

I can also change this to just expect a single `UDSPackage` or Package Name if we decide not to handle this edge scenario.

mjnagel: I'm okay with none of these changes for now - I think what you said makes sense:
1. Store list of names only: Can definitely see where storing more information would be useful. I know we discussed SSO client and VS uniqueness checks being possible/better with this. Initially I was thinking it might be nice to store separate maps with each of those, but I haven't thought through that too much. Probably better to leave the extra information and pull it out later if we get/store that in a different way.
2. Single vs list: That's a great callout for the upgrade. If this weren't a list, then your initial "add" function would have to pick one to set in the map (probably race condition) and might lead to some weird behavior on removal.

Going to resolve this comment - thanks for the explanation!

src/pepr/operator/controllers/packages/packages.ts
```diff
@@ -0,0 +1,51 @@
...
+    // This will run until the process is terminated or the watch is aborted
+    log.debug("Starting package watch...");
+    await watcher.start();
+  }
+}
+
+export function processPackages(pkg: UDSPackage, phase: WatchPhase) {
+  switch (phase) {
+    case WatchPhase.Added:
+      PackageStore.add(pkg);
```
mjnagel: May also want to have `WatchPhase.Modified` here? I think a rename/changing namespace would cause delete + add events, but adding modified would be more future proof for other checks we might add. If you do add it, we would also want/need to modify some of the `add` code to only add the package once:
```ts
  const existingValue = packageNamespaceMap.get(namespace);
  if (existingValue) {
    // Something to check if `pkg` is already in the `existingValue` list
    existingValue.push(pkg);
  } else {
```

docs/reference/configuration/UDS operator/package.md
slaskawi: Also please avoid the text going out of the boxes.

src/pepr/operator/controllers/packages/package-store.ts
```diff
@@ -0,0 +1,136 @@
...
+ * @param {UDSPackage} pkg - The package to be added. It should contain metadata with a namespace.
+ * @param {boolean} [logger=true] - Optional flag to enable logging. Defaults to true.
+ *
+ * This function retrieves the namespace from the package metadata and adds the package
+ * to the packageNamespaceMap. If the namespace is not present, it defaults to an empty string.
+ * For backwards compatibility, the function checks if the namespace has an existing package. If it does,
+ * the package to be added is appended to the existing list of Packages. If not, a new key is created
+ * in packageNamespaceMap for that namespace, containing a list of one element.
+ */
+function add(pkg: UDSPackage, logger: boolean = true): void {
```
slaskawi, line 39: Unless I'm mistaken, this won't survive Pod restarts. Can we use Pepr Store here?

src/pepr/operator/controllers/packages/packages.ts
slaskawi: I echo what @mjnagel said and I was wondering if this behaves correctly after a Pod restart?

src/pepr/operator/controllers/packages/package-store.ts
mjnagel: So we should actually be OK with this approach. When the pod restarts it will lose this in-memory store, but it will start up the watch again before it is even ready to serve validation requests. The first events the watch receives on startup will be "create" events (for everything already in cluster), so it will restore the previous state very quickly. This is actually the same approach we use for the policy exemptions today.

We've largely avoided the PeprStore for this use-case because it doesn't have the transactional, real-time speed needed for shared state across validation operations. There's some context in [this issue](https://github.com/defenseunicorns/uds-core/issues/409) and the related PR.

docs/reference/configuration/UDS operator/package.md
noahpb: Thanks for pointing these out! I have updated the diagrams with all of these considerations. LMK if that's better.

src/pepr/operator/controllers/packages/packages.ts
noahpb: Good callout. I wasn't sure if that config would be necessary here. Is there a default watch config we can override somewhere, or would you prefer setting config explicitly for each watch?

src/pepr/operator/controllers/packages/packages.ts
noahpb: I initially went down this path but ultimately decided to leave it out for this first implementation of the `PackageStore`. I can add it if we agree it's needed.

src/pepr/operator/controllers/packages/packages.ts
mjnagel: Pepr sets a [default watch config](https://github.com/defenseunicorns/pepr/blob/fede12ae9267a637b31fc9a5d150bbd98feaf289/src/lib/processors/watch-processor.ts#L58-L72), BUT that is only true when using Pepr's `.Watch()`. In this case we are setting up our own KFC watch, so we don't get the benefits of Pepr's watch config (and it reading those envs). So if we want the same watch config as Pepr's default, we have to explicitly set it here.

That being said, since this should be identical with the exemption one, we could pull that out to a common spot if we want. I don't feel strongly about that but it could reduce the feeling of copy/paste.

src/pepr/operator/controllers/packages/package-store.ts
noahpb: I can confirm that this is the case in my local testing. This is also true when switching over to this version of the UDS Operator as the `packageNamespaceMap` is automatically populated with `UDSPackages` on startup that existed prior.

I'm going to resolve this now, but feel free to re-open if there are additional concerns.

docs/reference/configuration/UDS operator/package.md
mjnagel: Diagram looks good from my perspective.

src/pepr/operator/controllers/packages/packages.ts
mjnagel: As mentioned elsewhere I don't believe this is strictly required for this validation, but in line with storing the full `Package` resource in our map, keeping the map up to date would better enable using this for future validations. I don't have a strong opinion on whether it needs to be done now, just adding context on why we would want it.

slaskawi:
LGTM on my side

src/pepr/operator/controllers/packages/packages.ts
noahpb: Copy/pasted the watch config from exemptions.

src/pepr/operator/controllers/packages/packages.ts
noahpb: Added an additional case to handle `Modified` Package CRs.

src/pepr/operator/controllers/packages/package-store.ts
```diff
@@ -0,0 +1,150 @@
...
+    if (index !== -1) {
+      // Update existing package
+      existingValue[index] = pkg;
+      if (logger) {
+        log.debug(
+          `Updating PackageStore for package ${pkg.metadata?.name} in namespace ${namespace}.`,
+        );
+      }
+    }
+    existingValue.push(pkg);
```
mjnagel: I believe there is a bug in this logic. When an existing package is found, it will be "updated in-place" using the index, but then also pushed to the array. `existingValue.push(pkg);` should be part of an else clause to ensure this doesn't happen.

noahpb: Yep, great catch! Fix pushed.

src/pepr/operator/controllers/packages/package-store.ts
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * A collection of functions related to watching UDSPackages
+ * Manages an in-memory map of UDSPackage resources
+ * Used in Pepr Validating Webhook Pods when vetting UDS Package resources for admission
+ */
+import { Component, setupLogger } from "../../../logger";
+import { UDSPackage } from "../../crd";
+const log = setupLogger(Component.OPERATOR_PACKAGES);
+
+export type PackageNamespaceMap = Map<string, UDSPackage[]>;
```
mjnagel: Discussed separately but it could simplify some of the logic here if we changed the type of this to be a nested map such as:
```ts
export type PackageNamespaceMap = Map<string, Map<string, UDSPackage>>;
```

This would enable us to just `set` the package in the map for both updates and new creations/additions (without needing to specifically check if it's an update unless we wanted to log accordingly).

src/pepr/operator/controllers/packages/packages.ts
```diff
@@ -0,0 +1,70 @@
...
+      break;
+
+    case WatchPhase.Modified:
+      PackageStore.add(pkg);
+      break;
```
mjnagel:
```suggestion
    case WatchPhase.Added:
    case WatchPhase.Modified:
      PackageStore.add(pkg);
      break;
```

Since these call the same function you can combine/slam together the cases.

mjnagel (dismissed):
LGTM from a functional perspective. Two comments on cleaning/simplifying logic, but the current approach does work as well.

src/pepr/operator/controllers/packages/packages.ts
noahpb: Updated

src/pepr/operator/controllers/packages/package-store.ts
noahpb: Refactored per suggestions.

mjnagel (approved):

Issue #1280, opened 2025-02-13
Make Istio Ambient components default in UDS Core


mjnagel:
Istio ambient components (CNI and Ztunnel) are currently present in uds-core but disabled by default. In order for core to fully adopt ambient these should be switched to `required: true`. We can also reorganize the `src/istio` folder as it makes sense to simplify the inheritance and values structure there.

Note: This may be pending on a decision regarding the Tetrate provided Ztunnel image (registry1 flavor).

mjnagel:
Dependent on https://github.com/defenseunicorns/uds-core/issues/1222

mjnagel:
May be some resource limitations on the runners that we need to work through to make sure all functional layers still can run their tests.

mjnagel:
@UnicornChance I don't think it drastically increases the size here, but I think it would be good to also evaluate/add the gateway api CRDs here too. They could be useful for end users looking to experiment with/use waypoints (to include our own Keycloak pacckage).

Issue #1248, opened 2025-02-03
Implementation for Istio ambient opt-in


mjnagel:
Follow-on/implementation for https://github.com/defenseunicorns/uds-core/issues/1247

Add/implement the ability to opt-in to ambient based on the outcome of the [design doc](https://www.notion.so/UDS-Core-Ambient-Opt-In-193e512f24fc80668c46f3a719fef969). We should have at least one test workload that opts-into ambient in CI.

mjnagel:
More details can/should be added based on the outcome of the design doc.

mjnagel:
May be some initial implementations for this in https://github.com/defenseunicorns/uds-core/pull/1142 and https://github.com/defenseunicorns/uds-core/pull/699

mjnagel:
This is temporarily on hold pending further review and decision of https://github.com/defenseunicorns/uds-core/issues/1245.

Initial draft implementation was pushed up to branches:
- [Draft work](https://github.com/defenseunicorns/uds-core/tree/ambient-opt-in): Does not encompass the network policy side, supports multiple Package CRs in one namespace
- ["Complete" work](https://github.com/defenseunicorns/uds-core/tree/ambient-opt-in-single-pkg): Only allows/supports a single package in the namespace

mjnagel:
Picking this back up with the assumption that it will only support/handle a single package CR in the namespace.

PR #1384, opened 2025-03-25
feat!: implement operator authorization policy management

### LINKED ISSUE #1245: Implementation of L4 network traffic restrictions in ambient mode
Follow-on / implementation for design doc from https://github.com/defenseunicorns/uds-core/issues/1028

Add/implement the auth policy generation in the operator based on the `network.allow` spec in the Package CR.

chance-coleman:
## Description

This PR implements dynamic generation of Istio AuthorizationPolicy resources in ambient mode based on the configuration defined in a UDSPackage. These policies enforce strict ALLOW-only traffic rules, supporting ingress controls, monitoring protection, and gateway exposure with per-port granularity.

## Key Features
- Allow Rules: Processed from spec.network.allow, with support for:
  - `remoteGenerated`, `remoteNamespace`, and `remoteServiceAccount` logic
  - Pod-level selector-based grouping for targeted policy generation
- Expose Rules: Derived from `spec.network.expose`, mapped to appropriate source namespaces based on gateway type
- Monitor Rules: One policy per monitor entry under `spec.monitor`, securing endpoints (e.g., Prometheus metrics)
- Policy Grouping:
  - Rules with selectors are grouped and merged into workload-specific policies
  - Rules without selectors are merged into a namespace-wide policy
- Port Handling:
  - Rules include `to` blocks when port(s) are provided
  - If no port is defined, only the `from` block is used
- Metadata Labels: Policies include `uds/package` and `uds/generation` labels for traceability
- Orphan Cleanup: Automatically purges outdated policies from previous generations

## Additional Context
Documentation is included to explain the policy generation flow, source derivation logic, grouping behavior, and real-world examples. Also see jest tests for additional UDS Core Package examples.

## Related Issue

Fixes #1245
Relates to #1248 #1280 

## Type of change

## Steps to Validate
- `uds run test-uds-core`
- validate authpols that are created for UDS packages
- apply additional packages to cluster and verify the generated authpols

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: ambient mesh authorization policy generation (https://github.com/defenseunicorns/uds-core/pull/1384)

BREAKING CHANGE: dynamic generation of Istio AuthorizationPolicy resources in ambient mode based on the configuration defined in a UDSPackage. ALL requests to a pod with authservice protection MUST have a jwt from Keycloak. More fine-grained protection can be done with additional DENY policies, but there will always be a requirement for a Keycloak JWT in addition.
END_COMMIT_OVERRIDE


.gitignore
```diff
@@ -32,3 +32,5 @@ uds-docs/**
 **.backup
 **/.playwright/**
 **/.playwright
+
+coverage/**
```
chance-coleman, line 36: `coverage` directory created by jest flag --coverage. Nice to have but not necessary if anyone is against adding it.

mjnagel: Don't see any problem from adding this - I've run `jest` with coverage before and just have to delete later.

bundles/k3d-slim-dev/uds-bundle.yaml
```diff
@@ -40,6 +40,8 @@ packages:
     # x-release-please-start-version
     ref: 0.38.0
     # x-release-please-end
+    optionalComponents:
+      - istio-ambient
```
mjnagel: I think this is probably okay, but may want to leave out for resource reasons for uds-packages (until we have the full opt-in option).

docs/reference/configuration/authorization-policies.md
```diff
@@ -0,0 +1,290 @@
+---
+title: Authorization Policies in Ambient Mode
+sidebar:
+  order: 2
+---
+
+## Overview
+
+In an ambient mode service mesh environment, security is enforced using Istio AuthorizationPolicies. The UDS‑Core operator automatically generates these **ALLOW** policies based on your UDSPackage configuration. The generated policies ensure that only explicitly permitted traffic is allowed, enforcing per‑port access control for your applications. In addition to handling custom ingress rules (allow/expose), the operator also processes monitor configurations to protect monitoring endpoints (such as Prometheus metrics).
```
mjnagel:
```suggestion
In an ambient mode service mesh environment, ingress network security is partially enforced using Istio AuthorizationPolicies. The UDS‑Core operator automatically generates these **ALLOW** policies based on your UDSPackage configuration. The generated policies ensure that only explicitly permitted traffic is allowed, enforcing per‑port access control for your applications. In addition to handling custom ingress rules (allow/expose), the operator also processes monitor configurations to protect monitoring endpoints (such as Prometheus metrics).
```

Feel free to change wording, just want to note that this is ingress network security essentially.

docs/reference/configuration/authorization-policies.md
```diff
@@ -0,0 +1,290 @@
...
+
+- Policies are generated for rules defined in the allow, expose, and monitor blocks, with each rule processed individually.
+
+- Allow rules determine the source based on remote namespace settings, remoteGenerated, and remoteServiceAccount (which specifies a principal).
+
+- Expose rules set the source using principals based on the specified gateway.
+
+- Monitor rules produce individual policies to secure monitoring endpoints.
+
+- Generated policy names incorporate an "ingress-" naming pattern along with metadata for traceability.
```
mjnagel: Open to other thoughts but I wonder if most of the content in this doc fits better as a dev doc? I don't think end users generally want/need to know how we translate the `Package` CR spec to other resources (we don't typically document this for servicemonitor, virtualservice for example).

For end user facing documentation I do see a few things that would definitely still be valuable:
- network security is enforced primarily with network policies
- to support istio ambient we also enforce network security (for ingress) with ALLOW authorization policies
- as a result, end user authorization policies should be implemented as DENY policies (could link to how Istio evaluates rules)
- for "best" security we recommend adding `remoteServiceAccount`

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -68,6 +68,7 @@ function jwtAuthZAuthorizationPolicy(
       namespace,
     },
     spec: {
+      action: IstioAction.Deny,
```
mjnagel, line 71: Let's make sure we mark this PR as a breaking change because of this, with a note on the impact. In short - authservice protection is absolute, all requests to a pod with authservice protection MUST have a jwt from Keycloak. More fine-grained protection can be done with additional DENY policies, but there will always be a requirement for a Keycloak JWT in addition.

It might also be good to update our [documentation for authservice](https://uds.defenseunicorns.com/reference/configuration/single-sign-on/auth-service/) to note that Authservice is intended for simple situations only (where absolute protection is okay). For more advanced scenarios auth should be implemented directly in the app or via a different means.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+ */
+function buildMonitorAuthPolicy(
+  pkgName: string,
+  pkgNamespace: string,
+  generation: string,
+  monitor: Monitor,
+): AuthorizationPolicy {
+  const mSelector = monitor.podSelector ?? monitor.selector;
+  const monitorRule: Rule = {
+    from: [{ source: { namespaces: ["monitoring"] } }],
```
mjnagel: This could/should be a service account (principal) since we know the source is prometheus (similar to the istio gateway rules for expose).

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+  if (!isEmpty(source)) {
+    ruleEntry.from = [{ source }];
+  }
+  if (ports.length > 0) {
+    ruleEntry.to = [{ operation: { ports } }];
+  }
+  return {
+    apiVersion: "security.istio.io/v1beta1",
+    kind: "AuthorizationPolicy",
+    metadata: {
```
mjnagel, line 184: I think we're missing ownerref in here (this is what would ensure these policies are deleted when the package CR is removed). Should be able to use the util function `getOwnerRef` for setting that up, although you may need to change this function's args to ensure that works.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+      } catch (err) {
+        log.error(
+          err,
+          `Error applying AuthorizationPolicy ${policy.metadata?.name} in namespace ${policy.metadata?.namespace}`,
+        );
+      }
+    }),
+  );
+
+  await purgeOrphans(generation, pkgNamespace, pkgName, AuthorizationPolicy, log);
```
mjnagel: I think this could potentially open a hole on authservice protection even if its brief. Since `purgeOrphans` strictly looks at the resource type, owner, and generation - it doesn't have the context of whether a given authpolicy is for our "networking" or authservice protection. Due to the [ordering of the reconciler](https://github.com/defenseunicorns/uds-core/blob/ambient-mesh-operator/src/pepr/operator/reconcilers/package-reconciler.ts#L75-L87), for an authservice-protected application this purge would delete the authservice authpolicies (since they would still have an older generation in metadata).

There's probably a few ways we could handle this:
- only purge allow policies (probably a one-off purge function)
- use an additional label like `uds/for: network` or `uds/for: authservice` and filter/purge based on that (maybe could do this in the `purgeOrphans`, or as a one-off)
- move where purge is called, or make it conditional, to remove any timing risks
- ?

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -585,6 +600,17 @@ export const v1alpha1: V1CustomResourceDefinitionVersion = {
...
+                  description: "Service mesh configuration for the package",
+                  type: "object",
+                  properties: {
+                    ambient: {
+                      description: "Enable ambient authentication for the service mesh",
+                      type: "boolean",
+                      default: false,
+                    },
+                  },
+                },
```
mjnagel: I think we can drop this from this PR since this toggle isn't implemented here (handled separately by https://github.com/defenseunicorns/uds-core/pull/1386) - and make sure that all the "child" files are regenerated (types, docs, schema).

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -105,6 +108,7 @@ export async function packageReconciler(pkg: UDSPackage) {
       endpoints,
       monitors,
       networkPolicyCount: netPol.length,
+      authorizationPolicyCount: authPol.length,
```
mjnagel: I think we might've discussed already - should this count also include the authservice authpolicies (if present)?

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+      log.debug(`Generated monitor authpol: ${authPolicy.metadata?.name}`);
+    }
+  }
+
+  // Apply policies concurrently.
+  await Promise.all(
+    policies.map(async policy => {
+      try {
+        await K8s(AuthorizationPolicy).Apply(policy, { force: true });
+        log.info(
```
mjnagel: nit: more of a debug, or even trace, log?

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+          `Applied AuthorizationPolicy ${policy.metadata?.name} in namespace ${policy.metadata?.namespace}`,
+        );
+      } catch (err) {
+        log.error(
+          err,
+          `Error applying AuthorizationPolicy ${policy.metadata?.name} in namespace ${policy.metadata?.namespace}`,
+        );
+      }
+    }),
+  );
```
mjnagel: I think this is likely more efficient than how we apply network policies, but I'm slightly concerned that we may overwhelm the API server this way - especially when Pepr cycles and is reconciling 10+ packages at the same time, each with 1-10 authpolicies likely? I'm not sure if we've seen any real time concerns with the one at a time approach of the netpols ([here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/policies.ts#L131)) - open to this route if you think it's safe though.

Regardless we should make sure we re-throw the error if one is caught so that this fails the `Package` CR reconciliation (and results in a retry). I'm not sure if that gets more confusing here with the `Promise.all` and when to throw/what to throw if multiple fail?

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+ */
+function processAllowRule(rule: Allow, pkgNamespace: string): { source: Source; ports: string[] } {
+  const ports: string[] = [];
+  if (rule.port !== undefined) ports.push(rule.port.toString());
+  if (rule.ports) ports.push(...rule.ports.map(p => p.toString()));
+
+  let source: Source = {};
+  if (rule.remoteCidr) {
+    source = { ipBlocks: [rule.remoteCidr] };
+  } else if (
```
mjnagel: I think `ipBlocks` would be used for a few of the `remoteGenerated` rules as well?
- `KubeAPI`: output of [this function](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/generators/kubeAPI.ts#L54)
- `KubeNodes`: output of [this function](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/generators/kubeNodes.ts#L54)
- `CloudMetadata`: [`META_IP`](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/generators/cloudMetadata.ts#L8)

I don't think any of these are super likely to have ingress rules for, but I have seen some in the wild (especially `KubeAPI`) so we likely should support this unless we want to explicitly block this config. Assuming we do support this, we should also make sure our node/kubeapi watches update the generated policies for those similar to netpols (see [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/generators/kubeAPI.ts#L147-L148) for example).

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+    rule.remoteNamespace === "" ||
+    rule.remoteNamespace === "*" ||
+    rule.remoteGenerated === RemoteGenerated.Anywhere
+  ) {
+    source = {};
+  } else if (rule.remoteGenerated === RemoteGenerated.IntraNamespace) {
+    source = { namespaces: [pkgNamespace] };
+  } else if (rule.remoteNamespace && rule.remoteNamespace !== "" && rule.remoteNamespace !== "*") {
```
mjnagel: nit: I think this is redundant since it's an else if after the second two conditions have been checked (in the inverse) already?
```suggestion
  } else if (rule.remoteNamespace) {
```

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+  } else {
+    source = {};
+  }
+  if (rule.remoteServiceAccount && rule.remoteServiceAccount.trim() !== "") {
+    const nsForSA =
+      rule.remoteNamespace && rule.remoteNamespace.trim() !== ""
+        ? rule.remoteNamespace
+        : pkgNamespace;
+    source = { principals: [`cluster.local/ns/${nsForSA}/sa/${rule.remoteServiceAccount}`] };
+  }
```
mjnagel: Since this is an "alternative" to the `from.namespaces` rules, would it make sense to include this in the `else if` flow above just to bring it in line with the other logic? Logically it should be a higher precedence than any of the namespace conditions.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+      const authPolicy = buildAuthPolicy(pkgName, pkgNamespace, generation, rule, source, ports);
+      policies.push(authPolicy);
+      log.debug(`Generated authpol: ${authPolicy.metadata?.name}`);
+    }
+  }
+
+  // Process monitor rules.
+  if (pkg.spec?.monitor) {
+    for (const monitor of pkg.spec.monitor) {
+      const authPolicy = buildMonitorAuthPolicy(pkgName, pkgNamespace, generation, monitor);
```
mjnagel: Probably just missing it, but is there a reason the monitor section is different than the other two? For allow/expose they both have a `process` function and then use a common `buildAuthPolicy` function. This one is "all in one" instead.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Generate and apply Istio Authorization Policies for a given UDSPackage.
+ */
+export async function generateAuthorizationPolicies(
+  pkg: UDSPackage,
+  overrideNamespace?: string,
+): Promise<AuthorizationPolicy[]> {
+  const pkgName = pkg.metadata?.name ?? "unknown";
+  const pkgNamespace = pkg.metadata?.namespace ?? overrideNamespace ?? "default";
```
mjnagel: It should be impossible to hit the fallback here (`default`) I think, but is there a reason that `overrideNamespace` isn't just directly used? I think we assume it exists for the netpols so thinking we could similar here?

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Derives a short name from a selector.
+ */
+function derivePolicyName(selector: { [key: string]: string }): string {
+  if (selector["app"]) return selector["app"].replace(/-pod$/, "");
+  if (selector["app.kubernetes.io/name"])
+    return selector["app.kubernetes.io/name"].replace(/-pod$/, "") + "-workload";
+  return "workload";
+}
```
mjnagel: Was this to keep names a shorter length in general? Or targetted on a specific usecase? I'd be worried that the `workload` fallback could cause name conflicts in some cases, when we could probably use `${Object.values(selector)}` or similar alternatively.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,269 @@
...
+    ruleEntry.from = [{ source }];
+  }
+  if (ports.length > 0) {
+    ruleEntry.to = [{ operation: { ports } }];
+  }
+  return {
+    apiVersion: "security.istio.io/v1beta1",
+    kind: "AuthorizationPolicy",
+    metadata: {
+      name: policyName,
```
mjnagel, line 185: I think we'll want to make sure we wrap this in `sanitizeResourceName` (and anywhere else we're making generated names). That just ensures that we truncate and trim anything that would violate k8s resource naming restrictions.

docs/reference/configuration/authorization-policies.md
chance-coleman: Yeah i agree with this, i had made a note of that and forgot to add it in the PR.

src/pepr/operator/controllers/network/authorizationPolicies.ts
chance-coleman: Yeah, i think that we should stick to the similar implementation as a few other places. I don't have strong feelings either way, and this is tried and true at the moment. If we hit any efficiency problems we can readdress.

src/pepr/operator/controllers/network/authorizationPolicies.ts
chance-coleman: yeah good eyes

src/pepr/operator/controllers/network/authorizationPolicies.ts
chance-coleman: yeah agreed

src/pepr/operator/controllers/network/authorizationPolicies.ts
chance-coleman: Good question, no reason 😄 refactored.

src/pepr/operator/controllers/network/authorizationPolicies.ts
chance-coleman: Check this again, i reworked the naming functions a bit. It still has a fallback to workload but i believe that should be impossible or really hard to actually get to.

src/pepr/operator/controllers/network/authorizationPolicies.ts
chance-coleman: this was OBE, however i think it would be good to double check this.

src/pepr/operator/controllers/network/authorizationPolicies.ts
chance-coleman: I went the route of adding a label for generated authpols as `uds/for: network` and added an additional parameter to the `pureOrphans` to filter out resources if that optional parameter was supplied.

docs/dev/authorization-policy-generation.md
```diff
@@ -0,0 +1,251 @@
+## Overview
+
+This guide describes how Istio AuthorizationPolicies are generated from the UDSPackage CR by the UDS‑Core operator. These **ALLOW** policies are primarily used to enable ingress security within an Istio Ambient Mesh environment.
+
+The code responsible for generating these policies can be found in the Pepr policy implementation and includes support for three rule types:
+- `allow`: Direct ingress rules for services.
+- `expose`: Gateway-based ingress exposure.
+- `monitor`: Restricts access to metrics endpoints.
+
```
noahpb: The wording here feels a bit off. I think the first sentence should be reworded to suggest that the code can be found in the `controllers/network` dir. As it reads now, it could be confused with the `pepr/policies` dir.

docs/reference/configuration/authorization-policies.md
```diff
@@ -0,0 +1,99 @@
+---
+title: How Authorization Policies Protect Your Services
+sidebar:
+  order: 3
+---
+
+In clusters running Istio Ambient Mesh, UDS‑Core enforces **ingress network security** using Istio **ALLOW** AuthorizationPolicies. These policies are automatically generated for each application package you define with a UDSPackage resource.
+
```
noahpb, line 8:
```suggestion
[UDS Package](https://uds.defenseunicorns.com/reference/configuration/uds-operator/package/)
```
I think we should try to be consistent with our naming of the UDS Package CR across our documentation. In the codebase we refer to UDS Packages as a `UDSPackage`, but it might not be immediately clear to someone when reading the docs that we are referring to the Package Custom Resource for the operator.

docs/reference/configuration/authorization-policies.md
```diff
@@ -0,0 +1,99 @@
...
+spec:
+  network:
+    allow:
+      - direction: Ingress
+        remoteNamespace: "external-app"
+        remoteServiceAccount: "my-client"
+        port: 8080
+```
+
+> This ensures that only a specific service account in another namespace can access your service.
```
noahpb:
```suggestion
> This ensures that only a workload running as this specific service account in another namespace can access your service.
```

docs/reference/configuration/authorization-policies.md
```diff
@@ -0,0 +1,99 @@
...
+- You should create your own DENY policies for more fine-grained control.
+
+More info: [Istio Authorization Policy Evaluation](https://istio.io/latest/docs/concepts/security/#authorization-policy)
+
+---
+
+## Summary
+
+- Ingress is denied by default.
+- You allow ingress by defining `allow` or `expose` rules in your UDSPackage.
```
noahpb:
```suggestion
- You allow ingress by defining `allow` or `expose` rules in your UDS Package resource definition.
```

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,259 @@
...
+  Source,
+} from "../../crd/generated/istio/authorizationpolicy-v1beta1";
+import { purgeOrphans, sanitizeResourceName } from "../utils";
+
+const log = setupLogger(Component.OPERATOR_NETWORK);
+
+// Constants for gateway principals.
+const ADMIN_INGRESS = "cluster.local/ns/istio-admin-gateway/sa/admin-ingressgateway";
+const TENANT_INGRESS = "cluster.local/ns/istio-tenant-gateway/sa/tenant-ingressgateway";
+
```
noahpb, line 28: Should we add a passthrough gateway principal here?

src/pepr/operator/controllers/network/authorizationPolicies.ts
mjnagel: Resolving this one, but I think the same applies (in a different spot) so I'll comment there.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,272 @@
...
+      policies.push(authPolicy);
+      log.debug(`Generated authpol: ${authPolicy.metadata?.name}`);
+    }
+  }
+
+  // Process monitor rules.
+  if (pkg.spec?.monitor) {
+    for (const monitor of pkg.spec.monitor) {
+      const selector = monitor.podSelector ?? monitor.selector;
+      const source: Source = { namespaces: ["monitoring"] };
```
mjnagel: Source here should be `principal`, with the prometheus service account (to be most specific/secure).

src/pepr/operator/controllers/network/authorizationPolicies.ts
mjnagel: Latest commit looks great to handle these for initial creation. I think for kubenodes and kubeapi we still need to make sure we handle updates when the kubeapi cidrs or node IPs change ([here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/index.ts#L57-L65) and [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/index.ts#L106-L114) is where we watch for those and call update functions). We currently handle this for networkpolicies by adding another generated label that we can filter on and make updates to those policies for (see [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/generators/kubeNodes.ts#L91-L95) for example).

mjnagel: LGTM on latest commit.

mjnagel (dismissed):

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -148,6 +152,78 @@ export async function updateKubeNodesNetworkPolicies() {
...
+export async function updateKubeNodesAuthorizationPolicies(): Promise<void> {
+  // Build the current set of node peers from nodeSet.
+  const newPeers = buildNodePolicies([...nodeSet]);
+  // Extract CIDR strings from the new peers.
+  const newIpBlocks = newPeers
+    .map(peer => peer.ipBlock?.cidr)
+    .filter((cidr): cidr is string => typeof cidr === "string");
+
+  // Query for AuthorizationPolicies labeled with uds/generated=KubeNodes.
+  log.debug("Calling K8s(AuthorizationPolicy).WithLabel(...).Get()");
```
mjnagel: nit: locally this was kind of noisy in the logs for me, maybe more of a trace log if we need to enable it for dev purposes?

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -148,6 +152,78 @@ export async function updateKubeNodesNetworkPolicies() {
...
+  const authPols = await K8s(AuthorizationPolicy)
+    .WithLabel("uds/generated", RemoteGenerated.KubeNodes)
+    .Get();
+  log.debug("Fetched AuthorizationPolicies:", authPols);
```
mjnagel: Similar comment here - probably a trace log?

src/pepr/operator/controllers/network/generators/kubeNodes.ts
chance-coleman: im going to remove this log. doesn't provide a whole lot of value.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,285 @@
...
+      const authPolicy = buildAuthPolicy(
+        policyName,
+        pkg,
+        rule.selector,
+        source,
+        ports,
+        additionalLabels,
+      );
+      policies.push(authPolicy);
+      log.debug(`Generated authpol: ${authPolicy.metadata?.name}`);
```
mjnagel: More broad comment on all of these logs, with the exception of `Starting policy generation...` - probably could bump them to trace logs. Otherwise it gets pretty noisy and hard to follow individual package CR reconciliation since there could be x generated authpol logs + x applied authpol logs.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -0,0 +1,285 @@
...
+ * Generate and apply Istio Authorization Policies for a given UDSPackage.
+ */
+export async function generateAuthorizationPolicies(
+  pkg: UDSPackage,
+  pkgNamespace: string,
+): Promise<AuthorizationPolicy[]> {
+  const pkgName = pkg.metadata?.name ?? "unknown";
+  const generation = pkg.metadata?.generation?.toString() ?? "0";
+  log.info(
+    `Starting policy generation for package "${pkgName}" in namespace "${pkgNamespace}" (generation ${generation}).`,
```
mjnagel:
```suggestion
    `Starting authorization policy generation for package "${pkgName}" in namespace "${pkgNamespace}" (generation ${generation}).`,
```

mjnagel (dismissed):

docs/dev/authorization-policy-generation.md
```diff
@@ -0,0 +1,251 @@
+## Overview
+
+This guide describes how Istio AuthorizationPolicies are generated from the UDSPackage CR by the UDS‑Core operator. These **ALLOW** policies are primarily used to enable ingress security within an Istio Ambient Mesh environment.
```
noahpb: This applies to all references of `UDS-Core` in this PR: let's use `UDS Core` when talking about the operator and save `uds-core` for when we're referring to the package artifact as it pertains to uds cli/zarf.

noahpb (commented):
One additional nit thing on docs. Apart from that LGTM!

docs/dev/authorization-policy-generation.md
mjnagel: I think for this reference in particular we typically refer to the operator as `UDS Operator` but may want to check other references to confirm that.

chance-coleman: updated the two `UDS-Core` and the `UDS Core operator` to `UDS Operator`

mjnagel (approved):

PR #1386, opened 2025-03-25
feat: add serviceMesh.mode in Package CR

### LINKED ISSUE #1248: Implementation for Istio ambient opt-in

mjnagel:
## Description

Adds `serviceMesh.mode` to the `Package` CR spec with associated reconciliation code to handle running packages in sidecar or ambient mode.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1248

## Type of change

## Steps to Validate

The podinfo package is switched to ambient mode in this PR, but you can also test with similar packages to switch to ambient and validate all works as expected.

## Checklist before merging


mjnagel:
Noting that this has some _loose_ dependencies (could be merged without these technically):
- https://github.com/defenseunicorns/uds-core/issues/1352 : Without this change the behavior in this PR will just follow a "last one wins" precedence for the istio mode (it essentially assumes a single CR and clobbers the mode). If that issue is not resolved first, this PR would be a breaking change partially since it removes some of the logic to handle multiple CRs on cleanup.
- https://github.com/defenseunicorns/uds-core/issues/1280 : Without this we should be okay as I added operator config/watch logic to fallback to sidecar if the ztunnel component isn't deployed. But may need to review if we want this logic in the interim or not.
- https://github.com/defenseunicorns/uds-core/issues/1245 : This is a harder dependency to make sure ambient can be safely added

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -142,6 +151,24 @@ export async function networkPolicies(pkg: UDSPackage, namespace: string) {
       policy.metadata.name = `allow-${pkgName}-${policy.metadata.name}`;
     }
 
+    // todo: validate if this is needed for out of mesh locations???
```
chance-coleman: noting this todo, still relevant?

src/pepr/operator/controllers/istio/namespace.ts
```diff
@@ -0,0 +1,275 @@
...
+  const updatedNamespace = {
+    metadata: {
+      name: pkg.metadata.namespace,
+      labels,
+      annotations,
+    },
+  };
+  if (
+    JSON.stringify(sourceNS.metadata?.labels) !== JSON.stringify(labels) ||
+    JSON.stringify(sourceNS.metadata?.annotations) !== JSON.stringify(annotations)
```
chance-coleman: Is there any concerns with key order being inconsistent in the labels and annotations? if yes, i think using JSON.stringify, will result in errors because of the comparison method it uses.

src/pepr/operator/controllers/istio/namespace.ts
```diff
@@ -0,0 +1,275 @@
...
+
+    // Get the UID of the owner of the pod or default to "other" (shouldn't happen)
+    const controlledBy = pod.metadata?.ownerReferences?.find(ref => ref.controller)?.uid || "other";
+    groups[controlledBy] = groups[controlledBy] || [];
+    log.debug(`Adding Pod ${ns}/${pod.metadata?.name} to ${controlledBy} deletion list.`);
+    groups[controlledBy].push(pod);
+  }
+
+  // Delete each group of pods
+  for (const group of Object.values(groups)) {
```
chance-coleman, line 188: Just a thought, could we use Promise.all here instead of deleting sequentially? I suppose it might not result in that much of a gain but could help if one pod is taking awhile.

src/pepr/operator/controllers/network/policies.ts
mjnagel: Yep - just remembered this yesterday. I'll do some validation and update and/or remove the comment.

src/pepr/operator/controllers/istio/namespace.ts
mjnagel: I'll take a look at this and see what makes sense - I think it was largely copy/paste with some edits from the old injection.ts file.

src/pepr/operator/controllers/istio/namespace.ts
mjnagel: Ditto to above comment, this was mostly a copy/paste. On this one I'm a bit more wary of editing since this has been there since early days and largely "just works" but I can see the potential performance gains here. I'm also a bit wary of cycling too many pods at once based on our experience with automating the native sidecar switch - might lean towards leaving this if you're okay with it.

chance-coleman: yeah totally fine by me! I would hope that the performance gain wouldn't be that much.

src/test/tasks.yaml
```diff
@@ -148,6 +148,28 @@ tasks:
             namespace: podinfo
             condition: "'{.status.phase}'=Ready"
 
+      - description: Verify podinfo namespace has ambient mode label
+        cmd: |
+          set -e
+          AMBIENT_LABEL=$(uds zarf tools kubectl get namespace podinfo -o jsonpath='{.metadata.labels.istio\.io/dataplane-mode}')
```
slaskawi: Would it be beneficial to move these tests to `jest`?

src/pepr/operator/index.ts
```diff
@@ -98,6 +98,25 @@ When(UDSPackage)
...
+    // todo: wait for keycloak and authservice to be running?
+    log.info("Istio Ambient deployed, operator configured to handle ambient packages.");
+    UDSConfig.isAmbientDeployed = true;
+  });
+When(a.DaemonSet)
+  .IsDeleted()
+  .InNamespace("istio-system")
+  .WithName("ztunnel")
+  .Watch(() => {
+    log.info("Istio Ambient deployed, operator will run all packages in sidecar mode.");
```
slaskawi: Do we need to wait for Keycloak and Authservice here as well?

docs/reference/configuration/UDS operator/package.md
```diff
@@ -10,6 +10,10 @@ sidebar:
 
 - **Enabling Istio Sidecar Injection:**
   - The operator facilitates the activation of Istio sidecar injection within namespaces where the CR is deployed.
+- **Support for Istio Ambient Mode:**
+  - Packages can now opt into Istio's Ambient mode by setting `spec.network.serviceMesh.mode: ambient`. This provides service mesh capabilities without sidecars, reducing resource overhead.
+  - If Ambient mode is not deployed in the cluster, the operator will automatically fall back to sidecar mode.
+  - Note that `Packages` with Authservice clients are not currently supported in ambient mode.
```
slaskawi: This is probably a crucial information and I also see we fail if a user tries to deploys an unsupported configuration. Can we somehow make it easier to spot in the documentation? Maybe we should put it in a "Limitations" paragraph or something similar.

src/pepr/operator/controllers/network/generators/ambientHealthprobes.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { Direction } from "../../../crd";
+import { generate } from "../generate";
+
+// See https://istio.io/latest/docs/ambient/usage/networkpolicy/#ambient-health-probes-and-kubernetes-networkpolicy
+export const allowAmbientHealthprobes = (namespace: string) =>
+  generate(namespace, {
+    direction: Direction.Ingress,
+    description: "Ambient Healthprobes",
+    remoteCidr: "169.254.7.127/32",
```
slaskawi, line 14: Are we sure this hardcoded cidr will be stable over time?

src/pepr/operator/controllers/istio/namespace.ts
```diff
@@ -0,0 +1,275 @@
...
+  }
+}
+
+/**
+ * Forces deletion of pods with the incorrect istio sidecar state
+ *
+ * @param ns The namespace to target
+ * @param wantSidecar Whether injection is being enabled
+ */
+export async function killPods(ns: string, wantSidecar: boolean) {
```
slaskawi, line 150: It seems we decided to kill the Pods manually. Wouldn't it be better to add for example an annotation to the Deployment/StatefulSet and let the controllers do this work for us?

src/pepr/operator/index.ts
mjnagel: Not sure I'm following - the intent with this watch (and the previous one) is to detect if Istio is "ambient enabled", or in other words if ztunnel is deployed so that we can run packages in ambient mode. The only thing we should care about in this check is the ztunnel daemonset I think?

src/pepr/operator/controllers/network/generators/ambientHealthprobes.ts
mjnagel: I'm not sure this is a suitable answer but this IP comes [Istio's documentation](https://istio.io/latest/docs/ambient/usage/networkpolicy/#ambient-health-probes-and-kubernetes-networkpolicy) where they state:
> by default ambient uses the link-local address 169.254.7.127 to identify and correctly allow kubelet health probe packets

From what I can tell this is an IP that Istio has chosen so I think we can expect it to be stable unless they make a breaking change?

src/pepr/operator/controllers/istio/namespace.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  // Handle labels based on ambient opt-in or sidecar default
+  if (pkg.spec?.network?.serviceMesh?.mode === Mode.Ambient) {
+    // Check if ambient mode is available
+    if (!UDSConfig.isAmbientDeployed) {
+      // Ambient mode requested but not available, fall back to sidecar mode
+      log.warn(
+        `Ambient mode requested for package ${pkg.metadata.name} but Ambient is not deployed. Falling back to sidecar mode.`,
+      );
+      await writeEvent(pkg, {
```
noahpb, line 61: Should we add a status to the Package that indicates this behavior as well? For the sake of being declarative, it might make sense to indicate that what is specified in the spec might not actually be true once fully reconciled.

src/pepr/operator/controllers/istio/namespace.ts
mjnagel: While I can see some benefits to this, I think it could cause unnecessary cycling later on and potentially add more complexity here. 

The first aspect to consider is that pods could be controlled by any number of owners - deployments, daemonsets, statefulsets, jobs, or CRs. While most of these could be standard code to patch a certain annotation spot, the CRs are likely more complexity to figure out as well. We could fallback to this kill logic is we don't recognize a standard ownerref to remove some of these considerations I think, but the patching on it's own might cause other "problems".

If we were to patch an annotation or similar on the owner that would allow the controllers to handle cycling the pods. However, on the next deploy/upgrade that annotation would be removed (since it's not in the helm chart baseline for the app), causing the pods to cycle again, even if there is no real app upgrade. There's again possible ways to mitigate this with mutations, but that adds to the logic and complexity here.

With that in mind - curious your thoughts. This logic was the initial approach we took for istio injection since pod cycling generally only happened once when you first deploy. With the ability to switch from sidecar -> ambient this might get called again. I mostly copied this logic over from how we did it before (see the injection.ts file) since that seemed proven to work, but always worth reevaluating if we're doing the right thing.

src/pepr/operator/index.ts
noahpb:
```suggestion
    log.info("Istio Ambient removed, operator will run all packages in sidecar mode.");
```

src/pepr/operator/controllers/istio/namespace.ts
```diff
@@ -0,0 +1,275 @@
...
+}
+
+/**
+ * Determine the proper istio mode (ambient or sidecar) and enable for the package namespace
+ *
+ * @param pkg
+ * @returns string mode
+ */
+export async function enableIstio(pkg: UDSPackage) {
+  if (!pkg.metadata?.namespace || !pkg.metadata.name) {
```
noahpb, line 35: Is this check redundant? Would k8s automatically reject a resource without a name?

src/pepr/operator/index.ts
slaskawi: ooh I see....

The comment in line 107 is the a bit misleading as it says:

```
// todo: wait for keycloak and authservice to be running?
```

So the question is how do we want to proceed here? Do we care only about ztunnel or do we want to ensure the Auth stack is also deployed when someone "flips the switch"?

src/pepr/operator/controllers/network/generators/ambientHealthprobes.ts
slaskawi: That makes sense. We should see a failure in the tests. 

Would you mind to add a small comment here with the link to the Istio doc for future maintainers?

src/pepr/operator/controllers/istio/namespace.ts
slaskawi: I OK with this approach. Perhaps it's worth consulting this with @soltysh as he's very familiar with writing this type of code.

src/pepr/operator/controllers/network/generators/ambientHealthprobes.ts
mjnagel: Docs link is above on line 9 🙂

src/pepr/operator/index.ts
mjnagel: Ah great catch. I copied that block from the Keycloak watch and forgot to clean up the comment (see line 88). I'll remove it - the auth stack shouldn't be relevant here.

src/pepr/operator/controllers/istio/namespace.ts
soltysh: > The first aspect to consider is that pods could be controlled by any number of owners - deployments, daemonsets, statefulsets, jobs, or CRs. 

That's actually something that should not happen. Every controller has a logic to adopt a pod (only orphans ie. not having owner refs are adopted) or not, and usually the pods are not shared (I can't say for the entire world), but definitely in core k8s you'll never see that happen. It's possible to have several layers, like Deployment managing ReplicaSets which manages Pods. But that doesn't mean  a pod is both managed by deployment and replicaset. 

> I OK with this approach. Perhaps it's worth consulting this with @soltysh as he's very familiar with writing this type of code.

In general this is ok approach. The controllers are designed to handle those situation nicely. The only suggestion I could make is around using eviction rather than straight deletion. The former will take PodDisruptionBudget into account, making sure your application is available for the entire time. The latter is just deletion, which might mean (eg. in a single replica app) downtime, or limited availability due to pod re-creation.

mjnagel: > That's actually something that should not happen. Every controller has a logic to adopt a pod (only orphans ie. not having owner refs are adopted) or not, and usually the pods are not shared (I can't say for the entire world), but definitely in core k8s you'll never see that happen. It's possible to have several layers, like Deployment managing ReplicaSets which manages Pods. But that doesn't mean a pod is both managed by deployment and replicaset.

Sorry that was definitely poor wording on my part - I meant that 10 pods in a namespace could be managed by 10 different controllers (in theory). Each individual pod should only be managed by a single owner (deployment, daemonset, etc) as you rightly stated. But we'd have to account for each of the different owner types/controllers (to include CRs) I think if we went this alternative route of editing/triggering the controller rather than deleting the pod.

src/pepr/operator/controllers/istio/namespace.ts
mjnagel: In theory I think you're right - this was a copy from the [current injection logic](https://github.com/defenseunicorns/uds-core/blob/122dc584c97fb789b6664683811dc5d33f7714bc/src/pepr/operator/controllers/istio/injection.ts#L22-L25). When I look at the typescript types it shows metadata as optional, but I think in practice that should be impossible.

I think this was likely originally added mostly for typescript type safety, if we add this check it simplifies the logic/code in the rest of this (otherwise we have to "account for undefined" for Typescript to compile happily).

mjnagel: I'll add a comment on ^

src/pepr/operator/controllers/istio/namespace.ts
mjnagel: As discussed synchronously for now I'm leaning towards leaving this just in the events, given that this is a temporary state (while ztunnel / ambient components are optional).

src/test/tasks.yaml
mjnagel: Migrated these to jest on the latest commit.

src/pepr/operator/controllers/network/policies.ts
mjnagel: After chatting synchronously and testing this out:
- For ingress from in-cluster remotes we need 15008 almost always (if either the source or destination is in ambient mode)
- For egress to in-cluster remotes we need 15008 almost always (same constraints, source or destination in ambient)
- For `remoteGenerated: Anywhere` we don't know if `Anywhere` is in or out of cluster/mesh (it could be both)
- For `remoteCidr` we also don't have a good way of knowing if this is in/out of cluster/mesh - while this is typically used for out of cluster it could cover both
- For the other 3 `remoteGenerated` types (CloudMetadata, KubeAPI, KubeNodes) there is no realistic scenario where these would be in-mesh (CloudMetadata is guaranteed out of the cluster as well)

Essentially we can only safely drop the `15008` port from policies using either of these 3 `remoteGenerated` endpoints. I'm testing out some code locally to see how easy/feasible it is to skip the port on these. In general those 3 targets tend to be used without port-based restrictions so I don't think the logic will be likely to get executed regardless and the net effect is allowing traffic to a specific istio port. If it seems overly complex to add I may skip this for now.

docs/reference/configuration/UDS operator/package.md
mjnagel: Added this to the Authservice limitations section in the SSO Docs - let me know if you think this or other lines should be present in a different location as well.

src/pepr/operator/controllers/istio/namespace.ts
soltysh: >  But we'd have to account for each of the different owner types/controllers (to include CRs) I think if we went this alternative route of editing/triggering the controller rather than deleting the pod.

Yes, that's the risk I agree with, that's why I'm saying the current approach is good. At most we might want to consider eviction.

mjnagel: For now going to leave this as deletion to not include more changes in this PR and given the limited circumstances this happens in (first deploy when in sidecar mode or when switching from sidecar -> ambient). Also made a feature request to KFC (https://github.com/defenseunicorns/kubernetes-fluent-client/issues/606) that could make this more compelling in the future.

src/pepr/operator/controllers/network/policies.ts
mjnagel: Resolved in https://github.com/defenseunicorns/uds-core/pull/1386/commits/372169ee9130b4a0d798d80a11249bb94dc88a9b

src/test/chart/templates/package.yaml
```diff
@@ -7,6 +7,16 @@ metadata:
   name: podinfo
   namespace: podinfo
 spec:
+  network:
+    serviceMesh:
+      mode: ambient
```
mjnagel, line 12: Open question: Should we put more/different workloads into ambient mode? I was wondering about the curl test pods, shifting those to ambient to validate the networking tests/restrictions in ambient mode, but may want to wait until we separate those out into different namespaces?

noahpb: I agree that we should, but don't want to delay this PR. We can address that in https://github.com/defenseunicorns/uds-core/pull/1372 if you'd like.

chance-coleman: I agree with @noahpb, i think we should get this out and have a follow on issue for fine tuning testing.

src/pepr/operator/controllers/istio/namespace.ts
```diff
@@ -0,0 +1,277 @@
...
+        message:
+          "Ambient mode requested but Ambient is not deployed. Falling back to sidecar mode.",
+        reason: "AmbientUnavailable",
+        type: "Warning",
+      });
+
+      // Fall back to sidecar mode
+      annotations[pkgKey] = "true";
+      if (originalInjectionLabel !== "enabled") {
+        labels[INJECTION_LABEL] = "enabled";
```
chance-coleman: i might be missing something here, but it seems like there is duplicate code between the `enableIstio` and `cleanupNamespace` functions with the setting and check of labels/annotations. Don't feel super strongly about it but could be helper function.

src/test/chart/templates/package.yaml
mjnagel: Going to resolve this - we should switch over some of the other test apps (probably the curl ones) in the future. I think we will want to maintain a mix of ambient/sidecar in the `src/test` packages to ensure we cover both.

src/pepr/operator/controllers/istio/namespace.ts
mjnagel: Good callout - I'll look at this when I rebase.

mjnagel: https://github.com/defenseunicorns/uds-core/pull/1386/commits/d5f55ba279bbe75b15a8f0e7e579ddf39be50c90 significantly refactored this. A few highlights:
- Pulled logic for detecting current istio state into `getCurrentIstioState`
- Pulled the istio label selection/setting functionality into `getIstioLabels` which reads in current and target state to return the right label set + whether pods need to be restarted (this is most directly your suggestion)
- Pulled namespace updates into `applyNamespaceUpdates`
- Pulled conditional pod restarts into `restartPodsIfNeeded`
- Cleaned up a bunch of unnecessary vars as a result of ^

src/pepr/operator/controllers/istio/namespace.ts
```diff
@@ -0,0 +1,315 @@
...
+  if (hasOtherPackages) {
+    // Keep existing labels if other packages are still present, don't cycle pods/change istio state
+    result = { labels, shouldRestartPods: false };
+  } else {
+    // Set labels based on the original state and determine if pods need to be restarted
+    result = getIstioLabels(labels, originalIstioState, currentState);
+
+    // Delete the annotation since we're restoring to original state
+    delete annotations[ISTIO_STATE_ANNOTATION];
+  }
```
mjnagel, line 128: Highlighting this behavior - I think we want this to ensure that we don't break cleanup of packages when multiple are present. This logic would only remove the pkg annotation from the namespace if other packages are still present, not performing any label changes or pod restarts. This allows removing a second package, but does not support any complex state changes since we don't plan to support multiple package CRs in a single namespace going forward.

mjnagel:
After testing the `podinfo` workload further I realized there is an issue with Prometheus scraping ambient workloads currently. Evaluating if we can fix this in the mutation logic/scrape config or if something else is needed.

mjnagel:
> After testing the `podinfo` workload further I realized there is an issue with Prometheus scraping ambient workloads currently. Evaluating if we can fix this in the mutation logic/scrape config or if something else is needed.

Documenting this as a limitation for now.

chance-coleman (approved):
pending CI, LGTM

noahpb (approved):
questions addressed synchronously - LGTM!

PR #1397, opened 2025-03-28
fix!: move disable registration fields to theme values

chance-coleman:
## Description
With the new theme customization work, moving the environment variable for disabling registration fields into the new `themeCustomizations` block since it's only theme related.

## Related Issue

[Fixes this issue](https://github.com/defenseunicorns/uds-identity-config/issues/395)
[Related to this PR](https://github.com/defenseunicorns/uds-identity-config/pull/396)

## Type of change

## Steps to Validate
- Steps to test are in the [identity config PR](https://github.com/defenseunicorns/uds-identity-config/pull/396)

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
fix!: move disable registration fields to theme values (#1397)

BREAKING CHANGE: Theming configuration for removing additional registration fields has moved under the `themeCustomizations` values (`settings.enableRegistrationFields`). If overriding `DISABLE_REGISTRATION_FIELDS` under `realmInitEnv`, you will need to switch to this new value.
END_COMMIT_OVERRIDE


slaskawi (dismissed):

src/keycloak/chart/values.yaml
```diff
@@ -55,21 +54,22 @@ realmAuthFlows:
...
-#        configmap:
-#          name: keycloak-theme-overrides
-#      - name: logo.svg
-#        configmap:
-#          name: keycloak-theme-overrides
-#      - name: favicon.svg
-#        configmap:
-#          name: keycloak-theme-overrides
+themeCustomizations:
+  DISABLE_REGISTRATION_FIELDS: false
```
slaskawi: Just something to think of - perhaps it would be better to move this configuration param into a subcategory. For example:

```yaml
themeCustomizations:
  realm:
    settings:
      DISABLE_REGISTRATION_FIELDS: false
```

This way it's more extensible in the future.

chance-coleman: yeah i think this makes sense in some capacity. i do think the realm.settings is a little confusing when we also have realmInitEnv and realmAuthFlows. I'll see what i can come up with for this naming schema.

slaskawi: Yeah.... if you have an idea how to structure this better, we could cut a ticket and then bring the proposal to the Delivery. Once we reach consensus - we'd just move things around

mjnagel: My preference here is one of the following:

This would keep it relatively simple, but switch the key to camelCase which is more standard for helm values (this was `SCREAMING_SNAKE_CASE` historically because the key was directly used in the secret/env):
```yaml
themeCustomizations:
  disableRegistrationFields: false
```

For more future-proofing, something similar to what @slaskawi suggested could be done, which allows us to grow this section with other settings:
```yaml
themeCustomizations:
  settings:
    disableRegistrationFields: false
```

Two other thoughts/nits that may be worth considering if we're making this a breaking change already
1. I generally prefer `enable` type fields instead of `disable` (it makes the interpretation a bit easier to understand.
2. Is `RegistrationFields` the right name for this? Should it be `extraRegistrationFields` (not sure if it's clear enough/if docs and comments are enough to describe what this is).

Ideas based on that:
```yaml
themeCustomizations:
  # Simple
  enableRegistrationFields: true
  # Maybe more standard?
  extraRegistrationFields:
    enabled: true
    # In theory could add more fields here
  # All of the above option
  settings:
    extraRegistrationFields:
      enabled: true
```

chance-coleman: yeah i think changing this to be `enable` + name would be good. I like keeping this simple and one line with the ability to maintain future proof. So i think we go with :
```
themeCustomization:
  settings:
    enableRegistrationFields: true
```
will add a small doc section explaining this env var and also add a comment to the values.yaml here, should mitigate the need for the `extra` portion and avoid an overly long name

slaskawi (approved):

mjnagel (approved):

PR #1428, opened 2025-04-04
feat!: make istio ambient components default in uds core

chance-coleman:
## Description
Istio ambient components (CNI and Ztunnel) are currently present in uds-core but disabled by default.

This PR:
  - Makes istio ambient default in uds-core. 
  - Implement Gateway API CRDs (future requirement for keycloak work with ambient)
  - Remove references to `istio-ambient` and update docs section

Breaking change: used to have `istio-ambient` component and `istio-controlplane`, now only have `istio-controlplane`. The two have been squashed together to improve developer experience and simplify the bundle/package experience.

## Related Issue

Fixes #1280

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Other (security config, docs update, etc)

## Steps to Validate
- Verify that deploy uds-core standard package is successful and creates ambient workloads as well as new CRDs
- `uds run test-uds-core`
- `kubectl get crd -A` and validate these four crds are present:
          - gateway.networking.k8s gatewayclasses
          - gateway.networking.k8s httproutes
          - gateway.networking.k8s grpcroutes
          - gateway.networking.k8s referencegrants

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md) followed

BEGIN_COMMIT_OVERRIDE
feat!: make istio ambient components default in uds core (#1428)

BREAKING CHANGE: Istio Ambient workloads are now included by default with UDS Core. These workloads are now part of the `istio-controlplane` component (previously part of the optional `istio-ambient` component) - any override values/configuration should target this component instead of `istio-ambient`.
END_COMMIT_OVERRIDE

PR #1442, opened 2025-04-08
fix: register crds in `dev-setup` task

noahpb:
## Description
After https://github.com/defenseunicorns/uds-core/pull/1428 was introduced, I noticed that `uds run dev-setup` began to fail due to the `Exemptions` Custom Resource not yet being registered on the target cluster. This PR introduces an additional step to `dev-setup` that calls the `registerCRDs` function to create the `Exemptions` CRD.

## Type of change

## Steps to Validate
```
$ uds run dev-setup
```

## Checklist before merging

### DIFF
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -30,7 +30,7 @@ tasks:
         cmd: npx ts-node -e "import { registerCRDs } from './src/pepr/operator/crd/register'; registerCRDs()"
         env:
           - "PEPR_MODE=dev"
-        
+
       # Note: the `registry-url` flag used here requires uds 0.19.2+
       - description: "Deploy the Istio source package with Zarf Dev"
         cmd: "uds zarf dev deploy src/istio --flavor upstream --registry-url docker.io --no-progress"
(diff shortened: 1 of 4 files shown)


tasks.yaml
```diff
@@ -28,7 +28,7 @@ tasks:
 
       # Note: the `registry-url` flag used here requires uds 0.19.2+
       - description: "Deploy the Istio source package with Zarf Dev"
-        cmd: "uds zarf dev deploy src/istio --flavor upstream --registry-url docker.io --no-progress"
+        cmd: "uds zarf dev deploy src/istio --deploy-set CREATE_EXEMPTIONS=false --flavor upstream --registry-url docker.io --no-progress"
```
mjnagel: What do you think about running [this register command](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/tasks.yaml#L11-L14) before the dev deploy here instead? That will install the CRDs in the cluster, without pepr fully up yet, allowing for CRs to be created. It would avoid us needing to make the exemptions optional, just for this dev usecase.

noahpb: That seems like a simpler approach. Let's do that.

mjnagel (approved):

PR #1445, opened 2025-04-08
feat!: opt prometheus stack into ambient

### LINKED ISSUE #1423: Opt prometheus-stack into ambient
Following completion of https://github.com/defenseunicorns/uds-core/issues/1248, use the implemented toggle to opt the prometheus-stack package into ambient mode.

Ensure that all prometheus functionality still works and any end-to-end testing works as expected. Also re-evaluate any "istio workarounds" if present (peerauthentications and headless services in particular) and see if they can be removed. Also ensure that all interactions with other core services (if applicable) continue to function as expected.

NOTE: Prometheus will require some rework, potentially impacting Pepr code as well, due to the metrics scraping setup assuming certs from the sidecar. This should be simplified but will require testing across upgrade scenarios to maintain confidence. Relevant code:
- https://github.com/defenseunicorns/uds-core/blob/main/src/prometheus-stack/values/values.yaml#L30-L46
- https://github.com/defenseunicorns/uds-core/blob/main/src/prometheus-stack/values/values.yaml#L69-L75
- https://github.com/defenseunicorns/uds-core/blob/main/src/prometheus-stack/values/values.yaml#L111-L118
- https://github.com/defenseunicorns/uds-core/blob/main/src/prometheus-stack/common/zarf.yaml#L24-L31
- https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/prometheus/index.ts

Particular things to validate:
- all core applications are scraped as expected (validated in CI)
- all core applications are scraped as expected across an upgrade (validated in CI?)
- additional applications are scaped as expected across an upgrade of core (without the app being redeployed/changed)

Definition of done: prometheus-stack runs in ambient mode within core, and functions as expected.

chance-coleman:
## Description
Opt the prometheus-stack package into ambient mode.

Cleanup ambient directory and combine with the common directory. Operator changes to allow prometheus port in Deny and CUSTOM authorizationpolicies, also remove unnecessary operator pieces.

## Related Issue

Fixes #1423

## Type of change

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: opt prometheus stack into ambient (#1445)

BREAKING CHANGE: ServiceMonitors and PodMonitors no longer require TLS configuration for Istio, and may fail to scrape metrics if TLS configuration is present. The UDS Operator will handle removing this configuration from monitors in most cases, but may not update your monitor if TLS configuration was directly added separate from the Operator's mutations. In addition, the `istio-certs` and `exempt` scrape classes are no longer supplied as part of the Prometheus setup and should not be set on your monitoring resources going forward.
END_COMMIT_OVERRIDE


mjnagel (commented):
Changes look good at a glance. I think we need to update/cleanup some documentation:
- Limitations on ambient: https://github.com/defenseunicorns/uds-core/blob/main/docs/reference/configuration/UDS%20operator/package.md?plain=1#L36 - this limitation should be gone
- Metrics docs: https://github.com/defenseunicorns/uds-core/blob/main/docs/reference/configuration/uds-monitoring-metrics.md?plain=1#L49-L51 - I think this can also be deleted? Metrics-server has permissive mTLS but is scraping fine, not sure if we should try to see what happens with a completely un-injected app though?
- Dev docs: https://github.com/defenseunicorns/uds-core/blob/main/docs/dev/monitoring-setup.md - this is mostly irrellevant now, maybe could update to note this was the previous setup for sidecar and some words about the new setup (which is very simple)?

Those were the ones I could find quickly, not sure if we might have other references. I did also find a few places where we have the `skip-sm-mutate` annotation that we may want to evaluate if it's necessary anymore:
- https://github.com/defenseunicorns/uds-core/blob/d36014ddbfcb90844e0bf984b3fa5ab8b0e33c96/src/metrics-server/chart/templates/service-monitor.yaml#L10
- https://github.com/defenseunicorns/uds-core/blob/d36014ddbfcb90844e0bf984b3fa5ab8b0e33c96/src/prometheus-stack/chart/templates/istio-monitor.yaml#L11
- https://github.com/defenseunicorns/uds-core/blob/d36014ddbfcb90844e0bf984b3fa5ab8b0e33c96/src/prometheus-stack/chart/templates/prometheus-pod-monitor.yaml#L11

docs/dev/monitoring-setup.md
```diff
@@ -1,5 +1,25 @@
...
+## Limited L7 Enforcement:
+Since ztunnel does not process L7 attributes by default, HTTP headers and JWT validation must be handled via a waypoint (or bypassed), which differs from sidecar mode where the sidecar enforces these rules.
+
+## Operator Mutation:
+The Prometheus operator mutates ServiceMonitor/PodMonitor resources to remove TLS configuration and reset the scrapeClass when a target is meant to run in ambient mode.
+
+</details>
+
+<details>
+<summary>Istio Sidecar Mode</summary>
```
mjnagel:
```suggestion
<summary>Istio Sidecar Mode [LEGACY]</summary>

> [!NOTE]  
> This setup is no longer used now that Prometheus is running in ambient mode. This documentation is here for historical/learning purposes.
```

Or similar, just want to make sure it's not confusing since this is "Prometheus running in ambient vs sidecar" not the individual applications.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -261,6 +263,25 @@ export async function generateAuthorizationPolicies(
     }
   }
 
+  // In Sidecar mode, Prometheus lacks mTLS credentials, causing metric scrapes on port 15020 to fail.
+  // Add an AuthorizationPolicy to allow all traffic on port 15020 for the package's namespace.
```
mjnagel: `Prometheus lacks mTLS credentials` is potentially the opposite of the actual issue here. I don't know that I can explain it well, but Prometheus _does_ have mTLS credentials/identity (otherwise you couldn't use the principal in this ALLOW policy).

I believe the issue is essentially that when Prometheus is in ambient mode, ALL traffic goes through the ztunnel proxy, and subsequently gets evaluated by the destination sidecar and blocked without a matching authpol. Previously in sidecar mode Prometheus didn't send any traffic through its sidecar (technically) and the destination accepted it as plaintext because it was the sidecar metrics rather than app metrics.

```suggestion
  // With Prometheus in Ambient mode, all traffic is sent over mTLS and the destination sidecar requires an ALLOW policy to expose sidecar metrics.
  // Add an AuthorizationPolicy to allow all traffic on port 15020 for the package's namespace.
```

Feel free to adjust wording as you see fit.

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -261,6 +263,25 @@ export async function generateAuthorizationPolicies(
...
+  if (istioMode === IstioState.Sidecar) {
+    const extraPolicyName = sanitizeResourceName(
+      `protect-${pkgName}-ingress-15020-metric-scraping`,
```
mjnagel:
```suggestion
      `protect-${pkgName}-ingress-15020-sidecar-metric-scraping`,
```

src/pepr/prometheus/index.ts
```diff
@@ -25,114 +25,121 @@ export const prometheus = new Capability({
...
+    log.info(
+      `Patching PodMonitor ${pm.Raw.metadata?.name}: clearing scrapeClass, setting endpoints to HTTP, and removing TLS config.`,
+    );
+    delete pm.Raw.spec!.scrapeClass;
+    if (pm.Raw.spec?.podMetricsEndpoints && Array.isArray(pm.Raw.spec.podMetricsEndpoints)) {
+      pm.Raw.spec.podMetricsEndpoints.forEach((endpoint: PodMonitorEndpoint) => {
+        endpoint.scheme = PodMonitorScheme.HTTP;
+        if (endpoint.tlsConfig) {
+          delete endpoint.tlsConfig;
+        }
```
mjnagel:
```suggestion
    log.info(
      `Patching PodMonitor ${pm.Raw.metadata?.name}: clearing scrapeClass and setting endpoints to HTTP.`,
    );
    delete pm.Raw.spec!.scrapeClass;
    if (pm.Raw.spec?.podMetricsEndpoints && Array.isArray(pm.Raw.spec.podMetricsEndpoints)) {
      pm.Raw.spec.podMetricsEndpoints.forEach((endpoint: PodMonitorEndpoint) => {
        endpoint.scheme = PodMonitorScheme.HTTP;
```

While this is unlikely to cause issues, our "sidecar mutations" [never actually set the tlsConfig for podmonitors](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/prometheus/index.ts#L105-L116). If one is present, that means the end user specified it and it should probably stay as is.

mjnagel (dismissed):
LGTM overall, validated upgrading with GitLab in cluster and all metrics worked as expected. Two final comments:
1. This might still be considered a breaking change if/because end users could create service monitors "for istio" (tls config) on top of core (without mutation) and this switch would break those. Likely an edge case, but I found a few packages around our org where that was the case. I think our un-mutate might still handle them for the ones I found, but there could be a weird combo of ignore annotations and pre-configured TLS config that would get missed. Open to opinions on this, maybe more something to note in the description/announcement than noting as breaking?
1. We should be able to delete the [prom podmonitor](https://github.com/defenseunicorns/uds-core/blob/opt-prometheus-stack-into-ambient/src/prometheus-stack/chart/templates/prometheus-pod-monitor.yaml) at this point since it will be redundant with the self monitoring service monitor.

mjnagel (approved):

PR #1456, opened 2025-04-14
docs: fix Velero doc link path

bm54cloud:
## Description

Fixes broken link to Velero documentation, as the path was changed

## Type of change

## Steps to Validate

## Checklist before merging

### DIFF
--- a/src/velero/README.md
+++ b/src/velero/README.md
@@ -129,5 +129,5 @@ velero restore create uds-restore-$(date +%s) \
 > before running the restore.
 
 > [!NOTE]
-> Additional configuration will be required to get CSI backed PVCs to be snapshotted. For more information on configuring CSI backed volumes, see https://github.com/defenseunicorns/uds-core/blob/main/docs/reference/configuration/backup%20and%20restore/vsphere-rke2-csi.md
+> Additional configuration will be required to get CSI backed PVCs to be snapshotted. For more information on configuring CSI backed volumes, see https://uds.defenseunicorns.com/reference/configuration/backup-and-restore/vsphere-rke2-csi/
 
--- a/src/velero/README.md
+++ b/src/velero/README.md
@@ -129,5 +129,5 @@ velero restore create uds-restore-$(date +%s) \
 > before running the restore.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


src/velero/README.md
```diff
@@ -129,5 +129,5 @@ velero restore create uds-restore-$(date +%s) \
 > before running the restore.
 
 > [!NOTE]
-> Additional configuration will be required to get CSI backed PVCs to be snapshotted. For more information on configuring CSI backed volumes, see https://github.com/defenseunicorns/uds-core/tree/main/docs/reference/configuration/csi-driver.md
+> Additional configuration will be required to get CSI backed PVCs to be snapshotted. For more information on configuring CSI backed volumes, see https://github.com/defenseunicorns/uds-core/blob/main/docs/reference/configuration/backup%20and%20restore/vsphere-rke2-csi.md
```
mjnagel:
```suggestion
> Additional configuration will be required to get CSI backed PVCs to be snapshotted. For more information on configuring CSI backed volumes, see https://uds.defenseunicorns.com/reference/configuration/backup-and-restore/vsphere-rke2-csi/
```

In general I think we try to link the doc site for any docs that are located there.

bm54cloud: yep great idea. How would the link look if docs run on local host? Like my local link is http://localhost:4321/reference/configuration/backup-and-restore/vsphere-rke2-csi/ and not sure if it's the same port for everyone

bm54cloud: Or I could just say something like "Navigate to reference/configuration/backup-and-restore/vsphere-rke2-csi of the UDS doc site"

mjnagel: So since this doc itself (velero readme) is not on the docs site, I think we'd generally just want to point to the published docs site (rather than assuming they have the local site up). 

If this README were part of the docs site we could make it a local reference (`/reference/configuration/backup-and-restore/vsphere-rke2-csi/`) which would work for both the published site and local dev site - but for now I think just pointing to the published makes sense?

bm54cloud: yes makes sense. TBH, didn't realize this was all in the published docs site also. Refresh and see if the new URL looks like what you're thinking

bm54cloud: actually, let me just change it to the backup restore folder....because there will be more docs there in the future

mjnagel (approved):

chance-coleman (approved):

PR #1458, opened 2025-04-14
feat: opt metrics-server into ambient

### LINKED ISSUE #1281: Opt metrics-server into ambient
Following completion of https://github.com/defenseunicorns/uds-core/issues/1248, use the implemented toggle to opt the metrics-server layer into ambient mode.

Ensure that all metrics-server functionality still works and any end-to-end testing works as expected. Also re-evaluate any "istio workarounds" if present (peerauthentications and headless services in particular) and see if they can be removed. Also ensure that all interactions with other core services (if applicable) continue to function as expected.

Definition of done: Metrics Server runs in ambient mode within core, and functions as expected.

noahpb:
## Description
Opts metrics-server into istio ambient mode via the Package CR. Also adds a new e2e test for metrics-server to validate that node and pod metrics are being collected.

## Related Issue
Fixes #1281

## Type of change

## Steps to Validate
```
$ uds run test-single-layer --set FLAVOR=upstream --set LAYER=metrics-server
```

## Checklist before merging


src/metrics-server/chart/templates/peerauthentication/metrics-api.yaml
```diff
@@ -1,20 +0,0 @@
-# Copyright 2024 Defense Unicorns
-# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
-
-{{- if .Capabilities.APIVersions.Has "security.istio.io/v1beta1" }}
-apiVersion: security.istio.io/v1beta1
```
slaskawi: Could you please tell me if removing the `PeerAuthentication` has any impact on the security posture? If it lowers it, do we have a remediation plan?

test/jest/metrics-server.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+describe("Metrics Server", () => {
+    const apiServerAddress = getApiServerAddress();
+    const certs = getCertsFromKubeConfig() || {};
+    const metricsApi = "/apis/metrics.k8s.io/v1beta1";
+    const https = require('https');
+
+    if (!apiServerAddress) {
+        console.error("API server address not found.");
```
chance-coleman: I might be misunderstanding this, if the apiServerAddress or the certs are missing we should fail the tests right? if so i think it would make more sense to throw an error. It would "fail fast", preventing jest from silently skipping test logic and being a bit more loud in CI if there is a failure.

```suggestion
        if (!apiServerAddress) {
            throw new Error("API server address not found.");
        }
```

test/jest/metrics-server.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  caCert?: string;
+  cert?: string;
+  key?: string;
+}
+
+function getApiServerAddress(): string | undefined {
+    try {
+      const currentContext = kc.getCurrentContext();
+      const context = kc.getContextObject(currentContext);
+
```
chance-coleman: There is a bunch of extra whitespace in this file.

chance-coleman: i realize now this comment is kinda hard to see unless you have plugin or something highlights that extra white space, sorry 😬. All of the extra spaces are in the `getCertsFromKubeConfig` and the `getApiServerAddress` functions which i think you'd be able to see if you highlighted them in your IDE, i use the `Trailing Spaces` vscode plugin for this.

noahpb: TIL, thanks for that tip!

src/metrics-server/chart/templates/peerauthentication/metrics-api.yaml
noahpb: Good question. It's my understanding that this `PeerAuthentication` is no longer required as previously it allowed `PERMISSIVE` ingress over port `10250` because calls to metrics server originate from the nodes themselves (and thus, outside of the mesh). In my testing, I found that ambient allowed cluster components to communicate with metrics-server without this workaround. 

Strict mTLS is still [enforced by default](https://github.com/defenseunicorns/uds-core/blob/3423176db0dbbcc17ac9be27a135f7d16939be5a/src/istio/common/chart/templates/peer-authentication.yaml#L13) across the entire mesh via the default `PeerAuthentication` that we create in the `istio-system` namespace, so there is no risk with removing this.

test/jest/metrics-server.spec.ts
noahpb: Good callout! Updated to `throw` errors instead of logging them.

noahpb:
~~Moving to draft - this is causing issues on non-k3d clusters.~~

Fixed. We needed the `PeerAuthentication` for multi node clusters (this is not the default setup for our CI or local testing).

src/metrics-server/chart/templates/peerauthentication/metrics-api.yaml
slaskawi: Perfect! Thanks for explaining!

slaskawi (dismissed):
LGTM assuming that testing issues are resolved

src/metrics-server/chart/templates/peerauthentication/metrics-api.yaml
noahpb: I've reverted this change as the `PeerAuthentication` is required for multi node clusters.

chance-coleman (dismissed):

slaskawi (dismissed):

test/jest/metrics-server.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+        return undefined;
+      }
+      return cluster.server;
+    } catch (err) {
+      throw new Error('Error getting API server address:');
+      return undefined;
+    }
+  }
+
+function getCertsFromKubeConfig(): KubeConfigCerts | undefined {
```
mjnagel: Only concern with this is validating/testing that it works on EKS/RKE2 since we run these tests there as well.

noahpb: Good callout. We can switch to using `client-node` for auth and use that for fetching pod metrics.

noahpb: Refactored to use `client-node`. Tested locally and works against EKS.

chance-coleman (approved):

mjnagel (approved):

PR #1466, opened 2025-04-16
feat: opt Grafana into ambient

### LINKED ISSUE #1422: Opt grafana into ambient
Following completion of https://github.com/defenseunicorns/uds-core/issues/1248, use the implemented toggle to opt the grafana package into ambient mode.

Ensure that all grafana functionality still works and any end-to-end testing works as expected. Also re-evaluate any "istio workarounds" if present (peerauthentications and headless services in particular) and see if they can be removed. Also ensure that all interactions with other core services (if applicable) continue to function as expected.

Definition of done: grafana runs in ambient mode within core, and functions as expected.

slaskawi:
## Description

This Pull Request introduces Grafana Package into Ambient Mesh

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1422

## Type of change

## Steps to Validate

## Checklist before merging


test/jest/prometheus.spec.ts
```diff
@@ -26,18 +26,34 @@ describe("Prometheus and Alertmanager", () => {
...
+      if (response.status === 200) {
+        const body = (await response.json()) as {
+          alerts: Array<{ status: { state: string } }>;
+        }[];
+
+        if (body[0]?.alerts[0]?.status.state === "active") {
+          success = true;
+          break;
+        }
+      }
```
mjnagel: Is there any specific status we might be able to check to differentiate between "it's broken" versus "we're waiting for it to fire"? I suspect probably not but figured I'd ask - if it's actually broken it would be nice to fail execution early rather than waiting 7 mins.

test/jest/prometheus.spec.ts
```diff
@@ -26,18 +26,34 @@ describe("Prometheus and Alertmanager", () => {
...
-    const response = await fetch(
-      `${alertmanagerProxy.url}/api/v2/alerts/groups?filter=alertname%3D%22Watchdog%22&silenced=false&inhibited=false&active=true`,
-    );
-    expect(response.status).toBe(200);
+    // The Watchdog configuration uses group_interval: 5m and group_wait: 30s. This test might be executed too quickly
+    // to catch this. Therefore, we retry this test over the course of 6 mins (36 retries of 10s each). The absolute
+    // test timeout is set to 7 minutes to accommodate extra execution delays.
+    const maxRetries = 36;
+    const delay = 10000;
+    let success = false;
```
mjnagel: I'm in general okay with this, although I don't love potentially adding 7 mins to our CI. Given the config you mentioned, should this really only retry for ~1 min? If the initial wait is only 30s, I'm wondering if we'd have confidence that there should be a first alert sent through within that first group? But I may be misunderstanding these parameters and how this works.

I also wonder if there's a more direct way/different way to check prom/alertmanager connectivity alternatively here? Just wary of potentially adding extra CI time, although I recognize that these retries should prevent "false failures", actually saving time in the end ideally.

test/jest/prometheus.spec.ts
slaskawi: I simplified this test and hopefully it will be more stable now.

test/jest/prometheus.spec.ts
slaskawi: Changed the approach as we discussed during the Slack chat.

mjnagel (approved):
LGTM - tested an HA grafana setup (with external DB) as well and didn't observe any issues.

PR #1469, opened 2025-04-16
feat: keycloak fips mode

slaskawi:
## Description

This Pull Request introduces Keycloak FIPS mode into the UDS Core. 

## Related Issue

Relates to https://github.com/defenseunicorns/uds-identity-config/issues/36

## Type of change

## Steps to Validate

N/A

## Checklist before merging


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -137,6 +144,11 @@ spec:
               value: DEBUG
             - name: QUARKUS_LOG_CATEGORY__ORG_KEYCLOAK_SERVICES_X509__LEVEL
               value: TRACE
+            # Crypto information, primarily for FIPS debugging
+            - name: QUARKUS_LOG_CATEGORY__ORG_KEYCLOAK_COMMON_CRYPTO__LEVEL
+              value: TRACE
+            - name: QUARKUS_LOG_CATEGORY__ORG_KEYCLOAK_CRYPTO__LEVEL
+              value: TRACE
```
mjnagel, line 151: Since the log level is being edited here figured I'd ask - should these all be defaults of TRACE? Or should they be DEBUG/INFO with a helm value to override? I know @ntwkninja has mentioned the noisiness of Keycloak logs before so just thinking about optimizations here for the default deployment.

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -178,12 +190,6 @@ spec:
                 secretKeyRef:
                   name: {{ include "keycloak.fullname" . }}-postgresql
                   key: password
-            # FIPS Mode
-            {{- if .Values.fips }}
-            # https://access.redhat.com/documentation/en-us/openjdk/11/html-single/configuring_openjdk_11_on_rhel_with_fips/index
-            - name: JAVA_TOOL_OPTIONS
-              value: "-Dcom.redhat.fips=true"
-            {{- end }}
```
mjnagel, line 193: Pending CI testing I think this is fine, just would want to make sure that dropping this/changing the effect of `fips` value doesn't have any backwards incompatible effects on FIPS or non-fips hosts.

src/keycloak/chart/templates/statefulset.yaml
slaskawi: I ran a testing scenario (with Keycloak running for a few minutes and deleting a Package CR) and managed to pull these figures:

- Total log lines: 12921
- The `org.keycloak.crypto` log lines: 4
- The `org.keycloak.common` log lines: 5

Regarding to the number of logs, we should be on the safe side. At the same time, these log lines contain some vital debugging information, in particular:

```
{"timestamp":"2025-04-23T18:15:55.820625845Z","sequence":13326,"loggerClassName":"org.jboss.logging.Logger","loggerName":"org.keycloak.common.crypto.CryptoIntegration","level":"TRACE","message":"Java security providers: [ \n KC(BCFIPS version 2.0 Approved Mode, FIPS-JVM: disabled) version 1.0 - class org.keycloak.crypto.fips.KeycloakFipsSecurityProvider, \n BCFIPS version 2.0 - class org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider, \n BCJSSE version 2.0019 - class org.bouncycastle.jsse.provider.BouncyCastleJsseProvider, \n SUN version 21 - class sun.security.provider.Sun, \n SunRsaSign version 21 - class sun.security.rsa.SunRsaSign, \n SunEC version 21 - class sun.security.ec.SunEC, \n SunJSSE version 21 - class sun.security.ssl.SunJSSE, \n SunJCE version 21 - class com.sun.crypto.provider.SunJCE, \n SunJGSS version 21 - class sun.security.jgss.SunProvider, \n SunSASL version 21 - class com.sun.security.sasl.Provider, \n XMLDSig version 21 - class org.jcp.xml.dsig.internal.dom.XMLDSigRI, \n SunPCSC version 21 - class sun.security.smartcardio.SunPCSC, \n JdkLDAP version 21 - class sun.security.provider.certpath.ldap.JdkLDAP, \n JdkSASL version 21 - class com.sun.security.sasl.gsskerb.JdkSASL, \n SunPKCS11 version 21 - class sun.security.pkcs11.SunPKCS11, \n]","threadName":"main","threadId":1,"mdc":{},"ndc":"","hostName":"keycloak-0","processName":"/usr/lib/jvm/java-21-openjdk-21.0.6.0.7-1.el9.aarch64/bin/java","processId":1}
```

The above line shows Bouncy Castle FIPS initialization in strict mode, which might be something we could write a test against (to validate if Keycloak really booted up in FIPS mode).

src/keycloak/chart/templates/statefulset.yaml
slaskawi: It took me a while but I think I have the answer!

The Red Hat Build of OpenJDK contains OpenJDK bits with an additional FIPS patch. You can download the source from the Red Hat Customer Portal (from [here](https://access.redhat.com/jbossnetwork/restricted/softwareDetail.html?softwareId=107772&product=core.service.openjdk&version=21.0.7&downloadType=distributions)). 

The code haven't changed much since JDK11, it looks very similar to [this](https://github.com/fedora-riscv/java-11-openjdk-portable/blob/9a3dc7ec9e57d5998996026eb012fbd626ec6abb/fips-11u-f93a863b56.patch#L490-L504):

```java
+        if (systemSecPropsEnabled) {
+            boolean shouldEnable;
+            String sysProp = System.getProperty("com.redhat.fips");
+            if (sysProp == null) {
+                shouldEnable = true;
+                if (sdebug != null) {
+                    sdebug.println("com.redhat.fips unset, using default value of true");
+                }
+            } else {
+                shouldEnable = Boolean.valueOf(sysProp);
+                if (sdebug != null) {
+                    sdebug.println("com.redhat.fips set, using its value " + shouldEnable);
+                }
+            }
+            if (shouldEnable) {
+                boolean fipsEnabled = SystemConfigurator.configureFIPS(props);
+                if (sdebug != null) {
+                    if (fipsEnabled) {
+                        sdebug.println("FIPS mode support configured and enabled.");
+                    } else {
+                        sdebug.println("FIPS mode support disabled.");
+                    }
```

Note, that in the absence of `com.redhat.fips`, the OpenJDK (at least the Red Hat's build of it) tries to enable the FIPS mode by default. The `SystemConfigurator.configureFIPS` is a JNI (Java Native Interface - calling C from Java). 

The other important thing from our angle is the `SystemConfigurator.enableFips` method that has the following [JavaDoc](https://github.com/fedora-riscv/java-11-openjdk-portable/blob/9a3dc7ec9e57d5998996026eb012fbd626ec6abb/fips-11u-f93a863b56.patch#L732C1-L746C9):

```java
+     * OpenJDK FIPS mode will be enabled only if the system is in
+     * FIPS mode.
+     *
+     * Calls to this method only occur if the system property
+     * com.redhat.fips is not set to false.
+     *
+     * There are 2 possible ways in which OpenJDK detects that the system
+     * is in FIPS mode: 1) if the NSS SECMOD_GetSystemFIPSEnabled API is
+     * available at OpenJDK's built-time, it is called; 2) otherwise, the
+     * /proc/sys/crypto/fips_enabled file is read.
```

The NSS is shipped as part of SSSD, which is part of the RHEL authentication stack and is present in the Keycloak image. According to [BZ852023](https://bugzilla.redhat.com/show_bug.cgi?id=852023), NSS brokers to `/proc/sys/crypto/fips_enabled`, which is usually what happens in container world (if the host OS uses FIPS mode it sets `/proc/sys/crypto/fips_enabled` to all containers it runs based on [BZ1957310](https://bugzilla.redhat.com/show_bug.cgi?id=1957310)). 

To sum it up - I believe it's safe to remove this property and let the Red Hat Build of OpenJDK do the default thing - try to enable FIPS. This will succeed if the underlying Host has FIPS turned on regardless to the distro (although UBI are expected to be more tested than anything else).

chance-coleman (commented):
just a general questions regarding fips, is there any kind of tests that we should implement to "show/prove" that fips is enabled and working? I don't know what this would look like..but using tests, where we can, to show proof of compliance seems like a good idea to me.

slaskawi:
> just a general questions regarding fips, is there any kind of tests that we should implement to "show/prove" that fips is enabled and working? I don't know what this would look like..but using tests, where we can, to show proof of compliance seems like a good idea to me.

@UnicornChance Once this (and it's UDS Identity Config counterpart) gets merged, I plan to modify `pull-request-conditionals.yaml` and introduce "keycloak_fips" parameter to the testing matrix. This way we'll be testing all flavors with and without Keycloak FIPS (cc @mjnagel)

As for the test itself, I have three ideas. One is to scan the logs and ensure we emit the `KC(BCFIPS version 2.0 Approved Mode` string. This one is emitted when Keycloak finishes FIPS configuration. The other one is to check supported certs and cryptographic keys. FIPS installations should use an approved subset of these. The final idea is to use the Keycloak Admin API and get Credentials for the Admin user. The default "argon2" shall not be used there (as it's not approved).

chance-coleman (approved):

chance-coleman:
@mjnagel and I were chatting about the testing approach. Adding fips/non-fips to the matrix in uds-core will mean at least 6 new jobs that need to run in our already super long testing pipeline. Micah suggested, and i agree, we should implement a fips/non-fips test in identity-config rather than core for this as it shouldn't impact other uds-core integrations. Especially if we make FIPS default.

PR #1472, opened 2025-04-17
feat: opt logging into ambient

### LINKED ISSUE #1396: Opt logging into ambient
Following completion of https://github.com/defenseunicorns/uds-core/issues/1248, use the implemented toggle to opt the logging layer (loki and vector) into ambient mode.

Ensure that all logging functionality still works (may want to test with external sinks as well like s3?) and any end-to-end testing works as expected. Also re-evaluate any "istio workarounds" if present (peerauthentications and headless services in particular) and see if they can be removed. Also ensure that all interactions with other core services (if applicable) continue to function as expected.

Definition of done: Logging runs in ambient mode within core, and functions as expected.

chance-coleman:
## Description
opt the logging layer (loki and vector) into ambient mode.

Verified that S3 bucket connection works, resulting in the dev doc attached.

## Related Issue

Fixes #1396 

## Type of change

## Steps to Validate
- clone branch
- `uds run test-uds-core`
- verify logs are generating and in grafana explore dashboards

## Checklist before merging


src/vector/values/values.yaml
```diff
@@ -4,6 +4,10 @@
 # Run as an agent daemonset
 role: "Agent"
 
+# Disable headless service for Ambient ServiceMesh
+serviceHeadless:
+  enabled: false
+
```
mjnagel: I tried to search docs/prs/issues and didn't see anything obvious on the intent of this headless service but worth noting that it's not one we created (i.e. for istio sidecar workaround purposes) and is enabled by default upstream. If we're not seeing any issues with disabling it then no objections to this change just wanted to call it out.

chance-coleman: Yeah totally fair, from what i could tell it was unused, no real reason to take it out either. I'm fine with either approach.

noahpb (approved):
Manually validated against a multi-node cluster (AKS) locally and all works as expected. LGTM. Thanks for the local dev writeup here too!

mjnagel (approved):

mjnagel (commented):
Also validated with Vector shipping to another sink (s3/minio) and all seemed to work as expected.

PR #1476, opened 2025-04-18
chore(ci): trigger iac workflows via label

### LINKED ISSUE #1495: Trigger IaC test workflows via PR label
Add the ability to trigger the EKS, RKE2 and AKS workflows by adding an optional PR label to an open PR. This would allow us to trigger e2e tests for this distros on demand - a functionality that we currently lack today.

noahpb:
## Description
This PR introduces new functionality to our GitHub CI Workflows as well as refactoring their structure. Notable changes and new features include:
- Introduces the ability to trigger IaC distro testing for a PR by the addition of a label
     - Supports triggering a distro-specific workflow via the following labels:
       - `test-aks`
       - `test-rke2`
       - `test-eks`
     - The below label triggers workflows for all IaC distros:
       - `test-all-iac`
- Switches all IaC workflows to be triggered via `workflow_call` and `schedule` only
     -  This also means that the `Lint` workflow must succeed before they are dispatched. Previously, IaC workflows would run regardless of the result of the `Lint` workflow.
- Prevents any files pertaining to IaC testing (`.tf` files and `.github/bundles`) from triggering the `all` packages filter. This allows devs to make changes to IaC without running unnecessary CI runs against UDS Core itself.
- New steps in the `Filter` workflow called `Schedule IaC <distro>` are responsible for calling IaC workflows. Workflows can be triggered either by the presence of a label (outlined above), by one of its related files being modified by the PR (defined [here](https://github.com/defenseunicorns/uds-core/blob/9fda9a38e026897a10acd7565be1bf5bebcec08e/.github/workflows/pull-request-conditionals.yaml#L101)), or both.

See an example of the new Workflow format in action [here](https://github.com/defenseunicorns/uds-core/actions/runs/15119493411?pr=1476).

Examples of workflows being triggered:
- [EKS](https://github.com/defenseunicorns/uds-core/actions/runs/15119493411/job/42498339185?pr=1476)
- [RKE2](https://github.com/defenseunicorns/uds-core/actions/runs/15119493411/job/42498339266?pr=1476)
- [AKS](https://github.com/defenseunicorns/uds-core/actions/runs/15119493411/job/42498339345?pr=1476)

## Related Issue
Fixes https://github.com/defenseunicorns/uds-core/issues/1495

## Type of change

## Steps to Validate

## Checklist before merging


noahpb:
note: still need to evaluate how we want status to be reported back for PRs where IaC related files are being modified.

edit: decided to add a new step that monitors the IaC workflows that were triggered and wait for them all to complete. While the output needs some adjustments, an example of this in action can be seen [here](https://github.com/defenseunicorns/uds-core/actions/runs/14764098521/job/41451623667).

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -104,6 +151,167 @@ jobs:
...
 
+  # Trigger IaC
+  run-iac-test:
+    needs: check-paths
+    name: Schedule IaC
+    runs-on: ubuntu-latest
+    if: ${{ needs.check-paths.outputs.distros != '[]' || contains(github.event.pull_request.labels.*.name, 'test-all-iac') || contains(github.event.pull_request.labels.*.name, 'test-aks') || contains(github.event.pull_request.labels.*.name, 'test-rke2') || contains(github.event.pull_request.labels.*.name, 'test-eks') }}
+    steps:
+      - name: Dispatch IaC Workflows
+        uses: actions/github-script@v7
```
noahpb: todo: pin to sha

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -104,6 +151,167 @@ jobs:
...
+        env:
+          DISTROS_MODIFIED: ${{ needs.check-paths.outputs.distros }}
+          HAS_ALL_DISTROS_LABEL: ${{ contains(github.event.pull_request.labels.*.name, 'test-all-iac') }}
+          HAS_AKS_LABEL: ${{ contains(github.event.pull_request.labels.*.name, 'test-aks') }}
+          HAS_RKE2_LABEL: ${{ contains(github.event.pull_request.labels.*.name, 'test-rke2') }}
+          HAS_EKS_LABEL: ${{ contains(github.event.pull_request.labels.*.name, 'test-eks') }}
+        with:
+          script: |
```
noahpb: to make this cleaner, we can reference a file here instead of running the script inline

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -104,6 +151,167 @@ jobs:
       test_type: ${{ matrix.test_type }}
     secrets: inherit # Inherits all secrets from the parent workflow.
 
+  # Trigger IaC
+  run-iac-test:
+    needs: check-paths
```
noahpb, line 157: right now the code loops over the iac distros to test. if we prefer, we can make this task a matrix instead

.github/workflows/test-aks.yaml
```diff
@@ -68,10 +67,11 @@ jobs:
       - name: Azure login
         uses: azure/login@a457da9ea143d694b1b9c7c869ebb04ebe844ef5 # v2
         with:
-          client-id: ${{ secrets.AZURE_GOV_CLIENT_ID }}
-          tenant-id: ${{ secrets.AZURE_GOV_TENANT_ID }}
-          subscription-id: ${{ secrets.AZURE_GOV_SUBSCRIPTION_ID }}
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
```
noahpb: had to switch our auth method for Azure (see [note](https://github.com/defenseunicorns/uds-core/issues/1495#issuecomment-2828799036))

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -104,6 +151,167 @@ jobs:
...
+                core.setFailed(`Workflow dispatch failed for ${distro}: ${error.message}`);
+              }
+            }
+
+  check-iac-test:
+    needs: [check-paths, run-iac-test]
+    name: Wait for IaC
+    runs-on: ubuntu-latest
+    # Only wait for status when the PR directly modifies IaC workflow related files
+    if: ${{ needs.check-paths.outputs.distros != '[]' }}
```
noahpb: right now this doesn't run if workflows are triggered via label. revisiting now, i think that is something we would likely want

.github/workflows/pull-request-conditionals.yaml
mjnagel: I'm probably missing something here, but why would we not just have this use `uses: ./.github/workflows/test-<distro>.yaml` + `workflow_call`? The script/`workflow_dispatch` feels overly complex here especially since `worklow_call` + `uses` is how we already run our other PR conditional tests.

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -70,20 +72,65 @@ jobs:
...
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+            all:
+              - "**"
+              - "!.github/workflows/test-aks.yaml"
+              - "!.github/workflows/test-eks.yaml"
+              - "!.github/workflows/test-rke2.yaml"
+              - "!.github/bundles/**"
```
mjnagel, line 87: I think this would cause some issues with our required checks unless I'm misunderstanding? If someone changes aks/eks/rke2 files it would run the IAC testing but not the required k3d install/upgrade tests?

.github/workflows/pull-request-conditionals.yaml
noahpb: We definitely could take that approach. However, I opted for this option as I would keep the IaC workflows "separate"/not nested underneath a single workflow like the package tests (I couldn't find a way to get the workflows to run on their own, but I could be missing something!). That's a small convenience that comes with the complexity of these scripts. Open to either route here really.

mjnagel: I might have to think more about it, but given the simplicity of the `workflow_call` compared to scripting out the trigger + wait, I'd lean that way if all we're gaining is a different visibility. I think your findings on not being able to run the workflows on their own makes sense, to "conditionally" trigger them we have to call them from another spot I think. We _could_ always "run" them, but conditionally execute the actual testing (`echo "Skipping testing"` unless the diffs or labels required a test) but that might just clutter up the PR status with more workflows "running" even when they aren't "truly running".

With the `workflow_call` approach I think we'd end up with:
- All/most PR workflows (including IAC) showing up under ["Filter"](https://github.com/defenseunicorns/uds-core/actions/workflows/pull-request-conditionals.yaml)
- Nightly/weekly IAC testing still showing up under the individual AKS/EKS/RKE2 tabs (since those are triggered via cron)
- Potentially wouldn't need the Azure cred changes which might be a plus?

That's my slight leaning to simplify this but open to other opinions.

.github/workflows/pull-request-conditionals.yaml
noahpb: That's correct. IMO I don't think we should run the required checks for k3d install/upgrade if we're not directly changing anything related to core apps or the k3d testing setup. I can adjust this if you prefer that we always run this - open to either!

.github/workflows/pull-request-conditionals.yaml
noahpb: Moving to `workflow_call` instead. Thanks for the thoughts!

.github/workflows/pull-request-conditionals.yaml
mjnagel: I don't have any objections to this (largely agree with you, if we're just changing IAC we might as well skip those). We would however need to shim out those workflows in these cases (such as how we do the docs shim) otherwise we'll be unable to merge PRs since the required checks won't have run.

.github/workflows/iac-shim.yaml
```diff
@@ -0,0 +1,34 @@
...
+  pull_request:
+    # milestoned is added here as a workaround for release-please not triggering PR workflows (PRs should be added to a milestone to trigger the workflow).
+    # labeled is added to support renovate-ready labelling on PRs
+    types: [milestoned, labeled, opened, reopened, synchronize]
+    paths:
+      - ".github/workflows/test-aks.yaml"
+      - ".github/workflows/test-eks.yaml"
+      - ".github/workflows/test-rke2.yaml"
+      - ".github/bundles/**"
+      - "!**"
```
mjnagel: Won't this negate everything, causing this to never run? [Ref docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) - `A matching negative pattern (prefixed with !) after a positive match will exclude the path.`

noahpb: Fixed! Thanks for the docs link, you're right. I think I just need the exact paths here.

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -70,20 +71,65 @@ jobs:
...
```
mjnagel, line 87: Should this also include `!.github/test-infra/**` and `!tasks/iac.yaml`? (if so also would apply to the shim in the reverse)

mjnagel (approved):

PR #1479, opened 2025-04-21
fix: remove dynamic client registration

### LINKED ISSUE #1388: Istio Ambient: Envoyfilter updates
Envoyfilters are not supported for pods in ambient mode / on waypoints (see [upstream issue](https://www.github.com/istio/istio/issues/43720) and the [decision to not implement](https://www.github.com/istio/istio/issues/43720#issuecomment-2740803785)).

In preliminary testing envoyfilters will still work when applied to the `GATEWAY` context (for example, our existing classification banner works on ambient workloads). The only envoyfilter currently applied on SIDECAR_INBOUND is [Keycloak's `block-path-parameters-in-non-final-segments` envoyfilter](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/templates/path-parameter-envoyfilter.yaml).

Note that switching to the envoyfilter being applied at the gateway may limit our ability to force this same restriction on in-cluster traffic and we should be careful to ensure that there is still adequate protection in-cluster, whether it is via an envoyfilter or other mechanism (potentially could be done in Keycloak).

Definition of done: The envoyfilter `block-path-parameters-in-non-final-segments` is updated to work as expected when Keycloak is running in ambient mode. Additional protections are enforced as needed to ensure this protection on our authorization policies is still in place.

### LINKED ISSUE #1499: Operator: Remove Keycloak DCR support
Now that we support client credentials in the Keycloak setup (for operator client registration), we should remove any operator code for dynamic client registration (since we have deprecated it already). In particular:
- operator code/types/config variables supporting this method
- authorization policy updates (pepr doesn't need access to this endpoint)

Note: We should also ensure that on the Keycloak side the DCR endpoint is still "protected" or inaccessible if we are changing istio envoyfilter/authorization policy logic.

Definition of done: DCR support is removed from UDS Core/Keycloak, endpoint for DCR is still "secure".

chance-coleman:
## Description
Keycloak dynamic client registration strategy is being deprecated in this release and this PR aims to remove the necessary pieces associated to that strategy and update documentation.

## Related Issue

Fixes #1499 

## Type of change

## Checklist before merging


src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -36,18 +35,6 @@ spec:
...
-            ports:
-              - "8080"
-            paths:
-              # Never allow anonymous client registration except from the pepr-system namespace
-              # This is another fallback protection, as the KC policy already blocks it
-              - "/realms/{{ .Values.realm }}/clients-registrations/*"
-      from:
-        - source:
-            notNamespaces:
-              - "pepr-system"
```
mjnagel, line 51: Do we need an upgrade/migration on the realm/identity-config side to make sure this endpoint is no longer accessible/turned on?

src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -19,7 +19,6 @@ spec:
               - "8080"
             paths:
               - "/admin*"
-              - "/realms/master*"
```
mjnagel: Was this removal intentional? I assume we'd still want to only expose the master realm via the admin gateway?

src/keycloak/chart/templates/istio-admin.yaml
slaskawi: No, the migration should happen within UDS Core v0.39.0, which pulled the v0.11.0. Here's the upgrade docs: https://uds.defenseunicorns.com/reference/uds-core/idam/upgrading-versions/#v0110

src/keycloak/chart/templates/istio-admin.yaml
slaskawi: Yeah, you are probably right. I think I removed too much!

src/keycloak/chart/templates/istio-admin.yaml
mjnagel: I knew client credentials got _enabled_ in that migration but I missed that the dynamic endpoint/access was also _disabled/removed_. I did validate this locally, when I try to hit the dynamic endpoint I end up getting this log/error: `{"error":"Client registration provider not found"}`, although I'm still a bit unsure where/how that changed (this is mostly curiosity, now that I have validated it's not accessible anymore).

slaskawi: This is very weird... 

It basically means that all `ClientRegistrationProvider` implementations have been dropped from the Realm. There should be at least 2 of them - the default and the OIDC one. 

This error definitely should not happen and it leads to a question if your UDS Realm has been properly initiated?

slaskawi:
LGTM assuming @mjnagel hit some weird upgrade issue and not a real problem.

src/keycloak/chart/templates/istio-admin.yaml
slaskawi: I'm resolving this as we clarified in Slack that `{"error":"Client registration provider not found"}` error comes from the wrong URL used. The URL needs to end with "default" as we're testing the "default" Dynamic Client Registration API.

slaskawi (dismissed):

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -19,16 +19,6 @@ spec:
...
-      - description: "UDS Operator"
-        direction: Ingress
-        selector:
-          app.kubernetes.io/name: keycloak
-        remoteNamespace: pepr-system
-        remoteSelector:
-          app: pepr-uds-core-watcher
-        remoteServiceAccount: pepr-uds-core
-        port: 8080
-
```
mjnagel, line 33: I think this is fine, since its covered by backchannel access (<insert always has been meme>). But just to call it out - Pepr still needs this ingress regardless of client registration method, it just is/has been covered by a different netpol.

src/pepr/operator/README.md
```diff
@@ -155,19 +155,13 @@ spec:
           bearer_only: clientField(bearerOnly)
 ```
 
-### Controlling how UDS Operator interacts with Keycloak
+### How UDS Operator interacts with Keycloak
 
-The UDS Operator can interact with Keycloak in two primary ways: using the Dynamic Client Registration or Client Credentials Grant. The method used is determined automatically or by specifying the environment variable `PEPR_KEYCLOAK_CLIENT_STRATEGY`.
+The UDS Operator authenticates with Keycloak using the OAuth 2.0 Client Credentials Grant. It uses the OAuth 2.0 Client Credentials Grant to authenticate against the `uds-operator` client defined in Keycloak. This special client has a limited control over managing Keycloak Clients for the UDS Operator.
```
mjnagel, line 160:
```suggestion
The UDS Operator authenticates with Keycloak using the OAuth 2.0 Client Credentials Grant. It authenticates against the `uds-operator` client defined in Keycloak, which has limited control to manage Keycloak Clients for UDS.
```

This wording felt weird/redundant. Not sure if my suggestion is better, just tried to simplify a bit?

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -80,7 +80,7 @@ export async function purgeSSOClients(pkg: UDSPackage, newClients: string[] = []
   const toRemove = currentClients.filter(client => !newClients.includes(client));
   for (const ref of toRemove) {
     try {
-      await deleteClient({ clientId: ref });
+      await credentialsDelete({ clientId: ref });
```
mjnagel, line 83: With the removal of the dynamic approach we may also want to add a "store cleanup". We did this in the past when we [migrated exemptions out of the store](https://github.com/defenseunicorns/uds-core/blob/v0.22.2/pepr.ts#L31-L47) just to clear things up for end users and not leave strange state lingering around.

If we do make this change we should:
- bump our operator version and check for the version when we do the migration
- ensure we are careful to only delete keys that exist - I'm not sure but we may be able to safely use `clear()` since we don't use the store for anything else now?
- careful test an upgrade where someone was using dynamic registration, to make sure everything goes smoothly (shouldn't be a problem since client-creds can manage the clients without those tokens, just would want to validate)

Okay making this a follow-on if we want, just want to keep it in mind.

mjnagel (commented):
LGTM overall - we may want to get the [identity-config change](https://github.com/defenseunicorns/uds-identity-config/pull/441) across the line first since that ensures we're still "securing" the dynamic endpoint.

src/pepr/operator/controllers/keycloak/client-sync.ts
chance-coleman: going to push this out, having some trouble getting this working as expected. created this [issue](https://github.com/defenseunicorns/uds-core/issues/1524) so that we can track and prioritize.

mjnagel (commented):
LGTM pending identity config release + merge into core.

mjnagel (approved):

PR #1482, opened 2025-04-22
feat: client credential registration default

### LINKED ISSUE #1388: Istio Ambient: Envoyfilter updates

chance-coleman:
## Description
In moving uds-core to ambient we need to rework how we implement the envoyfilter used to protect path parameters. In addition to that, in the coming release we will phase out the use of the dynamic client registration in favor client credential registration. This will improve security and decrease complexity when working with istio ambient.

This PR will focus on the following:
1. Change the default keycloak client registration strategy from `auto` to `client_credentials`

## Related Issue

Fixes #1388 

## Type of change

## Checklist before merging


slaskawi (dismissed):

src/keycloak/chart/templates/path-parameter-envoyfilter.yaml
```diff
@@ -23,17 +27,22 @@ spec:
...
-                  local invalid_semicolon = string.match(path, "^[^;]*;[^/]*/")
-                  if invalid_semicolon then
-                    request_handle:respond(
-                      { [":status"] = "400" },
-                      "Path parameters may only appear after the final path segment"
-                    )
+                -- Only apply to keycloak endpoints
+                if host and (
+                  string.match(host, "^sso%.uds%.dev$") or
+                  string.match(host, "^keycloak%.admin%.uds%.dev$")
```
mjnagel: This doesn't take the end-user's domain into account, so it would only work on our dev clusters. I think we could also have this just do a basic equality check with `==` rather than the match here where we have to make it more like a regex (may make it easier when pulling in the end-user domain config).

src/keycloak/chart/templates/path-parameter-envoyfilter.yaml
```diff
@@ -23,17 +27,22 @@ spec:
...
+                      request_handle:respond(
+                        { [":status"] = "400" },
+                        "Path parameters may only appear after the final path segment"
+                      )
+                    end
                   end
                 end
               end
-{{- end }}
+{{- end }}
```
mjnagel: Should we wait to move this to the gateway until https://github.com/defenseunicorns/uds-core/pull/1479, or even until we switch Keycloak to ambient? Until then we're just reducing the protection slightly here, when dynamic client cred could still be in use (even if deprecated).

src/keycloak/chart/templates/path-parameter-envoyfilter.yaml
chance-coleman: Yep good point. Updating to use the `domain` and `adminDomain` values.

src/keycloak/chart/templates/path-parameter-envoyfilter.yaml
chance-coleman: hmm yeah thats good callout..given that it's hopefully in the next release and already have a [PR](https://github.com/defenseunicorns/uds-core/pull/1479) up for removing dynamic client registration, i think it makes sense to move the envoyFilter change to Gateway to that PR.

src/keycloak/chart/templates/path-parameter-envoyfilter.yaml
```diff
@@ -8,12 +8,16 @@ apiVersion: networking.istio.io/v1alpha3
 kind: EnvoyFilter
 metadata:
   name: block-path-parameters-in-non-final-segments
-  namespace: {{ .Release.Namespace }}
+  namespace: istio-system
```
noahpb: Validated locally that on upgrade, helm will cleanup the old resource in the previous namespace.

src/keycloak/chart/templates/path-parameter-envoyfilter.yaml
chance-coleman: chatted in person, going to move envoyFilter changes out of this PR into the follow on PR when we completely remove the dynamic client registration.

mjnagel (approved):

mjnagel:
LGTM - can probably update the PR title/description to reflect that there are no envoyfilter changes in this PR anymore.

slaskawi (approved):

PR #1488, opened 2025-04-23
fix(ci): renovate readiness version loop fix

mjnagel:
## Description

Fixes a bug in the logic where "missing images" were also parsed, causing all flavor waits to get added to PRs missing one flavor update.

## Related Issue

N/A

## Type of change

## Steps to Validate

Jest tests updated to catch this and explicitly assert on ONLY the expected labels.

## Checklist before merging

### DIFF
--- a/.github/workflows/pull-request-conditionals.yaml
+++ b/.github/workflows/pull-request-conditionals.yaml
@@ -24,6 +24,7 @@ on:
       - release-please-config.json
       - CODEOWNERS
       - LICENSE
+      - scripts/** # scripts/hacks that are used specifically for non-testing workflows
 
 # Permissions for the GITHUB_TOKEN used by the workflow.
 permissions:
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request-conditionals.yaml
```diff
@@ -24,6 +24,7 @@ on:
       - release-please-config.json
       - CODEOWNERS
       - LICENSE
+      - scripts/** # scripts/hacks that are used specifically for non-testing workflows
```
slaskawi, line 27: Are you sure this won't block merges due to lacking required CI checks?

slaskawi (approved):
LGTM with one doubts that probably requires confirming

.github/workflows/pull-request-conditionals.yaml
mjnagel: The shim will run the required checks (triggered [here](https://github.com/defenseunicorns/uds-core/blob/renovate-readiness-fix/.github/workflows/docs-shim.yaml#L30)) - I actually missed this change previously so that ONLY the shim runs the required checks. The shim has [stubbed out](https://github.com/defenseunicorns/uds-core/blob/renovate-readiness-fix/.github/workflows/test-shim.yaml) testing that matches the required checks signatures, without actually running the full CI/taking a lot of time.

If this doesn't work as expected, we can always fix on the next PR of this code, but it should match the other ignored files for docs, etc now.

PR #1498, opened 2025-04-24
feat: opt neuvector into ambient

### LINKED ISSUE #1421: Opt runtime-security into ambient
Following completion of https://github.com/defenseunicorns/uds-core/issues/1248, use the implemented toggle to opt the runtime-security layer (neuvector) into ambient mode.

Ensure that all runtime-security functionality still works and any end-to-end testing works as expected. Also re-evaluate any "istio workarounds" if present (peerauthentications and headless services in particular) and see if they can be removed. Also ensure that all interactions with other core services (if applicable) continue to function as expected.

Definition of done: runtime-security runs in ambient mode within core, and functions as expected.

mjnagel:
## Description

Switches NeuVector to ambient mode, with supporting changes:
- Removes extra headless services (no longer necessary)
- Adds waypoint + mutation to ensure local auth is still blocked
- Adds network allow policy to ensure waypoint works (egress to istiod and ingress from gateway)
- Adds manager probes from upstream
- Removes enforcer and controller probe mutations - the previous probes were not doing anything due to the way [TCP probes work with sidecars](https://istio.io/latest/docs/ops/configuration/mesh/app-health-check/)
- Adds before/after deploy actions to handle upgrade scenarios and ensure that probes are set to the upstream defaults

NOTE: This PR switches all flavors to use a "self-managed" certificate as we have seen issues using the upstream cert job in some changes. In order to allow this transition we need to scale down the controller pre-upgrade (otherwise there will be conflicting certs used).

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1421

## Type of change

## Steps to Validate

CI tests should cover most relevant functionality for NeuVector.

## Checklist before merging


mjnagel:
May need some upstream changes to our images to fix some issues with healthprobes on the controller. Registry1 has a [fix in progress](https://repo1.dso.mil/dsop/neuvector/neuvector/controller/-/merge_requests/84) that uses a different base image where the `cat` process/path aligns with the upstream image. Waiting on Cgr for a unicorn image update/fix.

.gitignore
```diff
@@ -20,6 +20,7 @@ tmp-tasks.yaml
 cacert.b64
 run/
 extract-terraform.sh
+**/.terraform/**
```
mjnagel, line 23: Just to note this helped ensure that the license check task passed when I had local `.terraform` files present, otherwise I was hitting errors on those not having the license headers.

chance-coleman (approved):
everything is working as expected for me locally

noahpb (approved):
Validated by running an upgrade against AKS

Issue #1139, opened 2024-12-19
coredns scrape target down in hardened rke2


mjnagel:
### Environment
Device and OS: AWS Ec2
App version: 0.33.1 (snapshot)
Kubernetes distro being used: Rke2

### Steps to reproduce

Deploy an RKE2 cluster using one of the CIS profiles. When using these profiles RKE2 applies some default network policies that [block Ingress to kube-system](https://docs.rke2.io/security/hardening_guide#network-policies) for most services. As a result the service monitor that should monitor core-dns will appear as down.

### Expected result

Prometheus is able to scrape the configured targets in any environment.

### Actual Result

Prometheus cannot scrape coredns.

### Severity/Priority

Low - this may be environment specific and there is a clear workaround.

### Additional Context

I was able to workaround this by applying the following network policy:
```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-prometheus-to-kube-dns
  namespace: kube-system
spec:
  podSelector:
    matchLabels:
      k8s-app: kube-dns
  policyTypes:
    - Ingress
  ingress:
    - from:
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: monitoring
          podSelector:
            matchLabels:
              app: prometheus
      ports:
        - protocol: TCP
          port: 9153
```

This policy may belong in uds-core, behind a toggle? May warrant more conversation on whether this is a common problem or specific to hardened RKE2.

mjnagel:
Noting that we have a "fix" for this in our RKE2 CI (cc @noahpb) - we may just want to conditionally add this to our prometheus-stack chart.

PR #1501, opened 2025-04-25
feat: add conditional netpol for coredns

noahpb:
## Description
Adds a new conditional helm value to the `prometheus-stack` helm chart: `addRke2CorednsNetpol.enabled`. Disabled by default.

## Fixes
https://github.com/defenseunicorns/uds-core/issues/1139

## Type of change

## Steps to Validate
Validate against an RKE2 cluster

## Checklist before merging


src/prometheus-stack/chart/templates/coredns-netpol.yaml
```diff
@@ -0,0 +1,27 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+{{ if .Values.addCorednsNetpol }}
```
mjnagel: I think this might have been my initial suggestion, but is it better to name this `coreDnsNetpol.enabled` and/or tie it to RKE2 in the name?

Also might be nice to use this in our RKE2 testing instead of the one-off netpol we have now (this would ensure it's tested in CI too).

src/prometheus-stack/chart/templates/coredns-netpol.yaml
```diff
@@ -0,0 +1,26 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+{{- if (eq .Values.addRke2CorednsNetpol.enabled true) }}
```
mjnagel: nit: with helm we can just directly use the conditional unless I'm missing something special here?

```suggestion
{{- if .Values.addRke2CorednsNetpol.enabled }}
```

If you do make this change I might advocate for `rke2CorednsNetpol.enabled` (dropping the `add` as well, since it feels redundant with enabled).

tasks/iac.yaml
```diff
@@ -65,9 +65,9 @@ tasks:
               break
             fi
           done
-      - task: util:rke2-allow-prom-kube-dns
-        dir: .github/test-infra/aws/rke2/
-        maxTotalSeconds: 600
+      # - task: util:rke2-allow-prom-kube-dns
+      #   dir: .github/test-infra/aws/rke2/
+      #   maxTotalSeconds: 600
```
mjnagel: Any reason to leave this and the `rke2-allow-prom-kube-dns` task around, or can we clean those up?

mjnagel (dismissed):

tasks/iac.yaml
noahpb: Yeah let's clean that up

src/prometheus-stack/chart/templates/coredns-netpol.yaml
noahpb: I like this suggestion, I'll push up that change

mjnagel (approved):

PR #1504, opened 2025-04-28
feat: keycloak ambient opt in

### LINKED ISSUE #1424: Opt keycloak into ambient
Following completion of https://github.com/defenseunicorns/uds-core/issues/1248, use the implemented toggle to opt the keycloak package into ambient mode.

Ensure that all Keycloak functionality still works and any end-to-end testing works as expected. Also re-evaluate any "istio workarounds" if present (peerauthentications and headless services in particular) and see if they can be removed. Also ensure that all interactions with other core services (if applicable) continue to function as expected. Pay particular attention to the Pepr/operator interaction with Keycloak and validate it in both modes (dynamic and client-cred).

NOTE: Our current [L7 authpolicy](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/templates/istio-admin.yaml) will require a waypoint. This can be deployed as part of the UDS chart for Keycloak.

Definition of done: Keycloak runs in ambient mode within core, and functions as expected.

slaskawi:
## Description

Keycloak Ambient Opt In

## Related Issue

Fixes #1424

## Type of change

## Steps to Validate

1. Please check all unit and integration tests with `uds run test-single-layer --set FLAVOR=upstream --set LAYER=monitoring`
2. Next, run `uds run test-uds-core-ha` and check if Keycloak clustering works with PostgreSQL

## Checklist before merging


src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -29,6 +31,26 @@ spec:
         remoteServiceAccount: pepr-uds-core
         port: 8080
 
+      - description: "Waypoint health"
```
slaskawi, line 34: Note to reviewers: Without this, Prometheus fails to scrape metrics from Waypoint and it shows this target as failed.

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -71,8 +133,45 @@ spec:
...
+          - 15008
+          - 8080
+      - description: "Waypoint to Keycloak"
+        direction: Egress
+        remoteGenerated: IntraNamespace
+        ports:
+          - 15008
+          - 8080
+
+      - description: "Waypoint to Istio Control Plane"
```
slaskawi, line 99: Note to reviewers: Without this, the Waypoint fails to start. It needs to reach out to the Control Plane.

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -73,6 +93,28 @@ spec:
         {{- end }}
       {{- end }}
 
+      - description: "Keycloak to Waypoint"
```
slaskawi, line 88: Note to reviewers: These two seem to be the case for scraping metrics. Apparently Waypoint forwards the call to Keycloak.

src/keycloak/chart/templates/istio-admin.yaml
```diff
@@ -8,9 +8,10 @@ metadata:
   name: keycloak-block-admin-access-from-public-gateway
   namespace: {{ .Release.Namespace }}
 spec:
-  selector:
-    matchLabels:
-      {{- include "keycloak.selectorLabels" . | nindent 6 }}
+  targetRefs:
+    - kind: Service
+      group: ""
+      name: "keycloak-http"
```
mjnagel: Mostly noting for myself - it seems like [this is valid](https://istio.io/latest/docs/reference/config/security/authorization-policy/#AuthorizationPolicy) although I think I've seen more examples of targetting the Waypoint itself directly, rather than the service. I assume the difference here is more important if you have a single waypoint for a namespace, with multiple services - this would give you the granularity to target only a single service, rather than the full namespace.

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -37,6 +49,14 @@ spec:
...
         port: 8080
 
+      - description: "Keycloak ambient backchannel access"
+        direction: Ingress
+        selector:
+          app.kubernetes.io/name: keycloak
+        # Allow access from anything in cluster using an empty namespace selector
+        remoteNamespace: "*"
+        port: 15008
+
```
mjnagel: The operator should automatically add 15008 to the policy right above this, see [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/policies.ts#L154-L177) (so this shouldn't be necessary).

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -73,6 +93,28 @@ spec:
...
+        remoteGenerated: IntraNamespace
+        ports:
+          - 15008
+          - 8080
+      - description: "Waypoint to Keycloak"
+        direction: Egress
+        remoteGenerated: IntraNamespace
+        ports:
+          - 15008
+          - 8080
```
mjnagel: Similar comment on the ports here - we shouldn't need 15008 since the operator will add that ([ref](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/policies.ts#L154-L177)).

src/keycloak/chart/templates/waypoint.yaml
```diff
@@ -0,0 +1,19 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+{{- if .Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1" }}
+apiVersion: gateway.networking.k8s.io/v1
+kind: Gateway
+metadata:
+  name: waypoint
```
mjnagel: Definitely a nit but it might be worth naming this more specifically, `keycloak-waypoint` or something similar?

src/pepr/operator/index.ts
```diff
@@ -103,6 +103,14 @@ When(UDSPackage)
...
   });
 
+// See https://istio.io/latest/docs/ambient/usage/waypoint/#useawaypoint
+When(a.Namespace)
+  .IsCreatedOrUpdated()
+  .WithName("keycloak")
+  .Mutate(req => {
+    req.SetLabel("istio.io/use-waypoint", "waypoint");
+  });
+
```
mjnagel: Would it be better to label the keycloak service(s) and/or pod(s) ([ref](https://istio.io/latest/docs/ambient/usage/waypoint/#configure-a-service-to-use-a-specific-waypoint)) which we have control of in the helm chart?

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -73,6 +93,28 @@ spec:
         {{- end }}
       {{- end }}
 
+      - description: "Keycloak to Waypoint"
+        direction: Ingress
+        remoteGenerated: IntraNamespace
+        ports:
+          - 15008
```
UnicornChance, line 100: the 15008 port is [automatically added](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/policies.ts#L154) to netpols now, so we don't need to define that.

applicable to `keycloak to waypoint` and `waypoint to keycloak` netpols

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -37,6 +49,14 @@ spec:
         remoteNamespace: "*"
         port: 8080
 
+      - description: "Keycloak ambient backchannel access"
```
UnicornChance, line 52: I don't think this netpol is required. After pepr process the `Keycloak backchannel access` netpol it would include the 15008 port which is the only difference between this netpol and that one.

src/pepr/operator/index.ts
mjnagel: Noting as an update here - namespace-wide waypoint is certainly viable so just comes down to whether we'd rather set this up across the full namespace (using a mutate) or just for the pod/svc (using helm values). I tend to slightly lean towards the second since we only have the single Keycloak sts/svc but either is technically possible.

src/keycloak/chart/templates/uds-package.yaml
slaskawi: Fixed. Thanks for spotting this.

src/keycloak/chart/templates/uds-package.yaml
slaskawi: Fixed. Thanks for spotting this.

src/keycloak/chart/templates/waypoint.yaml
slaskawi: Fixed. Thanks for spotting this.

src/pepr/operator/index.ts
slaskawi: Annotating the Service seems to do the trick. I switched to it.

src/keycloak/chart/templates/istio-admin.yaml
slaskawi: This seems to work fine. Thanks for suggestion!

chance-coleman:
I think this looks good, not approving so that Micah and Noah can review before it gets auto-merged.

noahpb (approved):

chance-coleman (approved):

PR #1509, opened 2025-04-29
feat: pod anti-affinity for istiod, troubleshooting doc for webhook issue

### LINKED ISSUE #342: Pepr/Istiod Webhook chicken and egg
If nodes are killed without proper drain the pepr and istiod webhooks failing keeps any pods from scheduling to the cluster. istiod can't schedule from the pepr webhook failing and pepr can't schedule from the istiod webhook failing.

### Steps to reproduce
1. Have UDS core deployed across multiple nodes
2. Add new nodes
3. Kill old nodes that have workloads running without draining/cordoning

### Expected result

Pods, albeit slowly, schedule onto new nodes

### Actual Result

It's impossible for any pods to schedule onto the cluster going forward

### Severity/Priority

Med/High

### Additional Context

We were able to work around this by giving istiod as many replicas as there are nodes and a preferred podAntiAffinity so that istiod does it's best to be scheduled onto all nodes, but that doesn't feel like a proper solution.

chance-coleman:
## Description
When a node that contains both pepr and istiod is restarted, there can be circular dependencies between their webhooks on startup, resulting in failed state for pepr. This PR introduces pod anti-affinity to pepr and istiod so that if available (multi-node clusters) the pods will be spaced out across nodes. When this is done, the circular dependency is not an issue.

This is a soft affinity approach, meaning if there is not multiple nodes, the scheduling will still happen on the single node. 

- Also included is some new tasks for deploying a mutli-node cluster.
- A new trouble shooting doc for how to manually restart pepr and istiod if they are in a failed state.
- A dev doc for using multi-node clusters.

## Related Issue

Relates to #342

## Type of change

## Steps to Validate
There is a few things to test out in this PR:
- generally speaking using [this comment](https://github.com/defenseunicorns/uds-core/issues/342#issuecomment-2828571227) from the issue you can simulate a failed pepr/istiod reconciliation on a single node cluster
- then you can test the manual steps to get pepr and istiod healthy and happy again
- test out the dev doc for deploying uds-core on a multi-node cluster with the new task as well a k3d multi-node cluster only and adding layers/packages on top of that

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
chore: pod anti-affinity for istiod, troubleshooting doc for webhook issue (https://github.com/defenseunicorns/uds-core/pull/1509)
END_COMMIT_OVERRIDE


chance-coleman:
Important note, this PR doesn't completely fix this issue. Pepr is going to [implement an option eventually](https://github.com/defenseunicorns/pepr/issues/2077) to configure webhooks separately from watch webhooks so that we can ignore the istio-system for webhooks but not watch events.

bundles/k3d-slim-dev/uds-bundle.yaml
```diff
@@ -30,6 +30,9 @@ packages:
             - name: policies
               description: "Minio policies"
               path: policies
+            - name: K3D_EXTRA_ARGS
+              description: "Extra arguments for k3d cluster creation"
+              path: k3d_extra_args
```
mjnagel: I don't think this is correct/necessary? Where it's placed is under the `minio` chart so this wouldn't actually flow where you expect. Since `k3d_extra_args` is already exposed as a zarf var, we shouldn't need any override in the bundle itself.

src/pepr/values.yaml
```diff
@@ -7,9 +7,34 @@ watcher:
...
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 100
+          podAffinityTerm:
+            labelSelector:
+              matchExpressions:
+                - key: app.kubernetes.io/name
+                  operator: In
+                  values:
+                    - pepr
```
mjnagel: I'm not sure this is actually working the way we'd want it to? The label used here doesn't show up on the admission pods from what I can tell:
```yaml
  labels:
    app: pepr-uds-core
    pepr.dev/controller: admission
    pod-template-hash: f594d9cd4
    zarf-agent: patched
```

We probably want to key off of `app: pepr-uds-core`?

tasks/deploy.yaml
```diff
@@ -19,6 +19,11 @@ tasks:
       - description: "Deploy the UDS Core Standard Bundle"
         cmd: uds deploy bundles/k3d-standard/uds-bundle-k3d-core-demo-${UDS_ARCH}-${VERSION}.tar.zst --set INSECURE_ADMIN_PASSWORD_GENERATION=true --confirm --no-progress
 
+  - name: k3d-standard-bundle-multi-node
+    actions:
+      - description: "Deploy the UDS Core Standard Bundle"
+        cmd: uds deploy bundles/k3d-standard/uds-bundle-k3d-core-demo-${UDS_ARCH}-${VERSION}.tar.zst --set INSECURE_ADMIN_PASSWORD_GENERATION=true --set uds-k3d-dev.k3d_extra_args="--servers 3 --agents 2" --confirm --no-progress
```
mjnagel: Any objection to making `servers` and `agents` counts here inputs to the task (maybe in deploy and/or setup) so that someone could pass in `--with agents=1`, etc to customize the sizing a bit more?

tasks/setup.yaml
```diff
@@ -16,3 +16,9 @@ tasks:
       - description: "Initialize the cluster with Zarf"
         # renovate: datasource=github-tags depName=zarf-dev/zarf versioning=semver
         cmd: "uds zarf package deploy oci://ghcr.io/zarf-dev/packages/init:v0.53.0 --confirm --no-progress"
+
+  - name: create-k3d-cluster-multi-node
+    actions:
+      - description: "Create the multi-node K3d cluster"
+        # renovate: datasource=github-tags depName=defenseunicorns/uds-k3d versioning=semver
+        cmd: "uds zarf package deploy oci://defenseunicorns/uds-k3d:0.12.3 --confirm --no-progress --set k3d_extra_args='--servers 3 --agents 2'"
```
mjnagel: Kind of related to my [other comment](https://github.com/defenseunicorns/uds-core/pull/1509/files#r2066977275) - I think we could DRY this up by having an input to the existing `create-k3d-cluster` task that accepts `k3d_extra_args` or specific server/agent args?

noahpb (commented):
I think that adding this is a good idea. I think we could add another guarantee that this deadlock will be avoided by specifying an optional `podDisruptionBudget` for the Pepr admission pods. Distributing the workloads across different nodes ensures that the two webhooks would in theory never be taken offline at the same time in the event of a single node going offline, but specifying a PDB would ensure that the webhook is always available (would only be protected from voluntary evictions, such as during multi-node cluster upgrades).

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,88 @@
...
+During cluster restarts or after upgrades, Istiod and Pepr pods may fail to start properly due to a circular dependency between their admission webhooks.
+
+Both Istiod and Pepr depend on each other’s webhooks being available, which can cause a deadlock if one cannot admit the other’s workloads.
+
+This document explains how to manually resolve the deadlock.
+
+## Symptoms
+- Pepr pods are stuck in CrashLoopBackOff or Pending state.
+- Istiod pods are stuck in CrashLoopBackOff or Pending state.
+- Cluster logs show webhook admission errors like:
```
mjnagel: nit: I think this would be `events` not `logs` - likely showing on the replicasets?

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+## Why This Happens
+Both Pepr and Istiod register Kubernetes admission webhooks.
+When a cluster restarts, if the webhook targets (pods) aren’t available yet, admission fails, blocking the pods from being recreated — a chicken-and-egg problem.
+
+Temporarily relaxing webhook `failurePolicy` solves the startup deadlock.
+
+## Manual Recovery Procedure
+Temporarily disable the `failurePolicy` for the Pepr and Istiod webhooks, allowing pods to start.
```
mjnagel:
```suggestion
Temporarily modify the `failurePolicy` for the Pepr and Istiod webhooks, allowing pods to start.
```

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,88 @@
...
+
+```yaml
+failurePolicy: Ignore
+```
+Save and exit.
+
+### 2. Edit the Istiod Webhooks
+```bash
+kubectl edit mutatingwebhookconfiguration istio-sidecar-injector
+kubectl edit validatingwebhookconfiguration istio-validator-istio-system
```
mjnagel: I may be wrong, but I think you only need to edit the mutating webhook for Istio? Its validating webhook only targets Istio CRs - https://github.com/istio/istio/blob/master/manifests/charts/default/templates/validatingwebhook.yaml#L26-L38

If you do make this change, make sure to update wording, `webhooks` -> `webhook` on this step and `four webhooks` -> `three webhooks`.

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,88 @@
...
+
+Temporarily relaxing webhook `failurePolicy` solves the startup deadlock.
+
+## Manual Recovery Procedure
+Temporarily disable the `failurePolicy` for the Pepr and Istiod webhooks, allowing pods to start.
+
+### 1. Edit the Pepr Webhooks
+```bash
+kubectl edit mutatingwebhookconfiguration pepr-uds-core
+kubectl edit validatingwebhookconfiguration pepr-uds-core
```
mjnagel: Kind of a nit, but we should be able to provide `kubectl patch` commands for these rather than stepping through the edit interactively (more copy/paste).

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,88 @@
...
+
+### 4. Restore the Webhooks Back to Fail
+:::caution
+VERY IMPORTANT:
+After both Pepr and Istiod pods are running and healthy, revert your webhook configurations:
+Edit all four webhook configurations again:
+Set `failurePolicy: Fail` instead of `Ignore`.
+:::
+
+This restores strict security enforcement in admission control.
```
mjnagel, line 74: Is it also worth noting that they should take careful note of any new resources created while the webhooks are set to ignore as these could bypass istio injection and/or policy enforcement? The case where this could happen should be rare with the changes made in this PR, but it's still a definite risk to call out.

I think the alternative we discussed was excluding Istio/Pepr from each other's webhooks (without changing the policy). If that's something we could easily do with a patch command maybe it's worth trying? Also potentially should revisit if we need to change istio's webhook now that it is in ambient, or if certain label changes would resolve issues there (we may need to remove the istio injection label entirely?).

src/pepr/values.yaml
```diff
@@ -7,9 +7,34 @@ watcher:
...
+        - weight: 100
+          podAffinityTerm:
+            labelSelector:
+              matchExpressions:
+                - key: app
+                  operator: In
+                  values:
+                    - pepr-uds-core
+            topologyKey: kubernetes.io/hostname
+
```
mjnagel: This one is...fine I guess? But worth calling out this would schedule the watcher on a node that the admission pods are not on (watch has `app=pepr-uds-core-watcher` I believe). That's not necessarily required, but I don't think it hurts since this is just preferred?

mjnagel (commented):
Mostly docs comments - the changes here look good for the most part.

src/pepr/values.yaml
chance-coleman: also worth noting here, in talking to Case, they are implementing anti-affinity [on their side](https://github.com/defenseunicorns/pepr/pull/2083). So maybe we don't need to do this at all?

mjnagel: This was actually a good reminder to check the template...these values won't actually do anything today. If you look at the templates in the helm chart (or the source code [here](https://github.com/defenseunicorns/pepr/blob/main/src/lib/assets/helm.ts)) the values for `affinity` don't actually get plumbed through to the templates at all, despite being in the helm values. So for now we should definitely drop this since it doesn't have any effect - I went ahead and opened https://github.com/defenseunicorns/pepr/issues/2100 as well based on this.

docs/reference/troubleshooting/pepr-istiod-webhooks.md
chance-coleman: good idea, updated to use patches.

docs/reference/troubleshooting/pepr-istiod-webhooks.md
chance-coleman: validated this and you're right. only the mutating webhook was required.

docs/reference/troubleshooting/pepr-istiod-webhooks.md
chance-coleman: refactored the doc to use the patch commands that exclude istio/pepr from each others webhooks. remove the failure policy approach.

src/pepr/values.yaml
chance-coleman: refactored to remove this.

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+### 3. Restart the Pepr and Istiod Pods
+This isn't always required — typically the pods will retry and succeed once the webhooks have been patched.
+
+```bash
+kubectl rollout restart deployment pepr-uds-core -n pepr-system
+kubectl rollout restart deployment istiod -n istio-system
+```
+This forces Kubernetes to recreate them.
```
mjnagel: Not sure if it's important to callout but in my experience I saw pepr crash once (while istiod was still spinning up).

docs/reference/troubleshooting/pepr-istiod-webhooks.md
mjnagel: Resolving this one, see other comments.

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,116 @@
...
+```
+
+## Why This Happens
+Both Pepr and Istiod register Kubernetes admission webhooks.
+When a cluster restarts, if the webhook targets (pods) aren’t available yet, admission fails, blocking the pods from being recreated — a chicken-and-egg problem.
+
+## Manual Recovery Procedure
+Temporarily modify the Pepr mutating and validating webhooks, as well as the Istiod mutating webhook, to exclude each other:
+
+### 1. Patch the Pepr Webhooks to Exclude Itself and Istio
```
mjnagel, line 31: Comments on the commands and approach overall here:
- I don't think we actually need any edits to the istio webhooks now that pepr is in ambient mode? Worth validating this for yourself but I was able to get a cluster restored without modifying the istio webhooks.
- When modifying these with patches we need to be careful to not wipe out other config. For example, our default webhook config for pepr also ignores zarf and kube-system. End users could also ignore other namespaces beyond that.

Here's the commands I was able to test successfully with. The "cleanup" one isn't that pretty but it uses the tools we know they have (uds/kubectl) and makes sure we don't delete too much:
```console
# Setup
kubectl patch mutatingwebhookconfiguration pepr-uds-core --type='json' \
  -p='[{
    "op": "add",
    "path": "/webhooks/0/namespaceSelector/matchExpressions/0/values/-",
    "value": "istio-system"
  }]'

kubectl patch validatingwebhookconfiguration pepr-uds-core --type='json' \
  -p='[{
    "op": "add",
    "path": "/webhooks/0/namespaceSelector/matchExpressions/0/values/-",
    "value": "istio-system"
  }]'

# Cleanup
kubectl get mutatingwebhookconfiguration pepr-uds-core -o json | \
  uds zarf tools yq -p json -o json '.webhooks[0].namespaceSelector.matchExpressions[0].values |= map(select(. != "istio-system"))' | \
  kubectl apply -f -

kubectl get validatingwebhookconfiguration pepr-uds-core -o json | \
  uds zarf tools yq -p json -o json '.webhooks[0].namespaceSelector.matchExpressions[0].values |= map(select(. != "istio-system"))' | \
  kubectl apply -f -
```

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,116 @@
...
+kubectl patch validatingwebhookconfiguration pepr-uds-core --type='json' \
+  -p='[{"op": "remove", "path": "/webhooks/0/namespaceSelector"}]'
+
+kubectl patch mutatingwebhookconfiguration istio-sidecar-injector --type='json' \
+  -p='[{"op": "remove", "path": "/webhooks/0/namespaceSelector"}]'
+```
+This restores strict security enforcement in admission control.
+
+:::caution
+While the webhooks are excluding each other, any newly created resources (e.g., Pods) could bypass Istio sidecar injection or Pepr policy enforcement. This is rare but possible — identify and reapply configuration to any workloads started during this window.
```
mjnagel: Depending on the other comment might be able to rephrase this:
```suggestion
While the Pepr webhooks are excluding `istio-system`, any newly created resources in the `istio-system` namespace (e.g., pods/services) will bypass Pepr policy enforcement. It is important to audit any new/unexpected resources during this time.
```

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,116 @@
...
+:::
+
+## Alternative Ideas
+Rather than changing webhook failure policies, consider long-term approaches such as:
+
+- Adding a `namespaceSelector` to exclude the `istio-system` namespace from Pepr’s webhooks and vice versa.
+- Removing the Istio injection label from Pepr pods (especially if you're using Ambient Mesh).
+
+These strategies help prevent circular dependencies without relaxing webhook enforcement.
```
mjnagel: I think with the changes you made this might need to be updated since we're not changing the failure policy and the other two bullets are likely the exact approach we're taking here?

chance-coleman: removing altogether

docs/reference/troubleshooting/pepr-istiod-webhooks.md
chance-coleman: reworded line 79 to call out istiod being the determining piece for pods to retry and succeed.

docs/reference/troubleshooting/pepr-istiod-webhooks.md
noahpb: For cleanup, we could also do a `helm rollback` against the latest revision to essentially re-apply the latest release of the operator or do a `helm get manifest` + some `yq` selections to `kubectly apply` just the `MutatingWebhookConfiguration` and `ValidatingWebhookConfiguration` objects. FWIW, the cleanup commands above did work for me.

docs/dev/multi-node-cluster.md
```diff
@@ -0,0 +1,59 @@
...
+- k3d-uds-server-0, k3d-uds-server-1, k3d-uds-server-2 (control-plane nodes)
+- k3d-uds-agent-0, k3d-uds-agent-1 (worker nodes)
+
+Ideal for HA simulation and scaling tests.
+
+## Notes
+- The multi-node cluster is still a local k3d environment.
+- Networking is provided by Flannel.
+- No node taints are applied — all nodes are schedulable.
+- UDS Core packages will automatically distribute workloads across available nodes based on their scheduling rules.
```
noahpb: super nit: can we reword this to "UDS Core packages will automatically have their workloads distributed across the cluster by `kube-scheduler` based on their scheduling rules.

docs/reference/troubleshooting/pepr-istiod-webhooks.md
```diff
@@ -0,0 +1,78 @@
...
+Failed to call webhook: error calling webhook "pepr-uds-core"
+Failed to call webhook: error calling webhook "istiod-istio-system"
+```
+
+## Why This Happens
+Both Pepr and Istiod register Kubernetes admission webhooks.
+When a cluster restarts, if the webhook targets (pods) aren’t available yet, admission fails, blocking the pods from being recreated — a chicken-and-egg problem.
+
+## Manual Recovery Procedure
+Temporarily modify the Pepr mutating and validating webhooks, as well as the Istiod mutating webhook, to exclude each other:
```
mjnagel:
```suggestion
Temporarily modify the Pepr mutating and validating webhooks to exclude Istio:
```

mjnagel (dismissed):

mjnagel (approved):

Issue #728, opened 2024-09-05
HA configuration of bundle as part of nightly testing


mjnagel:
Related to https://github.com/defenseunicorns/uds-core/issues/213 , https://github.com/defenseunicorns/uds-core/issues/715

As part of our nightly testing we should configure our bundle with HA/prod ready configurations. This aligns with the general goal of nightly testing, being more focused on prod-like testing on full infrastructure deployments. A few pieces should be in scope here:
- External dependencies where possible: S3 for Loki/Velero, RDS for Keycloak, Elasticache for Authservice
- Multi-replica setup where possible (Loki/Keycloak are noted gaps currently)

[This doc](https://github.com/defenseunicorns/uds-core/blob/main/docs/configuration/resource-configuration-and-ha.md) may be a helpful reference for some of the suggested config to test with.

Out of scope for the initial pass:
- Affinity/anti-affinity
- ?

Definition of done: Nightly bundle(s)/configs (currently located [here](https://github.com/defenseunicorns/uds-core/tree/main/.github/bundles)) updated to deploy the required external dependencies and multiple replicas.

mjnagel:
Related to https://github.com/defenseunicorns/uds-core/issues/1188

mjnagel:
Closing as the first iteration of this is complete - HA bundles are being tested nightly.

PR #1518, opened 2025-05-01
feat!: enable Keycloak FIPS mode by default

### LINKED ISSUE #36: FIPS Support
#### At this time this is more of a discussion placeholder than an actual intent to implement.

FIPS is not supported at the moment. IF we want to move to supporting FIPS in our identity config we will need to spend some time understanding the different bouncy-castle libraries. This primarily impacts the [x509 plugin source code](https://github.com/defenseunicorns/uds-identity-config/tree/main/src/plugin), but will probably affect the other pieces as well.

At the moment we utilize the `bcpg-fips`  library, but based on bouncy-castle documentation that also requires at least a library for FIPS algorithms and potentially also using the `bc-fips` general FIPS library that isn't specific to OpenPGP.

Current bouncy-castle and Keycloak crypto libraries:
```      
        <dependency>
            <groupId>org.keycloak</groupId>
            <artifactId>keycloak-crypto-default</artifactId>
            <version>${keycloak.version}</version>
        </dependency>
        <dependency>
            <groupId>org.bouncycastle</groupId>
            <artifactId>bcpg-fips</artifactId>
            <version>1.0.7.1</version>
        </dependency>
```

Libraries that should be looked into:
```
        <dependency>
            <groupId>org.keycloak</groupId>
            <artifactId>keycloak-crypto-fips1402</artifactId>
            <version>${keycloak.version}</version>
        </dependency>
        <dependency>
            <groupId>org.bouncycastle</groupId>
            <artifactId>bc-fips</artifactId>
            <version>1.0.2.4</version>
        </dependency> 

```

- All testing should pass (integration tests, plugin unit tests, etc). 
- The provided identity config should be FIPS compliant, regardless of the surrounding environment.

### Testing Gotcha's
- While testing changes to libraries be aware that sometimes the order of the libraries in the pom.xml can cause errors. @rjferguson21 and @UnicornChance  experienced this behavior when importing the `keycloak-crypto-defualt` library after the `bcpg-fips` library. 

### Links
[Official Keycloak FIPS140-2 Docs](https://www.keycloak.org/server/fips)
[Maven Repo description of bcpg-fips](https://mvnrepository.com/artifact/org.bouncycastle/bcpg-fips)

### LINKED ISSUE #1549: Discuss the cadence of the nightly job and ensure we cover Keycloak clustering
## Summary

In the face of recent clustering failures, we need to agree what should be included in UDS Core nightly jobs and when to run them

slaskawi:
## Description

This Pull Request enables Keycloak FIPS mode by default.

In addition, this Pull Request introduces the RKE2 tests with Keycloak HA testing

## Related Issue

Fixes https://github.com/defenseunicorns/uds-identity-config/issues/36
Fixes https://github.com/defenseunicorns/uds-core/issues/1549
Relates to https://github.com/defenseunicorns/uds-core/issues/728

## Type of change

## Steps to Validate

Tested automatically 

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: enable Keycloak FIPS mode by default (#1518)

BREAKING CHANGE: UDS Core now uses Keycloak in FIPS (STRICT) mode by default (the `fips` Helm Chart flag is set to `true` by default). In some environments, this may be a breaking change that could result in the Keycloak Administrator account being locked out. Before upgrading, please ensure you have read and followed the [UDS Identity v0.14.0 upgrade guide](https://uds.defenseunicorns.com/reference/uds-core/idam/upgrading-versions/#v0140).
END_COMMIT_OVERRIDE

### DIFF
--- a/.github/workflows/test-nightly.yaml
+++ b/.github/workflows/test-nightly.yaml
@@ -1,0 +1,50 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+name: Test Nightly
+
+on:
+  schedule:
+    - cron: "0 0 * * *" # Every night at Midnight (UTC)
+  pull_request:
+    # milestoned is added here as a workaround for release-please not triggering PR workflows (PRs should be added to a milestone to trigger the workflow).
+    # labeled is added to support renovate-ready labelling on PRs
+    types: [milestoned, labeled, opened, reopened, synchronize]
+    paths:
+      - src/keycloak/**
+      - src/grafana/**
+
+permissions:
+  contents: read
+  id-token: write # This is needed for OIDC federation.
+  packages: read # Allows reading the published GHCR packages
+
+jobs:
+  test-nightly:
+    runs-on: uds-ubuntu-big-boy-8-core
+    timeout-minutes: 40
+    name: Test Nightly
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Environment setup
+        uses: ./.github/actions/setup
+        with:
+          registry1Username: ${{ secrets.IRON_BANK_ROBOT_USERNAME }}
+          registry1Password: ${{ secrets.IRON_BANK_ROBOT_PASSWORD }}
+          ghToken: ${{ secrets.GITHUB_TOKEN }}
+          chainguardIdentity: ${{ secrets.CHAINGUARD_IDENTITY }}
+
+      - name: Test UDS Core Upgrade
+        run: uds run test-uds-core-ha-upgrade --set FLAVOR=upstream --no-progress
+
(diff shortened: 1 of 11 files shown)


.github/workflows/test-nightly.yaml
```diff
@@ -0,0 +1,50 @@
+# Copyright 2025 Defense Unicorns
```
mjnagel, line 1: This workflow and corresponding task changes may overlap/conflict with https://github.com/defenseunicorns/uds-core/pull/1578 - okay merging either but maybe better to keep that separate rather than including in this PR focused on FIPS?

cc @UnicornChance - however we want to deconflict is fine, just for awareness.

mjnagel (commented):
LGTM overall pending CI, just the comment on the k3d testing and the parallel PR. If we do want to press with the k3d upgrade workflow in this PR, may further review here.

.github/workflows/test-nightly.yaml
slaskawi: Absolutely! Let me remove this from my PR

chance-coleman (approved):

PR #1522, opened 2025-05-03
feat: configurable tls redirect

JeffResc:
## Description
In some use cases (such as an AWS ALB), it is not possible to route traffic via https to the correct gateway server as the SNI information has been stripped from the connection resulting in `NR filter_chain_not_found`.

This PR allows the user to disable the http to https redirect, allowing use of http for inbound traffic. By default, the redirect will be enabled and will only be disabled by a server-level override.

## Type of change

## Steps to Validate

## Checklist before merging


noahpb (approved):

chance-coleman (approved):
LGTM, do wonder if this should be documented in some way for others. Also can create a backlog issue for doing that, and maybe a test to.

PR #1535, opened 2025-05-07
fix: add support for client name in pepr types

mjnagel:
## Description

Name has been supported in the SSO spec, but was missing from our client types in the pepr code.

## Related Issue

N/A

## Type of change

## Steps to Validate

Deploy with a package CR using sso.name and validate that the name is populated in Keycloak.

## Checklist before merging

### DIFF
--- a/src/grafana/chart/templates/uds-package.yaml
+++ b/src/grafana/chart/templates/uds-package.yaml
@@ -8,7 +8,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   sso:
-    - name: Grafana Dashboard
+    - name: Grafana
       clientId: uds-core-admin-grafana
       redirectUris:
         {{- if .Values.adminDomain }}
--- a/src/neuvector/chart/templates/uds-package.yaml
+++ b/src/neuvector/chart/templates/uds-package.yaml
@@ -19,7 +19,7 @@ spec:
   #     description: "Metrics"
 
   sso:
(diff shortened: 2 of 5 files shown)


src/grafana/chart/templates/uds-package.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   sso:
-    - name: Grafana Dashboard
+    - name: Grafana
```
mjnagel, line 11: Updated this because with this change this will now show on our login page and just Grafana seems better?

src/neuvector/chart/templates/uds-package.yaml
```diff
@@ -19,7 +19,7 @@ spec:
   #     description: "Metrics"
 
   sso:
-    - name: Neuvector
+    - name: NeuVector
```
mjnagel, line 22: Updated this to match the proper spelling upstream (to ensure the login page looks accurate).

chance-coleman:
I'm not entirely sure why this hasn't been included this whole time. I'm unaware of any decisions that were explicitly made to exclude it. This all looks good to me. The one thing that we should verify is that identity-config tests work as expected.

chance-coleman:
verified this works against identity-config tests as well. Pending @slaskawi i think we can move forward with this.

slaskawi (approved):
LGTM!

PR #1537, opened 2025-05-07
feat: keycloak fips

slaskawi:
## Description

This Pull Request introduces all changes needed to switch Keycloak to FIPS by default.

The FIPS execution paths are tested by https://github.com/defenseunicorns/uds-core/pull/1518

Once we confirm that everything works correctly, we'll switch Keycloak to FIPS by default. 

## Related Issue

Relates to https://github.com/defenseunicorns/uds-identity-config/issues/36

## Type of change

## Steps to Validate

- This PR is validated automatically

## Checklist before merging


src/keycloak/chart/values.yaml
```diff
@@ -124,6 +124,8 @@ smtp:
 
 # Configure FIPS mode for Keycloak
 fips: false
+# By default, Keycloak requires 14+ character passwords in FIPS mode. This switch disables this requirement.
```
chance-coleman, line 127: I could see either opinion on this, should this be documented in either of our FIPS docs? if so, should probably show the package override for setting it.

tasks.yaml
```diff
@@ -105,10 +105,45 @@ tasks:
   - name: test-uds-core-ha
```
chance-coleman, line 105: this task is not working for me anymore. I am getting this error from the postgres docker container:
```
2025-05-08 14:51:46.484 UTC [48] FATAL:  could not load server certificate file "/var/lib/postgresql/server.crt": Permission denied
2025-05-08 14:51:46.484 UTC [48] LOG:  database system is shut down
```
Im on linux, Ubuntu specifically. I was able to get this working using the following:

- A separate temporary container is used to:
  - Copy the certs
  - Fix their permissions and ownership (to UID 999 for Postgres)
- Postgres is then started normally in the background using -d, with the pre-fixed certs mounted.

```
      - description: "Cleanup previous PostgreSQL runs"
        cmd: docker stop postgres && docker rm postgres || true

      - description: "Create a network for the PostgreSQL container"
        cmd: docker network create k3d-uds || true

      - description: "Generate PostgreSQL certs"
        cmd: |
          mkdir -p build/certs
          openssl req -x509 -newkey rsa:4096 -sha256 -nodes \
            -keyout build/certs/server.key \
            -out build/certs/server.crt \
            -days 365 -subj "/CN=postgres" > /dev/null 2>&1 || {
              echo "Error: Failed to generate PostgreSQL certs"
              exit 1
            }
          chmod 600 build/certs/server.key
          chmod 644 build/certs/server.crt

      - description: "Prepare PostgreSQL certs inside a temp container"
        cmd: |
          docker run --rm \
            -v $(pwd)/build/certs:/certs \
            --entrypoint bash postgres:16 \
            -c "cp /certs/server.key /certs/fixed.key && \
                cp /certs/server.crt /certs/fixed.crt && \
                chown 999:999 /certs/fixed.key /certs/fixed.crt && \
                chmod 600 /certs/fixed.key && chmod 644 /certs/fixed.crt"

      - description: "Start PostgreSQL Docker container"
        cmd: |
          CONTAINER_NAME=postgres
          POSTGRES_VERSION=16
          TLS_VERSION=TLSv1.2

          docker kill $CONTAINER_NAME && docker rm $CONTAINER_NAME || true

          docker run -p 5432:5432 --network=k3d-uds --name $CONTAINER_NAME \
            -e POSTGRES_DB=keycloak \
            -e POSTGRES_USER=postgres \
            -e POSTGRES_PASSWORD='unicorn123!@#UN' \
            -v $(pwd)/build/certs/fixed.crt:/var/lib/postgresql/server.crt:ro \
            -v $(pwd)/build/certs/fixed.key:/var/lib/postgresql/server.key:ro \
            -d postgres:$POSTGRES_VERSION \
            -c ssl=on \
            -c ssl_cert_file=/var/lib/postgresql/server.crt \
            -c ssl_key_file=/var/lib/postgresql/server.key \
            -c ssl_min_protocol_version=$TLS_VERSION \
            -c ssl_max_protocol_version=$TLS_VERSION \
            -c log_connections=on \
            -c log_disconnections=on \
            -c log_min_messages=debug1 \
            -c log_line_prefix='%m [%p] %q%u@%d ' \
            -c log_statement=all

      - description: "Wait for PostgreSQL to be ready"
        cmd: |
          for i in {1..10}; do
            docker exec postgres pg_isready -U postgres && break
            sleep 1
          done

      - description: "Add Grafana database to PostgreSQL"
        cmd: docker exec postgres psql -U postgres -c "CREATE DATABASE grafana;"
```

definitely not a requirement to use this implementation.

src/keycloak/chart/values.yaml
slaskawi: It's actually both - in the upgrade guide as well here (just for a quick reference).

tasks.yaml
slaskawi: I took your solution mostly as is but also created https://github.com/defenseunicorns/uds-core/issues/1548 to revisit it later. Unfortunately my System76 was misbehaving with Fedora and I had to reinstall PopOS on it. It should be ready in a couple of days and then I will be able to pursue this task.

src/keycloak/common/zarf.yaml
```diff
@@ -48,7 +48,7 @@ components:
                   echo "Keycloak is not deployed in HA mode, skipping scale down."
```
noahpb: @UnicornChance @slaskawi I think we could remove this check entirely now as this was only needed for when upgrade issues were reported by consumers when upgrading to version 26.0.5. Unless it's considered best practice to manually scale down when upgrading at any version, we should remove this.

chance-coleman: Thats a good thought, i don't actually know what is best practice in this scenario. Hopefully Sebastian knows more. This is specific to HA mode so part of me thinks it would still be helpful to gracefully scale in those environments.

slaskawi: Keycloak 26 should support seamless upgrades, see:

* https://www.keycloak.org/2024/06/persistent-user-sessions-in-preview.html
* https://github.com/keycloak/keycloak/discussions/28271

So unless we discover another bug, we shouldn't need this anymore.

mjnagel (dismissed):

mjnagel (commented):
Did we want to enable FIPS on the EKS testing in this PR as well?

slaskawi:
> Did we want to enable FIPS on the EKS testing in this PR as well?

My plan is to implement this in the next release, when we'll be flipping FIPS on by default.

chance-coleman (approved):

PR #1578, opened 2025-05-19
chore(ci): add HA install and upgrade nightly tests

### LINKED ISSUE #1007: Test HA k3d setup in CI
Need to implement a test for the HA upgrade path. At this point we have an [HA test task](https://github.com/defenseunicorns/uds-core/blob/main/tasks.yaml#L94-L101) however we do not have an upgrade HA test.

Both existing HA test and HA upgrade test should probably be added in CI. 

Upgrading from uds-core 0.30.0 to 0.31.1 had errors with keycloak being able to upgrade properly since it was in HA mode. Having no testing for this scenario caused us to blindly push our upgrade into staging environment. 

Relates to #1009

chance-coleman:
## Description
To help address upgrade issues we've seen lately, adding a nightly HA install and upgrade test to CI. This will test all flavors against install and upgrade (from latest release).

## Related Issue

Fixes #1007 

## Type of change

- [] New feature (non-breaking change which adds functionality)

## Steps to Validate
- Run new task locally with `uds run test-uds-core-ha-upgrade`

## Checklist before merging

### DIFF
--- a/.github/workflows/upgrade-ha.yaml
+++ b/.github/workflows/upgrade-ha.yaml
@@ -1,0 +1,41 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+name: UDS Core Install/Upgrade HA Test
+
+on:
+  schedule:
+    # Runs every morning at 2:00 AM UTC
+    - cron: '0 2 * * *'
+
+permissions:
+  contents: read
+  id-token: write
+  packages: read
+
+jobs:
+  uds-core-ha-upgrade-nightly:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        flavor: [upstream, registry1, unicorn]
+
+    name: Run HA Install/Upgrade Test for ${{ matrix.flavor }}
+
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v4
+
+      - name: Environment setup
+        uses: ./.github/actions/setup
+        with:
+          registry1Username: ${{ secrets.IRON_BANK_ROBOT_USERNAME }}
+          registry1Password: ${{ secrets.IRON_BANK_ROBOT_PASSWORD }}
+          ghToken: ${{ secrets.GITHUB_TOKEN }}
+          chainguardIdentity: ${{ secrets.CHAINGUARD_IDENTITY }}
+
+      - name: Run UDS Core Install HA Test
+        run: uds run uds-core-upgrade --set FLAVOR=${{ matrix.flavor }} --no-progress
+
+      - name: Run UDS Core Upgrade HA Test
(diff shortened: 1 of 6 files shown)


noahpb (dismissed):
LGTM. Once https://github.com/defenseunicorns/uds-core/pull/1476 is merged, we should revisit this to integrate it with the `Filter` workflow so that we can trigger this test via a PR label.

.github/workflows/upgrade-ha.yaml
```diff
@@ -0,0 +1,41 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+name: UDS Core Install/Upgrade HA Test
+
+on:
+  schedule:
+    # Runs every morning at 2:00 AM UTC
+    - cron: '0 2 * * *'
```
mjnagel, line 9: Could we also run this on PR on changes to this specific workflow file + changes to the [ha config file](https://github.com/defenseunicorns/uds-core/blob/test-ha-k3d-setup-in-ci/bundles/k3d-standard/uds-ha-config.yaml)? Don't want to expand that too much, but I think at least those two would be good so that we validate workflow changes and updates to the config.

Edit: [This comment](https://github.com/defenseunicorns/uds-core/pull/1578/files#r2098690248) is a good example of something we might catch by running on PR.

.github/workflows/upgrade-ha.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+permissions:
+  contents: read
+  id-token: write
+  packages: read
+
+jobs:
+  uds-core-ha-upgrade-nightly:
+    runs-on: ubuntu-latest
```
mjnagel: Pretty sure this will not fit on the free ubuntu-latest runners, this likely needs to be one of our larger runners like `uds-ubuntu-big-boy-8-core`.

mjnagel: May also want to add `timeout-minutes` of 40+ here just to explicitly set that?

.github/workflows/upgrade-ha.yaml
```diff
@@ -0,0 +1,41 @@
...
+    strategy:
+      matrix:
+        flavor: [upstream, registry1, unicorn]
```
mjnagel, line 27: Maybe this falls into the bucket of "testing too much", but would we also want to matrix this with install/upgrade (to test both)? Open to either here - and the upgrade is probably the more important one to add initially.

.github/workflows/upgrade-ha.yaml
mjnagel: Maybe related - should we also run on changes to the standard bundle itself? This would ensure we run during the release PR which might be good?

chance-coleman: updated to run on changes to this upgrade workflow file and any changes the k3d-standard bundle directory

mjnagel (approved):

Issue #1440, opened 2025-04-07
Ambient testing/metrics data


darianc-github:
Conduct tests to answer common questions on performance, reliability, speed, etc. Tests should consider ambient v sidecar mode for UDS Core as a whole.

This data will be used for external blogs, videos, etc. 

Definition of done: Table comparing sidecar/ambient across at least 4 key metrics - deploy time, cpu usage, memory usage, and request latency. This can be added to the docs site or to an internal doc for now.

mjnagel:
Some key metrics to look at here (for full core):
- deploy time
- CPU/memory usage
- latency testing on request responses (we've used https://github.com/tsenart/vegeta before for this) - note that uds-k3d's LB setup adds latency 
- cost savings - taking cpu/memory gains and extrapolating to AWS instance sizes/costs?

For testing sidecar vs ambient we can likely use a release before ambient (0.38.0) against a release with full ambient (may need to be a snapshot initially).

noahpb:
repo started here:
https://github.com/defenseunicorns/uds-core-istio-benchmark

noahpb:
Additional metrics also captured [here](https://github.com/defenseunicorns/uds-core/tree/ambient-testing/istio-metrics).

PR #1571, opened 2025-05-15
chore(docs): update ambient docs

noahpb:
## Description
Updates our technical docs to include some more details about ambient mode and a summarization/comparison of ambient and sidecar.

## Related Issue

Fixes #1439

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Steps to Validate
n/a

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md) followed

PR #1582, opened 2025-05-20
feat: add grafana dashboard for istio mode comparison

noahpb:
## Description
Adds a custom grafana dashboard via a new `ConfigMap` in the `uds-grafana-config` helm chart.

The dashboard is straightforward - it displays a graph for both CPU and Memory consumption of all workloads in a given namespace. One metric on the graph displays the CPU and Memory consumption of that workload with istio sidecars included, while the other metric excludes the istio sidecar metrics - giving a user an idea of what their resource consumption would be for their workloads should they switch to ambient mode.

image

## Related Issue
Relates to #1440 

## Type of change

## Steps to Validate

## Checklist before merging


src/grafana/chart/templates/dashboards.yaml
```diff
@@ -0,0 +1,272 @@
...
+        "title": "Memory Usage - All Workloads (With vs Without Istio Sidecar)",
+        "type": "timeseries"
+        }
+    ],
+    "preload": false,
+    "refresh": "",
+    "schemaVersion": 40,
+    "tags": [
+        "istio",
+        "uds"
```
noahpb: Introducing a new tag here - `uds`. I figured that from this point forward, we can add this tag to any custom dashboards that we include.

noahpb:
If merged - #1571 should be updated to include a reference to this in the [istio-sidecar-vs-ambient.md](https://github.com/defenseunicorns/uds-core/pull/1571/files#diff-d19448b086d5ade16f5d5410b2d9f5b4669108ad9f707fa689073b95f3080bc3) doc.

mjnagel (commented):
LGTM overall! Small pieces of feedback I sent synchronously:
- Might be nice to add a description text block that explains how to view this and some caveats (explaining this is an estimate for what ambient might look like, etc)
- I think there's potentially edge cases with this dashboard for packages in ambient mode, especially with waypoints. Not sure how to handle this, but maybe we should filter out waypoints or ^ description can call this out.
- It may also be nice to display the "current mode" for people viewing the dashboard, or even hide the core packages, etc that are in ambient mode? Potentially could do this with namespace labels or pod/containers present.

src/grafana/chart/templates/dashboards.yaml
```diff
@@ -0,0 +1,413 @@
...
+                                    "id": "custom.hideFrom",
+                                    "value": {
+                                        "legend": false,
+                                        "tooltip": false,
+                                        "viz": true
+                                    }
+                                }
+                            ]
+                        }
+                    ]
```
mjnagel:
```suggestion
                    }
```

Might be missing why this was added, but it causes the CPU "with sidecar" to be hidden by default.

src/grafana/chart/templates/dashboards.yaml
```diff
@@ -0,0 +1,413 @@
...
+        "templating": {
+            "list": [
+                {
+                    "datasource": "Prometheus",
+                    "definition": "label_values(kube_pod_info, namespace)",
+                    "includeAll": false,
+                    "label": "Namespace",
+                    "name": "namespace",
+                    "query": {
+                        "query": "label_values(kube_pod_info{namespace!~\"kube-system|istio-system|istio-admin-gateway|istio-tenant-gateway|istio-passthrough-gateway|pepr-system|neuvector|loki|monitoring|grafana|vector|velero|keycloak|authservice\"}, namespace)",
```
mjnagel:
```suggestion
                        "query": "label_values(kube_pod_info{namespace!~\"kube-system|istio-system|istio-admin-gateway|istio-tenant-gateway|istio-passthrough-gateway|pepr-system|neuvector|loki|monitoring|grafana|vector|velero|keycloak|authservice|zarf|uds-dev-stack\"}, namespace)",
```

mjnagel (dismissed):
LGTM overall - in ambient/not-in-mesh namespaces the dashboard feels a bit weird. Don't know if that's just "the way it is" or if we should hide panels based on mode or put a description box at least?

mjnagel (approved):

PR #1589, opened 2025-05-23
fix: ensure uniqueness of sso client ids in cluster

mjnagel:
## Description

This PR uses the `PackageStore` in our validator to ensure that SSO client IDs are "globally" unique across the entire cluster (in addition to locally unique in the package CR).

## Related Issue

N/A

## Type of change

## Steps to Validate

Deploy a package CR with the same SSO client ID as another package. Ensure that it is blocked by the validator.

## Checklist before merging


src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -150,10 +150,34 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+    if (namespacesWithClientId.size > 0) {
+      const isOwnedByCurrentPackage = namespacesWithClientId.has(ns);
+
+      // If this client ID exists in other namespaces, deny the request
+      if (!isOwnedByCurrentPackage) {
+        // Get package names for each namespace with this client ID
+        const ownerInfo = Array.from(namespacesWithClientId)
+          .map(namespace => {
+            const pkgName = PackageStore.getPkgName(namespace);
+            return `${namespace}/${pkgName}`;
```
noahpb: super nit: Should we not include the namespace name in the output? The risk is likely minimal, but this deny message could unintentionally give users information about packages in other namespaces that they don't have access to.

mjnagel: I did consider this, but Keycloak doesn't treat client IDs as secret and we don't treat package CR names/namespaces as secret so it felt a bit restrictive to not output that information. Maybe could be considered `classification by compilation` so to speak though...

I can see both sides - if we remove the namespace, it feels like we may as well just output a generic message that "client id is already in use". I suppose if someone knows has access to all the Package CRs they can check which one has the conflicting client (may take a bit of searching), if they don't - they have no reason to know the name or namespace.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -150,10 +150,34 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
     }
     clientIDs.add(client.clientId);
+
+    // Check for global uniqueness (across all packages/namespaces)
+    const namespacesWithClientId = PackageStore.findPackagesWithSsoClientId(client.clientId);
+
+    // If we find namespaces with this client ID, make sure it's only the current namespace being updated
+    if (namespacesWithClientId.size > 0) {
+      const isOwnedByCurrentPackage = namespacesWithClientId.has(ns);
+
```
chance-coleman, line 165: double check my understanding here, but i believe we can save some computations with the following:

```suggestion

// if the only namespace using this clientId is the current namespace we can
// return early and skip the .map() and .join() computations
if (namespacesWithClientId.size === 1 && namespacesWithClientId.has(ns)) {
  return; // only this namespace uses the client ID — so there'd be no conflict
}

```

src/pepr/operator/crd/validators/package-validator.ts
mjnagel: Going to update to genericize this message.

src/pepr/operator/crd/validators/package-validator.ts
chance-coleman: OBE

mjnagel: Discussed synchronously - based on other changes this is no longer necessary.

chance-coleman (approved):

noahpb (approved):

PR #1617, opened 2025-05-29
fix: checkpoint with istio ambient

mjnagel:
## Description

The checkpoint package has been having issues since switching pepr/keycloak to Istio ambient mode. The root of the issue was that timing on pod startup caused issues during the restore. The changes to resolve this issue were primarily to scale down + scale up certain pods during the checkpoint creation and deployment. While this does add some time to the process, it also appears necessary to fix issues.

Additional changes in the PR:
- added iptables dump in save-logs
- added src/test e2e packages to slim-dev testing (and checkpoint)
- switched to ubuntu-24 arm runners
- documented an issue with ARM checkpoint on runners where network policies are not enforced

## Related Issue

Undocumented issue, but was causing problems in uds-common CI on https://github.com/defenseunicorns/uds-common/pull/478

## Type of change

## Steps to Validate

NOTE: Requires an amd64 machine.

Follow the checkpoint workflow and validate all creates and deploys as expected.

## Checklist before merging


src/test/tasks.yaml
```diff
@@ -42,36 +42,6 @@ tasks:
       - description: Deploy the test resources
         cmd: "uds zarf package deploy build/zarf-package-uds-core-test-apps-*.zst --confirm --no-progress"
 
-      - description: Wait for the admin app to be ready
```
mjnagel: Note: for all of these deletions this is mainly time savings + remove some flakiness caused by some of the ones waiting on pods. Zarf/helm will wait on all standard resources (deployments, etc) and also will wait on our Package CRs because we use kstatus conditions on them.

noahpb (approved):

PR #1653, opened 2025-06-16
chore(docs): sanitize docs directory naming

chance-coleman:
## Description

With the https://github.com/defenseunicorns/uds-docs/pull/243 we can rename our docs folders to by hyphenated instead of space separated.

## Type of change

## Checklist before merging


mjnagel (commented):
LGTM when testing. It would be nice if we had the `dev-docs` aligned but I think that would cause more of a hassle than it's worth. Should we update any docs/notes on the `dev-docs` task to indicate that the sidebar format might look different than the actual result?

mjnagel (approved):

Issue #1630, opened 2025-06-04
AWS ALB Support


slaskawi:
## Summary

The goal of this ticket is to provide the support for AWS ALB (AWS Application Load Balancer - L7) that handles TLS termination and passes the raw certificate into a `x-amzn-mtls-clientcert` header. 

This certificate cannot be easily interpreted by Keycloak as none of the Reverse Proxy Providers supports it. The Certificate needs to be at least URL decoded and Base64 encoded as highlighted in https://github.com/defenseunicorns/uds-identity-config/issues/470

As agreed with @eddiezane, @chance-coleman and @mjnagel over a meeting, the UDS Core will provide the full support for this scenario, including:

* Envoy Filter based implementation 
  * The https://github.com/defenseunicorns/uds-identity-config/issues/470 contains a starting point
* A high-level option that configures both Istio and Keycloak for this use case. 
  * Keycloak shall use the HAProxy Reverse Proxy Provider
* End-to-end tests with an EKS-based infrastructure
  * The biggest question mark is how to call CAC tests from the UDS Core?
* Documentation

LockedThread:
Looking forward to this.

mjnagel:
Also note https://github.com/defenseunicorns/uds-core/issues/1526 which may be required for some apps on top of core, such as GitLab, to ensure that the ALB `X-Forwarded` headers are trusted.

PR #1659, opened 2025-06-20
feat: set keycloak `KC_SPI_X509CERT_LOOKUP` vars from x509 lookup provider

JeffResc:
## Description
Currently, the environment variables `KC_SPI_X509CERT_LOOKUP_NGINX_SSL_CLIENT_CERT` and `KC_SPI_X509CERT_LOOKUP_NGINX_SSL_CLIENT_CERT_CHAIN_PREFIX` have `NGINX` hardcoded in the variable name. This has been updated to use the `x509LookupProvider` value, so that the environment variable will work with any configured x509 lookup provider.

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -154,10 +154,10 @@ spec:
             - name: KC_SPI_X509CERT_LOOKUP_PROVIDER
               value: {{ .Values.x509LookupProvider }}
             # Set nginx provider header name
-            - name: KC_SPI_X509CERT_LOOKUP_NGINX_SSL_CLIENT_CERT
+            - name: KC_SPI_X509CERT_LOOKUP_{{ .Values.x509LookupProvider | upper }}_SSL_CLIENT_CERT
               value: istio-mtls-client-certificate
             # Dumb value (not used in the nginx provider, but required by the SPI)
-            - name: KC_SPI_X509CERT_LOOKUP_NGINX_SSL_CLIENT_CERT_CHAIN_PREFIX
+            - name: KC_SPI_X509CERT_LOOKUP_{{ .Values.x509LookupProvider | upper }}_SSL_CLIENT_CERT_CHAIN_PREFIX
               value: UNUSED
           {{- if .Values.fips }}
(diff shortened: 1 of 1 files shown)


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -154,10 +154,10 @@ spec:
             - name: KC_SPI_X509CERT_LOOKUP_PROVIDER
               value: {{ .Values.x509LookupProvider }}
             # Set nginx provider header name
-            - name: KC_SPI_X509CERT_LOOKUP_NGINX_SSL_CLIENT_CERT
-              value: istio-mtls-client-certificate
+            - name: KC_SPI_X509CERT_LOOKUP_{{ .Values.x509LookupProvider | upper }}_SSL_CLIENT_CERT
+              value: {{ .Values.x509SSLClientCert }}
```
slaskawi: I would love @mjnagel to give it a peek

slaskawi (commented):
LGTM from the Keycloak perspective. Not approving to give @mjnagel a chance for review.

slaskawi:
This might also be related to https://github.com/defenseunicorns/uds-core/issues/1630 to a degree

src/keycloak/chart/templates/statefulset.yaml
mjnagel: I want to be really careful with complicating this setup or removing some security here. Currently our Keycloak virtualservice(s) strip any incoming `istio-mtls-client-certificate` and add that header, sourced from the [actual client cert](https://github.com/defenseunicorns/uds-core/blob/jeffr/configurable-x509-client-cert/src/keycloak/chart/values.yaml#L318) in envoy. I haven't tested this but I think you'd be able to pull different data from envoy there, such as a header with `%REQ(MY_CUSTOM_HEADER)%`. This way you'd maintain the same flow as the default setup and maintain the same header "sanitization" from Istio (not needing to rely on your LB to do that). Would that work, or is there an additional limitation here I'm not considering?

JeffResc:
After a discussion with @mjnagel, the existing helm value `mtlsClientCert` will work for my use-case. I have re-worked this PR to remove the configurable x509 client cert header.

mjnagel (approved):
LGTM -

PR #1670, opened 2025-06-25
feat: aws alb support

### LINKED ISSUE #1630: AWS ALB Support

### LINKED ISSUE #1719: [AWS ALB] Introduce necessary changes to Keycloak deployment
## Summary

The goal of this ticket is to introduce necessary changes to both Keycloak StatefulSet to work with the AWS ALB proxy

### LINKED ISSUE #1720: [AWS ALB] Ensure istio-mtls-client-certificate is trusted and is not trimmed down
## Summary

The goal of this ticket is to introduce necessary changes to the UDS Package so that it trusts the AWS ALB certificate passed in the header

### LINKED ISSUE #1724: [AWS ALB] Introduce an Envoy Filter that processes the x-amzn-mtls-clientcert header
## Summary

The goal of this ticket is to introduce an Envoy Filter that processes the AWS ALB header and makes it usable for Keycloak

### LINKED ISSUE #1727: [AWS ALB] Ensure the internal workloads cannot connect to the tenant gateway
## Summary

In order to harden the security posture by the AWS ALB support, we need to ensure that neither of the internal workloads can connect to the Tenant Gateway (or use another way to ensure the istio-mtls-client-certificate header cannot be spoofed).

slaskawi:
## Description

This Pull Request introduces the support for a AWS ALB connected to the Tenant Gateway.

The work in this PR requires also https://github.com/defenseunicorns/uds-identity-config/pull/541

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1630
Fixed https://github.com/defenseunicorns/uds-core/issues/1719
Fixes https://github.com/defenseunicorns/uds-core/issues/1720
Fixes https://github.com/defenseunicorns/uds-core/issues/1724
Fixes https://github.com/defenseunicorns/uds-core/issues/1727
Relates to https://github.com/defenseunicorns/uds-core/issues/1630

## Type of change

## Steps to Validate

This feature is tested via UDS Identity Config

## Checklist before merging


src/keycloak/chart/values.yaml
```diff
@@ -124,7 +124,7 @@ devMode: true
 debugMode: false
 
 # Enable JSON logging format for Keycloak
-jsonLogFormat: true
+jsonLogFormat: false
```
slaskawi: Remove this from the final PR

docs/reference/configuration/single-sign-on/application-load-balancer.md
```diff
@@ -0,0 +1,43 @@
...
+---
+
+UDS Core supports external TLS termination and custom client certificate headers for environments using external load balancers (e.g., AWS Application Load Balancer - ALB). This is achieved by overriding default Keycloak and Istio settings in your bundle configuration.
+
+The configuration requires disabling HTTPS redirects in Istio gateways and setting the number of trusted proxies to one, with the ALB acting as the trusted proxy. Changing this setting in runtime requires restarting the Istio Tenant Gateway Pods (using `uds zarf tools kubectl rollout restart deployment tenant-ingressgateway -n istio-tenant-gateway`). Keycloak configuration is then adjusted by deploying an additional `EnvoyFilter` that converts the Amazon ALB client certificate into a format that Keycloak can understand. The example below outlines the configuration needed to set up the [Amazon ALB](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/mutual-authentication.html) for UDS Core:
+
+```yaml
+packages:
+  - name: core
+    repository: oci://ghcr.io/defenseunicorns/packages/uds/core
```
slaskawi: Link this in the Istio Ingress docs

docs/reference/configuration/single-sign-on/application-load-balancer.md
```diff
@@ -0,0 +1,43 @@
...
+        uds-istio-config:
+          values:
+            - path: tls.servers.keycloak.enableHttpsRedirect
+              value: false
+            - path: tls.servers.tenant.enableHttpsRedirect
+              value: false
+      istio-controlplane:
+        istiod:
+          values:
+            - path: meshConfig.defaultConfig.gatewayTopology.numTrustedProxies
```
slaskawi: Let's implement a Pepr watcher for this.

slaskawi: And rotate admin Pods as well.

docs/reference/configuration/service-mesh/ingress.md
```diff
@@ -200,3 +200,13 @@ spec:
               number: 80
 ```
 This VirtualService matches requests to the root domain (`uds.dev`) with the path prefix `/my-app` and routes them to your service (`my-app-service` on port 80).
+
+### Using Application Load Balancers (ALBs) with UDS Core Gateways
+
+UDS Core Gateways can be used with Application Load Balancers (ALBs) connected to one of both Istio Gateways. This allows you to leverage the ALB's capabilities for routing, SSL termination, and other features while still using Istio for service mesh functionality. Such setups are highly dependent on the Cloud Provider configuration but usually require:
```
mjnagel: Just a nit on the phrasing here - `Application Load Balancer` is AWS specific. Azure uses `Application Gateway` I think? More generically this is support for a loadbalancer/access point that terminates TLS - not sure how to succinctly say that/name this. Same would apply for anywhere we use `ALB` acronym.

Maybe I'm misunderstanding though and this is meant to be AWS specific?

docs/reference/configuration/service-mesh/ingress.md
```diff
@@ -200,3 +200,13 @@ spec:
...
+
+* Disabling HTTPS redirects in Istio gateways.
+* Setting the number of trusted proxies to one, with the ALB acting as the trusted proxy.
+* Configuring Keycloak to accept the client certificate from the ALB.
+
+The most common setup has been covered in the [Application Load Balancer](https://uds.defenseunicorns.com/reference/configuration/single-sign-on/application-load-balancer/) part of the documentation.
```
mjnagel:
```suggestion
The most common setup has been covered in the [Application Load Balancer](/reference/configuration/single-sign-on/application-load-balancer/) part of the documentation.
```

Local refs are preferred to be able to view these pages on a local "dev-docs" deploy.

src/pepr/operator/controllers/istio/istio-configmap-sync.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export async function restartGatewayPods() {
+  const tenantGatewayPods = await K8s(kind.Pod).InNamespace(TENANT_GATEWAY_NAMESPACE).Get();
+  const adminGatewayPods = await K8s(kind.Pod).InNamespace(ADMIN_GATEWAY_NAMESPACE).Get();
+
+  log.info("Restarting {} pods to apply new configuration", TENANT_GATEWAY_NAMESPACE);
+  await evictStandalonePods(TENANT_GATEWAY_NAMESPACE, tenantGatewayPods.items, RESTART_REASON, log);
+
+  log.info("Restarting {} pods to apply new configuration", ADMIN_GATEWAY_NAMESPACE);
+  await evictStandalonePods(ADMIN_GATEWAY_NAMESPACE, adminGatewayPods.items, RESTART_REASON, log);
```
mjnagel: Any reason this doesn't use [`reloadPods`](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/secrets/reload-utils.ts#L26)? That function performs a more graceful restart by patching the controller (i.e. deployment).

src/pepr/operator/controllers/istio/istio-configmap-sync.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Component, setupLogger } from "../../../logger";
+import { evictStandalonePods } from "../secrets/reload-utils";
+
+export const TENANT_GATEWAY_NAMESPACE = "istio-tenant-gateway";
+export const ADMIN_GATEWAY_NAMESPACE = "istio-admin-gateway";
+export const RESTART_REASON = "Restarting Gateway Pods to apply new Istio configuration";
+
+const log = setupLogger(Component.OPERATOR_ISTIO);
+
+export async function restartGatewayPods() {
```
mjnagel: Any concern with this running "too often" and cycling pods for things that pods don't need to be cycled for? Maybe any change in the istio config should result in cycling these pods, just wasn't sure if we wanted to be more granular here (i.e. look for specific changes to specific keys).

src/keycloak/chart/values.yaml
```diff
@@ -329,3 +329,15 @@ mtlsClientCert: "%DOWNSTREAM_PEER_CERT%"
...
+# Third-party integration settings, such as Application Load Balancer support
+#thirdPartyIntegration:
+#  tls:
+#    # This would instruct Istio to allow unencrypted traffic at the Tenant Gateway
+#    # We could potentially configure numTrustedProxies when this is set to external
+#    tlsTermination: external
+#    # This would deploy the Envoy Filter and switch to HAProxy SSL Lookup Provider in Keycloak
+#    # For AWS, we'd use x-amzn-mtls-clientcert here
+#    tlsCertificateHeader: x-ssl-custom-cert
+#    # The format of the client certificate header (PEM or DER)
```
```suggestion
#    # The format of the client certificate header (PEM or AWS)
```

docs/reference/configuration/single-sign-on/application-load-balancer.md
```diff
@@ -0,0 +1,43 @@
...
+              value: 1
+  - name: identity-authorization
+    path: oci://ghcr.io/defenseunicorns/packages/uds/identity-config
+    ref: x.x.x
+    overrides:
```
mjnagel: I think this package reference is confusing. The first package (above this) is the full core package, which would include Keycloak. Is this intentionally separate?

docs/reference/configuration/single-sign-on/application-load-balancer.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Application Load Balancer
+---
+
+UDS Core supports external TLS termination and custom client certificate headers for environments using external load balancers (e.g., AWS Application Load Balancer - ALB). This is achieved by overriding default Keycloak and Istio settings in your bundle configuration.
+
+The configuration requires disabling HTTPS redirects in Istio gateways and setting the number of trusted proxies to one, with the ALB acting as the trusted proxy. Changing this setting in runtime requires restarting the Istio Tenant Gateway Pods (using `uds zarf tools kubectl rollout restart deployment tenant-ingressgateway -n istio-tenant-gateway`). Keycloak configuration is then adjusted by deploying an additional `EnvoyFilter` that converts the Amazon ALB client certificate into a format that Keycloak can understand. The example below outlines the configuration needed to set up the [Amazon ALB](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/mutual-authentication.html) for UDS Core:
```
mjnagel: If we stick with the watch, this can be updated to remove the requirement of `rollout restart`.

src/keycloak/chart/values.yaml
```diff
@@ -329,3 +329,15 @@ mtlsClientCert: "%DOWNSTREAM_PEER_CERT%"
...
+#    tlsCertificateFormat: PEM
```
mjnagel: Ok with this but my preference is to have all values uncommented and just nulled out/empty by default. Then conditionals (ideally) are simpler and just directly key off of the contents (i.e. `if .Values. thirdPartyIntegration.tls.tlsCertificateFormat`) rather than having to check each parent level for existence first.

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -165,26 +165,30 @@ spec:
...
         host: sso
         port: 8080
+        {{- if not (and .Values.thirdPartyIntegration (and .Values.thirdPartyIntegration.tls (eq .Values.thirdPartyIntegration.tls.tlsTermination "external"))) }}
         advancedHTTP:
           headers:
             request:
               remove:
                 - istio-mtls-client-certificate
               add:
                 istio-mtls-client-certificate: "{{ .Values.mtlsClientCert }}"
```
mjnagel, line 190: With the other changes here can we deprecate/remove `.Values.mtlsClientCert` (swap this with `"%DOWNSTREAM_PEER_CERT%"`)? It was necessary to set the `istio-mtls-client-certificate` header to the contents of a different header, but with these changes that is done by the `EnvoyFilter` for us already.

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -165,26 +165,30 @@ spec:
                 prefix: /metrics
           redirect:
             uri: "/realms/{{ .Values.realm }}/account"
+          {{- if not (and .Values.thirdPartyIntegration (and .Values.thirdPartyIntegration.tls (eq .Values.thirdPartyIntegration.tls.tlsTermination "external"))) }}
```
mjnagel: I think we _need_ a `tlsCertificateHeader` value, could we have the conditional go off of that instead of `tlsTermination`? `tlsTermination` almost feels like just an extra value if we already know they set `tlsCertificateHeader`.

mjnagel (commented):
This is on the right track I think - also wondering if we could deprecate/remove `x509LookupProvider` if we are standardizing on nginx for this?

docs/reference/configuration/single-sign-on/application-load-balancer.md
```diff
@@ -0,0 +1,43 @@
...
+          values:
+            - path: tls.servers.keycloak.enableHttpsRedirect
+              value: false
+            - path: tls.servers.tenant.enableHttpsRedirect
+              value: false
+      istio-controlplane:
+        istiod:
+          values:
+            - path: meshConfig.defaultConfig.gatewayTopology.numTrustedProxies
+              value: 1
```
mjnagel: I wonder if we need to document that this could change depending on setup? I think with certain LB setups you might have multiple hops and need this to be higher. cc @JeffResc

docs/reference/configuration/single-sign-on/application-load-balancer.md
```diff
@@ -0,0 +1,43 @@
...
+    repository: oci://ghcr.io/defenseunicorns/packages/uds/core
+    ref: x.x.x
+    overrides:
+      istio-tenant-gateway:
+        uds-istio-config:
+          values:
+            - path: tls.servers.keycloak.enableHttpsRedirect
+              value: false
+            - path: tls.servers.tenant.enableHttpsRedirect
+              value: false
```
mjnagel: Should this also be documented for the admin gateway?

slaskawi: Right now, there's a paragraph about both Gateways in `docs/reference/configuration/service-mesh/ingress.md`:

> UDS Core Gateways can be used with Application Load Balancers (ALBs) connected to one of both Istio Gateways. This allows you to leverage the ALB's capabilities for routing, SSL termination, and other features while still using Istio for service mesh functionality

Would you rather handle this separately in the docs somehow? I also think the ALB should cover only Tenant Gateway. Exposing the Admin Gateway with this approach is possible but questionable from the security perspective.

docs/reference/configuration/single-sign-on/application-load-balancer.md
slaskawi: Yes, that's 100% correct! This number reflects the number of Reverse Proxies (that transform `X-Forwarded-For` header) on the request way. With `Client -> ALB -> NLB -> Tenant Gateway` setup, this should be 2.

docs/reference/configuration/service-mesh/ingress.md
slaskawi: The robots suggested that the term "L7 Load Balancer" is the most technically correct and doesn't refer to any specific technology. I'll refactor the docs to use that instead.

src/pepr/operator/controllers/istio/istio-configmap-sync.ts
slaskawi: Updated to "reloadPods". Thanks for spotting this!

src/pepr/operator/controllers/istio/istio-configmap-sync.ts
slaskawi: Refactored. Thanks for reminding me about this.

docs/reference/configuration/single-sign-on/application-load-balancer.md
mjnagel: Maybe a good question for @JeffResc - would your setup use ALBs/TLS termination on the admin side as well? Maybe an incorrect assumption but if the TLS termination is to provide for a WAF, I'd imagine that might be used on the admin side too?

I might lean towards also documenting the overrides for the admin gateway (just a copy/paste of this block with the `istio-admin-gateway` component name). We could add a comment that says "if terminating TLS on the admin gateway" or something to that effect.

docs/reference/configuration/single-sign-on/application-load-balancer.md
mjnagel: In that case might be good to add a note right above this override indicating that, and/or linking to the upstream istio doc.

docs/reference/configuration/single-sign-on/application-load-balancer.md
slaskawi: No, it's just a copy-paste error. Thanks for spotting this - fixed.

docs/reference/configuration/single-sign-on/application-load-balancer.md
slaskawi: Absolutely! Removed.

src/keycloak/chart/values.yaml
slaskawi: Ok, this makes sense. Refactored the values.

docs/reference/configuration/single-sign-on/application-load-balancer.md
slaskawi: I updated the example and added a commented code for the Admin Gateway. I wouldn't like to make it look like a regular (default or expected) configuration.

docs/reference/configuration/single-sign-on/application-load-balancer.md
slaskawi: Added a bit more words to explain this.

src/keycloak/chart/templates/uds-package.yaml
slaskawi: Done!

slaskawi:
> This is on the right track I think - also wondering if we could deprecate/remove x509LookupProvider if we are standardizing on nginx for this?

@mjnagel Removed (cc @JeffRes @eddiezane)

src/keycloak/chart/templates/uds-package.yaml
slaskawi: Done!

slaskawi:
Pushed a new version of this PR. Here are the highlights:

- Reverted removal of `mtlsClientCert` and `x509LookupProvider`. As agreed with @JeffResc, the Mission Heroes need a safe way to revert to the previous mechanism as needed
- Added a paragraph about the infrastructure requirements for using an L7 Load Balancer

docs/reference/configuration/single-sign-on/l7-load-balancer.md
```diff
@@ -0,0 +1,57 @@
...
+#          values:
+#            - path: tls.servers.keycloak.enableHttpsRedirect
+#              value: false
+#            - path: tls.servers.tenant.enableHttpsRedirect
+#              value: false
+      istio-controlplane:
+        istiod:
+          values:
+            - path: meshConfig.defaultConfig.gatewayTopology.numTrustedProxies
+              value: 1
```
mjnagel:
```suggestion
            # Reminder: this should be set to the number of proxies in front of Istio, which may be more than 1 in some setups
            - path: meshConfig.defaultConfig.gatewayTopology.numTrustedProxies
              value: 1
```

docs/reference/configuration/single-sign-on/l7-load-balancer.md
```diff
@@ -0,0 +1,57 @@
+---
+title: L7 Load Balancer
+---
```
mjnagel, line 3: I think in reviewing this further I'd lean more towards this being part of the `service-mesh` section rather than SSO. While it does have the SSO specifics, it's more broadly tied to "how is ingress for the service mesh setup".

Open to disagreement here though 😄

docs/reference/configuration/single-sign-on/l7-load-balancer.md
```diff
@@ -0,0 +1,57 @@
...
+              # Supported formats are "AWS" and "PEM".
+              value: "AWS"
+```
+
+## Infrastructure Requirements
+
+When using an L7 Load Balancer, UDS Core completely trusts information passed through the Istio Gateways. In order to provide the necessary security guarantees, the following infrastructure requirements must be met:
+
+- All the network components between the public internet and the Istio Gateways must be hardened against HTTP header injection and spoofing attacks.
+- The Client Certificate header always needs to sanitized and ensure a client application cannot forge it (from both outside and inside the cluster).
```
mjnagel:
```suggestion
- The Client Certificate header always needs to be sanitized and ensure a client application cannot forge it (from both outside and inside the cluster).
```

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -165,26 +165,30 @@ spec:
                 prefix: /metrics
           redirect:
             uri: "/realms/{{ .Values.realm }}/account"
+          {{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader | default "" | eq "" }}
```
mjnagel:
```suggestion
          {{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader }}
```

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -165,26 +165,30 @@ spec:
...
                 istio-mtls-client-certificate: "{{ .Values.mtlsClientCert }}"
+          {{- end }}
 
       - description: "public auth access with optional client certificate"
         service: keycloak-http
         selector:
           app.kubernetes.io/name: keycloak
         host: sso
         port: 8080
+        {{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader | default "" | eq "" }}
```
mjnagel:
```suggestion
        {{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader }}
```

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -193,10 +197,12 @@ spec:
         gateway: admin
         host: keycloak
         port: 8080
+        {{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader | default "" | eq "" }}
```
mjnagel:
```suggestion
        {{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader }}
```

src/keycloak/chart/values.yaml
```diff
@@ -322,10 +322,27 @@ autoscaling:
...
+
+# DEPRECATED: Use `thirdPartyIntegration` instead
 mtlsClientCert: "%DOWNSTREAM_PEER_CERT%"
 
 # Custom migrations setting for UDS Core internal usage only
 migrations:
   # Deletes the generated truststore to enable recreating it by Keycloak
   deleteGeneratedTrustStore: false
+
+# Third-party integration settings, such as Application Load Balancer support
```
```suggestion
# Third-party integration settings, such as L7 Load Balancer support
```

src/pepr/operator/controllers/istio/istio-configmap-sync.ts
```diff
@@ -0,0 +1,51 @@
...
+      const tenantGatewayPods = await K8s(kind.Pod).InNamespace(TENANT_GATEWAY_NAMESPACE).Get();
+      const adminGatewayPods = await K8s(kind.Pod).InNamespace(ADMIN_GATEWAY_NAMESPACE).Get();
+
+      log.info("Restarting {} pods to apply new configuration", TENANT_GATEWAY_NAMESPACE);
+      await reloadPods(TENANT_GATEWAY_NAMESPACE, tenantGatewayPods.items, RESTART_REASON, log);
+
+      log.info("Restarting {} pods to apply new configuration", ADMIN_GATEWAY_NAMESPACE);
+      await reloadPods(ADMIN_GATEWAY_NAMESPACE, adminGatewayPods.items, RESTART_REASON, log);
+    }
+  }
```
mjnagel, line 50: Should any of this have error handling (try/catch, logs, retries, etc)?

src/keycloak/chart/templates/alb-envoyfilter.yaml
```diff
@@ -0,0 +1,63 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+{{- if eq .Values.thirdPartyIntegration.tls.tlsTermination "external" }}
+{{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader | default "" | ne "" }}
```
mjnagel: In the name of simplicity I think we could just make this:
```suggestion
{{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader }}
```

And then not even have `thirdPartyIntegration.tls.tlsTermination` as a value? But if that seems too magical we could have both and do something like:
```suggestion
{{- if and (eq .Values.thirdPartyIntegration.tls.tlsTermination "external") .Values.thirdPartyIntegration.tls.tlsCertificateHeader }}
```

I'm just unsure right now if there's a use-case for:
- setting external termination, but not setting a TLS cert header
- setting default termination, and setting a TLS cert header

With the current templating both of these scenarios would essentially just deploy the "defaults" (no envoyfilter, normal headers, etc).

src/pepr/operator/controllers/istio/istio-configmap-sync.ts
slaskawi: I don't think we need anything specific here. I would keep it simple.

docs/reference/configuration/single-sign-on/l7-load-balancer.md
slaskawi: The functionality here belongs to both but you may have a point that it probably belongs to Istio Ingress more. 

I'll switch the example and put the most part into the Istio Ingress and just reference it from Keycloak.

src/keycloak/chart/templates/alb-envoyfilter.yaml
slaskawi: I think there is - you might want to terminate TLS on the ALB level and just don't use x509 auth (CAC) at all. It's still a valid use case. Having said that, I'll revert your suggestion as someone may actually try to deploy it that way.

slaskawi: After thinking for a while, I had to revert all these changes. 

It's a perfectly valid use case to terminate the TLS at the edge and just don't forward Client Certificate to Istio. In this case, we just need to have x509 auth disabled (which has a corresponding configuration option).

Technically, this means, that:

* `tlsCertificateHeader` can be empty and we want to apply defaults in that case. 
* The `tlsTermination` should be set to external to activate the EnvoyFilter to do HTTP Client Cert Header modifications.

docs/reference/configuration/single-sign-on/overview.md
```diff
@@ -63,8 +63,9 @@ All groups are under the Uds Core parent group. Frequently a group will be refer
 2. [Device Flow Clients](/reference/configuration/single-sign-on/device-flow/)
 3. [Group Based Authorization](/reference/configuration/single-sign-on/group-based-auth/)
 4. [Keycloak Session Timeout](/reference/configuration/single-sign-on/keycloak-session-timeouts/)
-5. [Recovering lost Keycloak credentials](/reference/configuration/single-sign-on/recoving-lost-credentials/)
-6. [Service Account Roles Clients](/reference/configuration/single-sign-on/service-account/)
-7. [Client Attribute Validation](/reference/configuration/single-sign-on/sso-client-validation/)
-8. [Secret Templating](/reference/configuration/single-sign-on/sso-templating/)
-9. [Trusted Certificate Authority](/reference/configuration/single-sign-on/trusted-ca/)
+5. [L7 Load Balancer](/reference/configuration/single-sign-on/application-load-balancer/)
```
mjnagel:
```suggestion
5. [L7 Load Balancer](/reference/configuration/single-sign-on/l7-load-balancer/)
```

src/keycloak/chart/templates/alb-envoyfilter.yaml
mjnagel: I guess my observation is just that `tls.tlsTermination` set to `external` on its own has no effect. It's part of the envoyfilter conditional but it doesn't matter if you aren't also setting `tls.tlsCertificateHeader`. _Right now_ it just feels like an extra value to set with no purpose. 

If we envision using it to control other features going forward that's fine - I just want to make sure we aren't adding extra values unless they are necessary here for some decision making/logic around features. cc @chance-coleman too for thoughts.

src/keycloak/chart/templates/uds-package.yaml
mjnagel: Unresolving this one, but in the other spots that we have templates like this I think we're making it overly complex. My suggestion would be to simplify with the following:
- `| default "" | eq ""`:  `{{- if not .Values.thirdPartyIntegration.tls.tlsCertificateHeader }}`
- `| default "" | ne ""`: `{{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader }}`

Since the string value defaults to `""` in the `values.yaml` we don't need the default function here. Helm will also evaluate `""` as false for us, so we can just check the value directly (and use `not` to negate).

docs/reference/configuration/service-mesh/ingress.md
```diff
@@ -200,3 +200,61 @@ spec:
...
+            - path: thirdPartyIntegration.tls.tlsCertificateHeader
+              # This header is used by the ALB to pass the client certificate
+              value: "x-amzn-mtls-clientcert"
+            - path: thirdPartyIntegration.tls.tlsCertificateFormat
+              # This format is used by Keycloak to parse the client certificate.
+              # Supported formats are "AWS" and "PEM".
+              value: "AWS"
+```
+
+##### Infrastructure Requirements
```
chance-coleman:
```suggestion
#### Infrastructure Requirements
```

I think our docs will complain about the header depth here because the previous header (the parent of this section) has `###` and the headers have to increment by 1. Also using 4 #'s will mean that this header is in the sidebar navigation.

src/keycloak/chart/templates/alb-envoyfilter.yaml
```diff
@@ -0,0 +1,63 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+{{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader | default "" | ne "" }}
+{{- if eq .Values.thirdPartyIntegration.tls.tlsTermination "external" }}
+# This filter
```
chance-coleman: seems like an unfinished comment here describing the filter

docs/reference/configuration/service-mesh/ingress.md
```diff
@@ -200,3 +200,61 @@ spec:
...
+
+When using an L7 Load Balancer, UDS Core completely trusts information passed through the Istio Gateways. In order to provide the necessary security guarantees, the following infrastructure requirements must be met:
+
+- All the network components between the public internet and the Istio Gateways must be hardened against HTTP header injection and spoofing attacks.
+- The Client Certificate header always needs to be sanitized and ensure a client application cannot forge it (from both outside and inside the cluster).
+- All the traffic between edge and Istio Gateways must be secured (and preferably not reachable from both inside and outside the cluster).
+
+If any of these requirements cannot be met, it is not recommended to make any authentication decisions based on the Client Certificate header and use other MFA methods.
```
chance-coleman:
```suggestion
If any of these requirements cannot be met, it is not recommended to make any authentication decisions based on the Client Certificate header. We would recommend using other MFA methods instead.
```

This line reads funny and i think we want to be as explicit as possible when it comes to this security. This would be my suggestion but totally open to saying this other ways, maybe something that adds some more context why we don't recommend it.

src/pepr/operator/index.ts
```diff
@@ -138,3 +139,11 @@ When(a.Secret)
 
 // Watch for deleted secrets to clean up the checksum cache
 When(a.Secret).IsDeleted().WithLabel("uds.dev/pod-reload", "true").Reconcile(handleSecretDelete);
+
+// Istio Gateway Pods are not restarted automatically when the Istio ConfigMap is updated.
+// This approach helps with rapid progress and prototyping.
```
chance-coleman:
```suggestion
```

I think we can remove this comment and just keep the one above.

src/keycloak/chart/values.yaml
```diff
@@ -322,10 +322,27 @@ autoscaling:
...
   # Deletes the generated truststore to enable recreating it by Keycloak
   deleteGeneratedTrustStore: false
+
+# Third-party integration settings, such as L7 Load Balancer support
+thirdPartyIntegration:
+  tls:
+    # This setting controls deploying additional Envoy Filter that transforms Client Certificate into
+    # a format that Keycloak can use for authentication.
+    # Allowed settings: "default" and "external"
+    tlsTermination: "default"
```
chance-coleman: @mjnagel  pointed this out to me that this value is only used for determining if the envoyFilter should be created, which could also be done via the `tlsCertificateHeader` being present. I would advocate for removing this value if we don't have more use for it or future uses for it.

chance-coleman (commented):
Looking good!

src/keycloak/chart/templates/alb-envoyfilter.yaml
slaskawi: Fixed.

src/keycloak/chart/templates/alb-envoyfilter.yaml
slaskawi: I see your point now! At first I wanted to use it more for the future, but I guess you're right - let's remove it and just use `tlsCertificateHeader`. If someone wants to use the Envoy Filter without actually passing a Client Certs, he can always set it to "dummy" or "ignore" and get all the things setup this way.

src/keycloak/chart/templates/uds-package.yaml
slaskawi: Agreed!

src/keycloak/chart/values.yaml
slaskawi: Agreed!

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -165,26 +165,30 @@ spec:
                 prefix: /metrics
           redirect:
             uri: "/realms/{{ .Values.realm }}/account"
+          {{- if .Values.thirdPartyIntegration.tls.tlsCertificateHeader }}
```
mjnagel: Should these ones all be `if not`?

slaskawi: Yes, just found it in UDS Identity Config test pipeline! Thanks for spotting this!

mjnagel (approved):

PR #1671, opened 2025-06-25
docs: operator resource tree

### LINKED ISSUE #957: Diagram for Operator Resource Tree
Our operator has significant responsibility in the UDS stack and creates a number of child resources during reconciliation of a Package. We would like to document/diagram visually the resource stack that corresponds with pieces of a `Package` spec, i.e. a given `expose` entry will create a `virtualservice` and `serviceentry`. This diagram should be created in code ideally, or at least in an easily editable format so that it can be updated in the future when pieces of the Package spec change.

Stylistically it may be helpful to review [this diagram](https://raw.githubusercontent.com/defenseunicorns/pepr/63aebabb135099d523fe08dc220fab84027e1dc5/_images/pepr-arch.svg) from Pepr which has a similar style of drawing connections from the "code" (in our case `Package` spec) to the effect (in our case created resources).

Definition of done: Diagram of what is produced from the Package reconciliation, mapped visually to the spec of the Package CR.

mjnagel:
## Description

Adds an operator resource tree diagram created with d2 for maintainability. Due to how d2 integration works with Astro I went the route of committing the generated PNG rather than having the docs site build it.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/957

## Type of change


chance-coleman (approved):

slaskawi:
The changes LGTM but could I ask to open a Draft PR against UDS Documentation and change the UDS Core branch to this one? This way we could see the preview. I would love to double check if these colors play nicely with rest of the docs.

mjnagel:
> The changes LGTM but could I ask to open a Draft PR against UDS Documentation and change the UDS Core branch to this one? This way we could see the preview. I would love to double check if these colors play nicely with rest of the docs.

@slaskawi got this on a deploy preview here: https://deploy-preview-265--uds.netlify.app/reference/configuration/uds-operator/resource-tree/

There is a bit of an issue with timing on the merge + the preview because this doc references the image as if it were on main. If you inspect element and change the image ref/url to https://github.com/defenseunicorns/uds-core/blob/operator-resource-tree/docs/.images/diagrams/operator-resource-tree.png?raw=true (which has the branch rather than main) it should show up. I also normally locally run `uds run dev-docs` to test out my changes. One thing I will note on the colors - this uses a white background which blends seamlessly with light mode. In darkmode it's obviously a bit brighter than other content but I think this might be an acceptable compromise. This uses the [company color palette](https://www.notion.so/defense-unicorns/18ae512f24fc805bbf93ebb1269c0fde?v=18ae512f24fc80e593f5000ce312107b) and I almost exhausted all the color options with the various resource types 😅

slaskawi (approved):

PR #1672, opened 2025-06-26
chore(ci): check for autogenerated changes

### LINKED ISSUE #1625: CI checks for "autogenerated" changes
In UDS Core there are a few places where we have auto-generated files/changes. If a developer forgets to run certain tasks to regenerate these, we do not currently have an automated mechanism of checking for these items being missed. In particular:
- The pepr [gen-crds task](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/tasks.yaml#L8) which regenerates types and docs for our CRDs
- The keycloak [caccert task](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/tasks.yaml#L27) which updates our cacert value for the gateways

In each case we should add a CI check, likely to our [lint check](https://github.com/defenseunicorns/uds-core/blob/main/.github/actions/lint-check/action.yaml), that runs these tasks and then fails CI if any git diffs are detected. This will ensure that we have updated any autogenerated changes if CI is passing.

mjnagel:
## Description

Extends our lint check to also run the "generated changes" checks. For now this includes:
- gen-crds
- cacerts

After running these tasks, the git diffs are checked with `git status -s` and CI is failed if diffs are detected. To ensure that the schema generation worked as expected I also pinned KFC to the latest version.

Other changes:
- fixed docs generation (wrong folder path)
- added KFC to support-deps

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1625

## Type of change

## Steps to Validate

This is purely a CI change for the most part. It would be good to run through the changes to the action locally on a bash shell and make sure they behave as expected (for example if you forget to run one of the generated changes commands).

## Checklist before merging


.github/actions/setup/action.yaml
```diff
@@ -40,11 +40,11 @@ runs:
         # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
         version: v0.27.7
 
-    - name: Install Lula
-      uses: defenseunicorns/lula-action/setup@badad8c4b1570095f57e66ffd62664847698a3b9 # v0.0.1
-      with:
-        # renovate: datasource=github-tags depName=defenseunicorns/lula versioning=semver-coerced
-        version: v0.16.0
+    # - name: Install Lula
```
slaskawi: Shall we just remove this?

slaskawi (dismissed):

.github/actions/setup/action.yaml
mjnagel: Agreed, done!

chance-coleman (approved):

PR #1697, opened 2025-07-07
feat: ambient egress

### LINKED ISSUE #1601: Support istio egress control for workloads in ambient mode

In https://github.com/defenseunicorns/uds-core/pull/1331 we have added per-host + per workload egress controls with Istio to restrict the outbound traffic to a known host. This implementation was built out while ambient was still in progress so it did not support the way ambient workloads would need to have egress controls setup.

When I deploy a workload in ambient mode, I would like to be able to restrict egress to only a known hostname.

The operator should dynamically handle the right implementation based on the service mesh mode being used by the package.

I can switch my workload to sidecar mode instead, however ambient is ideal for resource gains and other reasons.

### Additional context

https://www.solo.io/blog/egress-gateways-made-easy

https://www.github.com/istio/istio/issues/52532

meganwolf0:
## Description

Adds egress optionality for ambient workloads per [Technical Design Doc](https://www.notion.so/defense-unicorns/Ambient-Egress-21ce512f24fc80598f59f73462dadbbb?source=copy_link) 

The PR includes: 
* Refactor of `egress` functionality into clearly named `egress-sidecar` and `egress-ambient` files for clarity in function organization and testing
* Added unit testing for ambient functionality and some unit test refactor for existing egress/egress-sidecar
* Addition of e2e test for ambient workload
* Documentation update to reflect optionality for ambient

Notes to reviewers:
* Currently re-arranged the `uds-istio-config` (ingress) and new `uds-istio-egress-config` into `src/istio/charts`. The global config exists elsewhere in `src/istio/common/chart`... This now feels like sort of a funny organization structure, but just calling it out
* Added ambient egress e2e app and package as a default test - sidecar egress is still optional/doesn't run for slim-dev.
* The network test for ambient was acting a little flakey, added a retry and "wait" prior to test execution because it seemed like the waypoint might not have actually been "ready"(?) but honestly had a hard time pinning it down.

## Related Issue

Fixes #1601 

## Type of change

### Simple Ambient Apply
* Apply package with ambient workload requesting egress

### Sidecar -> Ambient; Ambient -> Sidecar 
* Apply a package in sidecar mode with egress, validate egress comes up; Swap package to ambient mode; validate sidecar resources removed and ambient egress comes up

### Batch Apply
* Apply batch packages in ambient (see sidecar package examples); Manipulate (e.g., swap one to sidecar, delete some) and validate outcomes

## Checklist before merging


src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -69,6 +69,10 @@ export async function packageReconciler(pkg: UDSPackage) {
     await new Promise(resolve => setTimeout(resolve, backOffSeconds * 1000));
   }
 
+  // Migrate the package to the latest version
+  migrate(pkg);
```
chance-coleman: Probably just missing it, why is it necessary for `migrate(pkg)` to be called twice?

src/pepr/operator/controllers/utils.ts
```diff
@@ -212,3 +212,23 @@ export async function createEvent(
...
+      if (missingAllowed) {
+        return null;
+      } else {
+        throw e;
+      }
+    } else {
+      throw e;
+    }
+  }
+}
```
chance-coleman, line 234: I don't know if this is something that should be inlcuded in this PR but there is a few places that we could use this util instead of the direct k8s api calls.

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -4,13 +4,30 @@ title: Istio Egress
```
chance-coleman, line 6: won't let me make a comment on the line above, we need to add this:
```md
tableOfContents:
  maxHeadingLevel: 4
```

inside of the frontmatter so that the limitation sections are captured in the sidebar navigation.

src/pepr/operator/controllers/istio/waypoint.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+import { K8sGatewayFromType, K8sGateway } from "../../crd";
+import { getSharedAnnotationKey } from "./istio-resources";
+import { ambientEgressNamespace, sharedEgressPkgId } from "./egress-ambient";
+
+export const waypointName = "egress-waypoint";
+
+// Generate Waypoint for ambient egress
+export function generateWaypoint(pkgs: Set<string>, generation: number) {
```
chance-coleman: Can we make use of the ambient-waypoint.ts [createWaypointGateway](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/istio/ambient-waypoint.ts#L57) function for this, albiet with a refactor probably.

src/pepr/operator/controllers/istio/service-entry.ts
```diff
@@ -187,7 +200,12 @@ export function generateSharedServiceEntry(
   return serviceEntry;
 }
 
-function generateLocalEgressSEName(pkgName: string, portProtocol: PortProtocol[], host: string) {
+// TODO: add a test
```
chance-coleman: Still relevant?

src/pepr/operator/controllers/istio/service-entry.ts
```diff
@@ -93,6 +98,7 @@ export function generateSEName(pkgName: string, expose: Expose) {
   return name;
 }
 
+// TODO: Update test for ambient case
```
chance-coleman: still relevant?

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -110,16 +109,19 @@ export async function istioResources(pkg: UDSPackage, namespace: string) {
   return [...hosts];
 }
 
+// TODO: Update tests for this function
```
chance-coleman: Based on code coverage this comment seems to be relevant, ambient test case is needed.

chance-coleman (commented):
Still working my way through some of the bigger changes but had some early comments:

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -67,13 +68,17 @@ function getPeers(policy: Allow): V1NetworkPolicyPeer[] {
   } else if (policy.remoteCidr !== undefined) {
     peers = [remoteCidr(policy.remoteCidr)];
   } else if (policy.remoteHost) {
-    peers = [egressGateway];
+    if (istioMode === IstioState.Ambient) {
```
chance-coleman: Probably be good to add a test for this ambient functionality.

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
chance-coleman, line 1: topic for discussion: we've thrown around the idea in the past of adding ambient and sidecar directories in the controllers/istio folder. With these distinct paths for ambient and sidecar, maybe now is a good time to do this for a cleaner structure.

src/istio/charts/uds-istio-egress-config/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
# Copyright 2025 Defense Unicorns
```

src/istio/charts/uds-istio-egress-config/templates/configmap.yaml
```diff
@@ -0,0 +1,61 @@
+# Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
# Copyright 2025 Defense Unicorns
```

src/istio/charts/uds-istio-egress-config/values.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
# Copyright 2025 Defense Unicorns
```

src/istio/charts/uds-istio-egress-config/values.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+  # minAvailable and maxUnavailable are mutually exclusive, minAvailable takes precedence
+  podDisruptionBudget:
+    # minAvailable: 1
+    # maxUnavailable: 1
+
+  service:
+    # ports: []
+
+  serviceAccount:
```
joelmccoy: unclear to me by looking at this value what `serviceAccount` is for.  

is this the name of a predefined serviceAccount (if so, maybe we should have an option to create a service account in this chart as a default instead)

or 

is this the name of a service account that will be created by the operator logic when reading in the config.  if so, maybe we can clarify with a comment here that that is the case.

or

is this a full serviceAccount spec that will be deployed by the operator logic.  if so, it might be good to clarify with comments.

src/pepr/operator/controllers/istio/auth-policy.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
 * Copyright 2025 Defense Unicorns
```

src/pepr/operator/controllers/istio/auth-policy.ts
```diff
@@ -0,0 +1,58 @@
...
+      targetRef: {
+        group: "networking.istio.io",
+        kind: "ServiceEntry",
+        name: serviceEntryName,
+      },
+    },
+  };
+
+  return authPolicy;
+}
```
joelmccoy: there is a tiny bit of overlap with this and [here](https://github.com/defenseunicorns/uds-core/blob/27e482fc8236d57e727a5e19d16522e3ae4b481a/src/pepr/operator/controllers/network/authorizationPolicies.ts#L164-L204).  

I wonder if we could consolidate into a util function for building auth policies

src/pepr/operator/controllers/istio/auth-policy.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Copyright 2024 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
+import { V1OwnerReference } from "@kubernetes/client-node";
+import { IstioAuthorizationPolicy, IstioAction } from "../../crd";
+import { sanitizeResourceName } from "../utils";
+
+// Generate Authorization Policy for ambient egress
+export function generateAuthorizationPolicy(
```
joelmccoy: this function seems specific to generated auth policies for ambient egress. we generate auth policies elsewhere.  would recommend naming this more specifically like `generateAmbientEgressAuthorizationPolicy`.

src/pepr/operator/controllers/istio/auth-policy.ts
```diff
@@ -0,0 +1,58 @@
...
+      ownerReferences: ownerRefs,
+    },
+    spec: {
+      action: IstioAction.Allow,
+      rules: [
+        {
+          from: [
+            {
+              source: {
+                serviceAccounts: [serviceAccount],
```
joelmccoy: is there a reason why we chose to use `serviceAccounts` instead of the full [principal](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source-principals) like we do for `remoteServiceAccount`: https://github.com/defenseunicorns/uds-core/blob/27e482fc8236d57e727a5e19d16522e3ae4b481a/src/pepr/operator/controllers/network/authorizationPolicies.ts#L97

src/pepr/operator/controllers/istio/auth-policy.ts
```diff
@@ -0,0 +1,58 @@
...
+        kind: "ServiceEntry",
+        name: serviceEntryName,
+      },
+    },
+  };
+
+  return authPolicy;
+}
+
+function generateAuthorizationPolicyName(host: string, serviceAccount: string) {
```
joelmccoy: This function seems specific to generating names for egress AuthorizationPolicies. we generate AuthorizationPolicies elsewhere in the operator so it may be good to be a bit more specific with this function name.

joelmccoy: might be good to add some simple tests for this function

src/pepr/operator/controllers/istio/auth-policy.spec.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
 * Copyright 2025 Defense Unicorns
```

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
joelmccoy: I like that idea. probably not necessary for this PR as it would add too much noise.  I think separating them by filename is sane for now. and maybe we can add a follow on issue to refactor?

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
```diff
@@ -0,0 +1,219 @@
+/**
+ * Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
 * Copyright 2025 Defense Unicorns
```

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -0,0 +1,101 @@
+/**
+ * Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
 * Copyright 2025 Defense Unicorns
```

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -2,63 +2,71 @@
...
+import { IstioState } from "./namespace";
+import { HostPortsProtocol, HostResourceMap, PackageAction, PackageHostMap } from "./types";
+
+// Cache for in-memory sidecar-only shared egress resources from package CRs
 export const inMemoryPackageMap: PackageHostMap = {};
 
 // Lock to prevent concurrent updates to the inMemoryPackageMap
 let lock = false;
 // eslint-disable-next-line prefer-const
 let lockQueue: (() => void)[] = [];
```
joelmccoy: does it make sense to rename these variables to have `sidecar` in the name like the ambient ones below?

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -110,16 +109,19 @@ export async function istioResources(pkg: UDSPackage, namespace: string) {
...
+ * @param allowList
  * @param pkg
  * @param pkgName
  * @param namespace
  * @param generation
  * @param ownerRefs
+ * @param ambient
  */
 export async function istioEgressResources(
   hostResourceMap: HostResourceMap | undefined,
```
joelmccoy: nit: Consider consolidating hostResourceMap, allowList, pkgId, pkgName, etc., into a single argument derived from pkg. This would reduce repetition and simplify the function signature. You could move the logic from [lines 93–95](https://github.com/defenseunicorns/uds-core/blob/979d251e44e35784b86153e8df1c487cf3a6dc5c/src/pepr/operator/controllers/istio/istio-resources.ts#L93-L95) into the istioEgressResources function itself to encapsulate this logic.

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -129,53 +131,56 @@ export async function istioEgressResources(
...
-        host,
-        hostResourceMap[host],
+    if (istioMode === IstioState.Ambient) {
+      // Validate existing egress waypoint namespace
+      try {
+        await validateNamespace(ambientEgressNamespace);
+      } catch (e) {
+        let errText = `Unable to get the egress waypoint namespace ${ambientEgressNamespace}.`;
+        if (e?.status == 404) {
+          errText = `Egress waypoint is not enabled in the cluster. Please enable the egress waypoint and retry.`;
```
joelmccoy: Is this error message trying to give feedback to the user that the `istio-egress-waypoint` component may not be deployed (i.e. the one that creates this namespace)?  If so, does it make sense to word this like so
```suggestion
          errText = `The '${ambientEgressNamespace}' namespace was not found. Ensure the 'istio-egress-waypoint' component is deployed and try again.`;
```

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -0,0 +1,101 @@
...
+      true,
+    );
+
+    log.debug(serviceEntry, `Applying Service Entry ${serviceEntry.metadata?.name}`);
+
+    // Apply the ServiceEntry and force overwrite any existing resource
+    await K8s(IstioServiceEntry).Apply(serviceEntry, { force: true });
+  }
+
+  // Create Authorization Policy for service entry, if serviceAccount is specified
```
joelmccoy: this comment leads me to believe that we don't create a AuthorizationPolicy when a serviceAccount is not specified. However we do use the default service account if not specified correct?  Does this wording make more sense?
```suggestion
  // Create Authorization Policy for service entry, use specified serviceAccount or use "default" if no serviceAccount specified
```

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -0,0 +1,101 @@
...
+  for (const allow of egressRequested) {
+    const serviceAccount = allow.serviceAccount ?? "default";
```
joelmccoy: thinking out loud here... what would happen if the user specified a serviceAccount as `<namespace>/<serviceAccount>` like specified in the docs [here](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source-service_accounts).  I think we would want to prevent someone to specify a service account from a different namespace??  We would need some validation here or earlier in the chain.

All that being said... I think we should really generate the whole principal name and use [principals](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source-principals) like we do for our auth policies for remoteServiceAccount target [here](https://github.com/defenseunicorns/uds-core/blob/979d251e44e35784b86153e8df1c487cf3a6dc5c/src/pepr/operator/controllers/network/authorizationPolicies.ts#L97).

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -123,6 +123,12 @@ const allow = {
           Only valid for Ingress rules.",
         type: "string",
       },
+      serviceAccount: {
+        description:
+          "The service account to restrict outgoing traffic from within the package namespace. \
+          Only valid for Egress rules.",
```
joelmccoy, line 129: Based on this PR, isn't a more accurate description that we only apply the source `serviceAccount` for Egress rules with `remoteHost` targets?

We create [authpolicies](https://github.com/defenseunicorns/uds-core/blob/979d251e44e35784b86153e8df1c487cf3a6dc5c/src/pepr/operator/controllers/network/authorizationPolicies.ts#L164) for other egress rules but I don't think this PR adds the serviceAccount to that logic.  Maybe we should incorporate this field into that logic??  Thoughts here @mjnagel / @chance-coleman ?

joelmccoy (commented):
Haven't made it through all the code yet (or tested this locally), but wanted to drop some early comments on the code I have reviewed so far to provide some earlier iterative feedback.

src/pepr/operator/reconcilers/package-reconciler.ts
meganwolf0: that was a bad merge, I'll remove it

src/pepr/operator/controllers/istio/waypoint.ts
meganwolf0: hrm, looking at this function it looks different enough that I would think we'd possibly still want a separate function for the ambient egress waypoint? I can move the `generateWaypoint` function I have to this file though

docs/reference/configuration/service-mesh/egress.md
meganwolf0: added

src/pepr/operator/controllers/istio/service-entry.ts
meganwolf0: removed

src/pepr/operator/controllers/istio/service-entry.ts
meganwolf0: removed

src/pepr/operator/controllers/istio/istio-resources.ts
meganwolf0: added tests here

src/pepr/operator/controllers/network/generate.ts
meganwolf0: added tests here (for sidecar too since that was a previous gap)

src/istio/charts/uds-istio-egress-config/values.yaml
meganwolf0: yeah sorry this bit is maybe kinda confusing, basically this just gets templated through to the configmap, which is called by the waypoint and the resources it specifies will be created (deployment, service, serviceAccount are created by default). This is mainly so that some settings/configs can be passed through to those resources if desired. Tried to add more words at the top + doc link to hopefully better explain.

Also Micah asked for more specific configs for some of the elements (I skipped serviceAccount) so that's probably why that one in particular looks a little funny. If there are some suggestions on things that should be locked down for that resource, happy to add those in!

src/pepr/operator/controllers/istio/istio-resources.ts
meganwolf0: hm I could derive all those entries from pkg in `istioEgressResources` but I'd also be re-deriving several that are already in the calling function (albeit they aren't like costly calculations)... Might crowd source this one to see if others would also prefer that change

src/pepr/operator/controllers/istio/egress-ambient.ts
meganwolf0: yep nice catch! that comment reflects old logic

src/pepr/operator/controllers/istio/egress-ambient.ts
meganwolf0: > what would happen if the user specified a serviceAccount as <namespace>/<serviceAccount>

I don't think there'd be an issue, but yeah I think that's probably one of those things that's on the user to understand what they're specifying (like if someone specifies a selector that's inaccurate, there's no logic to like bounce that, it just won't work). Alternatively this does offer a pathway to allow another namespaced service account, if that was how someones stuff was set up 🤷‍♀️ 

> I think we should really generate the whole principal name and use [principals](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source-principals) like we do for our auth policies for remoteServiceAccount target

I did see the way you all were specifying principals, but you're still templating it out with namespace and serviceAccount, so it's almost exactly the same as just using the `serviceAccount` key itself, I think? If you were getting the full principal I could see why this would benefit, but since that's not the input solicited from the user/package CR I'm not sure I see the benefit. (There could very well be one but just curious if there is??)

src/pepr/operator/controllers/istio/istio-resources.ts
meganwolf0: updated

src/pepr/operator/controllers/istio/auth-policy.ts
meganwolf0: fixed

src/istio/charts/uds-istio-egress-config/values.yaml
meganwolf0: fixed

src/istio/charts/uds-istio-egress-config/templates/configmap.yaml
meganwolf0: fixed

src/istio/charts/uds-istio-egress-config/Chart.yaml
meganwolf0: fixed

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -64,6 +87,42 @@ Currently, only HTTP and TLS protocols are supported. The configuration will def
...
+  namespace: egress-gw-1
+spec:
+  network:
+    serviceMesh:
+      mode: ambient
+    allow:
+      - description: "Example Curl"
+        direction: Egress
+        port: 443
+        remoteHost: httpbin.org
```
slaskawi, line 108: Is it possible to use egress for internal services - for example pointing it to a Gitlab deployed inside of the cluster but without using Packages. I'm thinking about large clusters where not everything is controlled by UDS. I'm not sure if this is a valid approach through.

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -118,3 +173,12 @@ Additional security considerations to keep in mind when implementing egress:
 * Some potential vulnerabilities are introduced using TLS Passthrough - you’ll need to know what’s on the other side of that domain because of [domain fronting](https://en.wikipedia.org/wiki/Domain_fronting) - Essentially, this is only a safe feature for trusted hosts, or hosts you know are not vulnerable to domain fronting
 
 * We are not blocking DNS exfiltration
+
+## Debugging Tips
```
slaskawi: Nitpick: Maybe calling troubleshooting would be better?

src/istio/charts/uds-istio-egress-config/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
```
slaskawi, line 1: Do we need this file?

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -67,13 +68,17 @@ function getPeers(policy: Allow): V1NetworkPolicyPeer[] {
...
+    } else {
+      peers = [egressGateway];
+    }
   }
 
   return peers;
 }
 
-export function generate(namespace: string, policy: Allow): kind.NetworkPolicy {
+export function generate(namespace: string, policy: Allow, istioMode?: string): kind.NetworkPolicy {
```
slaskawi: Is it possible to check the Ambient/Sidecar mode deeper in the call stack and avoid adding it to the top-level functions? It feels this type of thing should be checked by some utility when it's needed to avoid polluting method signatures.

src/pepr/operator/controllers/istio/istio-resources.ts
slaskawi: I also have a similar comment to @joelmccoy here: https://github.com/defenseunicorns/uds-core/pull/1697/files#r2272979110

It seems the Ambient/Sidecar check pollutes the method signatures all over the place. Can we try to somehow limit that?

slaskawi (commented):
I've just reviewed the PR focusing on API and general usability. 

I think we're on-track with it and from my review I noticed the TypeScript code method  got polluted by the Sidecar/Ambient checks. There were also some comments about this from others and I think we might want to refactor this.

The other thing I was wondering (and there's an inline comment about this) is whether it is possible to use this functionality for in-cluster deployments.

Otherwise, it's almost LGTM!

docs/reference/configuration/service-mesh/egress.md
meganwolf0: Hm an interesting question... I would assume that if the service was registered in the cluster DNS then that should work. I can try to figure this out if that's something of interest! As for the validity of that approach, I suppose I'd think that was down to what the user was trying to achieve - for instance I'd think it could be valid if they want to track the egressed traffic from their slice of the cluster if it was like a multi-tenant environment

docs/reference/configuration/service-mesh/egress.md
meganwolf0: Good call, that's a much better header. Will update!

src/istio/charts/uds-istio-egress-config/.helmignore
meganwolf0: Possibly no, I was following the structure laid out in the existing uds-istio-config chart, and it also contains the `.helmignore`. Happy to remove though if the consistency thing is unimportant!

src/pepr/operator/controllers/istio/istio-resources.ts
meganwolf0: Sounds good, 2 beats 1, let me do a bit of refactor here :)

docs/reference/configuration/service-mesh/egress.md
slaskawi: Multi-tenant environments are quite interesting! So no pressure, but it would be great to at least know if this approach would work in-cluster or not and document it accordingly.

src/pepr/operator/controllers/istio/egress-ambient.ts
joelmccoy: I think there are really a couple small benefits of using the full principal IMO:
- Explicit trust domain established (not important now, but could be important for any multi-cluster work)
- Consistency in the code base for generation of auth policies
- We would template out the namespace to so that a user couldn't provide a namespace to the service account (i'm not sure there is a use case where the serviceaccount wouldn't be coming from the target namespace of the package)

But I don't think these are crazy huge advantages and would defer to others on this.

src/istio/charts/uds-istio-egress-config/values.yaml
joelmccoy: I think i understand the flow of how things are being loaded in with the operator.  But I'm just a little bit confused on the expected schema of what should be passed in the `serviceAccount` value here.  Is it a full ServiceAccount k8s manifest, just the `spec` of a ServiceAccount, the name of an existing `serviceAccount`, or certain properties for a new ServiceAccount?

I can dig through the operator code to see what can be passed in, but as an end user I'm a bit confused as to what needs to go in here.

So i guess my questions are:
1. What is exactly is expected to be passed into here.
2. Might be worth adding some clarity as comments explaining 1.

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -69,6 +69,7 @@ export async function packageReconciler(pkg: UDSPackage) {
     await new Promise(resolve => setTimeout(resolve, backOffSeconds * 1000));
   }
 
+  // Configure the namespace and namespace-wide network policies
```
joelmccoy: is this a rouge comment or am I missing some context here?

src/pepr/operator/controllers/istio/auth-policy.ts
joelmccoy, line 1: i would really recommend either:

1. merging this with our auth policy generation here:https://github.com/defenseunicorns/uds-core/blob/27e482fc8236d57e727a5e19d16522e3ae4b481a/src/pepr/operator/controllers/network/authorizationPolicies.ts#L164-L204
2. make this logic/naming specific to egress.

personally would prefer 1 if there is a good amount of overlap between the logic here.

src/pepr/operator/controllers/network/generate.ts
meganwolf0: Did a bit of refactor on the `istioResources` and `istioEgressResources` signatures that had the unnecessary `istioMode` ... however I'm not sure on the remediation for this function, since istioMode can't be derived otherwise. Open to suggestions!

src/istio/charts/uds-istio-egress-config/values.yaml
meganwolf0: Fair questions, I added a comment above it to address #1 - "Map of values that will be merged via Strategic Merge Patch strategy to the generated serviceAccount", and below it is a sample. Basically I think you can merge both metadata fields and spec ones. I'm truly not exactly sure what you'd want to change with the service account, so I left it fully open. However if you all want to lock it down (say only add annotations) that can be templated. I honestly don't have a ton of experience doing direct service account configurations so idk what would actually be appropriate/useful there to expose 🤷

src/pepr/operator/controllers/istio/istio-resources.ts
meganwolf0: updated this, take a look and let me know if this is what you all had in mind

src/pepr/operator/crd/sources/package/v1alpha1.ts
mjnagel: > We create [authpolicies](https://github.com/defenseunicorns/uds-core/blob/979d251e44e35784b86153e8df1c487cf3a6dc5c/src/pepr/operator/controllers/network/authorizationPolicies.ts#L164) for other egress rules but I don't think this PR adds the serviceAccount to that logic. Maybe we should incorporate this field into that logic??

I might be forgetting a different change, but current authpolicy logic is [entirely _Ingress_ focused](https://github.com/defenseunicorns/uds-core/blob/979d251e44e35784b86153e8df1c487cf3a6dc5c/src/pepr/operator/controllers/network/authorizationPolicies.ts#L225) I believe. The linked function for example is used to allow ingress from a remote service account -> a pod (with the destination being selected via labels rather than service account).

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -1,16 +1,35 @@
...
+  istio-egress-waypoint:
+    uds-istio-egress-config:
+      values:
+        - path: "config.deployment.replicas"
+          value: 4
+        - path: "config.horizontalPodAutoscaler.minReplicas"
+          value: 2
+```
+
+See the values.yaml for additional details and configuration options.
```
mjnagel: nit: link here to the values might be nice?

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -64,6 +87,42 @@ Currently, only HTTP and TLS protocols are supported. The configuration will def
...
+        direction: Egress
+        port: 443
+        remoteHost: httpbin.org
+        remoteProtocol: TLS
+        selector:
+          app: curl
+        serviceAccount: curl
+```
+
+When a Package CR specifies the `network.allow` field with, at minimum, the `remoteHost` and `port` or `ports` parameters, the UDS Core operator will create the necessary Istio resources to allow traffic to egress from the mesh. For ambient, the `serviceAccount` can be specified for an additional layer of security. The resources that are created include the following:
```
mjnagel:
```suggestion
When a Package CR specifies the `network.allow` field with, at minimum, the `remoteHost` and `port` or `ports` parameters, the UDS Operator will create the necessary Istio resources to allow traffic to egress from the mesh. For ambient, the `serviceAccount` can be specified for an additional layer of security. The resources that are created include the following:
```

small nit, I think we refer to the operator as UDS Operator in other docs

src/istio/charts/uds-istio-egress-config/values.yaml
```diff
@@ -0,0 +1,42 @@
...
+  podDisruptionBudget:
+    # minAvailable: 1
+    # maxUnavailable: 1
+
+  service:
+    # ports: []
+
+  # Stringified yaml that will be merged via Strategic Merge Patch strategy to the generated serviceAccount
+  # e.g.,
+  # serviceAccout: |
```
joelmccoy:
```suggestion
  # serviceAccount: |
```

docs/reference/configuration/service-mesh/egress.md
mjnagel: Given the below note `If no `serviceAccount` is specified, the `default` service account is used` - I might also rephrase "can be specified for an additional layer of security" to something more explicit about it being needed if your workload is not using the default service account?

packages/base/zarf.yaml
```diff
@@ -60,6 +60,11 @@ components:
     import:
       path: ../../src/istio
 
+  - name: istio-egress-waypoint
+    required: true
+    import:
+      path: ../../src/istio
+
```
mjnagel: very nit, but order wise might make sense to move this below passthrough to group all the ingress gateways first and egress gateway/waypoint after (both here and in standard package + istio zarf files)

src/istio/charts/uds-istio-egress-config/.helmignore
mjnagel: `.helmignore` is a pretty standard file. While it's probably not _needed_ here I'd just leave it since it comes default with `helm create chart` even.

src/istio/charts/uds-istio-egress-config/values.yaml
```diff
@@ -0,0 +1,42 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+# Egress Waypoint Configuration
+# The following configuration settings reference the resources created from adding an Istio Waypoint
+# These configurations will be overlaid on top of the generated resources using a Strategic Merge Patch strategy
+# See documentation: https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#automated-deployment
+config:
+
+  deployment:
```
mjnagel: just a nit, but I prefer values to show the "type" even when empty - i.e. add empty maps (`{}`) to all of these.

src/istio/charts/uds-istio-egress-config/values.yaml
mjnagel: I think the pipe is invalid here for the values (while it is the right format for the configmap, the values just expect a map rather than multi-line string)?

src/istio/zarf.yaml
```diff
@@ -112,11 +112,19 @@ components:
         namespace: istio-tenant-gateway
       - name: uds-istio-config
         version: 0.2.0
-        localPath: chart
+        localPath: charts/uds-istio-config
         namespace: istio-tenant-gateway
         valuesFiles:
           - "values/config-tenant.yaml"
 
+  - name: istio-egress-waypoint
```
mjnagel: I'm wondering if the naming of this will be confusing. Maybe documentation around these is sufficient but wondering if we should name them `egress-ambient` and `egress-sidecar` instead of the current waypoint/gateway? Would be a breaking change for the gateway/sidecar one though.

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    log.debug(serviceEntry, `Applying Service Entry ${serviceEntry.metadata?.name}`);
+
+    // Apply the ServiceEntry and force overwrite any existing resource
+    await K8s(IstioServiceEntry).Apply(serviceEntry, { force: true });
+  }
+
+  // Create Authorization Policy for service entry, use specified serviceAccount or use "default" if no serviceAccount specified
+  for (const allow of egressRequested) {
+    const serviceAccount = allow.serviceAccount ?? "default";
```
mjnagel: I'm not a _huge_ fan of the `default` service account fallback. Two alternatives to think through maybe:
- Fallback to a more open approach, using "from namespace" instead of principal. This is more open/less secure, but it does match how we handle this on the Ingress side for authpols. However - that was a different situation where we wanted backwards compatibility with netpols, etc.
- Make `serviceAccount` required at validation time if using ambient + egress + remoteHost. This would keep the same security (always have a principal/service account) but add a new required field. Given this feature is net-new, I think a new requirement is fine.

I would lean towards the second as a more secure approach, and being more explicit/clear that it is important to specify the service account. A "well architected" application hopefully isn't using the default service account, and the default fallback might be unexpected? Hopefully someone reads the docs and it's clear either way but wanted to open this thread for discussion.

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -7,19 +7,20 @@ import { V1NetworkPolicyPeer, V1NetworkPolicyPort } from "@kubernetes/client-nod
...
 import { remoteCidr } from "./generators/remoteCidr";
-import { egressGateway } from "./generators/egressGateway";
+import { egressGateway, egressWaypoint } from "./generators/egress";
 
 function isWildcardNamespace(namespace: string) {
   return namespace === "" || namespace === "*";
 }
 
-function getPeers(policy: Allow): V1NetworkPolicyPeer[] {
+function getPeers(policy: Allow, istioMode: string | undefined): V1NetworkPolicyPeer[] {
```
mjnagel: nit: `istioMode` should be `IstioState` type instead of string here

src/pepr/operator/controllers/network/generate.ts
mjnagel: nit: `istioMode` should be `IstioState` type instead of string here

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -1,64 +1,72 @@
...
+let ambientGeneration = 0;
 
-// reconcileEgressResources reconciles the egress resources based on the config
+// reconcileSharedEgressResources reconciles the egress resources based on the config
+// Handles mode transitions by updating both sidecar and ambient in-memory maps appropriately
 export async function reconcileSharedEgressResources(
   hostResourceMap: HostResourceMap | undefined,
   pkgId: string,
   action: PackageAction,
+  istioMode: string,
```
mjnagel: this should use the `IstioState` type (enum), not string

src/pepr/operator/crd/generated/gateway/gateway-v1.ts
```diff
@@ -0,0 +1,981 @@
+/**
+ * Copyright 2025 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
```
mjnagel: This file is a duplication of `src/pepr/operator/crd/generated/k8s/gateway-v1.ts` (just a different class name + copyright date). Can we nuke this and use the existing `K8sGateway` instead?

mjnagel: Looks like `NetworkGateway` isn't used, so maybe this is just a clean delete.

src/pepr/operator/controllers/network/generate.ts
mjnagel: After going through other files there's a mix of passing around the actual state sometimes and just a bool for ambient other times. Might be good to try and align the approach we take - I think state is nicer maybe, although with only 2 states a bool is fine too 🤷

src/pepr/operator/controllers/istio/egress-ambient.ts
mjnagel: I prefer principal just for consistency sake, but agree on the other two points here too - unless there's a strong argument for `serviceAccount`?

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  // Generate the waypoint payload
+  const waypoint = createEgressWaypointGateway(packageList, generation);
+
+  // Apply waypoint
+  log.debug(waypoint, `Applying Waypoint ${waypoint.metadata?.name}`);
+
+  // Apply the Waypoint and force overwrite any existing resource
+  await K8s(K8sGateway).Apply(waypoint, { force: true });
```
mjnagel: This might be okay, but this diverges a bit from some of the logic used for our waypoints with authservice where we also check health after applying ([ref](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/istio/ambient-waypoint.ts#L41-L42)). I think there'd be value in a quick call to [`waitForWaypointPodHealthy`](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/istio/ambient-waypoint.ts#L152)?

mjnagel (commented):
First pass review - will spend some more time playing with this locally as well. Overall approach looks aligned with design so hopefully these are smaller comments for the most part.

docs/reference/configuration/service-mesh/egress.md
meganwolf0: Ok, got a test set up going and proved that you can access other hosts in-cluster. I'm going to lay out what I did here so you can tell me if you think that's on or off track:
- Deployed modified uds-k3d with 8080 exposed -> added basic uds-slim-dev
- Added a spearate nginx controller / Load balancer component to listen on 8080
- Added an "echo" workload + Ingress for "echo.test" (using nginx-controller)
- Added a curl workload + UDS Package with egress allow to "echo.test", port 80
- Updated cluster coredns with the rewrite of "echo.test" to the nginx-controller service
- Successfully curled the host "echo.test" from my curl workload

If this does accurately mock what you were intending, I can add some additional notes to the egress.md

src/pepr/operator/controllers/istio/egress.spec.ts
```diff
@@ -1405,7 +1283,196 @@ describe("test validateProtocolConflicts", () => {
...
+      "example.com": {
+        portProtocol: [
+          { port: 443, protocol: RemoteProtocol.TLS },
+          { port: 80, protocol: RemoteProtocol.HTTP },
+        ],
+      },
+    });
+  });
+
+  // todo: multiple packages exist that use the port/protocol
```
joelmccoy: I think is done below? can we cleanup this todo?

src/pepr/operator/controllers/network/generators/egress.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
 * Copyright 2025 Defense Unicorns
```

packages/base/zarf.yaml
meganwolf0: ok! My thought process was default/not-default components, but can certainly reorder :)

src/test/app-egress-ambient-package.yaml
```diff
@@ -0,0 +1,53 @@
+# Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
# Copyright 2025 Defense Unicorns
```

src/test/app-egress-ambient.yaml
```diff
@@ -0,0 +1,169 @@
+# Copyright 2024 Defense Unicorns
```
joelmccoy:
```suggestion
# Copyright 2025 Defense Unicorns
```

joelmccoy (commented):
Finally made it all the way through the changes here.  Generally looking good to me.  Some high level things to consider based on comments throughout this PR:

- Consolidation of authorization policy generation into central place or rename new auth policy logic so that it is specific to egress functionality
- Swap to using Principal instead of ServiceAccount in the auth policy generation

Everything else is looking pretty solid. I haven't played around with this locally yet, but will once this PR reaches more of a steady state based on the feedback.

src/istio/charts/uds-istio-egress-config/values.yaml
meganwolf0: yep you're right, fixed and also updated the comment accordingly

packages/base/zarf.yaml
mjnagel: Yeah did consider that as well - ultimately shouldn't matter since there's no dependencies between these components so okay either way.

src/istio/zarf.yaml
meganwolf0: Yeah I think I generally agree with this sentiment, but I'm of two minds here. I'm torn about consistency with naming (e.g., istio-x-gateway components that are also in there). From an API standpoint I think it makes more sense to call the component by the functionality, however, idk that the same grace is given to zarf's ability to reconcile different components that share the same name (i.e., if you wanted to totally redefine what elements compose "egress-ambient" in the future, would that be a clean process for cluster resources). So by that logic might make a bit more sense to keep it tied to the actual cluster component? Food for thought, happy to do egress-ambient/egress-sidecar if you all think that's the right way forward!

src/pepr/operator/controllers/istio/egress-ambient.ts
meganwolf0: Yeah I agree that the second is more secure, I'd maybe also say that if a user wants to switch their CR from sidecar -> ambient, that's could possibly be a trip-wire. You all certainly set the vision for security vs usability though, so yeah other opinions are welcome! If default SA is not great, I'd personally be more inclined to use "from namespace" since that feels a bit more consistent to how UDS Operator handles other unspecified fields (like selector)

src/pepr/operator/controllers/network/generate.ts
meganwolf0: Yeahh to be fair, I was copying [you guys' lead](https://github.com/defenseunicorns/uds-core/blob/6aed2ac48ec14da6644f14192b359aa35ce3e538/src/pepr/operator/reconcilers/package-reconciler.ts#L92C24-L92C39) on the string thing 😄 
I do have one check for a bool ambient field that I can find, so I can likewise change that to state - I do agree that State is nicer - and will modify the strings to be IstioState as well!

mjnagel: Awkward - that was also my change specifically that added it as a string 😭.

src/pepr/operator/controllers/istio/auth-policy.ts
meganwolf0: Alright, so I went down the rabbit hole a bit on what it would take to implement 1 -
* I needed to change a few upstream function signatures, since "pkg" wasn't being passed down - that's fine but it also sorta impacts a bit of the consistency within istioResources (most resources are generated with a similar signature)
* The `selector` and `targetRefs` are mutually exclusive - so the updating the `buildAuthPolicies` function to handle both feels a little like it exceeds single responsibility, and could possibly lead to greater confusion down the line with testing/debugging
* I see the `uds/for` label, along with a purge routine at the end of "generateAuthorizationPolicies"... I feel like that might have some unintended consequences if it's potentially purging egress resources outside its scope

In a nutshell, after going down this route, I think it'd be cleaner and easier to understand/test if these were separate functions. Planning to just rename the function to `generateAmbientEgressAuthorizationPolicy` unless you all want to have additional discussion here/tell me why I'm wrong 😁

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
chance-coleman: opened this [Issue](https://github.com/defenseunicorns/uds-core/issues/1839) to track this work for the future

src/pepr/operator/controllers/utils.ts
chance-coleman: [created this issue to work this in the future](https://github.com/defenseunicorns/uds-core/issues/1840)

src/istio/zarf.yaml
mjnagel: I don't know that this necessarily leans more one way or the other but another consideration is that a waypoint is a gateway - just a K8s Gateway vs Istio Gateway. I'd also point out that the component itself doesn't actually include the waypoint, just a configmap for the waypoint - so it might be more accurately named istio-egress-waypoint-config if we wanted to get really specific with naming based on cluster component/included resources. Or maybe something like `istio-ambient-egress-config` would be okay?

I'm not totally following the concern with zarf reconciliation - there may be some expected issues with helm ownership when moving things between components, or removing them. Historically these have been easy to resolve with Zarf actions though to add the expected helm metadata or delete certain resources.

Ultimately I'm okay with either approach, as mentioned I think documentation can help here so that people know how to override this. I might slightly lean towards renaming the waypoint component to indicate ambient egress somehow, but also would be nice to avoid renaming the existing sidecar one. Not sure if that really helps though. Curious on @chance-coleman / @joelmccoy thoughts here?

docs/reference/configuration/service-mesh/egress.md
slaskawi: I think this approach is fine. Thanks for checking this. 

As this is a very niche use case, I'm leaving it up to you whether the docs should or should not be added.

src/pepr/operator/controllers/network/generate.ts
slaskawi: Ok, understood! I'm fine leaving it as is than.

slaskawi (commented):
This PR is LGTM! 

I'm not approving to give others a chance to review it.

src/pepr/operator/controllers/istio/auth-policy.ts
joelmccoy: Renaming is sufficient IMO.  closing this out

src/pepr/operator/controllers/istio/auth-policy.ts
joelmccoy: closing based on: https://github.com/defenseunicorns/uds-core/pull/1697/files#r2285353609

renaming this to be specific to ambient egress is sufficient.

src/istio/charts/uds-istio-egress-config/values.yaml
joelmccoy: I think the added comments clarify.  closing this out

src/pepr/operator/crd/sources/package/v1alpha1.ts
joelmccoy:
```suggestion
          Only valid for Egress rules using remoteHost.",
```

I believe this would be a more accurate description right?  Technically we could have a egress rule without `RemoteHosts` and this field would do nothing?  Maybe it should do something in the future?

Example- Create an egress rule to `remoteNamespace`, as the logic is written, I think the authpolicy would not enforce the originating serviceAccount correct?

If I am understanding this correctly, I think we should
1. Clarify that serviceAccount is only applicable for egress rules with `remoteHost`
2. Maybe we should create a follow up issue to add `ServiceAccount` in our existing auth policy logic? Would defer to @mjnagel on if this is desirable.

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -25,7 +25,7 @@ overrides:
           value: 2
 ```
 
-See the values.yaml for additional details and configuration options.
+See the [values.yaml](../../../../src/istio/charts/uds-istio-egress-config/values.yaml) for additional details and configuration options.
```
mjnagel: With the docs site this will not link as expected, so it should just be a direct github link on the main branch (will be broken initially, but work once merged).

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -105,7 +105,7 @@ export async function istioResources(pkg: UDSPackage, namespace: string) {
  */
 export async function istioEgressResources(pkg: UDSPackage, namespace: string) {
   // Get package data
-  const istioMode = pkg.spec?.network?.serviceMesh?.mode || Mode.Sidecar;
+  const istioState = getIstioStateFromPackage(pkg);
```
mjnagel: My fault on the comment before - I think we can eliminate `getIstioStateFromPackage` as a helper func and just use `Mode` similar to what you had here before:
```ts
   const istioMode = pkg.spec?.network?.serviceMesh?.mode || Mode.Sidecar;
```
I definitely did a bad typo/comment there saying `IstioState` - intent was more to make sure that our functions like `reconcileSharedEgressResources` have a `Mode` type param, rather than string, and then aligning all the places we check mode with that (rather than the bool for ambient in some spots).

meganwolf0: I updated so the function inputs are now of type `Mode`, hopefully we made it onto the same page there 😅

src/pepr/operator/crd/sources/package/v1alpha1.ts
meganwolf0: Yeah, I think maybe you're right and might be good to add an additional validation here to check if serviceAccount, that remoteHost is also defined, so it's clear they belong together? There is going to be an interesting intersection of logic for all those allow fields...

src/pepr/operator/controllers/istio/istio-resources.ts
mjnagel: LGTM, thanks!

src/istio/zarf.yaml
meganwolf0: > I'm not totally following the concern with zarf reconciliation

Yeah I think you probably got it with things that have to be done with actions to clean up old resources - essentially my thought process was if you had a package that had guts that changed (i.e., was composed of different components/charts) and you do a `zarf p deploy` of v1 then things change when you `zarf p deploy` v2, the v1 things don't necessarily get removed. I assume the user still needs some understanding of the composition of those things... but really just a thought I had with the concept of abstract naming in particular, might not be a big deal.

Also good to rename to `istio-ambient-egress-config` as well, you all let me know what you decide there :)

joelmccoy (commented):
Played around with this locally today and generally looking good. I did notice bug that will probably need addressed. I haven't dug into the code to figure out what the root cause is, but suspect it has to do with the name generation/cleanup of auth policies.

- Bug: When you add/remove the serviceAccount field to a package it doesn't cleanup the old AuthorizationPolicy: 

Start with:

```yaml
spec:
  network:
    allow:
    - description: Github API Curl
      direction: Egress
      ports:
      - 443
      remoteHost: api.github.com
      remoteProtocol: TLS
      selector:
        app: curl
    serviceMesh:
      mode: ambient
```

The following auth policy is created:

```yaml
api-github-com-egress ALLOW
```

Switch to add a service Account:

```yaml
spec:
  network:
    allow:
    - description: Github API Curl
      direction: Egress
      ports:
      - 443
      remoteHost: api.github.com
      remoteProtocol: TLS
      selector:
        app: curl
      serviceAccount: curl-sa
    serviceMesh:
      mode: ambient
```

The old authorization policy does not get cleaned up and allows traffic even though a serviceAccount is not attached to the pod.  I think this might have to do with the naming switch when adding/removing a service account

```yaml
api-github-com-curl-sa-egress  ALLOW
api-github-com-egress          ALLOW                
```

src/pepr/operator/reconcilers/package-reconciler.ts
joelmccoy, line 1: Also noticed when testing locally that we keep track of the number of authorization policies we have deployed in a package. see: https://github.com/defenseunicorns/uds-core/blob/ea670f1559dca603bfd42a77c0da4e63e84819c2/src/pepr/operator/reconcilers/package-reconciler.ts#L129

Should we also include the auth policies generated from the egress work in this count?  cc: @mjnagel

src/pepr/operator/crd/sources/package/v1alpha1.ts
joelmccoy: Yeah a validation actually might be a good idea.  Block use of ServiceAccount unless using remote host.  And then maybe in the future if we wanted to extend the serviceaccount logic we could tie that into our other egress authpolicies

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -1,16 +1,35 @@
...
-:::note
-This does not currently work with ambient mode enabled (`spec.network.serviceMesh.mode=ambient`) or with workloads that omit the sidecar proxy
-:::
-
 ## Configuring the Egress Workload
 
-The dedicated egress gateway is an *optional* component of UDS Core. To enable it in the UDS Bundle, add it to the `optionalComponents` as follows:
+### Ambient
+
+For workloads running in ambient mode, the dedicated egress gateway is a *default* component of UDS Core. It is enabled by default and deploys waypoint workloads to the `istio-egress-waypoint` namespace.
```
chance-coleman: nit: this reads funny to me with the double use of `default`
```suggestion
For workloads running in ambient mode, the dedicated egress gateway is automatically included in UDS Core. It comes pre-enabled and deploys waypoint workloads to the `istio-egress-waypoint` namespace.
```

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -64,6 +87,42 @@ Currently, only HTTP and TLS protocols are supported. The configuration will def
 Wildcards in host names are NOT currently supported.
 :::
 
+### Ambient Mode
+
+The following sample Package CR shows configuring egress to a specific host, "httpbin.org", on port 443.
```
chance-coleman: nit: remove whitespace

```suggestion
The following sample Package CR shows configuring egress to a specific host, "httpbin.org", on port 443.
```

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -64,6 +87,42 @@ Currently, only HTTP and TLS protocols are supported. The configuration will def
...
+* An Istio AuthorizationPolicy, in the package namespace, which is used to enforce that only traffic from workloads using the selected service account can egress. If no `serviceAccount` is specified, the `default` service account is used.
+* A shared Istio Waypoint, in the `istio-egress-waypoint` namespace, which is used to route the egress traffic.
+
+#### Limitations
+
+Due to Istio limitations and the selected implementation of shared waypoint egress, if two different egress requests in different packages specify the same host (e.g., pkg1 -> example.com AND pkg2 -> example.com) but request different port or protocols (e.g., pkg1 -> example.com:443/TLS AND pkg2 -> example.com:80/HTTP), the second request will be blocked during reconciliation and the package will fail to reconcile.
+
+### Sidecar Mode
+
 The following sample Package CR shows configuring egress to a specific host, "httpbin.org", on port 443.
```
chance-coleman: nit: not part of your changes, but cleanup whitespace

```suggestion
The following sample Package CR shows configuring egress to a specific host, "httpbin.org", on port 443.
```

src/istio/charts/uds-istio-egress-config/templates/configmap.yaml
```diff
@@ -0,0 +1,61 @@
...
+kind: ConfigMap
+metadata:
+  name: egress-waypoint-config
+  namespace: {{ .Release.Namespace }}
+data:
+  {{- with .Values.config.deployment }}
+  deployment: |
+    {{- with .podAnnotations }}
+    metadata:
+      annotations:
```
chance-coleman: nit: cleanup whitespace
```suggestion
      annotations:
```

chance-coleman (commented):
looking good, couple small nits is all i have. pending Joel's comments i'd approve!

src/istio/zarf.yaml
mjnagel: Let's go ahead and rename this component/namespace to `istio-egress-ambient`. For now to avoid overcomplicating this PR we can leave the existing `istio-egress-gateway` as-is. We will probably want to evaluate renaming that one or maybe even deprecating that functionality, but we can leave that to a future PR.

src/test/app-egress-ambient.yaml
```diff
@@ -0,0 +1,169 @@
...
+apiVersion: networking.istio.io/v1
+kind: DestinationRule
+metadata:
+  name: example-com-tls
+  namespace: egress-ambient-1
+spec:
+  host: example.com
+  trafficPolicy:
+    tls:
+      mode: SIMPLE
```
mjnagel: This looks to be unnecessary?

src/pepr/operator/reconcilers/package-reconciler.ts
mjnagel: Yeah - I think we should to keep that accurate. I don't know how much value the counts provide but if we have it we should update the logic to account for these ones as well.

src/pepr/operator/controllers/istio/istio-resources.ts
```diff
@@ -91,15 +85,7 @@ export async function istioResources(pkg: UDSPackage, namespace: string) {
...
-    getPackageId(pkg),
-    pkgName,
-    namespace,
-    generation,
-    ownerRefs,
-  );
+  await istioEgressResources(pkg, namespace);
 
   // Purge any orphaned resources
   await purgeOrphans(generation, namespace, pkgName, IstioVirtualService, log);
```
mjnagel, line 97: We're missing a purgeOrphans on authorizationpolicies here (cc @joelmccoy this is probably the cause of the issue you saw when changing to/from a service account). We will need to be careful that we only purge the authpols related to egress here though - @meganwolf0 you had asked me about the [`uds/for` label](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/authorizationPolicies.ts#L193) before and I'm just now remembering why that was relevant. It ensures that when we `purgeOrphans` we only purge the ones related to networking, so that we don't destroy the authservice authpols as well ([ref](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/authorizationPolicies.ts#L407-L409)). In this case we could take the same approach and use `uds/for: egress` or something, ensuring that all egress policies have that label and we only purge policies with that label.

src/istio/zarf.yaml
meganwolf0: renamed!

src/pepr/operator/crd/sources/package/v1alpha1.ts
meganwolf0: added a validation and a test for it

src/pepr/operator/reconcilers/package-reconciler.ts
meganwolf0: updated this to add an egress auth policy count - I wasn't sure if/how this should be tested (didn't see really any in the package-reconciler test module, but if there is an appropriate spot lmk. I did test it by modifying the package and watching the auth policy count change live

src/pepr/operator/controllers/istio/istio-resources.ts
meganwolf0: Added the `uds/for: egress` label on the auth policies and updated the purgeOrphans to key on that label only!

src/test/app-egress-ambient.yaml
meganwolf0: yep that is indeed unnecessary

src/pepr/operator/reconcilers/package-reconciler.ts
mjnagel: I think this is sufficient without any direct testing :+1:

src/pepr/operator/controllers/istio/istio-resources.ts
mjnagel: Looks good on the code side - going to resolve but will validate locally as well.

src/pepr/operator/controllers/istio/ambient-waypoint.ts
```diff
@@ -501,3 +504,56 @@ export async function reconcileExistingResources(
...
+
+  // Waypoint resource
+  const waypoint: K8sGateway = {
+    metadata: {
+      name: egressWaypointName,
+      namespace: ambientEgressNamespace,
+      annotations,
+      labels: {
+        "uds/package": sharedEgressPkgId,
+        "uds/generation": generation.toString(),
```
mjnagel, line 524: For future consideration - this generation label causes the pod(s) to update/cycle each time the generation changes. This means that multiple workloads requiring ambient egress could result in the waypoint cycling a few times. I'd say for now this is okay since the update policy will ensure 0 downtime, but may be an optimization for the future:
- debounce updates
- find a way to save generation on this without cycling the pod 🤔

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -154,9 +153,9 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
       return req.Deny("remoteHost does not support wildcard domains");
     }
 
-    // Ambient is not compatible with 'remoteHost'.
-    if (policy.remoteHost && networkSpec?.serviceMesh?.mode === Mode.Ambient) {
-      return req.Deny("remoteHost not supported in ambient mode");
+    // The 'serviceAccount' cannot be used if `remoteHost` is not specified.
+    if (policy.serviceAccount && !policy.remoteHost) {
+      return req.Deny("serviceAccount cannot be used without specifying remoteHost");
     }
```
mjnagel:
```ts
    if (policy.serviceAccount && (!policy.remoteHost || networkSpec?.serviceMesh?.mode !== Mode.Ambient)) {
      return req.Deny("serviceAccount is only valid for Ambient mode when using remoteHost");
    }
```

small nit, but I think we might want this check to be more like this? serviceAccount is only for ambient workloads with remoteHost? (also I did not test/format this so intentionally did not make this a suggestion)

mjnagel (commented):
LGTM overall, appreciate all the work that went into this.

src/pepr/operator/controllers/istio/ambient-waypoint.ts
meganwolf0: Yeah this is fair, I never had luck instrumenting the debouncing with those shared resources when working with sidecar, so just opted to take the slight performance hit. The issue I saw was that the pods wouldn't annotate the packages correctly with the packages that use them... but agree that it's worth trying to work more (in the future 😏)

src/pepr/operator/crd/validators/package-validator.ts
meganwolf0: Yep, that's a good call. Updated the logic for that validate statement and tested - remoteHost + ambient + serviceAccount (ok), remoteHost + sidecar + serviceAccount (denied), no remoteHost + ambient + serviceAccount (denied)

mjnagel (approved):

meganwolf0:
@joelmccoy this:
> Bug: When you add/remove the serviceAccount field to a package it doesn't cleanup the old AuthorizationPolicy

Should be addressed by
https://github.com/defenseunicorns/uds-core/pull/1697/commits/f2a25b55998e34e663a3418bd66c661f5ed69018
https://github.com/defenseunicorns/uds-core/pull/1697/commits/a0b02405e2acb26b3c12dd4988cd938af2fef5a7

chance-coleman (approved):

joelmccoy (approved):
LGTM. Thanks for addressing everything and getting this across the finish line!

PR #1713, opened 2025-07-08
feat: add secret pod reload

### LINKED ISSUE #354: Provide automated way of updating pods leveraging generated SSO secrets

### LINKED ISSUE #1361: Include support for external secrets
Secrets that have their lifecycle managed by supporting infrastructure require resources in the cluster to update the secret values when they change externally. The [External Secrets Operator](https://external-secrets.io/latest/) is one of the more common solutions for handling this issue.

I'd like to request the support for External Secrets in UDS Core, and also the ability to roll Deployments when the values change. [Reloader](https://github.com/stakater/Reloader) is a popular tool, but I assume there might be some way to handle this with the UDS Operator.

- **Given** a cluster application depends on an external secret (for example, AWS Secrets Manager)
- **When** that secret is changed at the infrastructure layer (for example, a managed RDS DB password rotates every seven days)
- **Then** the Secret internal to the cluster is updated and applications using that secret are recycled

- In cluster, we could also explore the [Secrets Store CSI driver](https://secrets-store-csi-driver.sigs.k8s.io/). This has a slightly different approach than External Secrets, but could potentially interface better with the existing UDS Operator.
- Only manage secrets with `uds run`, either on deploy time or on a schedule from a bastion/etc.

mjnagel:
## Description

Implements the new secret pod reload feature, based on the approved design doc. This functionality is also tested in 3 ways:
- unit testing across all functions
- end to end test of one of our test pods, validating it is rotated when a secret changes
- direct usage by neuvector for its SSO secret

Other changes to note:
- fix for PR conditionals when no packages change
- removal of all neuvector deploy actions - some were temporary upgrade actions, others were for SSO secret reloading
- update to neuvector tasks - due to improved timing around neuvector deployment (not cycling controllers) I reordered the job that cycles the scanners to after the e2e test that checks for scanner health
- change to handleFailure to properly bubble up k8s error response messages (into events)
- fix for vitest setup/teardown (https://vitest.dev/config/#globalsetup)

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/354

Fixes https://github.com/defenseunicorns/uds-core/issues/1361

## Type of change

## Steps to Validate

Deploy the runtime-security layer on top of slim-dev. Edit the neuvector-secret secret with a new value and confirm that neuvector controller pods are rotated.

## Checklist before merging


src/neuvector/common/zarf.yaml
```diff
@@ -37,42 +37,3 @@ components:
       #   gitPath: charts/monitor
       #   valuesFiles:
       #     - ../values/monitor-values.yaml
-    actions:
```
joelmccoy, line 40: nice 😎

src/pepr/operator/controllers/secrets/pod-reload.ts
```diff
@@ -0,0 +1,203 @@
...
+    if (usesSecretVolume) return true;
+
+    // Check for projected volumes that include the secret
+    const usesProjectedSecretVolume = pod.spec.volumes?.some(volume => {
+      if (!volume.projected || !volume.projected.sources) return false;
+
+      return volume.projected.sources.some(
+        source => source.secret && source.secret.name === secretName,
+      );
+    });
```
joelmccoy: nit: this is me trying to understand my TS syntax. I think this would be cleaner?
```suggestion
    const usesProjectedSecretVolume = pod.spec.volumes?.some(
      volume =>
        volume.projected?.sources?.some(
          source => source.secret?.name === secretName,
      ),
);
```

willswire:
I would suggest that this closes #1361 as well! Nice work

willswire (dismissed):

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
```diff
@@ -0,0 +1,532 @@
...
+  setupLogger: vi.fn(() => ({
+    debug: vi.fn(),
+    info: vi.fn(),
+    error: vi.fn(),
+    warn: vi.fn(),
+  })),
+  Component: {
+    OPERATOR: "OPERATOR",
+  },
+}));
```
joelmccoy: duplicate mock of `../../../logger`? I think the one with `Component: { OPERATOR_SECRETS: "OPERATOR_SECRETS" },` is the one you want right?

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
```diff
@@ -0,0 +1,532 @@
...
+          username: "dXNlcm5hbWU=",
+          password: "cGFzc3dvcmQ=",
+        },
+      };
+
+      // Should complete without throwing
+      await handleSecretUpdate(secret as kind.Secret);
+
+      // Verify that the function returns early and doesn't call reloadPods
+      expect(utils.reloadPods).not.toHaveBeenCalled();
```
joelmccoy, line 228: is it also worth checking that the proper error is logged here?

docs/reference/configuration/single-sign-on/sso-templating.md
```diff
@@ -48,3 +48,41 @@ spec:
...
           bearer_only: clientField(bearerOnly)
   ```
+
+## Secret Pod Reload
+
+UDS Core provides a Secret Pod Reload mechanism that can restart pods or deployments when secrets are updated. This is useful for SSO client secrets when they need to be updated.
+
+To enable automatic pod reload when a secret changes, add the `uds.dev/pod-reload: "true"` label to your secret.
+
+### Example
```
chance-coleman:
```suggestion
### Example Secret Configuration with Pod Reload
```

Replace the header and line 60 with this. I think this reads better personally than the duplicate "Example" also makes more sense in the sidebar when skimming the topic headers.

docs/reference/deployment/secret-pod-reload.md
```diff
@@ -0,0 +1,91 @@
...
+metadata:
+  labels:
+    uds.dev/pod-reload: "true"
+```
+
+### Targeting Pods for Restart
+
+There are two ways to specify which pods should be restarted when a secret changes:
+
+#### 1. Explicit Pod Selector
```
chance-coleman, line 34: I can't remember what the default max heading level is, for some reason i was thinking it was 3. if it is we can update the frontmatter:
```
---
title: Secret Pod Reload
tableOfContents:
  maxHeadingLevel: 4
---
```

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -175,15 +175,33 @@ export async function syncClient(
   if (!client.publicClient) {
     const generation = (pkg.metadata?.generation ?? 0).toString();
     const sanitizedSecretName = sanitizeResourceName(secretName || name);
+
+    // Prepare default labels
+    const secretLabels: Record<string, string> = {
+      "uds/package": pkg.metadata?.name || "",
```
chance-coleman: previously this was `pkg.metadata!.name` meaning it was required now its optional. Was that intentional?

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
```diff
@@ -0,0 +1,532 @@
...
+        Delete: vi.fn(),
+        Evict: vi.fn(),
+        Watch: vi.fn(),
+        Apply: vi.fn(),
+        WithField: vi.fn().mockReturnThis(),
+        Create: vi.fn(),
+        Patch: vi.fn(),
+        PatchStatus: vi.fn(),
+        Raw: vi.fn(),
+      }));
```
joelmccoy: nit: can this be put in a function? it looks like it is called multiple times through out these tests

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
```diff
@@ -0,0 +1,532 @@
...
```
joelmccoy: nit: DRY this with a function call?

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
```diff
@@ -0,0 +1,532 @@
...
```
joelmccoy: nit: DRY this with a function call?

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
```diff
@@ -0,0 +1,532 @@
...
```
joelmccoy: nit: DRY this with a function call?

src/pepr/operator/controllers/utils.ts
```diff
@@ -132,3 +132,294 @@ export async function retryWithDelay<T>(
...
+  for (const pod of pods) {
+    // Get the UID of the owner of the pod or default to "other"
+    const controlledBy =
+      pod.metadata?.ownerReferences?.find((ref: V1OwnerReference) => ref.controller)?.uid ||
+      "other";
+    groups[controlledBy] = groups[controlledBy] || [];
+    groups[controlledBy].push(pod);
+  }
+
+  // Evict each group of pods
```
chance-coleman: Maybe worth considering parallelization for the eviction of the pods?

```
  await Promise.all(
    pods.map(async (pod) => {
      log.info(`Evicting pod ${namespace}/${pod.metadata?.name} due to ${reason}`);

      try {
        await K8s(kind.Pod).InNamespace(namespace).Evict(pod.metadata!.name!);
        log.info(`Successfully evicted pod ${namespace}/${pod.metadata?.name}`);
      } catch (err) {
```

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
joelmccoy: maybe like this?
```ts
function createMockK8s({
  mockInNamespace,
  mockWithLabel,
  mockGet,
}: {
  mockInNamespace: any;
  mockWithLabel: any;
  mockGet: any;
}) {
  return {
    InNamespace: mockInNamespace,
    WithLabel: mockWithLabel,
    Get: mockGet,
    Logs: vi.fn(),
    Delete: vi.fn(),
    Evict: vi.fn(),
    Watch: vi.fn(),
    Apply: vi.fn(),
    WithField: vi.fn().mockReturnThis(),
    Create: vi.fn(),
    Patch: vi.fn(),
    PatchStatus: vi.fn(),
    Raw: vi.fn(),
  };
}

vi.mocked(K8s).mockImplementation(() =>
  createMockK8s({ mockInNamespace, mockWithLabel, mockGet })
);
```

src/pepr/operator/controllers/utils.ts
```diff
@@ -132,3 +132,294 @@ export async function retryWithDelay<T>(
...
+        // Try to use the Evict API
+        await K8s(kind.Pod).InNamespace(namespace).Evict(pod.metadata!.name!);
+        log.info(`Successfully evicted pod ${namespace}/${pod.metadata?.name}`);
+      } catch (err) {
+        // Fall back to Delete with grace period if Evict fails
+        log.warn(
+          `Failed to evict pod ${namespace}/${pod.metadata?.name} using Evict API, falling back to Delete: ${err.message}`,
+        );
+        try {
+          pod.metadata!.deletionGracePeriodSeconds = 30;
```
chance-coleman: just want to call this out that modifying the pod object here, is also modifying the original data passed into this function. At the moment we don't pass that data any further but if we did (say some enhancement/refactor in the future) this could lead to a bug/unexpected behavior.

At the moment it's not an issue, we could get ahead of that by using a spread operator and cloning the object rather than directly modifying the object.

src/pepr/operator/controllers/secrets/pod-reload.ts
```diff
@@ -0,0 +1,203 @@
...
+// Exported for testing purposes
+export const secretChecksumCache = new Map<string, string>();
+
+/**
+ * Computes a SHA256 checksum of the secret data
+ *
+ * @param data The secret data to compute the checksum for
+ * @returns A hex-encoded SHA256 checksum
+ */
+export function computeSecretChecksum(data: Record<string, string>): string {
```
chance-coleman, line 24: Probably a dumb question here, is the secret data base64 decoded at this point? I'm a bit fuzzy when a base64 encoded value update would result in a new value. if it updated and the base64 encoding didn't change, that could be a problem..maybe?

chance-coleman (commented):
hopefully some small changes 🤞🏼

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
```diff
@@ -0,0 +1,532 @@
...
+      expect(reloadPodCalls[2]).toBe("Secret test-secret change"); // reason
+
+      // Verify pods 1 and 2 were included and pod3 is NOT included
+      const reloaddPods = reloadPodCalls[1] as kind.Pod[];
+      expect(reloaddPods.length).toBe(2);
+      expect(reloaddPods.some(pod => pod.metadata?.name === "pod1")).toBe(true);
+      expect(reloaddPods.some(pod => pod.metadata?.name === "pod2")).toBe(true);
+      expect(reloaddPods.some(pod => pod.metadata?.name === "pod3")).toBe(false);
+    });
+  });
```
joelmccoy, line 519: can we add some tests for `discoverSecretConsumers`? it is pseudo tested with the above test, but specific tests for it should probably be here too?

the keys/cases that aren't covered in the above auto-discovery test include:
- initContainers
- projectedSecretVolumes

src/neuvector/chart/templates/uds-package.yaml
```diff
@@ -51,6 +51,10 @@ spec:
         - "https://neuvector.admin.{{ .Values.domain }}/openId_auth"
         {{- end }}
       secretName: neuvector-secret
+      secretLabels:
+        uds.dev/pod-reload: "true"
+      secretAnnotations:
+        uds.dev/pod-reload-selector: 'app=neuvector-controller-pod'
```
joelmccoy, line 57: for my understanding: Are you just adding this to neuvector to test a small sample first? Is there a reason you didn't add this to all sso secrets in core?  Or is neuvector some special case that we explicity only need to reload that sso secret?

src/neuvector/tasks.yaml
```diff
@@ -7,6 +7,12 @@ includes:
 tasks:
   - name: validate
     actions:
+      - description: Wait for NeuVector Job Completion
+        cmd: |
+          # On upgrades there may be cert rotation jobs or other jobs that run
+          if ./uds zarf tools kubectl get jobs -n neuvector --no-headers 2>/dev/null | grep -q .; then
+            ./uds zarf tools kubectl wait --for=condition=complete job --all -n neuvector --timeout=5m
+          fi
```
joelmccoy, line 15: note to self: get with @mjnagel to understand the change in logic for these validate tasks

src/pepr/operator/controllers/utils.ts
```diff
@@ -132,3 +132,294 @@ export async function retryWithDelay<T>(
...
+
+  log.info(`Processing ${pods.length} pods for restart/eviction in namespace ${namespace}`);
+
+  // Track which controllers we've already handled to avoid duplicate restarts
+  const handledControllers: Record<string, boolean> = {};
+  const standalonePodsToEvict: kind.Pod[] = [];
+
+  // First pass - identify controllers and standalone pods
+  for (const pod of pods) {
+    // Skip pods that already have a deletion timestamp
```
joelmccoy: I wonder if this is a condition we would want to skip the a pod:

- Pods that are a part of a job? I think succeeded or failed pods from jobs we don't want to restart unless the job tells it to right?
- Pods that have completed that have `restartPolicy` is `Never` or `OnFailure`

Or more generally I think this behavior comes down to checking the phase?:
```ts
const phase = pod.status?.phase;
if (
  phase === "Succeeded" ||
  phase === "Failed"
) {
  log.debug(`Ignoring Pod ${namespace}/${pod.metadata?.name} (phase: ${phase})`);
  continue;
}
```

src/pepr/operator/controllers/secrets/pod-reload.ts
mjnagel: This would be base64 encoded data still in this function. Since base64 is just a transformation of the text if the base64 changes we know the original changed, and vice versa. In other words - it shouldn't matter which we checksum as long as we are consistent.

docs/reference/deployment/secret-pod-reload.md
mjnagel: Good call - the default is 3. Updating to include this.

src/pepr/operator/controllers/utils.ts
```diff
@@ -132,3 +132,294 @@ export async function retryWithDelay<T>(
...
+  logger: Logger,
+): Promise<void> {
+  try {
+    // Get the ReplicaSet
+    const rs = await K8s(kind.ReplicaSet).InNamespace(namespace).Get(replicaSetName);
+
+    // Look for a Deployment owner
+    const deploymentOwner = rs.metadata?.ownerReferences?.find(ref => ref.kind === "Deployment");
+
+    if (deploymentOwner && deploymentOwner.name) {
```
joelmccoy: nit: 
```suggestion
    if (deploymentOwner?.name) {
```

src/pepr/operator/controllers/keycloak/client-sync.ts
mjnagel: While it wasn't totally intentional I don't think there's functionally a big difference here. The `!` [non-null operator](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-) is essentially just telling the compiler that `metadata` will always exist (never be null). In this case that is safe/true I think because with k8s objects metadata always exists.

The `?` on the other hand is just a short-circuit, if `metadata` were null it would immediately fall to the `|| ""`. If `metadata` is defined it would continue to try and access name.

The end result here should be the same in both cases, since `metadata` is always defined. I can just switch this back to `!` though for consistency with the older code.

src/pepr/operator/controllers/utils.ts
```diff
@@ -132,3 +132,294 @@ export async function retryWithDelay<T>(
...
+          namespace,
+          controller: controllerRef.kind,
+          controllerName: controllerRef.name,
+          error,
+        },
+        `Failed to handle controller for pod: ${reason}`,
+      );
+
+      // Fall back to direct pod eviction if controller handling fails
+      standalonePodsToEvict.push(pod);
```
joelmccoy: Are we sure we want this as desired behavior?  If we can't gracefully reload with the controller do we even want to reload?

For a HA deployment of an app we could call all pods to cycle at the same time taking down the service completely.

I would actually advocate falling back to not reloading if we can't handle it via the controller method.

src/pepr/operator/controllers/utils.ts
mjnagel: Hm, initially I wanted it to be sequential to potentially be more graceful (not killing all the pods at once). But there may be two mitigations that make that less necessary now:
- more graceful handling for most pods (deployment/sts are handled by a simple patch)
- usage of eviction rather than direct delete (so it honors PDB)

I don't have a strong feeling here, but might still lean towards having this sequential to avoid large batch evictions. It doesn't feel super likely to happen or cause problems though so I could go either way.

src/pepr/operator/controllers/utils.ts
mjnagel: I'm actually going to remove this line - I don't recall why it's there but I don't think it makes sense. Our delete of the pod won't be updating this since it just operates on the name of the pod (and we don't apply the update first or anything).

src/pepr/operator/controllers/keycloak/client-sync.ts
chance-coleman: No need to for changing unless you want too, i was just curious if something changed in the background that i was missing.

src/pepr/operator/controllers/utils.ts
```diff
@@ -132,3 +132,294 @@ export async function retryWithDelay<T>(
...
+  kindStr: string,
+  name: string,
+  reason: string,
+  logger: Logger,
+): Promise<void> {
+  try {
+    // Use proper controller kind based on string
+    let controllerKind: GenericClass;
+
+    switch (kindStr) {
```
joelmccoy: couldn't we just pass in the controller kind as the function parameter here instead of a string?  would remove the need for this switch statement?  Or am i missing something?

then you could check against the kinds and make sure it is one of the supported types.

```ts
  const supportedKinds = [kind.Deployment, kind.StatefulSet, kind.DaemonSet, kind.ReplicaSet];

  if (!supportedKinds.includes(controllerKind)) {
    const kindName = controllerKind?.name ?? "UnknownKind";
    throw new Error(`Unsupported controller kind: ${kindName}`);
  }
```

src/pepr/operator/controllers/utils.ts
chance-coleman: async conversation, we decided to stick with sequential eviction to try and keep things as graceful as possible. Also some concern that we could get too many api requests and trigger some of the catch clauses unintentionally.

src/pepr/operator/controllers/utils.ts
```diff
@@ -132,3 +132,294 @@ export async function retryWithDelay<T>(
...
+      `Successfully restarted ${kindStr} controller: ${reason}`,
+    );
+  } catch (error) {
+    logger.error(
+      { controller: kindStr, name, namespace, error },
+      `Failed to restart ${kindStr} controller: ${reason}`,
+    );
+    throw error;
+  }
+}
```
joelmccoy: what happens if the patch controller event is successful and the create event fails? (hopefully very unlikely)

The controller will still get restarted, but the entire function call will report an error causing potentially multiple restarts per controller if there are multiple pods that it owns.

Does it make to wrap each of these calls (patch/createEvent) in separate try/catch blocks and handle accordingly?

src/pepr/operator/controllers/utils.spec.ts
```diff
@@ -86,3 +106,372 @@ describe("retryWithDelay", () => {
...
+      Evict: mockEvict,
+      WithLabel: vi.fn().mockReturnThis(),
+      Get: mockGet,
+      Apply: mockApply,
+    }));
+
+    await reloadPods("default", pods as kind.Pod[], "Test eviction", mockLogger);
+
+    // The standalone pod should be evicted
+    expect(mockEvict).toHaveBeenCalled();
```
joelmccoy:
```suggestion
    expect(mockEvict)..toHaveBeenCalledWith("standalone-pod");
```

joelmccoy: more specific check for passing in the right name?

src/pepr/operator/controllers/utils.spec.ts
```diff
@@ -86,3 +106,372 @@ describe("retryWithDelay", () => {
...
+    await reloadPods("default", pods as kind.Pod[], "Test eviction", mockLogger);
+
+    // Should patch the Deployment with restart annotation
+    expect(mockPatch).toHaveBeenCalledWith([
+      expect.objectContaining({
+        op: "add",
+        path: "/spec/template/metadata/annotations/uds.dev~1restartedAt",
+        value: expect.any(String),
+      }),
+    ]);
```
joelmccoy: nit: is there a way to make sure this patch was called on the deployment and not the replicaset?

src/pepr/operator/controllers/utils.spec.ts
```diff
@@ -86,3 +106,372 @@ describe("retryWithDelay", () => {
...
+          kind: "ReplicaSet",
+          name: "standalone-replicaset",
+          namespace: "default",
+          // uid might be undefined in tests depending on mock implementation
+        }),
+      }),
+    );
+  });
+
+  it("should fall back to direct pod eviction if controller patching fails", async () => {
```
joelmccoy: Like mentioned in a previous comment, I am personally hesitant to fall back to evicting pods that the controller patching did not resolve.

For example if there is an HA deployment and the patching of the controller fails, we default to taking down all the pods at the same time without checking if the previous pod came up causing downtime.

src/pepr/operator/controllers/utils.spec.ts
```diff
@@ -86,3 +106,372 @@ describe("retryWithDelay", () => {
...
+      expect.stringContaining("Failed to handle controller for pod"),
+    );
+
+    // Verify pod eviction client was accessed
+    expect(mockPodClient.InNamespace).toHaveBeenCalledWith("default");
+
+    // Verify the fallback path logged the right message
+    expect(mockLogger.info).toHaveBeenCalledWith(expect.stringContaining("Directly evicting"));
+  });
+});
```
joelmccoy, line 241: Should we have a specific tests for `createEvent` or `restartController`? it is basically tested by proxy through reload pods. Although it might make sense to even add a very basic test for control flows of those functions.

test/vitest/secret-auto-reload.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+      // Ignore cleanup errors
+    }
+  });
+
+  test("should restart deployment when secret has explicit selector annotation", { timeout: 60000 }, async () => {
+
+    // Get the podinfo deployment to find its selector labels
+    const podinfoDeployment = await K8s(kind.Deployment)
+      .InNamespace(PODINFO_NAMESPACE)
+      .Get(PODINFO_DEPLOYMENT);
```
joelmccoy: note to self: figure out where in the code/ci/tasks actually deploys the podinfo deployment

test/vitest/secret-auto-reload.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+
+    if (!initialDeployment) {
+      throw new Error(`Podinfo deployment not found after secret creation`);
+    }
+
+    // Remove managedFields to avoid issues with future operations
+    if (initialDeployment.metadata) {
+      delete initialDeployment.metadata.managedFields;
+    }
```
joelmccoy: @mjnagel can you explain to me what is going on here? what issues are we trying to avoid?

test/vitest/secret-auto-reload.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+        targetKind: "Deployment",
+        targetName: PODINFO_DEPLOYMENT
+      }
+    );
+
+    // We should have either a secret change event or scaling events
+    expect(hasRequiredEvents).toBe(true);
+
+    // If we have secret change events, verify their properties
+    if (secretChangeEvents.length > 0) {
```
joelmccoy: shouldn't there always be a secretChangeEvent for this test?  shouldn't we have an expect statement here so it doesn't pass silently if there isn't a secretChangeEvent?

test/vitest/secret-auto-reload.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+      const event = secretChangeEvents[0];
+      expect(event.type).toBe("Normal");
+      expect(event.involvedObject?.namespace).toBe(PODINFO_NAMESPACE);
+      expect(event.message).toContain("Restarted due to:");
+    }
+
+    // If we have scaling events, verify they're for our deployment
+    if (scalingEvents.length > 0) {
```
joelmccoy: shouldn't there always be a scalingEvent for this test? shouldn't we have an expect statement here so it doesn't pass silently if there isn't a scalingEvent?

test/vitest/secret-auto-reload.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+    });
+
+
+    // Wait for resources to be fully created and registered
+    await new Promise(resolve => setTimeout(resolve, 5000));
+
+    // Verify the deployment is consuming the secret correctly
+    const testDeployment = await K8s(kind.Deployment)
+      .InNamespace(PODINFO_NAMESPACE)
+      .Get(TEST_DEPLOYMENT_NAME);
```
joelmccoy: this comment leads me to believe there is something missing here.  I don't see this variable `testDeployment` being used anywhere?

Or am I missing something here?

test/vitest/secret-auto-reload.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+
+    const initialRestartTimestamp = initialTestDeployment.spec?.template?.metadata?.annotations?.["uds.dev/restartedAt"];
+
+    // Update the auto-lookup secret to trigger controller restart
+    const newValue = uuidv4();
+
+    // Check if the controller is watching for the right secrets
+    const secretBeforeUpdate = await K8s(kind.Secret)
+      .InNamespace(PODINFO_NAMESPACE)
+      .Get(AUTO_LOOKUP_SECRET_NAME);
```
joelmccoy: I'm confused what is happening here.  Can you clarify the intent of this call? it doesn't look like this variable is being used again?

test/vitest/secret-auto-reload.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+      const currentPodNames = currentPods.items.map(pod => pod.metadata?.name).filter(Boolean);
+
+      // Check if pods have changed (evicted and replaced)
+      const podsChanged = initialPodNames.some(name => !currentPodNames.includes(name)) ||
+        currentPodNames.some(name => !initialPodNames.includes(name));
+
+      if (podsChanged) {
+        restartConfirmed = true;
+        continue;
+      }
```
joelmccoy: there are two exit conditions here:
1. timestamp of the annotation changed
2. pods evicted/cycled

Don't we want to make sure both of these conditions are true before continuing?

test/vitest/secret-auto-reload.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+
+      ++attempts;
+    }
+
+    // Test deployment should have been restarted or pods evicted after secret update
```
joelmccoy: i believe we should check for deployment restarted AND pods evicted (instead of or) right?

test/vitest/secret-auto-reload.spec.ts
```diff
@@ -0,0 +1,427 @@
...
+
+      const podEvictionEvents = podEvents.items?.filter(event =>
+        event.involvedObject?.kind === "Pod" &&
+        event.involvedObject?.name?.startsWith(`${TEST_DEPLOYMENT_NAME}-`) &&
+        event.message?.includes(AUTO_LOOKUP_SECRET_NAME)
+      ) || [];
+
+      // Should have pod eviction events as fallback
+      expect(podEvictionEvents.length).toBeGreaterThan(0);
+    }
```
joelmccoy: based on my comment earlier, I dont think we should fall back to pod eviction if the controller patching fails

src/pepr/operator/controllers/utils.ts
mjnagel: I am planning to add some sort of retry mechanism around some of these K8s calls to ensure we do a best effort to process secret updates. But otherwise as mentioned will leave sequential.

joelmccoy (commented):
Left comments for a first pass on this.  Haven't tested locally yet.  Some questions, some nits.

The most concerning thing I have is the fallback logic if the controller patching fails.  I personally think we should not manually evict pods if the controller patching fails.  If there is an HA deployment all the pods would get evicted (almost simultaneously) not honoring the deployment/rollout mechanisms of the controller to keep the app HA.

src/pepr/operator/controllers/secrets/pod-reload.ts
mjnagel: Agreed - this aligns closer with the approach for env as well. Will update.

src/neuvector/chart/templates/uds-package.yaml
mjnagel: The only SSO secrets in UDS Core right now are for Grafana and NeuVector I believe. NeuVector I know for sure needs to be reloaded when the secret updates. In the case of Grafana it's using a volume mount so there's the potential it doesn't need to be reloaded - less sure on how it loads the client details.

Might defer that change to a future issue/PR to evaluate if its necessary?

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
mjnagel: Definitely agree here - going to take an approach similar to what you noted.

src/neuvector/chart/templates/uds-package.yaml
joelmccoy: good with that! just wanted to make sure I understood why neuvector was singled out here.  it does seem to get picked on a lot :P

src/pepr/operator/controllers/utils.ts
mjnagel: This is a good catch - I had thought about Jobs earlier but forgot to handle this in any specific way. I think the approach of checking the phase this way makes a ton of sense and keeps it pretty simple.

src/pepr/operator/controllers/utils.ts
mjnagel: Discussed briefly with @chance-coleman and we agree on this. I'm also adding some retries around the patching/deletion that should help in a more direct and useful way rather than falling back to eviction. The main places where I might expect failure would be if/when there is k8s api instability, in which case a retry should succeed.

src/pepr/operator/controllers/utils.ts
mjnagel: Yep that's simpler here - taking a similar approach to your suggestion.

test/vitest/secret-auto-reload.spec.ts
mjnagel: Nope - I think you caught some things I forgot to cleanup here. I'm going to rewrite/work this a bit because there were some extra waits and such in here too from before.

test/vitest/secret-auto-reload.spec.ts
mjnagel: I think this was another thing I forgot to cleanup - in one iteration of this test I was pulling a deployment from the cluster, then applying a modified version. When applying you need to make sure managedFields are not present. With the current approach this is not necessary.

test/vitest/secret-auto-reload.spec.ts
mjnagel: Same thing - good catch, forgot to cleanup.

test/vitest/secret-auto-reload.spec.ts
mjnagel: The zarf package under [src/test](https://github.com/defenseunicorns/uds-core/tree/main/src/test) is deployed prior to these vitests running - which includes podinfo and some curl pods that we use for other networking tests in this folder.

src/neuvector/tasks.yaml
mjnagel: Vaguely noted in the description:
> update to neuvector tasks - due to improved timing around neuvector deployment (not cycling controllers) I reordered the job that cycles the scanners to after the e2e test that checks for scanner health

I'm actually not 100% sure this was entirely due to my changes but the timing around this on main is:
- we run the validate script, part of which includes triggering the updater job which cycles all scanner pods
- (on the single layer testing) we immediately after run the e2e test job which checks that the scanners are registered in the web console

Due to some lack of health checks on pods, there can be some poor timing around this where pods appear healthy but haven't _yet_ registered together in the NeuVector system. It seemed like the change removing the existing zarf actions (used for SSO secret management) may have made this issue more consistent. By reordering to run the job as part of the e2e-test, after our health check in the console, we have a more consistent result.

src/pepr/operator/controllers/utils.ts
mjnagel: Yeah this is a fair callout. One thing to note (with the current iteration of the code) is that there are no retries anywhere so there's not actually a scenario where there are multiple restarts caused by this. That's something I want to address by adding retries around specific calls with our `retryWithDelay` function. Essentially that will perform X number of retries on a specific function call (i.e. K8s Patch) and throw the error if all attempts fail (logging the failures in between if it eventually succeeds.

I may still adjust the try catch here since the log message could be misleading (`Failed to restart` when the event creation fails).

joelmccoy: I _think_ there is still a condition where we could have multiple restarts on the controller unintentionally as the code is written (and we might want to consider this scenario in the rewrite, although retries should make this pretty much very unlikely).  Here are the conditions that I think break this:

`Deployment A` has `Pod A` and `Pod B`

In `reloadPods` we iterate over all pods

On Pod A we realize that it has a controller so try to patch this.

We call `restartController` and this successfully patches the controller but for some random reason the follow on `createEvent` fails.

Since the patch and createEvent are wrapped in the try catch block, it exits to reloadPods and since there was an error thrown `handledControllers[controllerKey] = true;` (line 200) is never reached.

The next iteration Pod B is looked at and sees that it has a controller that has not been handled (because the `handledControllers[controllerKey] = true;` is never reached because of the createEvent failure)

This triggers another restart on the same controller.

Very specific and unlikely scenario and the actual consequences may be minimal 🤷 .  Just wanted to call this out.  Adding retries might make it obsolete. But also it might be good to split out the try catch logic to handle these calls differently

mjnagel: That's a good catch - I have this separated out locally and set so that the event creation logs the error but doesn't throw, whereas the patch failure will log + throw. That should address that issue/case, ensuring that we only "skip" marking a controller as true if the patch actually failed.

src/pepr/operator/controllers/secrets/reload-utils.ts
```diff
@@ -0,0 +1,282 @@
...
+    if (!controller.spec.template.metadata.annotations)
+      controller.spec.template.metadata.annotations = {};
+    controller.spec.template.metadata.annotations["uds.dev/restartedAt"] = new Date().toISOString();
+    // Clear managedFields before apply
+    delete controller.metadata?.managedFields;
+
+    // Update the annotation in the controller object and apply
+    async function applyControllerAnnotation() {
+      return K8s(controllerKind, { name, namespace }).Apply(controller);
+    }
```
mjnagel, line 188: Just to call this change out since the first iteration - a `Patch` on the controller would fail previously if the full parent path didn't exist (this is also true with `kubectl patch`). By using an apply and checking each parent path we ensure we won't hit unexpected errors here.

chance-coleman (dismissed):

src/pepr/operator/controllers/secrets/pod-reload.spec.ts
```diff
@@ -5,25 +5,37 @@
...
 } from "./pod-reload";
+import * as utils from "./reload-utils";
+
+// Create hoisted mocks
+const mocks = vi.hoisted(() => ({
+  mockDebug: vi.fn(),
+  mockInfo: vi.fn(),
+  mockWarn: vi.fn(),
+  mockError: vi.fn(),
+}));
```
joelmccoy, line 23: nice 😎

src/pepr/operator/controllers/secrets/reload-utils.spec.ts
```diff
@@ -0,0 +1,539 @@
...
+}
+
+function withRestartedAtAnnotation<T extends object>(obj: T): T {
+  const clone = JSON.parse(JSON.stringify(obj));
+  clone.spec.template.metadata.annotations = {
+    ...(clone.spec.template.metadata.annotations || {}),
+    "uds.dev/restartedAt": expect.any(String),
+  };
+  return clone;
+}
```
joelmccoy, line 150: nifty 😎

src/pepr/operator/controllers/secrets/reload-utils.spec.ts
```diff
@@ -0,0 +1,539 @@
...
+        lastControllerName = options.name;
+      }
+      return mockK8sClient;
+    });
+
+    // Execute the function under test
+    await reloadPods("default", pods as kind.Pod[], "Test eviction", mockLogger);
+
+    // Verify Apply was called and errored
+    expect(mockK8sClient.Apply).toHaveBeenCalledWith(withRestartedAtAnnotation(testStatefulSet));
```
joelmccoy: nit: the comment seems to hint that you check for an error in this line, but I don't think that is the case?

src/pepr/operator/controllers/secrets/reload-utils.spec.ts
```diff
@@ -0,0 +1,539 @@
...
+      expect.objectContaining({
+        controller: "StatefulSet",
+        controllerName: "test-statefulset",
+      }),
+      expect.stringContaining("Failed to handle controller for pod"),
+    );
+  });
+});
+
+describe("restartController", () => {
```
joelmccoy, line 427: ty for adding this in!

test/vitest/pod-reload.spec.ts
```diff
@@ -0,0 +1,354 @@
...
+    const replicas = deployment.spec?.replicas || 0;
+
+    // If initialGeneration is provided, wait for a new generation
+    const isNewGeneration = currentGeneration > initialGeneration;
+
+    // Deployment has been updated and all pods are available
+    if (isNewGeneration &&
+      observedGeneration === currentGeneration &&
+      updatedReplicas === replicas &&
+      availableReplicas === replicas) {
```
joelmccoy: This is fine, but I'm curious if there was a particular reason you just didn't check for the Deployment Condition to be `Available`?

joelmccoy (dismissed):
LGTM! I also tested this locally with a Keycloak HA deployment and confirmed expected behavior (confirming that a bad DB secret didn't take down the whole statefulset and that graceful rolling of pods was performed on the statefulset)

src/pepr/operator/controllers/secrets/reload-utils.spec.ts
mjnagel: Think I forgot to update the comment here - before this would throw an error...and then we caught that and fell back to pod eviction. Will update this comment.

test/vitest/pod-reload.spec.ts
mjnagel: No real reason - that's definitely simpler so will update.

joelmccoy (approved):

Issue #16, opened 2023-11-02
Istio sidecar cycling


mjnagel:
### Is your feature request related to a problem? Please describe.

When istio updates to a new version it does not automatically roll out new sidecars to every injected pod. The typical options for handling this include:
- Delete/cycle the pods manually yourself
- Add an annotation to each injected pod with the istio version, updating this annotation across pods should auto-cycle them
- Use a controller/operator to auto-cycle pods to the latest sidecar on update to istio
- ??? 

### Describe the solution you'd like

A new pepr capability seems like a great fit for this problem. The desired state would be:
```
When a new version of istio rolls out to my cluster, all istio injected pods gracefully cycle to get the latest sidecar image
```

OpenKruise has an interesting way that they handle in place sidecar updates - https://openkruise.io/docs/user-manuals/sidecarset - haven't dug into what they are doing but it could be interesting to look at how that works.

### Describe alternatives you've considered

Continue to handle this manually or manually add the annotation/label via values.

### Additional context

Big Bang umbrella "solved" this by adding the istio git version to every pod as an annotation.

mjnagel:
We haven't heard user reports of this being a problem since the start of UDS. While it's good to keep in mind, going to close this out as we look at Istio ambient (no sidecar to cycle) and other image cycling related issues (updating to the latest SHAs for a given tag).

Issue #1017, opened 2024-11-18
Add policy restricting usage of Istio annotations


mjnagel:
Istio provides a number of [resource annotations](https://istio.io/latest/docs/reference/config/annotations/) that can be used to adjust the sidecar's configuration. In particular there are a number of `traffic.sidecar.istio.io/` annotations that can be used to modify how traffic is captured by the sidecar.

These annotations should be evaluated for security impact and where necessary we should block annotations with a policy. This policy should allow an exemption, likely named something like `RestrictIstioAnnotations`.

mjnagel:
cc @bburky would be helpful to get your take on which annotations are particular important to block

bburky:
There's so many it's hard to say. Conservatively I'd say all, and exempt things one by one if we know they're safe.

I know it's a bit of hassle to create more Exceptions, but I'd also suggest all the per-port mTLS exemptions should be flagged and allowed by Exception. Similarly permissive TLS (but that's a whole separate CR, not an annotation.)

joelmccoy:
After a quick scrub of the istio annotations, would recommend the following specific annotations be blocked:

Annotation | Resource Type(s) | Decision | Reason
-- | -- | -- | --
traffic.sidecar.istio.io/excludeInboundPorts | Pod | Block | This can be used to prevent certain inbound ports from being captured by the sidecar, potentially allowing traffic to bypass policy enforcement and inspection.
traffic.sidecar.istio.io/excludeInterfaces | Pod | Block | This can exclude specific network interfaces from capture, which could be used to route traffic outside of the mesh and circumvent monitoring.
traffic.sidecar.istio.io/excludeOutboundIPRanges | Pod | Block | Allows specific outbound IP ranges to bypass sidecar redirection, which could be used to exfiltrate data without inspection.
traffic.sidecar.istio.io/excludeOutboundPorts | Pod | Block | Prevents redirection of specified outbound ports, enabling possible avoidance of egress controls or auditing.
traffic.sidecar.istio.io/includeInboundPorts | Pod | Block | Controls which inbound ports are redirected; misconfiguration could disable redirection entirely and weaken policy coverage.
traffic.sidecar.istio.io/includeOutboundIPRanges | Pod | Block | Can force redirect of specific outbound IP ranges, which if misused could expose unintended services or systems.
traffic.sidecar.istio.io/includeOutboundPorts | Pod | Block | Redirects outbound ports selectively; allows circumvention of full egress inspection.
[...]

Issue #101, opened 2024-01-12
istio-init container is being filtered out in all securityContext checks


rjferguson21:
### Describe what should be investigated or refactored

The existing policies all leverage a common function to return a list of container securityContexts.  This function currently excludes istio-init containers because historically istio-init was permitted to be in violation of specific polices.  As it is is written today it is excluded from checking policies that it would currently pass. 

It would probably be better if the istio-init exclusion was only applied to policies it was currently in violation of which I believe is only: 
* require-non-root(user/group) 
* restrict-capabilities (asks for NET_ADMIN and NET_RAW)

### Links to any relevant code
https://github.com/defenseunicorns/uds-core/blob/main/src/policies/common.ts#L37-L42

bburky:
A related issue is that I think it's possible to impersonate the `istio-init` container: you can add your own container with a name that matches the policy's checks. If we reduce the number of policy exceptions applied to it, there's less for an attacker to gain by doing this. NET_ADMIN and NET_RAW are still significant though.

It would be nice to validate that the `istio-init` container is actually Istio, but to make that meaningful we may also need to validate the whole sidecar's configuration (e.g. any [istio annotations](https://istio.io/latest/docs/reference/config/annotations/)) to make sure that even if it _is_ Istio, someone hasn't requested it to do some strange volume mounts to subvert it.

mjnagel:
I 💯% agree on both of the above mentioned issues but it's worth noting that validating something is actually istio would be a pretty in depth check (to be comprehensive I think you'd have to check the image/signature along with all the container spec pieces). Happy to be told there's an easier way though 😄.

Also just as an aside, all of the exemptions currently operate on relatively "simple" metadata (pod name regex + namespace) so there are more options than one that an attacker could abuse these (istio-init just might gain them the best privileges as is). This is very similar to how gatekeeper/kyverno typically operate but it would be great if we could make these exemptions more fool-proof.

mjnagel:
Related to https://github.com/defenseunicorns/uds-core/issues/500, if container level exclusions existed we could remove this from the pepr code and into an exemption CR (assuming we allowed regex for namespace). May be worth discussing which path/option is better. In code we can do more extensive checks (rather than just name), but the exemption CR would be more visible/auditable.

PR #1730, opened 2025-07-14
feat!: add policy restricting usage of istio user/group 1337

### LINKED ISSUE #1016: Add policy restricting usage of Istio user/group
Istio uses the user/group of 1337 as a special user/group meant for the sidecar only. Allowing access to run as this user/group can provide an pathway of attack/way to bypass the sidecar.

We should add a new policy to block usage of this user/group except by the istio proxy. This policy should also have an exemption for consistency, likely called something like `RestrictIstioUser`.

chance-coleman:
## Description
Istio uses the user/group of 1337 as a special user/group meant for the sidecar only. Allowing access to run as this user/group can provide an pathway of attack/way to bypass the sidecar.

We should add a new policy to block usage of this user/group except by the istio proxy. This policy should also have an exemption for consistency, likely called something like RestrictIstioUser.

## Related Issue

Fixes #1016 

## Type of change

## Steps to Validate
- `uds run test-uds-core --set FLAVOR=upstream`
- or manually run test file when pepr is deployed: `npx vitest run src/pepr/policies/security.spec.ts`

## Checklist before merging


src/pepr/policies/security.ts
```diff
@@ -16,6 +16,75 @@ import {
...
+
+    const pod = request.Raw as V1Pod;
+    const podSpec = pod.spec || ({} as V1PodSpec);
+    const podSecurityCtx = podSpec.securityContext || ({} as V1PodSecurityContext);
+
+    // Check if this is a ztunnel pod using multiple reliable labels
+    const isZtunnelPod =
+      pod.metadata?.labels?.["app"] === "ztunnel" &&
+      pod.metadata?.labels?.["app.kubernetes.io/name"] === "ztunnel" &&
+      pod.metadata?.labels?.["app.kubernetes.io/part-of"] === "istio";
```
mjnagel: For ztunnel I'd say we just add the exemption to the list in [here](https://github.com/defenseunicorns/uds-core/blob/user-1337-policy/src/istio/common/chart/templates/exemptions.yaml#L24), rather than putting something specific in code for it.

src/pepr/policies/security.ts
```diff
@@ -16,6 +16,75 @@ import {
...
+
+    // Check if this is a waypoint using multiple reliable indicators
+    const isWaypoint =
+      pod.metadata?.labels?.["gateway.istio.io/managed"] === "istio.io-mesh-controller" &&
+      pod.metadata?.labels?.["service.istio.io/canonical-name"]?.endsWith("-waypoint") &&
+      pod.metadata?.labels?.["istio.io/dataplane-mode"] === "none";
```
mjnagel: My preference here would be to check for:
- container args including "proxy"
- container name of "istio-proxy"
- container ports including "http-envoy-prom"

This would mirror the checks we do for [filtering istio init](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/policies/common.ts#L69-L75) and should cover both sidecars and waypoints. However, this would be a check at the container level so it should move into the loop below.

src/pepr/policies/security.ts
```diff
@@ -16,6 +16,75 @@ import {
...
+
+    // Skip validation for trusted Istio components
+    if (isZtunnelPod || isWaypoint) {
+      return request.Approve();
+    }
```
mjnagel: If you make the other suggested changes, this explicit approve can go away.

src/pepr/policies/security.ts
```diff
@@ -16,6 +16,75 @@ import {
...
+
+    // Check container security contexts
+    for (const container of containers(request)) {
+      const containerCtx = container.securityContext || {};
+
+      // Skip validation for istio-proxy containers
+      if (container.name === "istio-proxy") {
+        continue;
+      }
```
mjnagel: This should really be be replaced by the suggested check above.

docs/reference/configuration/pepr-policies.md
```diff
@@ -40,6 +40,7 @@ Mutations can be exempted using the same [Pepr Policy Exemptions](#pepr-policy-e
...
 |[Restrict SELinux Type](https://repo1.dso.mil/big-bang/product/packages/kyverno-policies/-/blob/main/chart/templates/restrict-selinux-type.yaml) | [`RestrictSELinuxType`](https://github.com/defenseunicorns/uds-core/blob/v0.27.0/src/pepr/policies/security.ts#L287-L327) | Implemented: ✅ <br> Subject: **Pod** <br> Severity: **high** <br><br> SELinux options can be used to escalate privileges. This policy ensures that the `seLinuxOptions` type field is undefined or restricted to the allowed list. |
+|[Restrict Istio User](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/policies/security.ts) | [`RestrictIstioUser`](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/policies/security.ts) | Implemented: ✅ <br> Subject: **Pod** <br> Severity: **high** <br><br> The Istio proxy user/group (1337) should only be used by trusted Istio components. This policy enforces that only ztunnel pods, Istio waypoint pods, or any container named 'istio-proxy' can run as UID/GID 1337. This prevents unauthorized pods from running with elevated privileges that could be used to bypass security controls. |
```
mjnagel:
```suggestion
|[Restrict Istio User](https://istio.io/latest/docs/ops/deployment/application-requirements/#pod-requirements) | [`RestrictIstioUser`](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/policies/security.ts) | Implemented: ✅ <br> Subject: **Pod** <br> Severity: **high** <br><br> The Istio proxy user/group (1337) should only be used by trusted Istio components. This policy enforces that only ztunnel pods, Istio waypoint pods, or Istio proxies (sidecars) can run as UID/GID 1337. This prevents unauthorized pods from running with elevated privileges that could be used to bypass security controls. |
```

The first link ideally should be an upstream link. I'm not sure if there's a better writeup somewhere on the "why" in terms of security impact but the suggested istio link at least notes that applications shouldn't run as 1337 and it is reserved for the sidecar 🤷

src/pepr/policies/security.spec.ts
```diff
@@ -451,3 +451,88 @@ describe("security policies", () => {
...
+  }
+
+  const expectIstioUserDenied = (e: unknown) => {
+    const error = e as K8sError;
+    const message = error.data?.message || error.message || "";
+    expect(message).toMatch(/use UID\/GID 1337 \(Istio proxy\)/);
+    return expect(error).toMatchObject({ ok: false });
+  };
+
+  it("should deny pod with runAsUser 1337", async () => {
```
mjnagel, line 470: Not really requesting any changes here now since this is already a pattern, but I think we should try to move any non-unit tests like this out of this folder and into our e2e folders. Most of our policy tests are like this, and require Pepr / webhooks to be actively deployed in a cluster. I went ahead and opened https://github.com/defenseunicorns/uds-core/issues/1735 to track this goal.

src/pepr/policies/security.spec.ts
```diff
@@ -451,3 +451,88 @@ describe("security policies", () => {
...
+              image: "127.0.0.1/fake",
+            },
+          ],
+        },
+      })
+      .then(failIfReached)
+      .catch(expectIstioUserDenied);
+  });
+
+  it("should deny pod with runAsGroup 1337", async () => {
```
mjnagel, line 493: Small nit on the testing - maybe worth adding a "waypoint"/"istio-proxy" test to validate those are not denied?

src/pepr/policies/security.ts
```diff
@@ -16,6 +16,62 @@ import {
...
+  .IsCreatedOrUpdated()
+  .Mutate(markExemption(Policy.RestrictIstioUser))
+  .Validate(request => {
+    if (isExempt(request, Policy.RestrictIstioUser)) {
+      return request.Approve();
+    }
+
+    const pod = request.Raw as V1Pod;
+    const podSpec = pod.spec || ({} as V1PodSpec);
+    const podSecurityCtx = podSpec.securityContext || ({} as V1PodSecurityContext);
```
mjnagel: Very nit but `const pod = request.Raw` isn't really needed here.

Only noticed comparing to our other policy checks where we generally have something like:
```ts
    const pod = request.Raw.spec!;
    pod.securityContext = pod.securityContext || {};
```

And then check `pod.securityContext.runAsUser`, etc.

src/pepr/policies/security.ts
```diff
@@ -16,6 +16,62 @@ import {
...
+
+    // Check pod-level security context for UID/GID 1337
+    if (
+      podSecurityCtx.runAsUser === 1337 ||
+      podSecurityCtx.runAsGroup === 1337 ||
+      podSecurityCtx.fsGroup === 1337 ||
+      podSecurityCtx.supplementalGroups?.includes(1337)
```
mjnagel, line 41: Maybe not for this PR (unless you want to add), but the check of `supplementalGroups` made me realize we aren't checking those for our runAsNonRoot policy and we probably should be.

src/pepr/policies/security.ts
```diff
@@ -16,6 +16,62 @@ import {
...
+        container.name === "istio-proxy" &&
+        container.ports?.some(p => p.name === "http-envoy-prom") &&
+        container.args?.some(arg => arg.includes("proxy"));
+
+      // Only check UID/GID 1337 if this is not an Istio proxy container
+      if (!isIstioProxy) {
+        const runAsUser = containerCtx.runAsUser ?? podSecurityCtx.runAsUser;
+        const runAsGroup = containerCtx.runAsGroup ?? podSecurityCtx.runAsGroup;
+
+        if (runAsUser === 1337 || runAsGroup === 1337) {
```
mjnagel: Another nit but the nullish check here isn't really necessary since we check the pod securityContext above.

```suggestion
        if (containerCtx.runAsUser === 1337 || containerCtx.runAsGroup === 1337) {
```

mjnagel (commented):
A few small nits on code/docs.

src/pepr/policies/security.ts
chance-coleman: added this.

src/pepr/policies/security.ts
```diff
@@ -16,6 +16,62 @@ import {
...
+
+    // Check container security contexts
+    for (const container of containers(request)) {
+      const containerCtx = container.securityContext || {};
+
+      // Check if this is an Istio proxy container
+      const isIstioProxy =
+        container.name === "istio-proxy" &&
+        container.ports?.some(p => p.name === "http-envoy-prom") &&
+        container.args?.some(arg => arg.includes("proxy"));
```
bburky, line 56: It's worth mentioning, this isn't a particularly secure check. A malicious pod could maybe match this. I think pod names are maybe unique in k8s, so it might also cause the istio injector to fail to inject too.

Worth testing some day. But I guess it's probably fine to leave like this, but maybe acknowledge it's a heuristic check in the docs.

src/pepr/policies/security.ts
```diff
@@ -16,6 +16,62 @@ import {
...
+ * It allows specific Istio components (waypoints, ztunnel, and sidecars) to use these IDs.
+ * This prevents unauthorized pods from running with elevated privileges that could be used to bypass security controls.
+ */
+When(a.Pod)
+  .IsCreatedOrUpdated()
+  .Mutate(markExemption(Policy.RestrictIstioUser))
+  .Validate(request => {
+    if (isExempt(request, Policy.RestrictIstioUser)) {
+      return request.Approve();
+    }
```
bburky, line 31: Not sure if we really want to handle this (and this comment probably applies to some of the other existing policies too):

but there are a number of other policies, that if you allow them, allow escalating to 1337. Like linux capabilities, privileged, or maybe running as root.

This PR probably isn't the place to discuss this, but honestly some of these almost belong in a hierarchy, or some dependencies between them

src/pepr/policies/security.ts
mjnagel: This is semi tracked/discussed in https://github.com/defenseunicorns/uds-core/issues/101 - this is (for now) the same heuristic we use in other places. As noted in that issue - definitely agree on this not being totally secure but I think complete security relies on a more comprehensive container spec check + image name/signature. Definitely a balance to make sure the check isn't too brittle and subject to frequent revising.

Definitely open to more concrete actions/checks here for how we could extend this heuristic meaningfully.

src/pepr/policies/security.spec.ts
chance-coleman: added a test for istio-proxy containers and ztunnel pods

mjnagel (approved):

src/pepr/policies/security.ts
brandtkeller: I came to drive-by and provide the same feedback - glad to hear there is more discussion on accuracy (and this is to reach parity). 

I need to digest what is happening - I assume that the policy does not provide specifics of which fields are disallowed on failure - meaning more generally that we have to balance what those errors look like more broadly. To vague and they're hard to debug - too defined and they could provide brute force attack paths. 

Much of the metadata for the proxy container could be spoofed - the image itself could be as well - but I would argue that is more difficult to perform. 

No changes requested here - just talking out loud.

mjnagel: Potential proposal of 3 alternative checks:
- undefined `command`: true istio proxies don't define a command, since the image entrypoint has it defined
- args starts with `proxy`: again based on the way istio-proxies are setup, they run the `proxy` arg as the first arg always
- image name matches istio proxy image: this one is the "trickiest" but manageable. 3 aspects to consider:
  - registry: ~this would need to be some sort of fuzzy match, or store the "zarf registry" for Pepr to trim off.~ This could probably be handled similarly to the repository below, storing the registry as a pod field -> environment variable.
  - repository: this changes depending on the flavor, so would need Pepr to be flavor aware. This might not be hard since we could inject the Pepr image into the [env of the admission pods](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables) and potentially detect flavor based off that image? Since Pepr and Istio are part of the same functional layer we are guaranteed they are the same flavor.
  - tag: I'd argue we just allow any tag. Tracking the latest proxy version might be a pain, and not all pods are guaranteed to be running the latest proxy (until/unless we resolve https://github.com/defenseunicorns/uds-core/issues/16)

With those 3 aspects we'd have reasonable confidence we have a valid istio image, running the entrypoint + `proxy` as an arg. While this doesn't 100% guarantee the image is "setup by Istio" I think most other heuristics are spoofable (metadata, volumes, etc). I think all 3 flavor images don't include a shell and are some level of "hardened" (distroless or otherwise) so ideally hard to access/exploit.

Edit: also noting this doesn't consider full image validation (checking signature, etc) so it's not entirely foolproof, but would be a relatively lightweight approach that makes this better for the time being in my opinion.

bburky: Something like "image name matches istio proxy image" would be nice, but it's quite tricky in practice. Make sure to remember that image names may change during an istio update. You temporarily have 2 different versions of the proxy running. And maybe you switch names too, not just versions, like switching onto Tetrate

mjnagel: On this aspect:
> You temporarily have 2 different versions of the proxy running

Might be oversimplifying but I think we'd actually be okay there. Any new pod startups _should_ use the new proxy image, and these policies are only enforced when a new pod is created/updated. So the existing pods (with older images) could continue to run as expected, and new pods would have the new proxy image and be allowed. 

Today someone could use the  `sidecar.istio.io/proxyImage` annotation to change the image being used but that probably should be blocked/unsupported (https://github.com/defenseunicorns/uds-core/issues/1017) so I'm not sure we need to consider that.

PR #1731, opened 2025-07-14
feat: add keycloak support for using pre-existing secrets

jasonwashburn:
## Description

Adds ability to alternatively use existing secret(s) for connection related (username, password, host) keycloak secrets rather than providing direct values which are currently used to create a new secret during package deployment.

## Related Issue

N/A

## Type of change

## Steps to Validate
- Run `uds run -f tasks/lint.yaml helm-unittest` to run keycloak chart unit tests

## Checklist before merging


.github/actions/setup/action.yaml
```diff
@@ -40,6 +40,19 @@ runs:
         # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
         version: v0.27.10
 
+    - name: Install helm
+      shell: bash
+      run: |
+        curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3
+        chmod +x get_helm.sh
+        # renovate: datasource=github-tags depName=helm/helm versioning=semver
+        DESIRED_VERSION=v3.17.0 ./get_helm.sh
```
mjnagel: We should pin the version of the script as well, and make sure this latest ideally:
```suggestion
        # renovate: datasource=github-tags depName=helm/helm versioning=semver
        HELM_VERSION=v3.18.4 ./get_helm.sh
        curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/${HELM_VERSION}/scripts/get-helm-3
        chmod +x get_helm.sh
        DESIRED_VERSION=${HELM_VERSION} ./get_helm.sh
```

src/keycloak/chart/README.md
```diff
@@ -35,6 +35,43 @@ If you are unable to open a browser to create this user (i.e. a headless install
...
+      name: "postgresql-password-secret"
+      key: "password"
+    host:
+      name: "postgresql-host-secret"
+      key: "host"
+  database: "keycloak"
+  port: 5432
+```
+
+**Note:** When using `secretRef`, you may use a mixture of both secretRef and direct values for username, password, and host. The `database` and `port` values are configured using direct values.
```
mjnagel: This docs content should probably be a part of https://github.com/defenseunicorns/uds-core/blob/main/docs/reference/configuration/resource-configuration-and-ha.md#keycloak instead, since that will surface on the docs site.

src/keycloak/chart/templates/_helpers.tpl
```diff
@@ -79,19 +79,157 @@ Create the service DNS name.
...
+    {{/* Validate that both name and key are set for the secretRef and are not set to "", otherwise fail*/}}
+    {{- if or (empty .Values.postgresql.secretRef.username.name) (empty .Values.postgresql.secretRef.username.key) -}}
+      {{- fail "Both \"postgresql.secretRef.username.name\" and \"postgresql.secretRef.username.key\" must be set when using secretRef." -}}
+    {{- else -}}
+      {{- true -}}
+    {{- end -}}
+  {{- else -}}
+    {{- false -}}
+  {{- end -}}
+{{- end -}}
```
mjnagel: This might just be preference but this feels overly complex to me. Since helm treats empty values as false we can make the conditionals slightly simpler:
```suggestion
{{/* Helper to determine if there is a valid configured secretRef for PostgreSQL username */}}
{{- define "keycloak.postgresql.username.validExistingSecretRef" -}}
{{- if and .Values.postgresql.secretRef.username.name .Values.postgresql.secretRef.username.key -}}
  {{- true -}}
{{- else if or .Values.postgresql.secretRef.username.name .Values.postgresql.secretRef.username.key -}}
  {{- fail "Both \"postgresql.secretRef.username.name\" and \"postgresql.secretRef.username.key\" must be set when using secretRef." -}}
{{- else -}}
  {{- false -}}
{{- end -}}
{{- end -}}
```

I swapped each of these `validExistingSecretRef` helpers locally to this approach and all tests passed.

src/keycloak/chart/templates/_helpers.tpl
```diff
@@ -79,19 +79,157 @@ Create the service DNS name.
...
+{{- $emptyPasswordValue := empty $.Values.postgresql.password -}}
+{{- $emptyHostValue := empty $.Values.postgresql.host -}}
+{{- $emptyDatabaseValue := empty $.Values.postgresql.database -}}
+{{- $emptyPortValue := empty $.Values.postgresql.port -}}
+
+{{- $usernameConfigured := or (not $emptyUsernameValue) $validUsernameSecretRef -}}
+{{- $passwordConfigured := or (not $emptyPasswordValue) $validPasswordSecretRef -}}
+{{- $hostConfigured := or (not $emptyHostValue) $validHostSecretRef -}}
+{{- $databaseConfigured := not $emptyDatabaseValue -}}
+{{- $portConfigured := not $emptyPortValue -}}
```
mjnagel: This feels odd to me. We're checking empty values, then `not`-ing that. Couldn't we just directly check the value?
```suggestion
{{- $usernameConfigured := or $.Values.postgresql.username $validUsernameSecretRef -}}
{{- $passwordConfigured := or $.Values.postgresql.password $validPasswordSecretRef -}}
{{- $hostConfigured := or $.Values.postgresql.host $validHostSecretRef -}}
{{- $databaseConfigured := $.Values.postgresql.database -}}
{{- $portConfigured := $.Values.postgresql.port -}}
```

.github/actions/setup/action.yaml
```diff
@@ -40,6 +40,19 @@ runs:
...
+
+    - name: Install helm unittest plugin
+      shell: bash
+      # renovate: datasource=github-tags depName=helm-unittest/helm-unittest versioning=semver
+      run: helm plugin install https://github.com/helm-unittest/helm-unittest.git --version v0.8.2
```
mjnagel: I'm not sure this would be tracked by renovate, unless I'm missing the match for this ([config ref](https://github.com/defenseunicorns/uds-common/blob/main/config/renovate.json5))? We may need to make an env here with `=` to [match](https://github.com/defenseunicorns/uds-common/blob/main/config/renovate.json5#L148-L164) the version.

We could also update this to 1.0.0 which just released 👀

.github/actions/setup/action.yaml
mjnagel: One other note here - I think the runners already have helm installed: https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#package-management

We may be able to just delete this section.

src/keycloak/chart/templates/_helpers.tpl
jasonwashburn: nice! ...also +1 for unit tests 🎉

src/keycloak/chart/templates/_helpers.tpl
jasonwashburn: 👍  ...originally I broke those variables all out the way they are now for readability in later conditional statements...I believe I was originally using those `empty` variables in other conditionals as well, but after a couple of refactors it looks like it's now just being used in the `configured` variables.

.github/actions/setup/action.yaml
jasonwashburn: I actually just yesterday learned that these renovate helpers weren't built in renovate magic to look for versions on the following line and were actually just from own our upstream regex configs.. 👍 on env and update to 1.0.0

.github/actions/setup/action.yaml
jasonwashburn: verified it's not required and removed helm install from setup action

src/keycloak/chart/README.md
jasonwashburn: Agreed, moved.

src/keycloak/chart/templates/_helpers.tpl
jasonwashburn: fixed

src/keycloak/chart/templates/_helpers.tpl
jasonwashburn: fixed.

.github/actions/setup/action.yaml
jasonwashburn: fixed

docs/reference/configuration/resource-configuration-and-ha.md
```diff
@@ -144,6 +144,41 @@ packages:
...
+              path: postgresql.secretRef.username.key
+            - name: KEYCLOAK_DB_DATABASE
+              path: postgresql.database
+            - name: KEYCLOAK_DB_PASSWORD_SECRETREF_NAME
+              path: postgresql.secretRef.password.name
+            - name: KEYCLOAK_DB_PASSWORD_SECRETREF_KEY
+              path: postgresql.secretRef.password.key
+```
+
+**Note:** When using secret references, you may use a mixture of both secret references and direct values for `username`, `password`, and `host`. The `database` and `port` values are configured using direct values.
```
mjnagel:
```suggestion
:::note
When using secret references, you may use a mixture of both secret references and direct values for `username`, `password`, and `host`. The `database` and `port` values are configured using direct values.
:::
```

Not required, but this would make the note block nicely formatted on the docs site.

mjnagel (dismissed):

mjnagel (approved):

PR #1732, opened 2025-07-14
chore!: remove uds-dev-stack namespace from default ignored namespaces

### LINKED ISSUE #1193: Remove `uds-dev-stack` namespace exclusion from Pepr defaults

Currently uds-core ignores the `uds-dev-stack` namespace entirely for any pepr code (validations, mutations, watches). While this does make sense for the deployments of the dev/demo bundle this really should not be set by default for all core deployments as it represents an extra area of risk/namespace to audit.

With the [changes upstream in pepr](https://github.com/defenseunicorns/pepr/pull/1641) we should be able to remove `uds-dev-stack` from our defaults, and instead move it to the helm value `additionalIgnoredNamespaces` as a bundle override for the dev/demo bundles. This may need special handling for single-layer tests since they do not deploy with a bundle.

This will be dependent on https://github.com/defenseunicorns/pepr/pull/1641 being released in a new Pepr version.

### Additional context

This will be a breaking change that should be noted as such since end users could be relying on this namespace being ignored. It may also be worth noting this bundle override in the documentation to help in scenarios where there are other system namespaces that need to be ignored. This should be done with a note of caution due to the security implications of making this change.

mjnagel:
## Description

Removes `uds-dev-stack` from the default ignored namespaces. This improved the baseline security while still allowing end users to set `additionalIgnoredNamespaces` for their own needs.

Note that due to the way single-layer tests work I added a "base-shim" bundle that gives us the power of overrides for those single layer tests. Once Zarf better supports helm overrides (of lists) this shouldn't be necessary anymore.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1193

## Type of change

## Steps to Validate

Deploy the dev/demo bundles and validate that the `uds-dev-stack` namespace is still ignored. Also run a single-layer deploy/test and validate the same.

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
chore!: remove uds-dev-stack namespace from default ignored namespaces (#1732)

BREAKING CHANGE: The `uds-dev-stack` namespace is no longer ignored by default for policies and operator reconciliation. If you wish to ignore this namespace or another namespace, you can continue to ignore namespaces by setting the `additionalIgnoredNamespaces` list in the `pepr-uds-core` component, `module` chart (via a bundle override). The dev and demo bundles will continue to ignore `uds-dev-stack`.
END_COMMIT_OVERRIDE


slaskawi (dismissed):

docs/reference/configuration/uds-operator/overview.md
```diff
@@ -11,6 +11,31 @@ The UDS Operator plays a pivotal role in managing the lifecycle of UDS Package C
...
+      pepr-uds-core:
+        module:
+          values:
+            - path: additionalIgnoredNamespaces
+              value:
+                - foo-system
+                - bar-system
+```
+
+In the example above policies would not be enforced on the `foo-system` and `bar-system` namespaces. In addition, any `Package` or `Exemption` custom resources in these namespaces would be ignored and not processed.
```
chance-coleman:
```suggestion
In the example above, policies would not be enforced on the `foo-system` and `bar-system` namespaces. In addition, any `Package` or `Exemption` custom resources in these namespaces would be ignored and not processed.
```

docs/reference/configuration/uds-operator/exemption.md
```diff
@@ -64,7 +64,7 @@ Default [policy exemptions](https://uds.defenseunicorns.com/reference/configurat
 - **Mitigation of Configuration Risks:**
   - By restricting exemptions to a specific namespace, the risk of unintentional misconfigurations in RBAC is significantly reduced. This ensures that cluster exemptions are only granted intentionally and within the confines of the designated namespace, minimizing the potential for security vulnerabilities resulting from misconfigured permissions.
 
-### Allow All Namespaces
+#### Allow Exemptions In All Namespaces
```
chance-coleman, line 69: nit: won't show up in side bar search unless you add the:
```
tableOfContents:
  maxHeadingLevel: 4
```

chance-coleman (dismissed):
couple really small docs changes

chance-coleman (approved):

joelmccoy (approved):

PR #1736, opened 2025-07-16
fix: dev-setup with uds-dev-stack exemption

mjnagel:
## Description

Fixes dev-setup to ensure that `uds-dev-stack` namespace is ignored from most policies. Previously this was globally true, but changed in https://github.com/defenseunicorns/uds-core/pull/1732. Due to the way `dev-setup` works, the exclusion/ignore looks a bit different than how this works for our bundles. For local dev purposes this is a close enough approximation though and ensures that pods like local-path-provisioner don't get blocked when creating PVCs.

## Related Issue

Follow on to https://github.com/defenseunicorns/uds-core/pull/1732

## Type of change

## Steps to Validate

```bash
# in your "main" terminal
uds run dev-setup 
# Ensure this succeeds with the new addition

# in another terminal, normally a Javascript debug terminal
npx pepr dev --confirm

# back in your main terminal
zarf dev deploy src/keycloak --flavor upstream
# ensure this finishes/keycloak starts up, previously it would hang because the PVC could not be created
```

## Checklist before merging

### DIFF
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -35,6 +35,35 @@ tasks:
         env:
           - "PEPR_MODE=dev"
 
+      - description: "Add Exemption for uds-dev-stack"
+        cmd: |
+          # Note: this behaves a bit differently than a normal dev/demo deploy where 'uds-dev-stack' is ignored at the webhook level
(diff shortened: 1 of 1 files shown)


chance-coleman (approved):
LGTM, tested locally. thanks for fixing this!

PR #1750, opened 2025-07-22
fix: upgrade tests uds-dev-stack ignore

### LINKED ISSUE #1752: HA upgrade test does not use the right flavor for initial install
Noticed this issue while working on https://github.com/defenseunicorns/uds-core/pull/1750

Our `uds-core-ha-upgrade` test currently deploys the [latest bundle release](https://github.com/defenseunicorns/uds-core/blob/v0.47.0/tasks/test.yaml#L166-L178) which deploys the [demo bundle tagged latest](https://github.com/defenseunicorns/uds-core/blob/v0.47.0/tasks/deploy.yaml#L68-L73). The problem with this approach is that the latest demo bundle is only published for the upstream flavor, so this task will be performing an upgrade from `upstream` -> `registry1`/`unicorn` in some cases.

While this may not cause errors, it does likely add CI time and was probably unintentional. We likely want to adopt a process similar to the shim bundle added in https://github.com/defenseunicorns/uds-core/pull/1750 (maybe just extending that shim bundle with the necessary variables to make it HA with a config file).

mjnagel:
## Description

The latest package release doesn't ignore the `uds-dev-stack` namespace by default. This causes [issues](https://github.com/defenseunicorns/uds-core/actions/runs/16445207958/job/46477491227?pr=1233) in our upgrade CI since we were deploying the package release on top of uds-k3d.

This PR adjusts the upgrade job so that it deploys the `core` package via a bundle, which is a copy of the `k3d-standard` bundle with dynamically set repo/ref based on the latest lookup already performed.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1752

Also fixes unwritten issue experienced in CI upgrades currently.

## Type of change

## Steps to Validate

Will be validated by CI in the upgrade jobs.

## Checklist before merging


chance-coleman (approved):

tasks/test.yaml
```diff
@@ -166,26 +167,28 @@ tasks:
...
-      - task: deploy:latest-package-release
+      - task: deploy:latest-bundle-release
+        with:
+          configFile: ${{ .inputs.configFile }}
       - task: create:standard-package
         with:
           create_options: "--skip-sbom"
-      - task: deploy:standard-package
+      - task: create:k3d-standard-bundle
+      # Note: we want to deploy from the bundle since we need overrides, but we don't want to redeploy k3d/zarf init so we pass `--packages` here
```
joelmccoy, line 187: very clever. TIL about the `--packages` flag. that's super useful.

joelmccoy (approved):

PR #1753, opened 2025-07-22
feat: add policies to warn usage of insecure istio annotation/label overrides

joelmccoy:
## Description

Introduce policies to warn for potentially dangerous Istio annotations (and label) that alter traffic interception, override sidecar behavior,, or modify ambient mesh secure configuration.  Annotations (and label) are grouped into three categories, each with a corresponding exemption name for fine-grained control.

These policies will eventually block these modifications in the future.

### Blocked Annotations/Labels

**Traffic and Interception Overrides**  
**Exemption:** `RestrictIstioTrafficInterceptionOverrides`

Annotations
- `traffic.sidecar.istio.io/excludeInboundPorts`
- `traffic.sidecar.istio.io/excludeInterfaces`
- `traffic.sidecar.istio.io/excludeOutboundIPRanges`
- `traffic.sidecar.istio.io/excludeOutboundPorts`
- `traffic.sidecar.istio.io/includeInboundPorts`
- `traffic.sidecar.istio.io/includeOutboundIPRanges`
- `traffic.sidecar.istio.io/includeOutboundPorts`
- `sidecar.istio.io/interceptionMode`
- `traffic.sidecar.istio.io/kubevirtInterfaces`

#### **Sidecar Configuration Overrides**  
**Exemption:** `RestrictIstioSidecarConfigurationOverrides`

Annotations
- `sidecar.istio.io/inject`
- `sidecar.istio.io/bootstrapOverride`
- `sidecar.istio.io/discoveryAddress`
- `sidecar.istio.io/proxyImage`
- `proxy.istio.io/config`
- `sidecar.istio.io/userVolume`
- `sidecar.istio.io/userVolumeMount`

Labels
- `sidecar.istio.io/inject`

#### **Ambient Mesh Overrides**  
**Exemption:** `RestrictIstioAmbientOverrides`

Annotations
- `ambient.istio.io/bypass-inbound-capture`

## Related Issue

Relates to #1017 

## Type of change

## Steps to Validate
- `uds run` (or more minimally `uds run dev-setup` and `npx vitest src/pepr/policies/istio.spec.ts`)

## Checklist before merging


src/pepr/operator/crd/sources/exemption/v1alpha1.ts
```diff
@@ -49,6 +49,10 @@ export const v1alpha1: V1CustomResourceDefinitionVersion = {
                         "RestrictExternalNames",
                         "RestrictHostPathWrite",
                         "RestrictHostPorts",
+                        "RestrictIstioAmbientOverrides",
+                        "RestrictIstioAuthorizationPolicyOverrides",
+                        "RestrictIstioSidecarConfigurationOverrides",
+                        "RestrictIstioTrafficInterceptionOverrides",
```
joelmccoy: not 100% sold on these names. open to feedback.  All these policies are pretty specific to using annotations to override default behavior. Is it worth adding `Annotation` to all these names to be very specific?

like: `RestrictIstioAmbientAnnotationOverrides`, `RestrictIstioAuthorizationPolicyAnnotationOverrides`, `RestrictIstioSidecarConfigurationAnnotationOverrides`, `RestrictIstioTrafficInterceptionAnnotationOverrides`

Or are the names good as is?

src/pepr/policies/istio.ts
```diff
@@ -0,0 +1,182 @@
...
+    const violations = Object.entries(annotations)
+      .filter(([key]) => {
+        // Ignore 'sidecar.istio.io/inject' annotation in 'istio-system' 'istio-tenant-gateway' 'istio-admin-gateway' 'istio-passthrough-gateway' namespace
+        const allowedIstioInjectDisableNamespaces = [
+          "istio-system",
+          "istio-tenant-gateway",
+          "istio-admin-gateway",
+          "istio-passthrough-gateway",
+          "istio-egress-gateway",
+        ];
```
joelmccoy: note: not ideal, but the istio helm charts do still apply the `sidecar.istio.io/inject` annotations to the pods via their helm charts that we use in core.  istio-system pods are set to false and the gateway pods are set to true.

ignoring these know namespaces in this policy may be sufficient enough?  

I think we assume istio inject would be controlled by the namespace label applied by the operator and adding this annotation to a pod could override that.  Having this blocked is still important, and end users can add an exemption if they really need to override a specific pod with an annotation

src/pepr/policies/istio.ts
```diff
@@ -0,0 +1,182 @@
...
+
+    const annotations = request.Raw?.metadata?.annotations || {};
+
+    const violations = Object.keys(annotations)
+      .filter(annotation => {
+        // Exempt if the custom inject template is specifically "gateway"
+        if (
+          annotation === "inject.istio.io/templates" &&
+          annotations[annotation]?.trim() === "gateway"
+        ) {
```
joelmccoy: note this is a default template that is needed for the gateway pods.  should be fine to use this and just ignore it

src/pepr/policies/istio.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+              image: "127.0.0.1/fake",
+            },
+          ],
+        },
+      })
+      .then(failIfReached)
+      .catch(expected);
+  });
+
+  it("should prevent multiple dangerous istio sidecard configuration annotations", async () => {
```
mjnagel:
```suggestion
  it("should prevent multiple dangerous istio sidecar configuration annotations", async () => {
```

src/pepr/policies/istio.ts
mjnagel: I think this makes sense - although it did make me think more about `inject.istio.io/templates`. The way it works is by adding a specific named template in the value (such as gateway). Those templates have to be already defined in the [Istio chart values](https://github.com/istio/istio/blob/master/manifests/charts/istio-control/istio-discovery/values.yaml#L148-L159). I can see this both ways, but if a cluster admin has defined a specific sidecar template maybe it's okay to just allow its use?

src/pepr/policies/istio.ts
```diff
@@ -0,0 +1,182 @@
...
+          "istio-admin-gateway",
+          "istio-passthrough-gateway",
+          "istio-egress-gateway",
+        ];
+        if (
+          key === "sidecar.istio.io/inject" &&
+          allowedIstioInjectAnnotationNamespaces.includes(namespace)
+        ) {
+          return false;
+        }
```
mjnagel: Couple comments on this one:
- We really should only block if they set `sidecar.istio.io/inject` to `"false"` I think? Presence on the annotation itself should be _okay_ if the value is `"true"`. I think this might also remove the need for filtering gateway namespaces since `sidecar.istio.io/inject=true` is set on the pods? Istio-system pods do set `sidecar.istio.io/inject=false` though so we'd still need to filter that namespace.
- I realized the annotation is actually deprecated and there is an equivalent (identical) [label](https://istio.io/latest/docs/reference/config/labels/#SidecarInject) that can be used now. We probably should block that as well - although there's some more to consider with that one since waypoints set it to `sidecar.istio.io/inject=false`, so we'd need to filter out waypoints.

src/pepr/policies/istio.ts
```diff
@@ -0,0 +1,182 @@
...
+When(AuthorizationPolicy)
+  .IsCreatedOrUpdated()
+  .Mutate(markExemption(Policy.RestrictIstioAuthorizationPolicyOverrides))
+  .Validate(request => {
+    if (isExempt(request, Policy.RestrictIstioAuthorizationPolicyOverrides)) {
+      return request.Approve();
+    }
+
+    const annotations = request.Raw?.metadata?.annotations || {};
+    const dryRunAnnotation = "istio.io/dry-run";
```
mjnagel: Thinking about this one more I'm on the fence - I think we'd want to validate that any UDS created authorization policies don't have this annotation, but we don't actually expose a way for end users to set annotations on those. For end user created auth policies `dry-run` could be legitimate for those to test them out before enforcing. My main concern initially was bypassing something like authservice, but that should only be possible with manual editing of authpols.

Okay either way on this one I guess - just throwing some thoughts out there. My slight leaning might be to not check this annotation.

mjnagel (commented):
Discussed in Slack but I think we'll probably want to switch these to warnings + approve for this first release. That will give end users time to add exemptions or update their resources before we start enforcing.

src/pepr/policies/istio.ts
joelmccoy: good call. note to self action items:
- update to only check if `sidecar.istio.io/inject` is set to false
- remove check for gateway namespaces because of above
- add check for label but filter out on waypoint pods (will need to find a sane way to id a waypoint pod)

src/pepr/policies/istio.ts
joelmccoy: yeah. that added context makes me lean towards not checking for this as well.  going to remove this for now.

src/pepr/operator/crd/sources/exemption/v1alpha1.ts
mjnagel: I think I prefer without `Annotation` as it gives us more flexibility to expand these categories (i.e. like the `sidecar.istio.io/inject` _label_).

Okay with the names as-is, I tend to prefer shorter names although I feel like we could lose some specificity that way. If you were going to change, my suggestion might be to simplify these to `Restrict` + `Istio` + `<single word for category>` + `Overrides`, such as:
- `RestrictIstioAmbientOverrides`
- `RestrictIstioSidecarOverrides` (drop configuration)
- `RestrictIstioTrafficOverrides` (drop interception)

While it might be slightly more vague/general that could be a benefit if we need to add new annotations in the future?

mjnagel (commented):
We'll also want to make sure we document the new policy(ies) and their respective exemptions - https://github.com/defenseunicorns/uds-core/blob/main/docs/reference/configuration/pepr-policies.md

src/pepr/policies/istio.ts
joelmccoy: yeah i had this thought as well.  the templates annotation here is really less dangerous that the description lends it to be (since these templates need to be defined ahead of time in the istio config).  Fine with leaving this one out

src/pepr/operator/crd/sources/exemption/v1alpha1.ts
joelmccoy: took the recommendation to shorten the names as written

src/pepr/policies/istio.ts
joelmccoy: remove this check for templates as it is not inherently dangerous due to needing to have templates pre-loaded in.

src/pepr/policies/istio.spec.ts
joelmccoy: fixed

src/pepr/policies/istio.ts
joelmccoy: remove this

src/pepr/policies/istio.ts
joelmccoy: addressed action items

src/pepr/policies/istio.ts
```diff
@@ -0,0 +1,208 @@
...
+      return request.Approve([
+        [
+          "The following istio annotations can modify secure sidecar configuration and should be removed/exempted: ",
+          violations.join(", "),
+        ].join(""),
+      ]);
+      // TODO: Uncomment this line to block the request
+      // return request.Deny(
+      //   `The following istio annotations can modify secure sidecar configuration and are not allowed: ${violations.join(", ")}`,
+      // );
```
mjnagel: Quick question - I know that the warnings param here is an array, but is there a reason this uses the second array + join? In other words, why not:
```ts
      return request.Approve([
        `The following istio annotations can modify secure sidecar configuration and should be removed/exempted: ${violations.join(", ")}`,
      ]);
```

Similar to what is in the commented out deny message, just wrapped in `[]` to fit the warnings list.

chance-coleman (dismissed):
LGTM

src/pepr/policies/istio.ts
joelmccoy: No good reason. just decided to make it more complicated 😂 .  Fixed this in the three places where it was located

mjnagel (approved):

PR #1756, opened 2025-07-24
feat: support root domain in package cr

### LINKED ISSUE #1641: Support Root Domain in Expose Key Package CR
UDS Core does not currently support setting up endpoint access at the root domain of the deployment via the package CRD.  For example if the root domain is `uds.dev` I can not expose Application A at `uds.dev`.  It would be nice to support this the `expose` key for the package CR.  You could maybe set `host: .` to indicate root domain?

It might look something like:
```yaml
spec:
  network:
    expose:
      - service: uds-registry
        selector:
          app: uds-registry
        host: . # maybe "." is reserved key for the root domain?
        gateway: tenant
        port: 8080
        targetPort: 8080
```

- **Given** a deployment of UDS Core
- **When** a package CR is created to expose an app at the root domain
- **Then** the app is accessible at the root domain url

Manually create the VS to expose the app (this works fine, but it would be nice to have this in operator logic)

### Additional context
This is a use case that is popping up for the UDS Registry public service

mjnagel:
## Description

Adds support for using the root/apex domain in a Package CR expose entry. The root domain can be used by setting host to the reserved `.`.

Note: I also included a uds-k3d update in this PR to make testing easier (https://github.com/defenseunicorns/uds-k3d/pull/240) since the nginx config previously did not route root domain requests to the gateway. Instructions included under `Steps to Validate` for testing this locally.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1641

## Type of change

## Steps to Validate

<details>
    <summary>Testing steps</summary>

For testing, modify slim-dev to add these overrides (note: add them to the existing `istio-tenant-gateway` section, make sure not to duplicate that + `uds-istio-config`):
```yaml
      istio-tenant-gateway:
        uds-istio-config:
          values:
            - path: rootDomain.enabled
              value: true
            - path: rootDomain.tls.mode
              value: SIMPLE
            - path: rootDomain.tls.credentialName
              value: "gateway-tls"
```

Then deploy slim-dev:
```console
uds run slim-dev --set flavor=unicorn # pick your favorite flavor
```

Deploy a package CR + app that uses the root domain (gist provided to make this easy):
```console
kubectl apply -f https://gist.githubusercontent.com/mjnagel/e3ba035d03a598ed56c4930273150ce4/raw/5658a55d4eae029530fc01416892b58a28079f49/root-domain.yaml
```

Hit https://uds.dev/ and you should see `Hello from port 8080`. 

</details>

## Checklist before merging

### DIFF
--- a/tasks/setup.yaml
+++ b/tasks/setup.yaml
@@ -10,7 +10,7 @@ tasks:
     actions:
       - description: "Create the K3d cluster"
         # renovate: datasource=github-tags depName=defenseunicorns/uds-k3d versioning=semver
-        cmd: "uds zarf package deploy oci://defenseunicorns/uds-k3d:0.14.3-airgap --confirm --set K3D_EXTRA_ARGS='${{ .inputs.K3D_EXTRA_ARGS }}' --no-progress"
+        cmd: "uds zarf package deploy oci://defenseunicorns/uds-k3d:0.15.0-airgap --confirm --set K3D_EXTRA_ARGS='${{ .inputs.K3D_EXTRA_ARGS }}' --no-progress"
 
   - name: k3d-test-cluster
     actions:
--- a/tasks/setup.yaml
+++ b/tasks/setup.yaml
@@ -10,7 +10,7 @@ tasks:
     actions:
       - description: "Create the K3d cluster"
         # renovate: datasource=github-tags depName=defenseunicorns/uds-k3d versioning=semver
-        cmd: "uds zarf package deploy oci://defenseunicorns/uds-k3d:0.15.0-airgap --confirm --set K3D_EXTRA_ARGS='${{ .inputs.K3D_EXTRA_ARGS }}' --no-progress"
+        cmd: "uds zarf package deploy oci://defenseunicorns/uds-k3d:0.15.1-airgap --confirm --set K3D_EXTRA_ARGS='${{ .inputs.K3D_EXTRA_ARGS }}' --no-progress"
 
   - name: k3d-test-cluster
     actions:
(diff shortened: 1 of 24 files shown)


tasks/setup.yaml
```diff
@@ -3,14 +3,10 @@
...
-    inputs:
-      K3D_EXTRA_ARGS:
-        default: ""
-        description: "Extra args to pass to k3d"
     actions:
       - description: "Create the K3d cluster"
-        # renovate: datasource=github-tags depName=defenseunicorns/uds-k3d versioning=semver
-        cmd: "uds zarf package deploy oci://defenseunicorns/uds-k3d:0.14.3-airgap --confirm --set K3D_EXTRA_ARGS='${{ .inputs.K3D_EXTRA_ARGS }}' --no-progress"
+        # renovate: datasource=docker depName=ghcr.io/defenseunicorns/packages/uds-k3d versioning=docker
+        cmd: "uds zarf package deploy oci://defenseunicorns/uds-k3d:0.15.1-airgap --confirm --no-progress"
```
mjnagel, line 9: Note: This is to fix the renovate matching for uds-k3d. Issue seen in https://github.com/defenseunicorns/uds-core/pull/1766 where the versioning was not tracked loosely as docker versioning (semver) causing update issues. In addition noticed that `K3D_EXTRA_ARGS=` caused some extra failures since that matches a different Renovate regex. Since we aren't using this anywhere in CI I just removed it.

chance-coleman (approved):
could move the FQDN piece into a util since its used twice but not a big deal! LGTM

mjnagel:
> could move the FQDN piece into a util since its used twice but not a big deal! LGTM

Since it's relatively small (4 lines) and simple logic I'll leave it where it is for now. If we had more overlap with other logic in that file might be more justification to pull into a common util.

PR #1774, opened 2025-07-29
fix: cleanup kubenodes target policies if node ip changes

### LINKED ISSUE #1141: KubeNodes target does not fully handle node IP changes
https://github.com/defenseunicorns/uds-core/pull/1119 introduces a new `KubeNodes` target for network policies. There is an edge case in the current logic where a node could change IPs and the new IP will get added but the old IP will not be removed.

This is low impact as the network policies will allow the proper connectivity still (since the new node IP was added) but it could be confusing or accidentally allow more egress traffic than desired. Note that when Pepr cycles (specifically a restart of the watcher pod) the IP list would be completely refreshed so there is a path to handle this but it is not automated.

Possible paths forward:
- Use Pepr's [OnSchedule](https://docs.pepr.dev/v0.42.1/user-guide/onschedule/) to refresh the list of nodes every X interval: This wouldn't be immediate but would clean up network policies as often as whatever cadence we are comfortable with
- Rather than tracking a list of IPs, track a list of node names/ids along with the corresponding IP: This would be one more param to track but would give us a way to identify if a node with the same name has switched IPs and replace the IP rather than adding a new one (this should be relatively safe since renaming a node is uncommon/hard to do successfully)
- On each node update, refresh the entire node list (do a k8s.get call): This was one of the initial implementations in the PR, but should be careful to not cause strain on the controlplane

Regardless of solution the desired outcome is that when a node changes IPs, the old IP should be removed from the list. Ideally this should also be covered by a unit test since reproducing this behavior with k3d/e2e tests may be difficult.

joelmccoy:
## Description

Adds logic to cleanup KubeNodes target when node ips change.  

## Related Issue

Fixes #1141 

## Type of change

## Steps to Validate
- This is validated with a new unit test: `npx vitest src/pepr/operator/controllers/network/generators/kubeNodes.spec.ts`

## Checklist before merging

### DIFF
--- a/src/pepr/operator/controllers/network/generators/kubeNodes.spec.ts
+++ b/src/pepr/operator/controllers/network/generators/kubeNodes.spec.ts
@@ -285,7 +285,7 @@ describe("kubeNodes module", () => {
       );
 
       const notReadyNode = {
-        metadata: { name: "node2" },
+        metadata: { name: "node1" },
         status: {
           addresses: [{ type: "InternalIP", address: "10.0.0.1" }],
           conditions: [{ type: "Ready", status: "False" }],
@@ -332,6 +332,40 @@ describe("kubeNodes module", () => {
 
       // the apply should not have been called
       expect(mockApply).not.toHaveBeenCalled();
+    });
+
+    it("should remove the old node IP if the node is updated with a new IP", async () => {
+      const oneNodeList = {
+        items: [
+          {
+            metadata: { name: "node1" },
+            status: {
+              addresses: [{ type: "InternalIP", address: "10.0.0.1" }],
+              conditions: [{ type: "Ready", status: "True" }],
+            },
+          },
+        ],
+      };
+
+      mockK8sGetNodes.mockResolvedValueOnce(oneNodeList);
+      await initAllNodesTarget();
+      const cidrsBeforeUpdate = kubeNodes();
+      expect(cidrsBeforeUpdate).toHaveLength(1);
+      expect(cidrsBeforeUpdate[0].ipBlock?.cidr).toBe("10.0.0.1/32");
+
+      // Update the node with a new IP
+      const updatedNode = {
+        metadata: { name: "node1" },
+        status: {
+          addresses: [{ type: "InternalIP", address: "10.0.0.2/32" }],
+          conditions: [{ type: "Ready", status: "True" }],
+        },
+      };
+
+      await updateKubeNodesFromCreateUpdate(updatedNode);
+      const cidrsAfterUpdate = kubeNodes();
+      expect(cidrsAfterUpdate).toHaveLength(1);
+      expect(cidrsAfterUpdate[0].ipBlock?.cidr).toBe("10.0.0.2/32");
     });
   });
 
--- a/src/pepr/operator/controllers/network/generators/kubeNodes.spec.ts
+++ b/src/pepr/operator/controllers/network/generators/kubeNodes.spec.ts
@@ -357,7 +357,7 @@ describe("kubeNodes module", () => {
       const updatedNode = {
         metadata: { name: "node1" },
         status: {
-          addresses: [{ type: "InternalIP", address: "10.0.0.2/32" }],
+          addresses: [{ type: "InternalIP", address: "10.0.0.2" }],
           conditions: [{ type: "Ready", status: "True" }],
         },
       };
--- a/src/pepr/operator/controllers/network/generators/kubeNodes.ts
+++ b/src/pepr/operator/controllers/network/generators/kubeNodes.ts
@@ -18,6 +18,9 @@ const log = setupLogger(Component.OPERATOR_GENERATORS);
 // Maintain a set of all node internal IPs
 const nodeSet = new Set<string>();
 
+// Maintain a map of node names to their internal IPs
+const nodeNameToIPMap = new Map<string, string>();
+
 // Track whether AuthorizationPolicies are available yet (Pepr installs before Istio)
 let authorizationPolicyExists = false;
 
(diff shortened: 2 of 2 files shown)


src/pepr/operator/controllers/network/generators/kubeNodes.spec.ts
```diff
@@ -285,7 +285,7 @@ describe("kubeNodes module", () => {
       );
 
       const notReadyNode = {
-        metadata: { name: "node2" },
+        metadata: { name: "node1" },
```
joelmccoy, line 288: note: just updated this test to use proper node with associated ip

src/pepr/operator/controllers/network/generators/kubeNodes.spec.ts
```diff
@@ -333,6 +333,40 @@ describe("kubeNodes module", () => {
...
+      await initAllNodesTarget();
+      const cidrsBeforeUpdate = kubeNodes();
+      expect(cidrsBeforeUpdate).toHaveLength(1);
+      expect(cidrsBeforeUpdate[0].ipBlock?.cidr).toBe("10.0.0.1/32");
+
+      // Update the node with a new IP
+      const updatedNode = {
+        metadata: { name: "node1" },
+        status: {
+          addresses: [{ type: "InternalIP", address: "10.0.0.2/32" }],
```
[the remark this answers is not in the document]
joelmccoy: good bot

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -71,8 +81,20 @@ export function kubeNodes(): V1NetworkPolicyPeer[] {
...
   const ip = getNodeInternalIP(node);
+  const nodeName = node.metadata?.name;
   if (ip) nodeSet.add(ip);
 
+  // Account for a node being updated with a new IP
+  if (nodeName && ip) {
+    const oldIP = nodeNameToIPMap.get(nodeName);
+    // If the IP changed, remove the old IP from the set and map
+    if (oldIP && oldIP !== ip) {
+      nodeNameToIPMap.set(nodeName, ip);
```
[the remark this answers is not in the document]
joelmccoy: good bot

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -71,8 +81,20 @@ export function kubeNodes(): V1NetworkPolicyPeer[] {
...
+      nodeSet.delete(oldIP);
```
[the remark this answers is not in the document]
joelmccoy: dont think the order matters here

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -66,13 +76,27 @@ export function kubeNodes(): V1NetworkPolicyPeer[] {
...
 
 /**
- * When a node is created or updated, if it's Ready, add its IP to the set,
+ * When a node is created or updated, if it's Ready, add its IP to the nodeSet and nodeNameToIPMap,
  * rebuild the policies, and update the NetworkPolicies.
  */
 export async function updateKubeNodesFromCreateUpdate(node: kind.Node) {
   const ip = getNodeInternalIP(node);
+  const nodeName = node.metadata?.name;
   if (ip) nodeSet.add(ip);
```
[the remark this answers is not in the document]
joelmccoy: think this is still fine. wanted to separate the logic here from updating the set and handling the updates to the map

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -43,7 +46,14 @@ export async function initAllNodesTarget() {
...
         const ip = getNodeInternalIP(node);
-        if (ip) nodeSet.add(ip);
+        const nodeName = node.metadata?.name;
+
+        if (ip) {
+          nodeSet.add(ip);
+        }
+        if (nodeName && ip) {
+          nodeNameToIPMap.set(nodeName, ip);
+        }
```
mjnagel: I'm pretty sure this is just the types throwing me off - `nodeName` should always be set right? Or is there a way to have a k8s node without a name actually set? We could use a non-null assertion here on metadata/name (`!`) if we are confident this will be set.

Otherwise the code currently read as if there might be an edge case where a node doesn't have a name. Later when a name gets added, it wouldn't already be part of `nodeNameToIPMap` so we'd fail to remove the old IP. Again - don't think that is possible here, just the way the types are?

src/pepr/operator/controllers/network/generators/kubeNodes.ts
```diff
@@ -66,12 +76,24 @@ export function kubeNodes(): V1NetworkPolicyPeer[] {
...
- * When a node is created or updated, if it's Ready, add its IP to the set,
+ * When a node is created or updated, if it's Ready, add its IP to the nodeSet and nodeNameToIPMap,
  * rebuild the policies, and update the NetworkPolicies.
  */
 export async function updateKubeNodesFromCreateUpdate(node: kind.Node) {
   const ip = getNodeInternalIP(node);
-  if (ip) nodeSet.add(ip);
+  const nodeName = node.metadata?.name;
+
+  if (ip && nodeName) {
```
mjnagel: Related to my other comment - maybe the easiest approach is just to use this same conditional in `initAllNodesTarget`? Currently on initialization we add all nodes to the set (regardless of name), but during new creations/updates we only add them if a name is present.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
joelmccoy: this is me learning silly TS intricacies. I was getting type errors when using the `!` non null assertion before, but realized i need to use it on both properties like so `const nodeName = node.metadata!.name!` instead of `const nodeName = node.metadata!.name`

This makes more sense.  Will update the in the places that need this

src/pepr/operator/controllers/network/generators/kubeNodes.ts
joelmccoy: I think we can actually use the non null assertion on the node name in this scenario as well and in the `initAllNodesTarget` and we only have to check if an is present here.

Update the logic here with the non null assertion statement.

src/pepr/operator/controllers/network/generators/kubeNodes.ts
joelmccoy: updated to use the non null assertion statement in all the places for node name

mjnagel (approved):

PR #1777, opened 2025-07-30
chore: improved istio container detection in policies

### LINKED ISSUE #101: istio-init container is being filtered out in all securityContext checks

### LINKED ISSUE #1738: Privileged Container Detection Heuristics (Istio Proxy, User 0, etc.)
## Background
Our current and planned policies, such as RestrictIstioUser (blocking UID/GID 1337 except for legitimate Istio proxies) and our user 0/root enforcementm all rely on heuristics to determine which pods/containers are “allowed” to run as privileged users or groups. Recent code reviews (see [PR #1730](https://github.com/defenseunicorns/uds-core/pull/1730#discussion_r2210843284)) have highlighted limitations and improvement areas for these heuristics.

### Heuristics can be spoofed:
Whether matching on metadata, args, or even image name, a determined attacker can craft a pod that “looks like” a privileged container unless we use stronger validation (e.g., image signatures).

### Consistency needed across policies:
Both Istio proxy and user 0/root enforcement would benefit from using a consistent set of detection checks and future-proofing for new privileged user policies.

### Error messaging needs balance:
Avoid error messages that are either too vague (hard to debug) or too specific (helpful to attackers).

### Image and environment handling:
- Image registry/repository can change due to mirroring, vendor change, or flavor (e.g., Zarf, Tetrate).
- Tags/versions can drift during upgrades.
- Special annotations (e.g., sidecar.istio.io/proxyImage) could allow image override.

### No supply-chain enforcement:
Even if we match on image name or args, we aren’t actually verifying the image came from a trusted builder.

### Short-Term: Enhanced Heuristics (applies to all privileged container policies)
Undefined command: True proxies/privileged containers should rely on their image entrypoint.

Args starts with known value: E.g., "proxy" for Istio proxies, nothing for containers that should have no args.

### Image matching: 
 - Registry/repository: Normalize or allow fuzzy match, and consider Pepr flavor-awareness (using env or config).

### Tag: Allow any tag (given rolling upgrades, version skew).

Clearly document the heuristic/best-effort nature and acknowledge known limitations in code and user docs.

Centralize heuristic logic where possible so all relevant policies benefit and stay consistent.

### Long-Term: Supply-Chain Verification and Policy Hardened
https://github.com/zarf-dev/proposals/pull/40

mjnagel:
## Description

This PR includes a few changes:
- Migrate `RestrictIstioUser` policy (and associated tests) to the Istio policy file: While not a required change I think it makes sense to co-locate these.
- Make `securityContextContainers` _conditionally_ filter out Istio init containers: Previously this was done for all policies, changed to be conditional and only done for specific policies (restrict capabilities and require non-root)
- Better validation around istio init and istio proxy checks:
  - Maintained existing checks around pod annotations, container names, and ports
  - Changed arg check to more strictly require the first arg to be as expected
  - Added check that command is undefined
  - Added basic validation of image (`validateIstioImage`) -  this validation is strictly based on the image name but validates
    - The repository is a known/allowed repository for the proxy image - any of the 3 flavors
    - The registry is either the hosted registry based on the flavor (docker.io, registry1.dso.mil, or quay.io) or the zarf registry. By supporting both we allow for pods that may not be mutated by zarf to pull the connected proxy image.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1738

Also fixes https://github.com/defenseunicorns/uds-core/issues/101

## Type of change

## Steps to Validate

CI should perform some solid validation on this PR. It may also be useful to test a mixture of sidecar workloads, waypoints, and gateways, and ensure that all are still able to come up as expected.

## Checklist before merging


src/pepr/policies/common.ts
```diff
@@ -51,43 +146,67 @@ export function securityContextMessage(
...
 /**
- * Returns true if the container looks like an istio init container
+ * Returns true if the container looks like an istio-init container
  *
  * @param request the request to check
  * @param container the container to check
- * @returns
+ * @returns boolean
  */
 export function isIstioInitContainer(
```
chance-coleman, line 155: i'd advocate for some further unit tests on this function, somewhat tested for happy path via other functions already, but not sad paths.

src/pepr/policies/common.ts
```diff
@@ -51,43 +146,67 @@ export function securityContextMessage(
...
+}
 
-  // Exit if no Istio proxy is found in initContainers
-  if (!hasInitContainerSidecar) {
+/**
+ * Check if a container is an Istio proxy container
+ * @param container the container to check
+ * @returns boolean
+ */
+export function isIstioProxyContainer(container: V1Container): boolean {
```
chance-coleman, line 192: i'd advocate for some further unit tests on this function, somewhat tested for happy path via other functions already, but not sad paths.

chance-coleman (commented):
this looks good to me, just a couple more tests for added confidence in edge cases with istioInit and istioProxy containers.

src/pepr/policies/common.ts
mjnagel: added, I think with full coverage of this function.

src/pepr/policies/common.ts
mjnagel: added, I think with full coverage of this function.

chance-coleman (approved):

joelmccoy (approved):

PR #1778, opened 2025-07-30
feat: introduce new keycloak grafana dashboards

### LINKED ISSUE #1623: Introduce Keycloak Grafana Dashboards
## Summary

In order to leverage the User Metrics in Keycloak (enabled by https://github.com/defenseunicorns/uds-core/pull/1614), the UDS Core Grafana should provide proper Dashboards out of the box. 

Keycloak Operator provides a few of them out of the box:

* https://github.com/keycloak/keycloak-grafana-dashboard
* https://www.keycloak.org/observability/grafana-dashboards
* https://www.keycloak.org/observability/event-metrics#_enable_event_metrics

The goal of this ticket is to provide these dashboards out of the box.

chance-coleman:
## Description
Two new dashboards providing:
- **Keycloak Capacity Planning Dashboard**
  - Focuses on authentication request metrics and events
  - Tracks authentication success/failure rates
  - Monitors token issuance and validation metrics
  - Helps with understanding authentication workload patterns

- **Keycloak Troubleshooting Dashboard**
  - Comprehensive cluster health monitoring
  - Tracks JVM metrics (memory, GC, threads)
  - Monitors database connection pool and performance
  - Provides system resource utilization metrics
  - Helps identify and diagnose cluster-wide issues
  
Broke out exsiting ambient vs sidecar dashboard into a dashboard directory for cleaner maintenance. Added new playwright e2e tests for monitoring that the dashboards are working.

For two reasons i didn't introduce any functionality to update the dashboards via a command or check with ci:
1. No github releases for the dashboards, so checking would be kinda ugly
2. Dashboard json's don't work out of the box, they needed some small tweaks for the datasource. [related issue](https://github.com/keycloak/keycloak-grafana-dashboard/issues/2) 

Due to the small tweaks it wasn't trivial to scripting to be able to check version updates or pull in new json files.

## Related Issue

Fixes #1623

## Type of change

## Steps to Validate
- deploy core + keycloak + monitoring package and do some user activities signing in and logging out of sso.uds.dev to see dashboard metrics
- during my testing i wrote a really ugly script to make sure all panels were populated on the capacity dashboard for refresh tokens, failed authentications, user without proper permissions, users with unfinished registration stages, etc.

## Checklist before merging


joelmccoy:
As far as the auto updating... Couldn't we steal this task [here](https://github.com/keycloak/keycloak-benchmark/blob/5e410fbeb27d0fff7ac37f4b30d0b70354e28c69/provision/common/Taskfile.yaml#L246-L269) as a UDS task for substituting things? This was mentioned in this [issue](https://github.com/keycloak/keycloak-grafana-dashboard/issues/2) as an alternative. Or is there something more complicated that needed to happen when you modified the dashboards for use of the grafana chart?

I worry about maintenance of this dashboard as things update.  Having a way to pull down from upstream would be desirable.

src/grafana/chart/templates/dashboards.yaml
```diff
@@ -9,380 +9,9 @@ metadata:
...
-        "description": "A summary of CPU and Memory usage by namespace. Provides estimated resource savings when switching from Sidecar to Ambient mode.",
-        "uid": "cemp0hgd41ssge",
-        "version": 1
-    }
+  # Load dashboards from files
+  {{- $files := .Files }}
+  {{- range $path, $bytes := .Files.Glob "dashboards/*.json" }}
+  {{ base $path }}: |-
+    {{ $files.Get $path | nindent 4 }}
+  {{- end }}
```
mjnagel: I might lean towards putting the new Keycloak dashboards into the Keycloak chart, similar to what we do with [Loki](https://github.com/defenseunicorns/uds-core/blob/1623-introduce-keycloak-grafana-dashboards/src/loki/chart/templates/loki-dashboards.yaml). 

I can see arguments for both approaches though, and it may matter less here since Grafana requires Keycloak so there's no situation where you'd have the dashboard without Keycloak itself.

test/playwright/grafana.test.ts
```diff
@@ -91,6 +91,63 @@ test("validate loki dashboard", async ({ page }) => {
...
+
+      // Navigate to dashboards page
+      await page.goto("/dashboards");
+
+      // Search for the dashboard
+      await page.getByPlaceholder("Search for dashboards and folders").fill("Keycloak");
+      await page.click(`text="${dashboard.name}"`);
+
+      // Wait for dashboard to load
+      await page.waitForLoadState("networkidle");
```
mjnagel: Playwright docs say:
```
`'networkidle'` - **DISCOURAGED** wait until there are no network connections for at least `500` ms. Don't use
```

I know sometimes it's hard to avoid - can we wait on `domcontentloaded` instead, or a specific request that we expect? I get a bit wary of `networkidle`, especially when we're loading dashboards that might autorefresh.

test/playwright/grafana.test.ts
```diff
@@ -91,6 +91,63 @@ test("validate loki dashboard", async ({ page }) => {
...
+
+      // Wait for panels to load
+      await page
+        .waitForSelector(".panel-loading", { state: "hidden", timeout: 15000 })
+        .catch(() => console.log("No loading indicator found or already loaded"));
+
+      // Additional wait for any async operations
+      await page.waitForTimeout(2000);
+
```
mjnagel: Maybe connected to https://github.com/defenseunicorns/uds-core/pull/1778/files#r2245901235 - could we wait on a specific call(s) here?

chance-coleman: removed this as the previous line should await the panels being present.

test/playwright/grafana.test.ts
chance-coleman: updated to use `domcontentload` and checking for a specific selector being present.

src/grafana/chart/templates/dashboards.yaml
chance-coleman: few updates here, moved istio related dashboard to src/istio from src/grafana. updated istio, loki, keycloak values.yaml files to expose the dashboardannotations. also updated all these directories to use the same syntax for pulling dashboard json files to be dynamic.

chance-coleman:
Added new task to the src/keycloak/tasks.yaml to update the keycloak dashboard files. This would be updated if/when the grafana playwright e2e tests fail or anytime someone wants to manually run the command.

src/keycloak/tasks.yaml
```diff
@@ -52,3 +52,55 @@ tasks:
...
+        cmd: |
+          cd .tmp-keycloak-grafana-dashboard/dashboards
+          for F in *; do
+            cp "${F}" copy.json
+            DS_PROMETHEUS=Prometheus envsubst '${DS_PROMETHEUS}' < copy.json > "${F}"
+            rm copy.json
+          done
+      - description: "Copy updated dashboards into uds-core"
+        cmd: |
+          cp .tmp-keycloak-grafana-dashboard/dashboards/keycloak-capacity-planning-dashboard.json src/grafana/chart/dashboards/keycloak-capacity-planning-dashboard.json
```
joelmccoy:
```suggestion
          cp .tmp-keycloak-grafana-dashboard/dashboards/keycloak-capacity-planning-dashboard.json src/keycloak/chart/dashboards/keycloak-capacity-planning-dashboard.json
```

src/keycloak/tasks.yaml
```diff
@@ -52,3 +52,55 @@ tasks:
...
+          cp .tmp-keycloak-grafana-dashboard/dashboards/keycloak-troubleshooting-dashboard.json src/grafana/chart/dashboards/keycloak-troubleshooting-dashboard.json
```
joelmccoy:
```suggestion
          cp .tmp-keycloak-grafana-dashboard/dashboards/keycloak-troubleshooting-dashboard.json src/keycloak/chart/dashboards/keycloak-troubleshooting-dashboard.json
```

src/keycloak/tasks.yaml
```diff
@@ -52,3 +52,55 @@ tasks:
...
+                .current = {"selected":true,"text":"keycloak","value":"keycloak"} |
+                .options = [{"selected":true,"text":"keycloak","value":"keycloak"}] |
+                .datasource = "Prometheus"
+              elif .name == "realm" then
+                .current = {"selected":true,"text":"uds","value":"uds"} |
+                .options = [{"selected":true,"text":"uds","value":"uds"}] |
+                .datasource = "Prometheus"
+              else . end
+            )
+          ' src/grafana/chart/dashboards/keycloak-capacity-planning-dashboard.json > tmp.json && mv tmp.json src/grafana/chart/dashboards/keycloak-capacity-planning-dashboard.json
```
joelmccoy:
```suggestion
          ' src/keycloak/chart/dashboards/keycloak-capacity-planning-dashboard.json > tmp.json && mv tmp.json src/keycloak/chart/dashboards/keycloak-capacity-planning-dashboard.json
```

src/keycloak/tasks.yaml
```diff
@@ -52,3 +52,55 @@ tasks:
...
+        cmd: |
+          jq '
+            .templating.list |= map(
+              if .name == "namespace" then
+                .current = {"selected":true,"text":"keycloak","value":"keycloak"} |
+                .options = [{"selected":true,"text":"keycloak","value":"keycloak"}] |
+                .datasource = "Prometheus"
+              else . end
+            )
+          ' src/grafana/chart/dashboards/keycloak-troubleshooting-dashboard.json > tmp.json && mv tmp.json src/grafana/chart/dashboards/keycloak-troubleshooting-dashboard.json
```
joelmccoy:
```suggestion
          ' src/keycloak/chart/dashboards/keycloak-troubleshooting-dashboard.json > tmp.json && mv tmp.json src/keycloak/chart/dashboards/keycloak-troubleshooting-dashboard.json
```

joelmccoy (changes requested):

test/playwright/grafana.test.ts
```diff
@@ -91,6 +91,61 @@ test("validate loki dashboard", async ({ page }) => {
   });
 });
 
+// If these tests are failing, may indicate the dashboards need to be updated (use src/keycloak/tasks.yaml update-keycloak-grafana-dashboards to update)
+// originally brought in from: https://github.com/keycloak/keycloak-grafana-dashboard
+test.describe("validate Keycloak Dashboards", () => {
+  const keycloakDashboards = [
+    { name: "Keycloak capacity planning dashboard", id: "keycloak-capacity-planning" },
+    { name: "Keycloak troubleshooting dashboard", id: "keycloak-troubleshooting" },
+  ];
```
joelmccoy: nit: I don't think the `id` is being used at all in this test.  you can probably just reduce this to a list of string and iterate over that

joelmccoy (dismissed):
LGTM. Tested the task and dashboards locally.  Left one nit comment in the playwright test that might be worth cleaning up

test/playwright/grafana.test.ts
joelmccoy: or if you meant to check something against the id here, i don't think it is present in the test

src/istio/chart/templates/istio-dashboards.yaml
```diff
@@ -0,0 +1,19 @@
...
+    grafana_dashboard: "1"
+  annotations:
+    {{- toYaml .Values.dashboardAnnotations | nindent 4 }}
+data:
+  # Load dashboards from files
+  {{- $files := .Files }}
+  {{- range $path, $bytes := .Files.Glob "dashboards/*.json" }}
+  {{ base $path }}: |-
+    {{ $files.Get $path | nindent 4 }}
+  {{- end }}
```
mjnagel: I think this is actually the wrong chart to add this to - this is chart used for each gateway, so we'd end up with 2-3x copies of this dashboard (unsure if Grafana deconflicts, but there would be one per gateway namespace).

I think this should be moved to https://github.com/defenseunicorns/uds-core/tree/1623-introduce-keycloak-grafana-dashboards/src/istio/common/chart/templates

test/playwright/grafana.test.ts
chance-coleman: yeah good catch, that was a remnant from earlier testing. cleaned that up

joelmccoy (dismissed):

mjnagel (approved):

PR #1786, opened 2025-08-04
fix(ci): snapshot release creation fix

mjnagel:
## Description

The snapshot release has been broken since the introduction of the base-shim bundle - https://github.com/defenseunicorns/uds-core/actions/workflows/snapshot-release.yaml

This PR makes the base-shim bundle optional with a new input, since it shouldn't be created when we are making the base layer for the purpose of other bundles (i.e. slim-dev).

## Related Issue

N/A

## Type of change

## Steps to Validate

Validated by running snapshot on this branch: https://github.com/defenseunicorns/uds-core/actions/runs/16724528903

## Checklist before merging

### DIFF
--- a/tasks/create.yaml
+++ b/tasks/create.yaml
@@ -48,6 +48,7 @@ tasks:
         task: single-layer
         with:
           layer: base
+          shim_bundle: "false" # Don't create the shim bundle since we are making the slim-dev bundle
 
       - description: "Create identity-authorization package"
         task: single-layer
@@ -80,6 +81,9 @@ tasks:
       create_options:
         default: ${CREATE_OPTIONS}
         description: "Additional options passed in when creating Zarf packages. For example: --skip-sbom"
+      shim_bundle:
+        default: "true"
+        description: "Whether to create the shim bundle or not (base layer only)"
     actions:
       # Note: Maru does not support `or` for this conditional task so we just duplicate it for each case
       - task: pepr-build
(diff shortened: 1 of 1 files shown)


tasks/create.yaml
```diff
@@ -97,7 +101,12 @@ tasks:
           architecture: ${ZARF_ARCHITECTURE}
       - description: "Create the shim bundle (Base only)"
         if: ${{ eq .inputs.layer "base" }}
-        cmd: "uds create bundles/base-shim --confirm --no-progress --architecture=${ZARF_ARCHITECTURE}"
+        cmd: |
+          if [ ${{.inputs.shim_bundle}} = "true" ]; then
+            uds create bundles/base-shim --confirm --no-progress --architecture=${ZARF_ARCHITECTURE}
+          else
+            echo "Skipping base-shim bundle creation."
+          fi
```
joelmccoy: nit: when adding script logic like this it might be good idea to specify the shell like so we get consistency across envs
```suggestion
        cmd: |
          if [ ${{.inputs.shim_bundle}} = "true" ]; then
            uds create bundles/base-shim --confirm --no-progress --architecture=${ZARF_ARCHITECTURE}
          else
            echo "Skipping base-shim bundle creation."
          fi
        shell:
          linux: bash
          darwin: bash
```

joelmccoy (dismissed):
lgtm with a nit

tasks/create.yaml
mjnagel: I had been under the impression that the default shell is `sh` when not specified based on [Zarf docs](https://docs.zarf.dev/ref/actions/#cmd-action-configuration) (which these actions are based on/share a common library for). When I added this I made sure that it was `sh` specific (single `=` and `[]`) and tested locally (mac) plus in CI (linux).

 I did go ahead and set this to `bash` and update the conditional as well (to use a more specific bash format) anyways - but I think this might be okay either way?

joelmccoy: ahh yeah that is true. i just saw the commit for the wrong shell logic and thought you might have run into an issue where it wasn't working locally for you.  it should be fine either way.

joelmccoy (approved):

tasks/create.yaml
mjnagel: So I do think bash is a safer option here. That commit was because locally it worked but in CI it failed (technically worked, but evaluation of the conditional threw an error log). It seems like that's due to a difference in `sh` on Linux/Mac ([ref](https://stackoverflow.com/questions/19943767/why-does-bin-sh-behave-differently-on-mac-and-ubuntu)). I've definitely experienced this in the past where certain things work locally on `sh` for me (on my Mac) but definitely are not valid in dash on my Ubuntu machine. `bash` should hopefully be a more standard experience for the most part.

PR #1804, opened 2025-08-07
chore(ci): slack pipeline alerting for nightly tests

### LINKED ISSUE #1633: nightly pipeline slack alerting
As we are starting to introduce new tests for uds-core atop different k8s distros, we should send alerts to slack for pipeline failures.

Definition of done: Alerts are created and sent via slack when pipelines fail.

chance-coleman:
## Description
Configure pipeline alerting on nightly testing. Only sends alerts to workflows that run on a schedule (nightly) and that workflow fails. Alerts are sent to the `uds-foundation-ci-alerts` channel.

## Related Issue

Fixes #1633 

## Type of change

## Checklist before merging


.github/actions/slack-alert/action.yaml
```diff
@@ -0,0 +1,56 @@
...
+  using: "composite"
+  steps:
+    - name: Send Slack notification
+      uses: slackapi/slack-github-action@91efab103c0de0a537f72a35f6b8cda0ee76bf0a # v2.1.1
+      with:
+        webhook: ${{ inputs.webhook-url }}
+        webhook-type: 'incoming-webhook'
+        payload: |
+          {
+            "channel": "#uds-foundations-ci-alerts",
```
mjnagel: nit: wonder if we might want this as a repo variable so that we could change as needed (without a PR)?

mjnagel (dismissed):

.github/actions/slack-alert/action.yaml
chance-coleman: yeah i like this idea. Updated to use an input with a default value for the current channel.

mjnagel (approved):

PR #1805, opened 2025-08-08
fix: servicemonitor mutation logic cleanup

### LINKED ISSUE #1791: Prometheus monitor incorrect tls config in 0.40.0 upgrade
### Environment

UDS Core upgrade from 0.39.0 -> 0.40.0

### Problem

Logic inside the [mutation](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/prometheus/index.ts) checks the istio status on the namespace(s) and if the namespace has already switched from sidecar -> ambient mode this code would result in certain "cleanup" being missed. Specific lines that are problematic:
- https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/prometheus/index.ts#L75
- https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/prometheus/index.ts#L125

We may want to take a different more broad approach to clean these references up when found:
- Servicemonitor: if the TLS config is set, and matches the istio tls config - delete it + switch endpoints to http
- Podmonitor: if scrape class is `istio-certs`, switch endpoints to http

The podmonitor case is a bit trickier as we want to make sure that it was set to https by Pepr, not by the end user. This is why the conditionals were written the way they are, but at this point the scrape class has likely been cleaned up by previous mutations so we may not have much to go off of.

### Severity/Priority

Medium - specifically for some monitors in the `monitoring` namespace this could cause relatively silent failures and loss of metrics data

### Additional Context

Actual issue is persistent across subsequent version updates and can only be remediated with manual edits. This is not a consistent problem and seems to potentially be a race condition based on when the UDS Package CR reconciles in relation to when specific monitors are updated.

joelmccoy:
## Description

Fixes mutation logic for Prometheus ServiceMonitor resources.  Previously, there was a potential bug/race condition that led to ServiceMonitors in ambient namespaces to not have the TLSConfigs in the ServiceMonitor Endpoints cleaned up (which was necessary for older versions of core when the monitoring namespace was not on ambient).  

- Removed the skip condition for `!isIstioInjected` for ServiceMonitor mutation
- Default to removing any Istio TLS Config for ServiceMonitorEndpoints as well as setting the scheme to http

Also:
- refactored the mutation logic into separate functions
- broke out `isIstioInjected` into a utils file for easier testing/mocking
- Added unit tests for all this logic

## Related Issue

Fixes #1791 

## Type of change

## Steps to Validate
- Deploy k3d-demo-bundle and try to apply the following ServiceMonitor
```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: test
  namespace: monitoring
spec:
  endpoints:
  - enableHttp2: true
    path: /metrics
    port: http-web
    scheme: https
    tlsConfig:
      caFile: /etc/prom-certs/root-cert.pem
      certFile: /etc/prom-certs/cert-chain.pem
      insecureSkipVerify: true
      keyFile: /etc/prom-certs/key.pem
  - path: /metrics
    port: reloader-web
    scheme: https
    tlsConfig:
      caFile: /etc/prom-certs/root-cert.pem
      certFile: /etc/prom-certs/cert-chain.pem
      insecureSkipVerify: true
      keyFile: /etc/prom-certs/key.pem
  fallbackScrapeProtocol: PrometheusText0.0.4
  namespaceSelector:
    matchNames:
    - monitoring
  selector:
    matchLabels:
      app: kube-prometheus-stack-alertmanager
      release: kube-prometheus-stack
      self-monitor: "true"
```
Ensure that the scheme gets set to http and the tlsconfig is removed on apply.

## Checklist before merging


src/pepr/prometheus/index.ts
```diff
@@ -25,118 +25,117 @@ export const prometheus = new Capability({
...
+
+  log.info(
+    `Patching PodMonitor ${pm.Raw.metadata?.name}: clearing scrapeClass, setting endpoints to HTTP.`,
+  );
+  delete pm.Raw.spec!.scrapeClass;
+  if (pm.Raw.spec?.podMetricsEndpoints && Array.isArray(pm.Raw.spec.podMetricsEndpoints)) {
+    pm.Raw.spec.podMetricsEndpoints.forEach((endpoint: PodMonitorEndpoint) => {
+      endpoint.scheme = PodMonitorScheme.HTTP;
+    });
+  }
```
mjnagel, line 138: This might be an edge case but it's possible there are pod monitors that are HTTPS scheme, but not "istio https". This unconditionally sets monitors to http, even if the end user explicitly requested an https one. Maybe it's safe to assume that they would've been using the skip annotation in the past for this, but they may have removed that in subsequent versions of core, allowing this case to get hit.

I'm really not sure how to safely set the scheme here, this is where `isIstioInjected` _might have_ been useful before - but it's definitely not perfect as we identified in the issue. Erring on the side of caution I might lean towards keeping the `isIstioInjected` check on the pod monitor, unless we can find a way to modify this and account for `HTTPS` pod monitors. Given how many UDS versions back this change was introduced I'd be worried we're "overcorrecting" here and introducing new issues.

joelmccoy: I see the issue now.  Since the original issue we experienced in the environment was specific to the servicemonitors, I am going to leave the podmonitor logic alone.  I added back in the isIstioInjected function.

Did a little refactor here and added tests for the isIstioInjected as well.

src/pepr/prometheus/index.ts
```diff
@@ -24,119 +24,123 @@ export const prometheus = new Capability({
...
+When(PrometheusServiceMonitor)
+  .IsCreatedOrUpdated()
+  .Mutate(async sm => await mutateServiceMonitor(sm));
 
 /**
+ * Mutate a PodMonitor resource by applying UDS conventions.
+ *
  * PodMonitor mutation logic:
  * - If a custom scrapeClass is set (not "istio-certs" or "exempt"), update fallback only.
  * - Else if skip conditions apply (skip annotations, not istio-injected, or scrapeClass is "exempt"),
```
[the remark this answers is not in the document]
joelmccoy: bad bot. this is not true. I added it back in the skip condition

src/pepr/prometheus/index.ts
```diff
@@ -24,119 +24,123 @@ export const prometheus = new Capability({
...
-      return true;
-    }
-  }
-  return false;
-}
+// Import isIstioInjected from utils
+import { isIstioInjected } from "./utils";
+
+// Re-export for backwards compatibility
+export { isIstioInjected };
```
mjnagel: Not sure this is used, probably can be removed.

mjnagel (approved):

PR #1808, opened 2025-08-11
fix: netpol naming

chance-coleman:
## Description
Bug fix for netpol name generation when not in ambient mode. When none-ambient package has multiple expose policies, the naming would clash and override each other resulting in a single netpol rather than individual netpols.

Update naming and fix tests as well as add new more specific test for this scenario.

## Type of change

## Checklist before merging

### DIFF
--- a/src/pepr/operator/controllers/network/policies.spec.ts
+++ b/src/pepr/operator/controllers/network/policies.spec.ts
@@ -359,17 +359,12 @@ describe("networkPolicies", () => {
     // Verify we have the expected number of policies (default policies + exposed services)
     expect(policies.length).toBeGreaterThanOrEqual(3);
 
-    // Find policies for each exposed service by name pattern
-    const frontendPolicies = policies.filter(p => p.metadata?.name?.includes("frontend"));
-
-    const backendPolicies = policies.filter(p => p.metadata?.name?.includes("backend"));
-
-    const apiPolicies = policies.filter(p => p.metadata?.name?.includes("api-3000"));
-
-    // Each service should have at least one policy
-    expect(frontendPolicies.length).toBeGreaterThan(0);
-    expect(backendPolicies.length).toBeGreaterThan(0);
-    expect(apiPolicies.length).toBeGreaterThan(0);
+    // Define expected service policies
+    const servicePolicies = [
+      { name: "frontend", ports: [8080, 15008] },
+      { name: "backend", ports: [8080, 15008] },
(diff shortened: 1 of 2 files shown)


src/pepr/operator/controllers/network/policies.spec.ts
```diff
@@ -278,43 +279,175 @@ describe("findMatchingClient", () => {
...
+    expect(policies.length).toBeGreaterThanOrEqual(3);
+
+    // Find policies for each exposed service by name pattern
+    const frontendPolicies = policies.filter(p => p.metadata?.name?.includes("frontend"));
+
+    const backendPolicies = policies.filter(p => p.metadata?.name?.includes("backend"));
+
+    const apiPolicies = policies.filter(p => p.metadata?.name?.includes("api-3000"));
+
+    // Each service should have at least one policy
```
joelmccoy: nit (non-blocking): we already check that the length for each of these 1 below. might be redundant

joelmccoy (dismissed):
lgtm

src/pepr/operator/controllers/network/policies.spec.ts
chance-coleman: cleaned that up with a small refactor 👍🏼

joelmccoy (approved):

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -133,12 +133,7 @@ export async function networkPolicies(pkg: UDSPackage, namespace: string, istioM
         app: `${gateway}-ingressgateway`,
       },
       port: policyPort,
-      description: getGatewayPolicyDescription(
-        policyPort!,
-        matchingClient?.clientId || undefined,
-        gateway,
-        !!waypointName,
-      ),
+      description: getGatewayPolicyDescription(serviceName, policyPort!, gateway, !!waypointName),
```
mjnagel, line 136: Previously (pre-ambient authservice change) this was:
```ts
      // Use the port, selector, and gateway to generate a description for VirtualService derived policies
      description: `${policyPort}-${Object.values(selector)} Istio ${gateway} gateway`,
```

I'm not sure I'm seeing the value of `getGatewayPolicyDescription` considering it just appends `(ambient)` if in ambient mode? If ambient mode is set, but this isn't an authservice protected service that is kind of irrelevant? I might lean towards just going back to what this was before since we knew that was unique and won't cause any further changes to the naming.

Issue #883, opened 2024-10-09
Enable ability to add group auth to Grafana and NeuVector keycloak clients


brianrexrode:
### Is your feature request related to a problem? Please describe.
I'd the ability to capture authentication failures in a centralized location i.e. Keycloak (IdP). This will allow for me to provide accurate and consistent authentication logs.

Currently, Grafana and NeuVector keycloak clients do not provide group auth, therefore any authentication failures are captured in their respective logs and may not provide the same verbosity or error handling as the Keycloak logging.

### Describe the solution you'd like

- **Given** a user
- **When** during failed login
- **Then** consistent failure message and log is generated in Keycloak (IdP) logging

### Describe alternatives you've considered
(optional) A clear and concise description of any alternative solutions or features you've considered.

mjnagel:
On the core side the main changes we would want to make to support this would be adding the `groups` section to the SSO spec for NeuVector and Grafana ([ref docs](https://uds.defenseunicorns.com/reference/configuration/uds-operator/#package)). We may want to make this configurable in case someone overrides the allowed groups, as well as updating [the docs](https://uds.defenseunicorns.com/reference/configuration/uds-user-groups/) to indicate the configuration if we do expose it.

mjnagel:
Especially for NeuVector this would provide a much better UX around missing the required group. I'd advocate for setting this by default (to the default groups for neuvector/grafana) and ensuring we have overrides available for someone to add other groups.

mjnagel:
This may also close https://github.com/defenseunicorns/uds-identity-config/issues/372

mjnagel:
@joelmccoy I think this can be closed?

joelmccoy:
ah yes. forgot to put fixes/closes in the PR description.

PR #1809, opened 2025-08-11
feat!: add sso group auth grafana/neuvector

joelmccoy:
## Description

Adds a group auth check to the Grafana and Neuvector Keycloak clients.  This adds the group check at the Keycloak level.  The chart defaults to using the default groups `/UDS Core/Admin` and `/UDS Core/Auditor` and can be overridden with value overrides.

This also adds some basic helm unittests for the new helm templating logic this PR introduces following the pattern established in the Keycloak chart.

## Related Issue

#883 

## Type of change

## Steps to Validate
- Deploy standard bundle.  Create/sign in user with no groups.  Login to Grafana/Neuvector and ensure Keycloak displays error message (and not the target app)

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: add sso group auth grafana/neuvector (#1809)

BREAKING CHANGE: Grafana and NeuVector now have [group auth protection](https://uds.defenseunicorns.com/reference/configuration/single-sign-on/group-based-auth/) provided by Keycloak. If you have been allowing different groups access to these applications (beyond the default Admin/Auditor groups), you will need to provide additional overrides to ensure Keycloak allows these groups to access the applications (see [docs](https://uds.defenseunicorns.com/reference/configuration/single-sign-on/overview/#applications)).
END_COMMIT_OVERRIDE


src/grafana/chart/templates/uds-package.yaml
```diff
@@ -66,8 +74,6 @@ spec:
         remoteNamespace: monitoring
         remoteSelector:
           app.kubernetes.io/name: alertmanager
-        selector:
-          app.kubernetes.io/name: grafana
```
joelmccoy, line 70: helm unittest caught this extra key here 😎

src/grafana/chart/values.yaml
```diff
@@ -29,6 +29,12 @@ postgresql:
     remoteSelector: {}
     remoteNamespace: ""
 
+sso:
+  # Groups that are allowed to use the Keycloak Client for Grafana
+  groups:
+    - "/UDS Core/Admin"
+    - "/UDS Core/Auditor"
```
mjnagel, line 36: Wondering if we need documentation on this? Specifically if someone changes this, it will only allow a group through _Keycloak's_ group protection. They would also need to modify the grafana upstream [`role_attribute_path`](https://github.com/defenseunicorns/uds-core/blob/main/src/grafana/values/values.yaml#L52) to map that group to a specific role.

Potentially that could live [here](https://uds.defenseunicorns.com/reference/configuration/single-sign-on/overview/#user-groups)?

joelmccoy: good call. added configuration docs for both grafana and neuvector

src/grafana/chart/values.yaml
```diff
@@ -29,6 +29,12 @@ postgresql:
     remoteSelector: {}
     remoteNamespace: ""
 
+sso:
+  # Groups that are allowed to use the Keycloak Client for Grafana
+  groups:
```
slaskawi, line 34: Once this gets merged, we'll probably need a follow-up Pull Request for the UDS Identity Config that removes this task: https://github.com/defenseunicorns/uds-identity-config/blob/main/tasks.yaml#L132-L139

slaskawi (approved):

mjnagel (approved):

PR #1816, opened 2025-08-12
chore: exclude istio-system from policy checks

### LINKED ISSUE #1795: Improved handling of webhook circular dependency

Currently both Pepr and Istio have webhooks that "depend on" / process through each other. On single node clusters (common for edge deployments) this conflict is consistent on node restarts.

While this problem has been documented with a manual recovery solution - https://uds.defenseunicorns.com/reference/troubleshooting/pepr-istiod-webhooks/ - this is a painful manual process each time that a cluster restarts in these environments.

Current behavior:
- istiod, istio-cni, and ztunnel pods fail to start due to contacting the Pepr webhook
- pepr pods fail to start due to the istio control/data planes not being ready, so these pods cannot join the mesh and network as expected

When istio and pepr pods restart simultaneously, all pods are able to come up healthy without manual intervention.

Potential options:
- Run Pepr admission pods outside of the mesh: This tends to be a common approach for policy webhooks from other projects. This would still ensure our watcher runs inside the mesh, while removing the dependency on the Istio control/data plane for pod startup of our admission pods. It is worth noting that currently the Pepr admission pods already allow permissive traffic on the webhook port due to the way webhook calls work, so this would fundamentally only change metrics flow/mTLS. Since Pepr is in ambient we would need to use the proper ambient label here (`istio.io/dataplane-mode=none`?)
- Run istio outside of policy controls (ignore in our webhook configuration): This would ensure that Istio could start up regardless of Pepr status. Since Istio already runs with some privileges this is viable, we would just want to ensure we have a solid, secure baseline being removing it from policy checks. This could be done with `admission.alwaysIgnore.namespaces`.

mjnagel:
## Description

This PR adds `istio-system` to the ignored namespaces for Pepr admission only. Additional context is contained in the linked issue, but also in [the ADR](https://github.com/defenseunicorns/uds-core/blob/pepr-admission-no-istio/adrs/0003-exclude-istio-from-policy-engine.md). 

Note: With [some upstream Pepr changes](https://github.com/defenseunicorns/pepr/issues/2546) we could switch this to more of a "best effort enforcement" of policies on Istio - using a `failurePolicy` of `Ignore` _ONLY_ for Istio. We could also implement some CI checking for "drift" if Pepr supported [dry-runs of policies](https://github.com/defenseunicorns/pepr/issues/2552).

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1795

## Type of change

## Checklist before merging


adrs/0003-exclude-istio-from-policy-engine.md
```diff
@@ -0,0 +1,48 @@
...
+
+Security-wise, both the Istio CNI and Ztunnel pods are already excluded from a [number of policies](https://github.com/defenseunicorns/uds-core/blob/v0.48.1/src/istio/common/chart/templates/exemptions.yaml) since service mesh needs to run privileged to configure the proper host networking. We also ship Istio with a hardened baseline (hardened images and reduced privileges where possible).
+
+## Decision
+
+The `istio-system` namespace will be excluded from Pepr’s admission webhooks removing policy engine interference with service mesh start-up, recovery, and upgrades. While this does have the potential to slightly reduce security, it is important to note that `istio-system` contains privileged, system/platform workloads, rather than end-user workloads. Our policy engine is primarily meant to prevent end-user workloads ("mission apps") from running with excessive privilege or the ability to escape any controls we have enforced with the service mesh and UDS Operator.
```
slaskawi: The risk I'm seeing here is that someone may try to abuse the `istio-system` namespace to schedule untrusted workloads. 

Would it be possible to create a "special policy" and add Pods we expect to some sort of allow list (while denying anything else)? Or maybe there are some other ways to protect this namespace?

slaskawi (commented):
Added one question that may result in some follow-up work. 

Otherwise, LGTM!

adrs/0003-exclude-istio-from-policy-engine.md
mjnagel: Definitely agreed that there is a small risk here. Unfortunately I'm not sure there's (currently) a good answer.

The main limitation here is that we have no real options on the policy/code side. In the situation we are trying to prevent, Pepr pods are down so nothing we could write in code would help. We need to make sure that the webhook configuration itself filters out/allows Istio when Pepr is down. At the webhook config level we have 2 options (1) `objectSelector` config for selecting/deselecting objects and (2) `namespaceSelector` config for selecting/deselecting namespaces. Unfortunately these are separate selectors, so we don't have the granularity of "pod with X label in Y namespace", we are stuck with "pod with X label" (anywhere) and "pods in Y namespace" as separate filters. Even if there was a way to combine these, it would only provide basically no extra security since labels can easily be added to pods.

As an alternative way to protect the namespace, [Pod Security Admission](https://kubernetes.io/docs/concepts/security/pod-security-admission/) is an interesting idea (I think you have mentioned before). I spent a bit of time looking more at it and don't think it would be the best fit here unfortunately. Given our privieleged workloads for Istio we would need to add some "exemptions", however with PSA this is done via api server configuration ([ref](https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-admission-controller/#configure-the-admission-controller)) which we don't have control of in UDS. Applying a more open PSS policy like `Privileged` would "work" but provides not difference from not having it at all.

All that being said I think there's a couple things to consider here:
1. Currently in core we have exemptions for Istio pods (CNI, Ztunnel) that allow significant privileges. These exemptions are purely based on pod name, so if someone has permissions in the `istio-system` namespace, it's already easy to abuse this for malicious workloads. I say this mostly to identify that we already have an assumption of proper RBAC around "system namespaces" I think.
2. I opened an upstream feature request for Pepr to allow more [flexible webhook config](https://github.com/defenseunicorns/pepr/issues/2546). In particular this would allow us to enforce policies on `istio-system`, but set a `failurePolicy` of Ignore (allowing requests in when Pepr pods are down) - exclusively for `istio-system`. This should narrow the window of malicious activity being possible, essentially providing a "best effort" policy enforcement, without blocking pods in critical downtime situations.

I'll go ahead and add some words around this to the ADR, especially under consequences, to capture this risk and some of the above comment to note things that were considered. I'd lean towards not making any changes to the current decision/code though unless there's things I'm glossing over here.

chance-coleman (approved):
lgtm

joelmccoy (approved):

adrs/0003-exclude-istio-from-policy-engine.md
slaskawi: Thanks a million for the explanation!

slaskawi (approved):

PR #1838, opened 2025-08-19
feat: limit the number of concurrent sessions

slaskawi:
## Description

This Pull Request limits the number of concurrent Sessions and Authentication Sessions (in-flight login attempts) per user.

This Pull Request requires: https://github.com/defenseunicorns/uds-identity-config/pull/591

## Related Issue

Relates to https://github.com/defenseunicorns/uds-identity-config/issues/574

## Type of change

## Steps to Validate

Tested automatically in UDS Identity Config

## Checklist before merging


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -111,6 +109,7 @@ spec:
             {{- else }}
             - "--features=preview"
             {{- end }}
+            - "--spi-authentication-sessions--map--auth-sessions-limit={{ .Values.realmInitEnv.SSO_SESSION_MAX_IN_FLIGHT_LOGINS_PER_USER | default 300 }}"
```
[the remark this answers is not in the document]
slaskawi: Nope, this is correct. Ignoring.

docs/reference/configuration/single-sign-on/keycloak-session-management.md
```diff
@@ -0,0 +1,108 @@
+---
+title: Keycloak Session Management
```
chance-coleman, line 2: When we change the title of the doc it changes the URL on the doc site. In the past we were concerned about people bookmarking a specific page and then being confused by it disappearing so we updated the [redirects in the uds-docs](https://github.com/defenseunicorns/uds-docs/blob/611851bedfb2982ed4e7a60ed0ed611127119462/astro.config.mjs#L12). I'm not sure if we should care or not...

mjnagel: Yeah no concerns on my end with renaming the doc, but let's open a PR for adding a redirect. We can merge that shortly after so that the content sticks around. There's 2 issues:
1. People have links they could bookmark or have for reference in chat, etc.
2. Search engines index these pages (and reindexing may not be super quick).

To just make it as graceful as possible we've maintained redirects - we could reevaluate a better way to handle if we want too.

src/keycloak/chart/values.yaml
```diff
@@ -43,6 +43,10 @@ realmInitEnv:
   # EMAIL_VERIFICATION_ENABLED: true
   # TERMS_AND_CONDITIONS_ENABLED: true
   # X509_OCSP_FAIL_OPEN: true
+  # Maximum number of concurrent active sessions per user. Set to 0 to disable the limit (unlimited).
+  SSO_SESSION_MAX_PER_USER: "0"
+  # Maximum number of in-flight (ongoing) login attempts per user to prevent login storms.
+  SSO_SESSION_MAX_IN_FLIGHT_LOGINS_PER_USER: "300"
```
mjnagel: Maybe a bit of a nit, but the placement of this under `realmInitEnv` I think could be confusing since it's actually always valid to set/update (not just as part of init). I know we had confusion on this in the past when things were under `realmInitEnv` (or not) and how the upgrade was expected to work. I'm not sure of a better placement unless we made a new top level key (`realmConfig`, etc), but wanted to note this.

docs/reference/configuration/single-sign-on/keycloak-session-management.md
slaskawi: Thanks a million @chance-coleman for adding the redirect: https://github.com/defenseunicorns/uds-docs/pull/281

I'll resolve this comment than.

src/keycloak/chart/values.yaml
slaskawi: I agree with the argument that it's a bit misleading to have the `SSO_SESSION_MAX_IN_FLIGHT_LOGINS_PER_USER` here as it's used in the StatefulSet rather than the Realm. This is caused by the inconsistency in Keycloak itself - Authentication Session limit is set through an SPI, whereas User Session limit is set in the Authentication Flows. 

I went ahead and put it here as all other Session limits and timeouts (such as `SSO_SESSION_LIFESPAN_TIMEOUT`, `SSO_SESSION_MAX_LIFESPAN` and `SSO_SESSION_MAX_PER_USER`) are set here. The current approach favors consistency.

So I have two proposals and I'm curious about your opinion:

1. We can leave it as
2. We can move `SSO_SESSION_MAX_IN_FLIGHT_LOGINS_PER_USER` into a separate key in the Helm chart. Maybe that `realmConfig` is a good idea as subelements are extensible.

slaskawi: After thinking about this for a bit more, I actually implemented your suggestion and introduced `realmConfig`.

src/keycloak/chart/values.yaml
```diff
@@ -43,6 +43,13 @@ realmInitEnv:
   # EMAIL_VERIFICATION_ENABLED: true
   # TERMS_AND_CONDITIONS_ENABLED: true
   # X509_OCSP_FAIL_OPEN: true
+  # Maximum number of concurrent active sessions per user. Set to 0 to disable the limit (unlimited).
+  SSO_SESSION_MAX_PER_USER: "0"
+
+# Realm-level configuration
+realmConfig:
```
chance-coleman, line 50: With the addition of the new `realmConfig` value, should we add some docs to [this section](https://github.com/defenseunicorns/uds-identity-config/blob/main/docs/reference/uds-core/idam/customization.md#customizing-realm)? Even for our own sake of remember the difference between `realmInitEnv` and `realmConfig` would probably be helpful.

slaskawi: Addressed in UDS Identity Config PR

chance-coleman (approved):

PR #1863, opened 2025-08-28
fix: clean up testing methodologies

### LINKED ISSUE #1735: Clear divide between unit and e2e tests in `src/pepr`

Currently we have a large amount of `vitest` tests under `src/pepr` to test everything from policies to the operator functionality.

Some of this testing requires an active cluster, with Pepr/Core deployed. I think it would benefit us to keep a more clear divide between unit and e2e testing, so that all of the unit tests could be run without a cluster.

A few places that require a cluster:
- https://github.com/defenseunicorns/uds-core/blob/main/vitest.setup.js
- https://github.com/defenseunicorns/uds-core/tree/main/src/pepr/policies (almost all policy testing executes live applies)
- https://github.com/defenseunicorns/uds-core/tree/main/src/pepr/operator/controllers/keycloak/authservice (unsure exactly which testing, but I have encountered errors with Authservice before when running the tests without a cluster)

### Additional context

Potential ways to resolve this:
- Move e2e tests into the `test/vitest` area
- Improve mocking/function setup so that policy tests can be tested without an active cluster (i.e. separate validation logic into functions that can be tested)

chance-coleman:
### Context
Want to make a clear deliniation of unit vs e2e testing. at the moment both are present in the same areas,
making it confusing. E2E is defined as need a cluster and unit is defined as standalone test that doesnt
require a cluster. Unit tests should continue to live next to the source code that its written for and E2E
should be moved to the `test/vitest` directory.

### What has been done
- moved four e2e tests to the `test/vitest/pepr-policies` directory because they are E2E tests not unit tests:
    - `istio.spec.ts`
    - `network.spec.ts`
    - `storage.spec.ts`
    - `security.spec.ts`
- rework the mocking of `authservice.spec.ts` to not require a cluster
- rework the mocking of `client-secret-sync.spec.ts` to not require a cluster
- move the `vitest.setup.js` from the root of the project to the `test/vitest` e2e tests directory and update the `vitest.config.js` at the root of the project to not use a cluster and the `vitest.config.js` in the `test/vitest` to use a cluster since its e2e tests specifically. Also update the root `vitest.config.js` to not run e2e vitest files.
- new readme for explaining where tests go and how to run them
- add folders to test/vitest directory for better organization

## Related Issue

Fixes #1735 

## Type of change

## Checklist before merging


src/pepr/operator/controllers/keycloak/authservice/authservice.spec.ts
```diff
@@ -26,45 +26,57 @@ vi.mock("../../istio/ambient-waypoint", () => ({
...
+        Get: vi.fn().mockResolvedValue({ items: [] }),
+        // Add other chainable/query methods as needed
+      })),
+    }),
+  kind: {
+    Secret: "Secret",
+    // Add more kinds if needed
+  },
+  R: {
+    filter: <T>(fn: (value: T, index: number, array: T[]) => boolean, arr: T[]): T[] =>
```
mjnagel: I don't think we should need to mock `R` here, might just need to use the `actual` like in this example: https://github.com/defenseunicorns/uds-core/blob/8686ceb69306e049c20cd50fe422132f7c42a1ff/src/pepr/prometheus/utils.spec.ts#L12-L22

test/vitest/loki.spec.ts
```diff
@@ -4,7 +4,7 @@
 
 import * as net from "net";
 import { afterAll, beforeAll, describe, expect, test } from "vitest";
-import { closeForward, getForward } from "./forward";
+import { closeForward, getForward } from "./helpers/forward";
```
mjnagel, line 7: I wonder if we can just use the proxy feature from KFC here now ([ref](https://github.com/defenseunicorns/kubernetes-fluent-client/pull/792))? I'm not sure it's as flexible as we might need...so might be a no.

mjnagel (commented):
Maybe not for this PR, but it would be nice to setup an approach for unit testing our policies. Essentially if we pulled the logic for validations into standalone functions that return an approve/deny and message instead of putting that all in the `When...Validate` directly, then we'd be able to test those independent of a cluster for fast feedback on policy changes.

test/vitest/loki.spec.ts
chance-coleman: Looking into this i think this would be a bad idea, it would require a significant rewrite since the KFC based proxy is for single requests. Since it's not a persistent reusable tunnel its not really suitable for how our tests currently executing with multiple requests.

src/pepr/operator/controllers/keycloak/authservice/authservice.spec.ts
chance-coleman: yup agreed, removed this mock.

src/pepr/operator/controllers/keycloak/authservice/authservice.spec.ts
```diff
@@ -26,42 +26,41 @@ vi.mock("../../istio/ambient-waypoint", () => ({
...
+// --- Pepr K8s and R utility mock ---
 const mockGet = vi.fn();
 const mockUpdate = vi.fn();
 const mockApply = vi.fn();
 const mockList = vi.fn();
 
-vi.mock("kubernetes-fluent-client", async importOriginal => {
-  const actual = await importOriginal<typeof import("kubernetes-fluent-client")>();
+vi.mock("pepr", async importOriginal => {
+  const actual = (await importOriginal()) as Record<string, unknown>;
```
mjnagel: nit: in other places I think we have this typed more specifically:
```suggestion
vi.mock("pepr", async () => {
  const actual = await vi.importActual<typeof import("pepr")>("pepr");
```

mjnagel (approved):

mjnagel:
Love that we can run the pepr validations without a cluster now ❤️

PR #1870, opened 2025-09-02
fix: avoid port conflicts on the host

slaskawi:
## Description

This Pull Request avoids port conflict on the host CI machines such as https://github.com/defenseunicorns/uds-core/actions/runs/17391613494/job/49366233589

## Type of change

## Checklist before merging

### DIFF
--- a/src/test/podinfo-values.yaml
+++ b/src/test/podinfo-values.yaml
@@ -9,5 +9,5 @@ securityContext:
 
 service:
   enabled: true
-  type: NodePort
-  hostPort: 1000
+  # Using a Load Balancer (rather than a NodePort) avoids port conflicts on the host.
+  type: LoadBalancer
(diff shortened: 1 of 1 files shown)


src/test/podinfo-values.yaml
```diff
@@ -9,5 +9,5 @@ securityContext:
 
 service:
   enabled: true
-  type: NodePort
-  hostPort: 1000
```
mjnagel: Originally this was set to node/host port to test exemptions (see [here](https://github.com/defenseunicorns/uds-core/blob/fix_pod_info_nodeport_service/src/test/chart/templates/exemption5.yaml) and [here](https://github.com/defenseunicorns/uds-core/blob/fix_pod_info_nodeport_service/src/test/chart/templates/exemption4.yaml)). I've definitely seen the same issue as the linked CI run so I'd be okay with this change - although we might just want this as a standard clusterip service instead. We should also clean up those exemptions.

mjnagel (approved):

src/test/podinfo-values.yaml
slaskawi: Verified. Zero forces allocating a free port. Resolving.

PR #1872, opened 2025-09-02
chore: cleanup egress ambient test wait/delay

mjnagel:
## Description

Removes a delay in the egress ambient test - during PR review the Pepr code was updated to wait on the waypoint so this shouldn't be necessary.

## Related Issue

N/A

## Type of change


mjnagel:
Thinking about this one a bit more and looking at the egress behavior/timing.

PR #1877, opened 2025-09-02
chore: remove upgrade mutations for prometheus monitors

### LINKED ISSUE #1828: Remove Extra Prometheus ServiceMonitor and PodMonitor Mutation Logic

When transitioning to Istio ambient for the monitoring layer `0.39.X` -> `0.40.0`, there was added logic to do cleanup on ServiceMonitors and PodMonitors that existed pre-ambient. 

Hypothetically most of the logic in the mutation should not be necessary anymore if people have migrated to `0.40.0`.  We should evaluate and cleanup the unnecessary logic.

- https://github.com/defenseunicorns/uds-core/blob/3efb97cc1c03792e1c17044806bc2d853faa2129/src/pepr/prometheus/index.ts#L47-L88
- https://github.com/defenseunicorns/uds-core/blob/3efb97cc1c03792e1c17044806bc2d853faa2129/src/pepr/prometheus/index.ts#L112-L138

### Additional context

There was a bug in the cleanup logic addressed in this PR: https://github.com/defenseunicorns/uds-core/pull/1805

mjnagel:
## Description

Primarily code removal, made a small addition to the unit tests as well.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1828

## Type of change

## Steps to Validate

Unit tests validate the mutation behavior that we are keeping. Removal of other behavior should have 0 effect in CI as this was to support upgrades to 0.40.0 originally.

## Checklist before merging


slaskawi (commented):
The code LGTM but I wanted to double check if we are 100% convinced we can remove it. 

I suspect there might be some Mission Heroes who perform a long upgrade jumps. Would everything work correctly if they performed pre-ambient to the latest UDS Core version (even though they shouldn't)?

mjnagel:
> I suspect there might be some Mission Heroes who perform a long upgrade jumps. Would everything work correctly if they performed pre-ambient to the latest UDS Core version (even though they shouldn't)?

@slaskawi yeah I think this is a fair point. So as long as they make one stop in between 0.39.0 and 0.52.0 (what this will be in) - they should be fine. The mutation just needs to run once. Now if they do an upgrade jump from 0.39.0 or before, all the way to 0.52.0+ there will be _some_ issues. Some metrics/service monitors would not have the correct config (they'd still have istio sidecar related config). This is something that could be manually cleaned up so they wouldn't be entirely stuck - if we want to be extra cautious we could note this in the release notes. Ultimately version jumps are something we [document](https://uds.defenseunicorns.com/reference/deployment/upgrades/#upgrade-strategies) as untested with cautions so I think this is okay?

chance-coleman (approved):

slaskawi:
@mjnagel Yes! documenting this somehow in the release notes would be perfect! Would you mind adding the release notes override? Once done, I'm happy to approve.

mjnagel:
> @mjnagel Yes! documenting this somehow in the release notes would be perfect! Would you mind adding the release notes override? Once done, I'm happy to approve.

@slaskawi we can add/update the release notes after the fact in this case. Our only avenues for editing them now would be via the PR name (and resulting commit) or marking this as a breaking change which doesn't seem fitting imo?

joelmccoy (approved):

PR #1880, opened 2025-09-03
fix: pepr policy unit testing

chance-coleman:
## Description
Previously we had E2E testing of the pepr policies, which meant that during this PR those tests were moved to the E2E vitest directory. Because of the complexity of testing functionality inside of a the `When` `Mutate` and `Validate` we decided to refactor the code to move the uds-core business logic to its own functions so that we could avoid needing to mock all of the pepr functions. This creates straightforward simple unit testing scenarios. 

No cluster is required for these tests to run. 47 test files, 704 tests, complete under 2 seconds.

## Related Issue

Relates to #1863

## Type of change

## Checklist


src/pepr/policies/storage.ts
```diff
@@ -24,39 +25,73 @@ When(a.Pod)
...
 
-      // If the volume type is not in the allowed list, deny the request
-      if (!allowedVolumeTypes.includes(volumeType)) {
-        return request.Deny(
-          `Volume ${volume.name} has a disallowed volume type of '${volumeType}'.`,
-        );
-      }
+    if (!isValid && invalidVolume) {
+      return request.Deny(
+        `Volume '${invalidVolume.name}' has a disallowed volume type of '${invalidVolume.type}'.`,
```
mjnagel: Single quotes added here are causing the e2e test failure - will just need to update the expected text in the e2e test (or revert the single quotes).

tasks/test.yaml
```diff
@@ -121,6 +121,13 @@ tasks:
       - description: remove test resources
         task: test-resources:remove
 
+  - name: unit-tests
```
mjnagel, line 124: Not sure exactly where but wondering if we should add this task to the docs somewhere? Contributing guide or a dev doc?

chance-coleman: yeah that's a good idea.

chance-coleman: updated the contributing doc to point out unit testing and how its run.

.github/actions/unit-tests/action.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright 2024 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+name: unit-tests
+description: "Run all Core Unit Tests"
+
+runs:
+  using: composite
```
slaskawi, line 8: Just to educate me - why do we need composite here?

.github/actions/unit-tests/action.yaml
```diff
@@ -0,0 +1,15 @@
...
+  steps:
+    - name: Environment setup
+      uses: ./.github/actions/setup
+
+    - name: Run Formatting Checks
```
slaskawi: I think it's a copy-paste error and should be "Run unit tests" ?

slaskawi (dismissed):
The code LGTM. 

There's one nitpick that might be worth fixing and one question from my side.

.github/actions/unit-tests/action.yaml
chance-coleman: The `composite` type is necessary when you want to combine multiple steps into a single action. If the action only had one `run` command then it wouldn't be required but since the setup action also has `run` commands we have to use `composite`

.github/actions/unit-tests/action.yaml
chance-coleman: good catch!

.github/actions/unit-tests/action.yaml
slaskawi: Perfect! Thank you!

slaskawi (dismissed):

CONTRIBUTING.md
```diff
@@ -120,6 +123,8 @@ Depending on the scope and focus of your changes you may find other specific typ
 
 Each individual component of UDS Core contains lightweight validations in its own `src/<component>/tasks.yaml` file. These validations focus on the bare minimum functionality, typically covering pod health and endpoint accessibility.
 
+Unit test files are placed next to the source code they are testing. These tests should be designed to run without dependencies like a running cluster or mocking higher level functions, for example we want to avoid mocking pepr functions where possible.
```
mjnagel:
```suggestion
Unit test files are placed next to the source code they are testing and are used for Pepr code (under `src/pepr/`). These tests should be designed to run without requiring a running cluster.
```

I'm unsure on the last part of this doc line - I understand the intent but I think that more comes down to how the actual function code is written than the test itself?

src/pepr/policies/istio.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+  });
+
+  it("should find containers using Istio UID/GID", () => {
+    const containers = [
+      createContainer("container-1", { runAsUser: 1000 }),
+      createContainer("container-2", { runAsUser: 1337 }),
+      createContainer("container-3", { runAsGroup: 1337 }),
+    ];
+
+    expect(findContainerUsingIstioUserID(containers as V1Container[])).toBe("container-2");
```
mjnagel, line 64: This test case is a very interesting one - is there any reason we wouldn't have the `findContainerUsingIstioUserID` function return both `container-1` and `container-2` here? Right now it just returns the first match which is fine from a pure approve/deny perspective, but we could improve the denial message if we noted both containers?

src/pepr/policies/istio.ts
```diff
@@ -195,36 +203,52 @@ When(a.Pod)
...
+  for (const container of podContainers) {
+    // Check if this is an Istio proxy container
+    const isIstioProxy = isIstioProxyContainer(container);
+
+    // Only check UID/GID 1337 if this is not an Istio proxy container
+    if (!isIstioProxy) {
+      const containerCtx = container.securityContext || {};
+      if (containerCtx.runAsUser === 1337 || containerCtx.runAsGroup === 1337) {
+        return container.name;
+      }
```
mjnagel, line 250: Connected to the test comment - I wonder if this should return a list of containers if multiple are using 1337? I'm okay leaving that for after this PR since this is just restructuring to add unit tests currently.

src/pepr/policies/networking.ts
```diff
@@ -28,18 +29,29 @@ When(a.Pod)
...
+/**
+ * Checks if a pod spec is using host namespaces
+ */
+export function checkHostNamespaces(pod: V1PodSpec): boolean {
+  // If the pod is using the host network, IPC, or PID namespaces, it's invalid
+  if (pod.hostNetwork || pod.hostIPC || pod.hostPID) {
+    return false;
+  }
+  return true;
+}
```
mjnagel: This is definitely a nit, but the function name and return value here feel a bit off? `checkHostNamespaces` returns true when host namespaces aren't used 🤔. Maybe just aligning with the others where it's `checkNo`/`checkNot` would make sense here?

src/pepr/policies/networking.ts
```diff
@@ -84,13 +104,26 @@ When(a.Service)
...
   });
 
+/**
+ * Checks if a service is using ExternalName type
+ */
+export function checkNotExternalNameService(serviceSpec: V1ServiceSpec | undefined): boolean {
+  if (!serviceSpec) {
+    return true; // No spec means no ExternalName type
+  }
+  return serviceSpec.type !== "ExternalName";
```
mjnagel: nit: slight reduction in lines of code (aligns with what we did before)
```suggestion
  return serviceSpec?.type !== "ExternalName";
```

src/pepr/policies/networking.ts
```diff
@@ -107,10 +140,22 @@ When(a.Service)
...
   });
+
+/**
+ * Checks if a service is using NodePort type
+ */
+export function checkNotNodePortService(serviceSpec: V1ServiceSpec | undefined): boolean {
+  if (!serviceSpec) {
+    return true; // No spec means no NodePort type
+  }
+  return serviceSpec.type !== "NodePort";
```
mjnagel: same small nit:
```suggestion
  return serviceSpec?.type !== "NodePort";
```

mjnagel (commented):
Working through security/storage - few small comments.

src/pepr/policies/istio.spec.ts
chance-coleman: yeah i think this would be a good addition, like you say it works as is but its technically only a partial check and half the full picture.

src/pepr/policies/istio.ts
chance-coleman: agreed, this make sense to pull our of this PR.

CONTRIBUTING.md
chance-coleman: updated to suggested.

src/pepr/policies/networking.ts
chance-coleman: went with `checkNoHostNamespaces`

src/pepr/policies/networking.ts
chance-coleman: updated

src/pepr/policies/networking.ts
chance-coleman: updated

src/pepr/policies/istio.ts
mjnagel: Cool - resolving, something to think about later.

src/pepr/policies/security.ts
```diff
@@ -97,68 +118,24 @@ When(a.Pod)
...
-      pod.securityContext.runAsUser = 1000;
-    }
-
-    // Set the runAsGroup field to 1000 if it is undefined
-    if (pod.securityContext.runAsGroup === undefined) {
-      pod.securityContext.runAsGroup = 1000;
-    }
-
+    setNonRootUserSettings(request.Raw.spec!, request.Raw.metadata!);
     annotateMutation(request, Policy.RequireNonRootUser);
```
mjnagel, line 122: Maybe for a future PR - I think there's the potential that this doesn't actually mutate anything and shouldn't be annotated? Could return a true/false from `setNonRootUserSettings` to track that. This is just maintaining the current behavior on `main` so probably shouldn't change in this PR.

mjnagel (approved):
LGTM overall - great work on this.

src/pepr/policies/security.ts
mjnagel: Opened https://github.com/defenseunicorns/uds-core/issues/1900 to track.

PR #1881, opened 2025-09-03
docs: add doc on trusting private pki

### LINKED ISSUE #593: Document how to configure private PKI
Currently there is no global configuration for uds-core to configure trusted certificate bundles when using private PKI.  As a stop-gap to an automated solution we should have a document that covers how each package needs to be configured to support private PKI.

Related to:
* https://github.com/defenseunicorns/uds-core/issues/465
* https://github.com/defenseunicorns/uds-core/issues/464

Note that while additional UDS Packages may also require this, the scope for this issue is only the UDS Core components. It may be useful to add some generic notes for external applications (i.e. usage of volumes/volume mounts + typical locations to place CA certificates). This documentation may also influence how we address the related issues depending on complexity and any specific places we can improve this experience so keep this in mind while working the issue.

joelmccoy:
## Description

Adds doc on trusting private PKI.

## Related Issue

Fixes #593 

## Type of change

## Steps to Validate
- Deploy uds core locally with self signed CA and use bundle overrides provided in doc to confirm.  Detailed manual testing methodology will be dropped in PR comments

## Checklist before merging

### DIFF
--- a/docs/reference/configuration/private-pki.md
+++ b/docs/reference/configuration/private-pki.md
@@ -11,7 +11,7 @@ If your UDS Core environment connects to services using **self-signed certificat
 If you only use certificates from public, trusted CAs (e.g., Let's Encrypt, DigiCert), you do **not** need these steps.
 :::
 
-:::warning Security Consideration
+:::caution[Security Consideration]
 Mounting additional volumes and certificates can introduce minimal security risks by expanding the attack surface. Only mount trusted CA certificates from verified sources and regularly audit your certificate configurations.
 :::
 
--- a/docs/reference/configuration/private-pki.md
+++ b/docs/reference/configuration/private-pki.md
@@ -47,7 +47,7 @@ Authservice can be configured with an additional trusted CA bundle when UDS Core
 
 To configure, set `UDS_CA_CERT` as an environment variable with a Base64 encoded PEM formatted CA bundle that can be used to verify the certificates of the tenant gateway.
 
-See [trusted-ca.md](./single-sign-on/trusted-ca.md) for complete Authservice configuration details.
+See [trusted CA SSO doc](/reference/configuration/single-sign-on/trusted-ca) for complete Authservice configuration details.
 
 ### Grafana
 
--- a/docs/reference/configuration/private-pki.md
+++ b/docs/reference/configuration/private-pki.md
@@ -125,22 +125,28 @@ This approach replaces the system CA bundle for all Loki components. Ensure your
 
 ### Keycloak
 
-Keycloak needs to validate certificates when connecting to external identity providers or LDAP servers. Configure Keycloak to trust private certificates using its built-in truststore mechanism:
+Keycloak needs to validate certificates when connecting to external identity providers or LDAP servers. Configure Keycloak to trust private certificates using additional truststore paths:
 
 ```yaml
 values:
   - path: extraVolumeMounts
     value:
       - name: ca-certs
-        mountPath: /opt/keycloak/conf/truststores/private-ca
+        mountPath: /tmp/ca-certs
         readOnly: true
   - path: extraVolumes
     value:
       - name: ca-certs
         configMap:
           name: private-ca
+  - path: truststorePaths
+    value:
+      - "/tmp/ca-certs"
 ```
 
-Keycloak automatically scans the `conf/truststores` directory for certificate files and adds them to its system truststore, preserving existing Java default certificates.
+This configuration:
+- Mounts your private CA ConfigMap to `/tmp/ca-certs`
+- Configures Keycloak to scan this directory using the `--truststore-paths` startup argument
+- Preserves Keycloak's default truststore while adding your private certificates
 
 For additional details on Keycloak truststore configuration, see the [upstream Keycloak documentation](https://www.keycloak.org/server/keycloak-truststore#_configuring_the_system_truststore).
--- a/docs/reference/configuration/private-pki.md
+++ b/docs/reference/configuration/private-pki.md
@@ -1,0 +1,146 @@
+---
+title: Private Certificate Authority (CA) Configuration
+---
+
+Some UDS Core components need to connect to external services over TLS. By default, they trust the well-known public certificate authorities (CAs) that come with their container images. If your environment uses self-signed certificates or certificates issued by a private CA, these components will not trust those endpoints unless you explicitly provide the CA bundle.
+
+This guide explains how to configure UDS Core components to recognize and trust your private CA certificates. Not every component requires this configuration — only those that make outbound TLS connections (for example, to identity providers, object storage, or other HTTPS endpoints).
+
+:::tip Who should use this guide?
+If your UDS Core environment connects to services using **self-signed certificates** or certificates issued by a **private CA**, you’ll need to follow this configuration.  
+If you only use certificates from public, trusted CAs (e.g., Let’s Encrypt, DigiCert), you do **not** need these steps.
+:::
+
+## Prerequisites
+
+Before configuring private PKI, you'll need the following:
+
+1. The trusted CA bundle in PEM format that your certificates are signed by
+2. A ConfigMap containing the trusted CA bundle from (1), available in each namespace (a tool like [trust-manager](https://cert-manager.io/docs/trust/trust-manager/) can help automate this)
+
+:::note
+For the examples in this guide, we assume you have a ConfigMap named `private-ca` with a key `ca.pem` that contains your CA bundle.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: private-ca
+  namespace: <your-namespace>
+data:
+  ca.pem: |
+    -----BEGIN CERTIFICATE-----
+    MIIDeTCCAmGgAwIBAgIUQj...<snip>...hQ==
+    -----END CERTIFICATE-----
+```
+:::
+
+## Component Configuration
+
+### Authservice
+
+Authservice can be configured with an additional trusted CA bundle when UDS Core ingress gateways are deployed with private PKI.
+
+To configure, set `UDS_CA_CERT` as an environment variable with a Base64 encoded PEM formatted CA bundle that can be used to verify the certificates of the tenant gateway.
+
+See [trusted-ca.md](./single-sign-on/trusted-ca.md) for complete Authservice configuration details.
+
+### Grafana
+
+Grafana may need to access external https services.  If these services use certificates signed by a private CA, Grafana needs to be configured to trust these certificates.
+
+Configure Grafana to trust private certificates by mounting and overriding the CA bundle via UDS Bundle overrides:
+
+```yaml
+values:
+  - path: extraConfigMapMounts
+    value:
+      - name: ca-certs
+        mountPath: /etc/ssl/certs/ca-certificates.crt # For Debian/Ubuntu images
+        # For RedHat-based images (registry1 flavor), use:
+        # mountPath: /etc/pki/tls/certs/ca-bundle.crt
+        configMap: private-ca
+        readOnly: true
+        subPath: ca.pem
+```
+
+:::caution
+Mounting to `/etc/ssl/certs/ca-certificates.crt` replaces the system CA bundle. Ensure your CA bundle includes both your private CA certificates and any public CAs that Grafana needs to trust.
+:::
+
+For additional details on Grafana private CA configuration, see the [upstream Grafana documentation](https://grafana.com/docs/grafana/latest/setup-grafana/installation/helm/#configure-a-private-ca-certificate-authority).
+
+### Velero
+
+Velero may need to access external backup storage services. If these services use certificates signed by a private CA, Velero needs to be configured to trust these certificates.
+
+Configure Velero to trust private certificates by mounting the CA bundle via UDS Bundle overrides:
+
+```yaml
+values:
+  - path: extraVolumeMounts
+    value:
+      - name: ca-certs
+        mountPath: /etc/ssl/certs/ca-certificates.crt # For Debian/Ubuntu images
+        # For RedHat-based images (registry1 flavor), use:
+        # mountPath: /etc/pki/tls/certs/ca-bundle.crt
+        subPath: ca.pem
+        readOnly: true
+  - path: extraVolumes
+    value:
+      - name: ca-certs
+        configMap:
+          name: private-ca
+```
+
+:::caution
+Mounting to `/etc/ssl/certs/ca-certificates.crt` replaces the system CA bundle. Ensure your CA bundle includes both your private CA certificates and any public CAs that Velero needs to trust.
+:::
+
+### Loki
+
+Loki components need to access external storage backends. To configure Loki to use S3-compatible storage with private certificates, you need to provide the CA bundle to all components.
+
+Configure Loki to trust private certificates by overriding the system CA bundle using global volume mounts:
+
+```yaml
+values:
+  - path: global.extraVolumeMounts
+    value:
+      - name: ca-certs
+        mountPath: /etc/ssl/certs/ca-certificates.crt # For Debian/Ubuntu images
+        # For RedHat-based images (registry1 flavor), use:
+        # mountPath: /etc/pki/tls/certs/ca-bundle.crt
+        subPath: ca.pem
+  - path: global.extraVolumes
+    value:
+      - name: ca-certs
+        configMap:
+          name: private-ca
+```
+
+:::caution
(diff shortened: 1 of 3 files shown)


docs/reference/configuration/private-pki.md
joelmccoy, line 1: Not sure if this is the best place or filename for this doc. Let me know if you have better suggestions.

chance-coleman: i think the name of the file is good 👍🏼. I do wonder if it would make more sense to live in the `configuration/service-mesh/` directory though, not something i feel strongly about but that's where i would look first personally.

joelmccoy: is this doc relevant to the service mesh though?  I think this doc is more for setting up individual applications to trust additional private self signed certs.  It doesn't really modify service mesh behavior, more so the individual applications.

mjnagel: I agree that this doc itself probably doesn't belong in the service mesh section, but I think it would be good to add a link to this new doc in the [note block here](https://github.com/defenseunicorns/uds-core/blob/593_document_private_pki/docs/reference/configuration/service-mesh/ingress.md?plain=1#L107).

docs/reference/configuration/single-sign-on/trusted-ca.md
```diff
@@ -4,14 +4,21 @@ title: Trusted Certificate Authority
...
+packages:
+  - name: core
+    repository: ghcr.io/defenseunicorns/packages/uds/core
+    ref: x.x.x
+    overrides:
+      uds-operator-config:
+        uds-operator-config:
+          values:
+            - path: cluster.expose.caCert
+              value: <base64 encoded certificate authority>
```
mjnagel: Out of curiosity what was the intent with this change? The previous example should be valid since it is still exposed as a zarf var ([here](https://github.com/defenseunicorns/uds-core/blob/593_document_private_pki/src/pepr/uds-operator-config/values.yaml#L19)). Maybe not a strong feeling, but I'd lean towards using zarf vars where available instead of bundle overrides.

docs/reference/configuration/private-pki.md
```diff
@@ -0,0 +1,152 @@
+---
+title: Private Certificate Authority (CA) Configuration
+---
+
+Some UDS Core components need to connect to external services over TLS. By default, they trust the well-known public certificate authorities (CAs) that come with their container images. If your environment uses self-signed certificates or certificates issued by a private CA, these components will not trust those endpoints unless you explicitly provide the CA bundle.
+
+This guide explains how to configure UDS Core components to recognize and trust your private CA certificates. Not every component requires this configuration — only those that make outbound TLS connections (for example, to identity providers, object storage, or other HTTPS endpoints).
+
+:::tip Who should use this guide?
```
mjnagel:
```suggestion
:::tip[Who should use this guide?]
```

docs/reference/configuration/private-pki.md
```diff
@@ -0,0 +1,156 @@
...
+If your UDS Core environment connects to services using **self-signed certificates** or certificates issued by a **private CA**, you'll need to follow this configuration.
+If you only use certificates from public, trusted CAs (e.g., Let's Encrypt, DigiCert), you do **not** need these steps.
+:::
+
+:::warning Security Consideration
```
mjnagel:
```suggestion
:::caution[Security Consideration]
```

docs/reference/configuration/private-pki.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Component Configuration
+
+### Authservice
+
+Authservice can be configured with an additional trusted CA bundle when UDS Core ingress gateways are deployed with private PKI.
+
+To configure, set `UDS_CA_CERT` as an environment variable with a Base64 encoded PEM formatted CA bundle that can be used to verify the certificates of the tenant gateway.
+
+See [trusted-ca.md](./single-sign-on/trusted-ca.md) for complete Authservice configuration details.
```
mjnagel: This link will need to be a link to the doc from the docs-site perspective:
```suggestion
See [trusted CA SSO doc](/reference/configuration/single-sign-on/trusted-ca) for complete Authservice configuration details.
```

joelmccoy:
Here is how I tested this:

- Created a self signed CA and certs for `udstest.dev`. Helper script:
```
#!/bin/bash

set -euo pipefail

# === Local CA (root of trust) ===
openssl genrsa -out uds-local-ca.key 4096

cat > uds-local-ca.cnf <<'EOF'
[req]
prompt = no
distinguished_name = dn
x509_extensions = v3_ca
[dn]
CN = UDS Local CA
[v3_ca]
subjectKeyIdentifier = hash
authorityKeyIdentifier = keyid:always,issuer
basicConstraints = critical, CA:true
keyUsage = critical, keyCertSign, cRLSign
EOF

openssl req -x509 -new -nodes -key uds-local-ca.key \
  -days 3650 -sha256 \
  -out uds-local-ca.crt \
  -config uds-local-ca.cnf

# === udstest.dev (covers udstest.dev and *.udstest.dev) ===
openssl genrsa -out udstest.dev.key 2048

cat > udstest.dev.cnf <<'EOF'
[req]
prompt = no
distinguished_name = dn
req_extensions = v3_req
[dn]
CN = udstest.dev
[v3_req]
subjectAltName = @alts
basicConstraints = critical, CA:false
keyUsage = critical, digitalSignature, keyEncipherment
extendedKeyUsage = serverAuth
[alts]
DNS.1 = udstest.dev
DNS.2 = *.udstest.dev
EOF

openssl req -new -key udstest.dev.key -out udstest.dev.csr -config udstest.dev.cnf
openssl x509 -req -in udstest.dev.csr \
  -CA uds-local-ca.crt -CAkey uds-local-ca.key -CAcreateserial \
  -out udstest.dev.crt -days 825 -sha256 \
  -extensions v3_req -extfile udstest.dev.cnf

# === udstest.admin.dev (covers udstest.admin.dev and *.udstest.admin.dev) ===
openssl genrsa -out udstest.admin.dev.key 2048

cat > udstest.admin.dev.cnf <<'EOF'
[req]
prompt = no
distinguished_name = dn
req_extensions = v3_req
[dn]
CN = udstest.admin.dev
[v3_req]
subjectAltName = @alts
basicConstraints = critical, CA:false
keyUsage = critical, digitalSignature, keyEncipherment
extendedKeyUsage = serverAuth
[alts]
DNS.1 = udstest.admin.dev
DNS.2 = *.udstest.admin.dev
EOF

openssl req -new -key udstest.admin.dev.key -out udstest.admin.dev.csr -config udstest.admin.dev.cnf
openssl x509 -req -in udstest.admin.dev.csr \
  -CA uds-local-ca.crt -CAkey uds-local-ca.key -CAcreateserial \
  -out udstest.admin.dev.crt -days 825 -sha256 \
  -extensions v3_req -extfile udstest.admin.dev.cnf

# === Base64 encode for Kubernetes ===
echo "===== Kubernetes Secret values for udstest.dev ====="
echo "tls.crt:"
base64 < udstest.dev.crt | tr -d '\n'; echo
echo
echo "tls.key:"
base64 < udstest.dev.key | tr -d '\n'; echo
echo "===================================================="

echo "===== Kubernetes Secret values for udstest.admin.dev ====="
echo "tls.crt:"
base64 < udstest.admin.dev.crt | tr -d '\n'; echo
echo
echo "tls.key:"
base64 < udstest.admin.dev.key | tr -d '\n'; echo
echo "=========================================================="

# === CA bundle for clients (trust this CA) ===
cp uds-local-ca.crt ca-bundle.pem
```
- Deployed k3d-standard with these cert values for the tenant/admin gateway
- Setup local DNS to point udstest.dev to localhost
- Deployed the self signed CA bundle as configmaps to all relevant namespaces
- Deployed k3d-standard core with the overrides in this doc that points to the configmaps
- Execed into each pod that has the overrides and used curl to try to hit sso.udstest.dev (validated the cert was trusted)
  - Did this for all flavors and ensured I had the proper proper path based on the base OS (registry1 utilizes redhat based images and has a different path)
  - For keycloak, I just ensured the additional cert was loaded into the trust store with `keytool -list -keystore /opt/keycloak/data/keycloak-truststore.p12  -storetype PKCS12`

docs/reference/configuration/private-pki.md
```diff
@@ -0,0 +1,156 @@
+---
+title: Private Certificate Authority (CA) Configuration
+---
+
+Some UDS Core components need to connect to external services over TLS. By default, they trust the well-known public certificate authorities (CAs) that come with their container images. If your environment uses self-signed certificates or certificates issued by a private CA, these components will not trust those endpoints unless you explicitly provide the CA bundle.
+
+This guide explains how to configure UDS Core components to recognize and trust your private CA certificates. Not every component requires this configuration — only those that make outbound TLS connections (for example, to identity providers, object storage, or other HTTPS endpoints).
```
mjnagel: Maybe this is clear enough, but I wonder if we should explicitly list 2 scenarios where this might be required:
1. Keycloak/your domain cert is self-signed or private PKI: At least Grafana and NeuVector? would need the CA I believe for your domain cert.
2. External dependencies (object storage maybe being the most likely) use private PKI: Velero, Loki (object storage) and then Grafana/Keycloak potentially for DBs, datasources, external IdPs?

Maybe this is all laid out in per-component sections well enough, but might be nice to have some up front clarification especially on the identity (Keycloak) one. Might also be worth noting that this covers UDS Core components, but you may need to take similar steps for other packages on top of core.

mjnagel (commented):
Generally good with this - I added some format comments on a few issues during preview of the doc. `uds run dev-docs` locally should show you the same preview issues I was seeing.

I'm also curious - do we have a way to make NeuVector (manager) trust a private CA? I expect this might be needed if Keycloak is using a private PKI cert?

docs/reference/configuration/single-sign-on/trusted-ca.md
joelmccoy: ahh i was totally confused here. I thought this variable block was for a uds-bundle helm variable.  Missed the fact that this is really a zarf variable. Going to revert this change.  Agreed with the sentiment of leaning towards the zarf variable here.

mjnagel:
> I'm also curious - do we have a way to make NeuVector (manager) trust a private CA? I expect this might be needed if Keycloak is using a private PKI cert?

I deployed with self-signed certs just a bit ago and did not seem to need to add the CA to NeuVector which is very interesting 🤔

joelmccoy:
> > I'm also curious - do we have a way to make NeuVector (manager) trust a private CA? I expect this might be needed if Keycloak is using a private PKI cert?
> 
> I deployed with self-signed certs just a bit ago and did not seem to need to add the CA to NeuVector which is very interesting 🤔

This is also what I am noticing as well.  I guess the built in http library that does the SSO exchange doesn't verify TLS certs or maybe everything happens client side in this SSO exchange (i.e. Neuvector doesn't actually make a request to the sso endpoint, it just provides the redirect to the browser client)?  Would it be safe to leave Neuvector out of this doc (especially if it works fine and also the pending replacement of Neuvector)?

docs/reference/configuration/private-pki.md
joelmccoy: added

docs/reference/configuration/private-pki.md
joelmccoy: added a blurb to cover this. lmk if it is sufficient.

mjnagel:
> Would it be safe to leave Neuvector out of this doc (especially if it works fine and also the pending replacement of Neuvector)?

Yeah I think that's fine for now.

docs/reference/configuration/private-pki.md
mjnagel: LGTM! We can't exhaustively anticipate scenarios, but I think this + the ingress doc link might help provide a quick up front understanding of when this is useful.

docs/reference/configuration/private-pki.md
```diff
@@ -0,0 +1,160 @@
...
+Grafana may need to access external https services.  If these services use certificates signed by a private CA, Grafana needs to be configured to trust these certificates.
+
+Configure Grafana to trust private certificates by mounting and overriding the CA bundle via UDS Bundle overrides:
+
+```yaml
+values:
+  - path: extraConfigMapMounts
+    value:
+      - name: ca-certs
+        mountPath: /etc/ssl/certs/ca-certificates.crt # For Debian/Ubuntu images
```
mjnagel: just a small nit: should this specify `(unicorn or upstream flavor)` to be explicit about when to use this mount path based on flavor?

docs/reference/configuration/private-pki.md
```diff
@@ -0,0 +1,160 @@
...
+Loki components need to access external storage backends. To configure Loki to use S3-compatible storage with private certificates, you need to provide the CA bundle to all components.
+
+Configure Loki to trust private certificates by overriding the system CA bundle using global volume mounts:
+
+```yaml
+values:
+  - path: global.extraVolumeMounts
+    value:
+      - name: ca-certs
+        mountPath: /etc/ssl/certs/ca-certificates.crt # For Debian/Ubuntu images
```
mjnagel: same nit here: should this include the flavors where this is relevant?

docs/reference/configuration/private-pki.md
```diff
@@ -0,0 +1,160 @@
...
+        configMap: private-ca
+        readOnly: true
+        subPath: ca.pem
+```
+
+:::caution
+Mounting to `/etc/ssl/certs/ca-certificates.crt` replaces the system CA bundle. Ensure your CA bundle includes both your private CA certificates and any public CAs that Grafana needs to trust.
+:::
+
+For additional details on Grafana private CA configuration, see the [upstream Grafana documentation](https://grafana.com/docs/grafana/latest/setup-grafana/installation/helm/#configure-a-private-ca-certificate-authority).
```
mjnagel: This is a good caution and something we'll need to consider if we expand features around CA certs - in some cases it may be desirable to only trust a single private CA, in others you might want both private PKI + standard public PKI certs.

Currently the CA_CERT value we require/use for Authservice is documented as just including the domain CA. If we were to build off of that we might want to find a way to include public PKI CAs separately and join them conditionally based on end user "requests" (maybe something in the Package CR, etc).

One thing I found interesting reviewing the upstream doc is that they actually reference a different path for mounting an extra CA (`/etc/ssl/certs/ca.pem`). Does that behave different than the approach in this doc (adding a CA vs replacing all the system CAs), and might there be cases where we'd suggest that instead?

mjnagel (commented):
LGTM overall - a few last comments/nits here.

docs/reference/configuration/private-pki.md
joelmccoy: had to do some digging on where `/etc/ssl/certs/ca.pem` came from and i traced it to the go crypto library happens to check this directory for additional certs: https://github.com/golang/go/blob/go1.25.1/src/crypto/x509/root_linux.go#L21

So for grafana this is why they specify this directory as a quick drop in replacement that doesn't overwrite your system CA.  

I could see the value in specifying to a user that you can add a CA cert OR overwriting the system CA to be more specific.  I will update this with both paths.  I can easily test this for grafana with the SSO integration.

But for the other applications I might need to do some digging.  If they are standard go applications this path should just work?

mjnagel: Thanks for diving into this - I like the idea of (at least for grafana) documenting both options. Both Loki and Velero are go based so I suspect they might be similar?

I also found [this doc](https://velero.io/docs/main/self-signed-certificates/) from Velero upstream which lead me to [this value](https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/values.yaml#L388-L389) in the helm chart. I think it probably makes more sense to try and align with the other apps (mounting the CA to be trusted by the container itself) especially since I'm not seeing the same value for volume snapshots.

docs/reference/configuration/private-pki.md
joelmccoy: added

docs/reference/configuration/private-pki.md
joelmccoy: added

docs/reference/configuration/private-pki.md
joelmccoy: added a small section to clarify the CA overwrite vs adding an additional cert.  Also updated all the references for paths in the configs for grafana, velero, and loki to list out the options for overrides including `/etc/ssl/certs/ca.pem`.  lmk if this reads well

mjnagel (dismissed):

docs/reference/configuration/private-pki.md
mjnagel: Looks good :+1:

mjnagel (approved):

PR #1882, opened 2025-09-03
feat: add helm values for extra volumes and truststorepaths to keycloak

joelmccoy:
## Description

- Adds the ability to attach extra volumes and volume mounts to the Keycloak statefulset.  
- Adds a value to set the `--truststore-paths` to provide additional truststore directories you would like keycloak to read from (ref: https://www.keycloak.org/server/keycloak-truststore#_configuring_the_system_truststore)

This is useful for injecting additional private CA bundles for Keycloak to trust.

## Related Issue

Relates to #593 

## Type of change

## Steps to Validate
- create private CA cert bundle and put it in a configmap called `private-ca` in the keycloak namespace
- deploy slim-dev bundle with these overrides:
```yaml
values:
  - path: extraVolumeMounts
    value:
      - name: ca-certs
        mountPath: /tmp/ca-certs
        readOnly: true
  - path: extraVolumes
    value:
      - name: ca-certs
        configMap:
          name: private-ca
  - path: truststorePaths
    value:
      - "/tmp/ca-certs"
```
- exec into the keycloak pod and run `keytool -list -keystore /opt/keycloak/data/keycloak-truststore.p12  -storetype PKCS12` and verify the CA you added is in this output

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -109,6 +109,9 @@ spec:
             {{- else }}
             - "--features=preview"
             {{- end }}
+            {{- if .Values.truststorePaths }}
+            - "--truststore-paths={{ join "," .Values.truststorePaths }}"
+            {{- end }}
             - "--spi-authentication-sessions--map--auth-sessions-limit={{ .Values.realmConfig.maxInFlightLoginsPerUser | default 300 }}"
             - "--proxy-headers=xforwarded"
             - "--http-enabled=true"
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -292,6 +292,9 @@ spec:
               readOnly: true
             - name: client-secrets
               mountPath: /var/run/secrets/uds/client-secrets
+            {{- with .Values.extraVolumeMounts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
       enableServiceLinks: {{ .Values.enableServiceLinks }}
       restartPolicy: {{ .Values.restartPolicy }}
       {{- with .Values.nodeSelector }}
@@ -372,4 +375,7 @@ spec:
                       path: tc.txt
               {{- end }}
               {{- end }}
+        {{- end }}
+        {{- with .Values.extraVolumes }}
+        {{- toYaml . | nindent 8 }}
         {{- end }}
--- a/src/keycloak/chart/values.yaml
+++ b/src/keycloak/chart/values.yaml
@@ -371,3 +371,9 @@ thirdPartyIntegration:
     # The Client Certificate format used by the L7 Load Balancer.
     # Allowed settings: "PEM" and "AWS"
     tlsCertificateFormat: PEM
+
+# Extra volumes to mount in the StatefulSet
+extraVolumes: []
+
+# Extra volume mounts to mount in the Keycloak container
+extraVolumeMounts: []
--- a/src/keycloak/chart/values.yaml
+++ b/src/keycloak/chart/values.yaml
@@ -377,3 +377,7 @@ extraVolumes: []
 
 # Extra volume mounts to mount in the Keycloak container
 extraVolumeMounts: []
+
+# Additional truststore paths for Keycloak to scan for certificates
+# These paths will be added to the --truststore-paths startup argument
+truststorePaths: []
--- a/src/keycloak/chart/tests/kc_extra_volumes_test.yaml
+++ b/src/keycloak/chart/tests/kc_extra_volumes_test.yaml
@@ -1,0 +1,46 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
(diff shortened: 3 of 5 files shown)


slaskawi:
I think we're mostly fine with this approach but in case of Keycloak it opens the door for abuse and we need to carefully document it. 

Keycloak uses a single Truststore for everything - including CAC validation. This approach will likely change in the future but there's nothing in the roadmap for short and mid term that can help us with this. I would love to hear @bburky 's opinion here but at least we should provide a big red warning in the docs to ensure our Mission Heros know the risk of using this feature.

joelmccoy:
Reference the keycloak instructions included in the docs in this PR for more context: https://github.com/defenseunicorns/uds-core/pull/1881

src/keycloak/chart/values.yaml
```diff
@@ -371,3 +371,13 @@ thirdPartyIntegration:
     # The Client Certificate format used by the L7 Load Balancer.
     # Allowed settings: "PEM" and "AWS"
     tlsCertificateFormat: PEM
+
+# Extra volumes to mount in the StatefulSet
+extraVolumes: []
+
+# Extra volume mounts to mount in the Keycloak container
+extraVolumeMounts: []
```
mjnagel, line 379: I don't have any immediate concerns here but just to ask - are there any issues with exposing volumes generically here? Or would we want to restrict the directory or something to tailor this for tmp dirs, etc only?

It seems like a pretty generic function to include in a helm chart but I know we're generally cautious around exposing too many knobs (like the args) to make sure we don't have easy security holes.

joelmccoy: I personally see it as not a big concern.  

This is something that is exposed on other upstream helm charts for other applications in core and is a pretty common pattern.

I'm not sure the immediately security value gain if we limited which directories we could mount to.  At the end of the day you can control the content of what gets mounted.  Doesn't matter where the directory is located IMO.

But would definitely defer to others opinion/vote on this.

mjnagel: I generally agree - wasn't sure if there might be directories we'd want to limit due to Keycloak automatically loading content from there (such as [the ones we leverage for identity-config](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/sync.sh#L9-L13)). Overall I think this is fine and probably enables more interesting config beyond the initial intent - just wanted to drop a note in case @slaskawi or someone else could think of an issue with this.

Ultimately we want UDS to be "more secure" but always a balance in security and configurability - ensuring we have a good secure baseline but also generally exposing _some knobs_ for more flexibility.

slaskawi: I'm fine with this but I'm insisting on a corresponding warning in the docs and I would love to have @bburky 's thumbs up on this PR as well.

bburky: If you're configuring Keycloak, your're a cluster admin. So I'm not too worried security-wise about someone using arbitrary volumes. Assuming it isn't too error prone or something.

And I think all the DoD certs can exceed the size of a single ConfigMap if I remember. So mounting two here is nice

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -109,6 +109,9 @@ spec:
             {{- else }}
             - "--features=preview"
             {{- end }}
+            {{- if .Values.truststorePaths }}
+            - "--truststore-paths={{ join "," .Values.truststorePaths }}"
+            {{- end }}
```
bburky, line 114: A non-issue, but FYI: "By default, Keycloak uses the System Truststore to validate certificates" (mTLS x509 CAC). This isn't really an issue at all as long as we continue to use Istio to do cert validation before the certificate reaches Keycloak.

https://www.keycloak.org/server/mutual-tls#_using_a_dedicated_truststore_for_mtls

src/keycloak/chart/values.yaml
mjnagel: > And I think all the DoD certs can exceed the size of a single ConfigMap if I remember. So mounting two here is nice

Tangent but so far we've been okay putting all the CAs into a single secret (we do this for the gateway CA). Looks like maybe [credentialNames](https://istio.io/latest/docs/reference/config/networking/gateway/#ServerTLSSettings-credential_names) could be an option once we hit that threshold (I think we're close).

joelmccoy: Sounds like we are ok with this from a security perspective.  I added an additional warning callout in the WIP private pki doc here: https://github.com/defenseunicorns/uds-core/pull/1881

@slaskawi Does that seem sufficient?

src/keycloak/chart/templates/statefulset.yaml
joelmccoy: Ack - It might be nice to allow this override in the future if we wanted to allow a full override on what CA certs Keycloak trusts.  For now, having the truststore paths allows additional certs to be injected into the default trust store.  Allowing Keycloak external connections to trust additional self-signed/private pki.

mjnagel (approved):

src/keycloak/chart/values.yaml
slaskawi: Looks good on my side. I'm OK to resolve.

slaskawi (approved):
LGTM!

PR #1891, opened 2025-09-09
chore: bump uds-k3d to 0.17.0

mjnagel:
## Description

Separating out from https://github.com/defenseunicorns/uds-core/pull/1876 since the uds-cli update will require more work there.

This update includes k3s 1.33.4 as well as some other dependency updates and new features.


mjnagel:
Seems like we're encountering the same issue where healthprobes are accidentally being captured by ztunnel and blocked due to STRICT mTLS. In some local testing this seemed better with Istio 1.27 but not super confident that's the answer.

PR #1898, opened 2025-09-11
feat: keycloak sizing

### LINKED ISSUE #1748: Decrease the maximum amount of Keycloak Pods
## Summary 

Currently, Keycloak's HPA can spin up up to 10 Pods, which seems too high.

Looking at the Keycloak performance figures I have at hand, Keycloak roughly performs at 160 req/s (since Authorization Code Flow is 3 steps, this is roughly 50 user authentications/s) in a 3-node setup. This goes roughly to 3k users per minute and seems sufficient for all the deployments out there.

The proposed solution is to decrease the maximum number of Keycloak Pods to 5, which is already way above the needs.

slaskawi:
## Description

Based on the upstream conversation (https://github.com/keycloak/keycloak/discussions/42681), this Pull Request adjusts CPU and Memory size to Keycloak deployments leaving the HPA support as is. 

The highlights of the upstream discussion:

* Starting from 26.4.0 release, the improved JDBC_PING2 protocol helps addressing split brain scenarios. This mean there's no longer a need to keep the odd number of replicas.
* Keycloak doesn't use `DENY_READ_WRITES` for the default cache configuration, so it won't go into the degraded mode. Again - this helps addressing Split Brain scenarios. 
* Starting from 26.4.0, Multi AZ becomes de-facto standard deployment model.

<details>

<summary>Memory and CPU measurements</summary>

M2 MacOS (Keycloak Pods)

image

M2 MacOS (Keycloak Waypoint)

image
 
M2 MacOS (Keycloak Pods)

image

M2 MacOS (Keycloak Waypoint)

image

Fedora (aggregated CPU)

obraz

Fedora x86 (Intel i9) (aggregated memory)

obraz

**Note that Linux figures are lower both for memory and CPU. Therefore the default settings have been calibrated to MacOS**

</details>

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1748

## Type of change

## Steps to Validate

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/values.yaml
+++ b/src/keycloak/chart/values.yaml
@@ -184,8 +184,8 @@ podLabels: {}
 # Pod resource requests and limits
 resources:
   requests:
-    cpu: "500m"
-    memory: "512Mi"
+    cpu: "100m"
+    memory: "700Mi"
   limits:
     cpu: "1"
     memory: "1Gi"
--- a/bundles/k3d-slim-dev/uds-bundle.yaml
+++ b/bundles/k3d-slim-dev/uds-bundle.yaml
@@ -157,7 +157,7 @@ packages:
             - name: KEYCLOAK_WAYPOINT_MEMORY_REQUEST
               description: "Memory Requests for the waypoint"
               path: "waypoint.deployment.requests.memory"
-              default: "128Mi"
+              default: "64Mi"
             - name: INSECURE_ADMIN_PASSWORD_GENERATION
               description: "Generate an insecure admin password for dev/test"
               path: insecureAdminPasswordGeneration.enabled


src/keycloak/chart/values.yaml
```diff
@@ -319,45 +321,27 @@ prometheusRule:
...
+  enabled: false
+  # Upper bound for replicas when HA mode is enabled (should be an odd number in practice)
+  maxReplicas: 3
+
+# Deprecated, use `autoscaling` instead
 autoscaling:
-  # If `true`, an autoscaling/v2beta2 HorizontalPodAutoscaler resource is created (requires Kubernetes 1.18 or above)
-  # Autoscaling seems to be most reliable when using KUBE_PING service discovery (see README for details)
-  # This disables the `replicas` field in the StatefulSet
+  # Deprecated, use `autoscaling.enabled` instead
```
```suggestion
  # Deprecated, use `highAvailability.enabled` instead
```

src/keycloak/chart/values.yaml
```diff
@@ -319,45 +321,27 @@ prometheusRule:
...
 # Istio waypoint configuration
 waypoint:
   deployment:
-    # Larger requests by default since Keycloak manages all SSO for the cluster
     requests:
-      cpu: 250m
-      memory: 256Mi
-    # Upstream defaults for limits
+      cpu: 20m
+      memory: 60Mi
```
mjnagel: We need to be very careful setting these requests so low. It might be fine with 0 load, but as soon as there is load we're likely to hit the 90% of 20m CPU very quickly causing the HPA to scale for the waypoint.

I would prefer to leave both the requests and limits where they were, or reduce the requests by a smaller margin.  In my previous testing using [vegeta](https://github.com/tsenart/vegeta) and hitting the `sso.uds.dev` endpoint, I would see the waypoint hit 100-150m CPU usage (don't recall memory). We may be okay to drop the CPU requests to 150m or 200m. Unsure what is reasonable for memory - that one we might be okay to drop more since the HPA doesn't consider it.

The limits for reference are from the upstream defaults. I'd be hesitant to drop them?

joelmccoy: agreed with this sentiment. i'd prefer they be left as is. I noticed the same spikes testing locally with load.  And having low resource requests will trigger the HPA scaling pretty quickly

src/keycloak/chart/values.yaml
```diff
@@ -187,8 +189,8 @@ podLabels: {}
 # Pod resource requests and limits
 resources:
   requests:
-    cpu: "500m"
-    memory: "512Mi"
+    cpu: "100m"
+    memory: "750Mi"
```
joelmccoy: I think it would be good to compare your local testing with what we see in the registry envs for these request numbers to ensure these numbers make sense.

src/keycloak/chart/values.yaml
```diff
@@ -319,45 +321,27 @@ prometheusRule:
...
 
+# High availability options for deployments where pod count is controlled outside HPA
+highAvailability:
+  # If `true`, enables HA-recommended settings without using HPA
+  enabled: false
+  # Upper bound for replicas when HA mode is enabled (should be an odd number in practice)
+  maxReplicas: 3
+
+# Deprecated, use `autoscaling` instead
 autoscaling:
```
joelmccoy, line 319: Is the plan to remove this value at some point?  If so, might be good to cut a ticket to clean this up so we don't forget about it

src/keycloak/chart/values.schema.json
```diff
@@ -140,6 +140,14 @@
     "autoscaling": {
       "$ref": "#/definitions/autoscalingSchema"
     },
+    "highAvailability": {
+      "type": "object",
+      "additionalProperties": false,
+      "properties": {
+        "enabled": { "type": "boolean" },
+        "maxReplicas": { "type": "number" }
```
joelmccoy:
```suggestion
        "maxReplicas": { 
           "type": "number",
           "not": {
            "multipleOf": 2
           }
        }
```

I think you can also add this little check to make sure people aren't deploying even number of instances

joelmccoy (commented):
Generally looks good!  Add a couple comments for consideration

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -12,6 +12,11 @@ metadata:
     {{- printf "%s: %s" $key (tpl $value $ | quote) | nindent 4 }}
     {{- end }}
 spec:
+  {{- if and (not .Values.autoscaling.enabled) (not .Values.highAvailability.enabled) }}
+  replicas: 1
+  {{- else }}
+  replicas: {{ max (int .Values.highAvailability.maxReplicas) (int .Values.autoscaling.maxReplicas) }}
```
briantwatson: I'm thinking through this and am I accurate we are keeping the usage of the of the `autoscaling.maxReplicas` here for backwards compatibility?

Would using minReplicas be better?  Just any system that has these values set that's referencing this chart would automatically scale up to max.  Perhaps that's what we want.

src/keycloak/chart/values.yaml
```diff
@@ -319,45 +321,27 @@ prometheusRule:
   #   labels:
   #     severity: warning
 
+# High availability options for deployments where pod count is controlled outside HPA
+highAvailability:
```
briantwatson: Instead of `highAvailability.maxReplicas`, could we consider `replicaCount` that is used for replicas?  We could add the logic that @joelmccoy suggests to ensure an odd number.  I think `replicaCount` is fairly standard in the helm landscape and that's what this value is doing rather than an "upper bound". 

If I'm reading correctly, as is a user could technically set `highAvailability.maxReplicas` to `1` and thereby not be highly available. I think if we're keeping `highAvailability.enabled` for user experience,  defaulting to 3, and ensuring both odd and above 3 would be good checks.

I like the way istio uses some conditional checks like right here in their gateway chart:  https://github.com/istio/istio/blob/master/manifests/charts/gateway/templates/deployment.yaml#L13-L17

briantwatson (commented):
Liking the direction we're headed here.  I have a couple comments below, I may be missing some context on decisions made.

src/keycloak/chart/values.yaml
joelmccoy: second this. I had this as a follow up comment but forgot to submit.  `maxReplicas` implies autoscaling to me and I think `replicaCount` is the better value here.

src/keycloak/chart/templates/statefulset.yaml
mjnagel: Yeah I believe intent was backwards compatibility - so that people currently setting `autoscaling.enabled` would enjoy the benefits of the new HA mode (since previously autoscaling was the only way to get HA). That being said - I'm a bit unsure on the usage of `maxReplicas` too - I see the intent, but definitely agree that it might scale too high if someone overrode this to 9 replicas for example.

We may want to consider just tearing off the bandaid here and only maintaining `autoscaling.enabled` backwards compatibility (but not the replica counts)?

src/keycloak/chart/values.yaml
slaskawi: Fixed. Thanks!

src/keycloak/chart/values.schema.json
slaskawi: Great suggestion! Added!

src/keycloak/chart/templates/statefulset.yaml
slaskawi: Yes, we're keeping the `autoscaling` primarily for backwards compatibility - to give our Mission Heros a bit more time to migrate. I created https://github.com/defenseunicorns/uds-core/issues/1909 to track the effort of removing this element entirely.

slaskawi: > We may want to consider just tearing off the bandaid here and only maintaining autoscaling.enabled backwards compatibility (but not the replica counts)?

I was also thinking about this approach and seeing it suggested here, I'll go this way.

src/keycloak/chart/values.yaml
slaskawi: Done!

src/keycloak/chart/values.schema.json
slaskawi: I implemented this in a slightly different place as this part of the schema also applies to the Waypoint (where we don't have these restrictions).

src/keycloak/chart/values.yaml
slaskawi: Great idea! https://github.com/defenseunicorns/uds-core/issues/1909

src/keycloak/chart/values.yaml
slaskawi: I used `siege` for load testing and these were the results:

image

slaskawi: I adjusted the Memory Requests to 80Mi based on these measurements.

src/keycloak/chart/templates/statefulset.yaml
briantwatson: Would it be simpler to just add a `replicaCount` and drop the `highAvailability` altogether?  It's not a pattern I'm familiar with. Do we think there are users of the helm chart who don't equate greater than 1 replica with increased availability? I think we could just add a comment next to the default value if we wanted to further clarify.

Also I'm trying to read up on history of the issue and are we confident the HPA doesn't work in _any_ deployment as expected or is this largely influenced by our experience running Registry?

If we're unsure, I'd be inclined to leave the autoscaling intact, and document the recommendation in the values.  We could use a pattern like the istio chart I linked earlier to set the `replicas` in the statefulset only when `autoscaling.enabled` is false

mjnagel: > Also I'm trying to read up on history of the issue and are we confident the HPA doesn't work in any deployment as expected or is this largely influenced by our experience running Registry?

On this piece I think we were going off of both the registry environment and our nightly CI where we saw scaling issues as well. Although I do find it interesting that the [upstream documents](https://www.keycloak.org/getting-started/getting-started-scaling-and-tuning#_horizontal_autoscaling) the exact HPA setup we included by default 🤔. I wonder if there's something else that we should dig into rather than pure replica numbers and autoscaling.

I do generally agree with plain `replicaCount` although I think it might be hard to adopt in a clean backwards compatible way (since previously HA was locked behind `autoscaling.enabled`). Your suggestion of leaving both options with a recommendation might be a clean approach.

src/keycloak/chart/values.yaml
```diff
@@ -319,45 +321,27 @@ prometheusRule:
...
     requests:
-      cpu: 250m
-      memory: 256Mi
-    # Upstream defaults for limits
+      cpu: 200m
+      memory: 80Mi
     limits:
       cpu: 2
-      memory: 1Gi
+      memory: 100Mi
```
mjnagel: This seems like a pretty drastic drop for the waypoint memory request compared to the upstream default (1Gi). Might be reasonable but I would feel a bit more comfortable if the memory limit was higher since our HPA scales based on CPU usage and OOM will actually kill the waypoint (vs CPU just throttling).

src/keycloak/chart/values.yaml
```diff
@@ -319,45 +321,27 @@ prometheusRule:
   #   labels:
   #     severity: warning
 
+# High availability options for deployments where pod count is controlled outside HPA
+highAvailability:
+  # If `true`, enables HA-recommended settings without using HPA
+  enabled: false
+  # Upper bound for replicas when HA mode is enabled (should be an odd number)
```
joelmccoy: @slaskawi was trying to find a reference to the fact that HA keycloak should run as an odd number of instances and couldn't find anything.  Is there some context/docs around this fact that you could share?

slaskawi: This is less about Keycloak and more about Infinispan - the caching layer used by Keycloak.

You can find the reference materials here:

* https://infinispan.org/docs/stable/titles/tuning/tuning.html#performance-partition-handling_deployment-planning
* https://docs.redhat.com/en/documentation/red_hat_data_grid/7.2/html/administration_and_configuration_guide/handling_network_partitions_split_brain

src/keycloak/chart/values.yaml
slaskawi: Based on UDS Registry performance figures as well as in my local testing, this oscillates in less than 50 Mi. But since we don't have a UDS Core Performance Tests in place, maybe it's better to be a bit more defensive here. I'll set it up to the previous value - 1 Gi.

src/keycloak/chart/templates/statefulset.yaml
slaskawi: In general, I'm not a big fan of top-level settings as they are very limiting in terms of future development. 

Just to illustrate this, let's assume we'd introduce a support for an external Infinispan cluster. These settings could be aggregated within the `highAvailability` as follows:

```
highAvailability:
  enabled: true
  externalInfinispan: true <-- The new setting
```

In this case, all the setting can be easily aggregated into `highAvailability` as they expose settings in the same "theme". Introducing top level settings like `replicaCount` doesn't give us this flexibility and the `values.yaml` quickly become bloated with loosely-coupled settings. 

> Do we think there are users of the helm chart who don't equate greater than 1 replica with increased availability? I think we could just add a comment next to the default value if we wanted to further clarify.

Although it's technically a good fit, I think using `replicaCount` for HA configuration might be confusing to the users as it requires a bit of technical knowledge that you should set it to an odd number to minimize the Split Brain possibility. Even with the odd number validation, I find it too low-level for the users.

To sum it up - I think the `highAvailability` is better for configuration scalability and provides better high-level setting for the user. Unless anyone objects, I would leave it as is.

src/keycloak/chart/values.yaml
slaskawi: Yes, we already did that and there should be a good match!

src/keycloak/chart/values.yaml
slaskawi: Resolving this thread for the sake of cleaning up. The approach has been changed here completely.

src/keycloak/chart/templates/statefulset.yaml
slaskawi: Resolving this thread for the sake of cleaning up. The approach has been changed here completely.

src/keycloak/chart/values.yaml
slaskawi: Resolving this thread for the sake of cleaning up. The approach has been changed here completely.

mjnagel (approved):

joelmccoy (approved):

PR #1903, opened 2025-09-11
feat: add falco as an optional zarf package for runtime-security

### LINKED ISSUE #1709: Build a zarf package for Falco
Once a runtime security tool is chosen, build out a zarf package (following the UDS Core structure) that includes the deployment of Tetragon across all 3 flavors (upstream, registry1, unicorn).

Keep in mind sane default values and configuration for the tooling.

### LINKED ISSUE #1710: UDS integration for Falco
Once https://github.com/defenseunicorns/uds-core/issues/1709 is created, integrate the tool with UDS - to include networking aspects, service mesh, monitoring, etc.

joelmccoy:
## Description

Adds Falco as an optional zarf package to UDS Core in the runtime-security layer.

- Deploys Falco / Falco Sidekick
- Integrates FalcoSidekick to ship Falco events to Loki
- Added docs for runtime-security
- e2e test for security event -> Falco Sidekick

## Related Issue

Fixes #1709

Fixes #1710 

## Type of change

## Steps to Validate
- Can simulate Falco security events locally (easy one is interactive shell into a container)
- View security events in Falco sidekick pod logs
- View security events in loki with the query `{priority=~".+"}`
- View security events and metrics in Falco Grafana dashboards 

## Checklist before merging


packages/runtime-security/zarf.yaml
```diff
@@ -14,11 +14,17 @@ metadata:
...
   - name: neuvector
     required: true
     import:
       path: ../../src/neuvector
+
+  # Falco
+  - name: falco
+    required: false # Optional component until we swap out Neuvector
+    import:
+      path: ../../src/falco
```
mjnagel, line 30: I think optional is the right move here, but I assume this means someone either gets just NeuVector or both for now (no option for just Falco). Probably fine for this overlap/testing period, just want to note it.

joelmccoy: Yeah, in my head the transition period would allow having both tools deployed at once with an opt in for Falco.  I've been deploying and testing these simultaneously with no issues.

mjnagel: One thing we'll want to be conscious of is whether NeuVector will alert on Falco or vice-versa. May want to filter out these alerts for now to not create spam (assuming they detect each other as "privileged"/malicious - that may not happen).

joelmccoy: Good call. I have had clusters running for some time and haven't noticed Falco barking at Neuvector at all.  But some extra thought and analysis would probably be a good idea.

tasks.yaml
```diff
@@ -134,7 +134,7 @@ tasks:
   - name: dev-deploy
     description: "Deploy the given core layer with Zarf Dev"
     actions:
-      - cmd: "uds zarf dev deploy packages/${LAYER} --flavor ${FLAVOR} --no-progress"
+      - cmd: "uds zarf dev deploy packages/${LAYER} --flavor ${FLAVOR} --components '*' --no-progress"
```
joelmccoy, line 98: default to deploying all components for dev-deploys. useful for deving falco as it is an optional component right now

tasks/deploy.yaml
```diff
@@ -71,8 +71,10 @@ tasks:
           # Create temp directory
           mkdir -p tmp/core-shim
 
-          # Copy k3d-standard bundle
-          cp -r bundles/k3d-standard/uds-bundle.yaml tmp/core-shim/
+          LATEST_CORE_VERSION=$(echo ${LATEST_VERSION} | cut -d'-' -f1) # Extract version before any hyphen (e.g., "0.52.0" from "0.52.0-upstream")
+
+          # Download latest release bundle
+          curl -sSL https://raw.githubusercontent.com/defenseunicorns/uds-core/v${LATEST_CORE_VERSION}/bundles/k3d-standard/uds-bundle.yaml -o tmp/core-shim/uds-bundle.yaml
```
joelmccoy, line 84: ran into an issue where the old zarf package does not have falco, but the new bundle definition includes it as an optional component: https://github.com/defenseunicorns/uds-core/actions/runs/17652758981/job/50167749565#step:6:399

I think for our upgrade tests it makes sense to use the bundle definition from the latest release version instead of the one in main for the latest release deploy

test/vitest/falco.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      throw new Error("No falco pod found");
+    }
+
+    await execAndWait(
+      exec,
+      "falco",
+      falcoPod.metadata!.name!,
+      ["find", `/tmp/test-${randomString}`, "-name", "id_rsa"],
+      "falco",
+    );
```
joelmccoy: right now i am execing into the falco pod which works fine (and the falco pod is guaranteed to be up for the falco tests).  but if someone feels strongly I can create another workload to test this?  just wanted to keep it simple

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -0,0 +1,189 @@
...
+
+- Privilege escalation attempts
+- Unauthorized file access
+- Suspicious network activity
+- Container breakout attempts
+
+See more about default rules in the [Falco documentation](https://falco.org/docs/reference/rules/default-rules/).
+
+:::caution[Sandbox/Incubating Rules]
+UDS Core currently does not include the Falco sandbox/incubating ruleset by default. These rules are experimental and may generate false positives.
```
mjnagel: If we plan to include these rules (#1908) should we change the wording here at all? The second part reads like we would discourage using them, but I think we'd plan to incorporate them once we resolve the linked issue?

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -0,0 +1,189 @@
...
+
+1. **Falco Metrics Dashboard** - Shows Falco performance and rule statistics
+2. **Falco Sidekick Metrics Dashboard** - Displays Sidekick forwarding performance
+3. **Falco Sidekick Events Dashboard** - Visualizes security events and logs
+
+These dashboards are automatically available in Grafana when Falco is deployed and can be accessed through the standard UDS Core Grafana interface.
+
+### External Alert Forwarding
+
+While Loki integration provides centralized logging of Falco events, it's recommended to configure external alert forwarding using [Falco Sidekick's native output forwarding](https://github.com/falcosecurity/falcosidekick#outputs) for real-time notifications.  It is generally a good idea to send these alerts to a messaging platform like Slack, Microsoft Teams where these security events can be more visbile to relevant teams.
```
mjnagel, line 49: Do we want to document/encourage direct integration with external destinations (slack, etc)? Or would this be another thing we want to centralize with alertmanager?

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -0,0 +1,189 @@
...
+                # ref: https://uds.defenseunicorns.com/reference/configuration/custom-resources/packages-v1alpha1-cr/#allow
+                # Allow egress to your.remotehost.com on port 443 using TLS
+                - direction: Egress
+                  selector:
+                    app.kubernetes.io/name: falcosidekick
+                    ports:
+                      - 443
+                    remoteHost: your.remotehost.com # set to the hostname where you want to send events
+                    remoteProtocol: TLS
+                    description: "Allow egress Falco Sidekick to your.remotehost.com" # update description as needed
```
mjnagel, line 75: nit: pending the other question, might be nice to just incorporate this into each of the below examples?

bundles/k3d-standard/uds-bundle.yaml
```diff
@@ -43,6 +43,7 @@ packages:
       - istio-passthrough-gateway
       - istio-egress-gateway
       - metrics-server
+      - falco
```
mjnagel, line 46: Should we also include in the IaC bundles to get early feedback on whether it works on EKS/RKE2/AKS? Could also have that as a follow-on PR, but would like confidence in it there before it becomes default.

src/falco/chart/values.yaml
```diff
@@ -0,0 +1,15 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+domain: "###ZARF_VAR_DOMAIN###"
+adminDomain: "###ZARF_VAR_ADMIN_DOMAIN###"
```
mjnagel: nit: don't think domain is used in the chart

src/falco/common/zarf.yaml
```diff
@@ -0,0 +1,26 @@
...
+    charts:
+      - name: uds-falco-config
+        namespace: falco
+        version: 0.1.0
+        localPath: ../chart
+        valuesFiles:
+          - ../chart/values.yaml
+      - name: falco
+        url: https://falcosecurity.github.io/charts
+        version: 6.2.5
```
mjnagel: super cool updates in 6.3.0 - https://github.com/falcosecurity/charts/releases/tag/falco-6.3.0 😄

src/falco/tasks.yaml
```diff
@@ -0,0 +1,30 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+includes:
+  - common-setup: https://raw.githubusercontent.com/defenseunicorns/uds-common/v1.18.0/tasks/setup.yaml
```
mjnagel: don't think this includes is used

src/falco/values/values.yaml
```diff
@@ -0,0 +1,65 @@
...
+  kind: modern-bpf
+
+# work around for: https://github.com/falcosecurity/charts/issues/888
+falco:
+  config_files:
+    - /dummy
+
+falcosidekick:
+  enabled: true
+
```
mjnagel, line 21: The [default replicaCount is 2](https://github.com/falcosecurity/charts/blob/master/charts/falcosidekick/values.yaml#L6) - is that something we might want to change? Either here or in the demo bundle? Might also make sense to document that in the [resource config/HA doc](https://uds.defenseunicorns.com/reference/configuration/resource-configuration-and-ha/).

src/falco/zarf.yaml
```diff
@@ -0,0 +1,62 @@
...
+  description: "UDS Core Falco"
+  url: https://falco.org/docs/
+
+variables:
+  - name: DOMAIN
+    description: "Cluster domain"
+    default: "uds.dev"
+
+  - name: ADMIN_DOMAIN
+    description: "Domain for admin services, defaults to `admin.DOMAIN`"
```
mjnagel: nit: don't think these are used

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -0,0 +1,189 @@
...
+{rule="Search Private Keys or Passwords"}
+```
+
+### Grafana Dashboards
+
+The upstream Falco helm charts include some Grafana dashboards out of the box for visualizing security events:
+
+1. **Falco Metrics Dashboard** - Shows Falco performance and rule statistics
+2. **Falco Sidekick Metrics Dashboard** - Displays Sidekick forwarding performance
+3. **Falco Sidekick Events Dashboard** - Visualizes security events and logs
```
mjnagel: nit: the names of these dashboards (from what I deployed at least) seem to be `Falco`, `Falcosidekick`, and `Falco logs`. Might be good to align the references here with those names?

src/falco/values/values.yaml
```diff
@@ -0,0 +1,65 @@
...
+      hostport: "http://loki-gateway.loki.svc.cluster.local:80"
+      format: json
+      # Enables the Grafana dashboard for visualizing Falco events
+      grafanaDashboard:
+        enabled: true
+
+# Enable Grafana dashboard for Falco metrics
+grafana:
+  dashboards:
+    enabled: true
```
mjnagel: In testing it seems like these dashboards may have some issues, or at least strange behavior that might need explanation? Worth considering if we need to disable them for now and contribute upstream?

mjnagel (commented):
Let's also make sure we add a new renovate grouping for Falco - similar to [NeuVector's](https://github.com/defenseunicorns/uds-core/blob/main/renovate.json#L41-L45).

docs/reference/configuration/runtime-security/alerting.md
joelmccoy: good call. addressed the wording here.  we can update this later when we add those rules in.

docs/reference/configuration/runtime-security/alerting.md
joelmccoy: as discussed in person, it's probably a good idea to encourage sending these events directly to your end notification destination.  these events aren't events that are firing/not firing. more of one off events. and sending them through alertmanager feels unnecessary

docs/reference/configuration/runtime-security/alerting.md
joelmccoy: i added the snippet to each of the below examples

bundles/k3d-standard/uds-bundle.yaml
joelmccoy: good call. added

src/falco/chart/values.yaml
joelmccoy: removed

src/falco/common/zarf.yaml
joelmccoy: updated 😎

src/falco/tasks.yaml
joelmccoy: removed

src/falco/values/values.yaml
joelmccoy: hmm. sidekick just does stateless event routing.  i'm almost partial to defaulting to HA (2 replicas). the only downside is the extra resource consumption of 2 replicas, but the resource consumption is so minimal that i don't think it is a big deal.

src/falco/zarf.yaml
joelmccoy: removed

docs/reference/configuration/runtime-security/alerting.md
joelmccoy: good call. i removed the other dashboards and just called out `Falco logs` here

src/falco/values/values.yaml
joelmccoy: removing for now by commenting out.  i played around with these and I am a bit confused why the dashboard queries are breaking down.  will need some further investigation

src/falco/values/values.yaml
mjnagel: I think that's fair. Would we want to:
- reduce to 1 replica for the demo bundle (maybe not necessary but we like to keep that demo bundle small-ish)
- document replica config in the resource config/HA doc (assuming it makes sense to scale it sometimes)

test/vitest/falco.spec.ts
joelmccoy: i changed this to use the nginx pod in the uds-dev-stack namespace.  falco unicorn images don't let you exec in (which is probably a good thing). 

i think we can assume the nginx pod is always going to be in the uds-dev-stack namespace.  I am trying to avoid adding a manifest to create a pod just for this test, but if someone feels this is necessary I'd be happy to add in.

src/falco/values/values.yaml
joelmccoy: set the bundles to 1 and added a section in the HA docs.

joelmccoy:
blocked until we get some registry1 fixes: https://github.com/defenseunicorns/uds-core/issues/1940

test/vitest/falco.spec.ts
joelmccoy: jk. the nginx pod will break down on our IAC tests.  I opted to just create a pod in the kube-system namespace to avoid zarf/pepr mutations

docs/reference/configuration/runtime-security/overview.md
```diff
@@ -0,0 +1,38 @@
+---
+title: Runtime Security Overview
+sidebar:
+  order: 1
```
chance-coleman: i may have missed a conversation about the ordering of these docs. 

At the moment they are like the screenshot below. This is a really confusing to explain but its a limitation of astro.js. The runtime security `Overview` doc needs the sidebar ordering but the `Alerting` doc does not, this is also the pattern we've used elsewhere, otherwise every doc will need to be specifically numbered. We order only one file in each configuration subfolder (`runtime-security`, `single-sign-on`, `uds-operator`, etc.), typically the overview.md. The only reason we order it is so that the subfolder itself is rendered on the doc site in a specific order.

So long story short, if you set the runtime-security/overview.md to `order: 7` it will show up below `external-dependencies` on the sidebar (easiest solution). Or you could put it directly after `Single-Sign-On` which means you set `order: 5` and then increment `Service-Mesh`, `Backup-Restore`, and `External-Dependencies`.

You can use the `uds run dev-docs` to play around with this locally.

Screenshot 2025-09-24 at 07 52 20

joelmccoy: good callout. i just got pointed out this weird behavior in another PR as well 😂 .  adjusted runtime security to `order: 7` so that it appears after external-dependencies: 
image

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -0,0 +1,217 @@
+---
+title: Runtime Security Alerting
+tableOfContents:
+  maxHeadingLevel: 4
```
mjnagel, line 4:
```suggestion
```

Per other thread.

mjnagel (commented):
Doing a local deploy/test but this is looking good at this point. One follow-on we may want to track is updating the docs site references from NeuVector -> Falco (or for now adding Falco there). [Quick search](https://github.com/search?q=repo%3Adefenseunicorns%2Fuds-docs%20neuvector&type=code) shows a few spots to track.

packages/runtime-security/zarf.yaml
mjnagel, line 30: I think we'll also want to import the falco validate/e2e into [this task file](https://github.com/defenseunicorns/uds-core/blob/1709_falco-zarf-package/packages/runtime-security/tasks.yaml) to make sure they run on single layer tests.

mjnagel, line 30: Also the readme: https://github.com/defenseunicorns/uds-core/tree/1709_falco-zarf-package/packages/runtime-security

We may need to list the logging layer as a dependency for falco? If not deployed Falco will spam errors hitting Loki.

joelmccoy: hmm. the logging layer as a dependency is a good question... I would call it a soft dependency and I am hesitant to list it in the docs/readme as actual dependency.  Things will function without the logging layer.  You will just get some sidekick log errors.  And you could always disable the loki integration with overrides.  I don't want to mislead people in that in order to deploy runtime-security you need the logging layer.

Is it acceptable to not document it as a dependency based on that?

mjnagel: I would probably list it as a soft dependency with a qualifying statement similar to https://github.com/defenseunicorns/uds-core/tree/main/packages/logging?

Something like:
- logging (optional for Falco to send logs to Loki)

docs/reference/configuration/runtime-security/alerting.md
joelmccoy: am i missing something here? when i remove this I lose a couple headings here
image
vs
image

mjnagel: 🤦 sorry - thought this was the page ordering key. This is fine 😅

packages/runtime-security/zarf.yaml
joelmccoy: updated

joelmccoy: updated

mjnagel (dismissed):

joelmccoy:
> Doing a local deploy/test but this is looking good at this point. One follow-on we may want to track is updating the docs site references from NeuVector -> Falco (or for now adding Falco there). [Quick search](https://github.com/search?q=repo%3Adefenseunicorns%2Fuds-docs%20neuvector&type=code) shows a few spots to track.

Draft PR: https://github.com/defenseunicorns/uds-docs/pull/298

chance-coleman (dismissed):

chance-coleman (approved):

mjnagel (approved):

PR #1912, opened 2025-09-15
feat: add support for extra gateways in expose spec

### LINKED ISSUE #1170: Allow `expose` section to use gateways other than admin, tenant, passthrough

Currently the UDS Package CR only allows the selection of admin, tenant and passthrough for gateway choice. As the variety of UDS packages expand, and additional gateways are added by developers, it would be nice to be able to have Expose create the virtual service for custom gateways instead of having to provide both.

Given the expose key with `gateway: <custom>` this should be supported.

mjnagel:
## Description

Adds support for using custom / extra gateways in the `expose` spec, follows the approved design.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1170

## Type of change

## Steps to Validate

A full e2e test is a bit annoying to setup with uds-k3d, but there are unit tests included for the new/changed operator logic. In addition you can perform some basic tests following the docs:
- Follow the docs to create and deploy a custom gateway
- Follow the docs to create and deploy a package CR using that custom gateway: You may also want to play with the behavior when you do/don't include domain and when the gateway name does/doesn't include `admin` or `passthrough`
- Validate full routing through the custom gateway: this is where there's a bit more to setup - you'd need to ensure that nginx from the uds-dev-stack routes to your new gateway

## Checklist before merging


docs/reference/configuration/service-mesh/custom-gateways.md
```diff
@@ -0,0 +1,94 @@
...
+        version: 1.26.2 # This should match the Istio version currently in UDS Core
+        releaseName: custom-ingressgateway
+        namespace: istio-custom-gateway
+      - name: uds-istio-config
+        version: v0.52.0 # This should match the version of UDS Core you are deploying
+        url: https://github.com/defenseunicorns/uds-core.git
+        gitPath: src/istio/charts/uds-istio-config
+        namespace: istio-custom-gateway
+        valuesFiles:
+          - "config-custom.yaml"
```
joelmccoy, line 53: Love this pattern of referencing our chart. they would get the benefits of any updates we include in that chart.  And we can ensure consistency in configuration if they use our upstream chart.

docs/reference/configuration/service-mesh/custom-gateways.md
```diff
@@ -0,0 +1,94 @@
...
+      - service: foobar
+        selector:
+          app.kubernetes.io/name: foobar
+        gateway: custom
+        domain: mydomain.dev
+        host: foobar
+        port: 8080
+```
+
+Make sure that the `gateway` name lines up with the name you used in your gateway deployment (`custom` in our example). For `domain`, you will want to set this if exposing your service on a domain separate
```
joelmccoy:
```suggestion
Make sure that the `gateway` name lines up with the name you used in your gateway deployment (`custom` in our example). For `domain`, you will want to set this if exposing your service on a domain different from the default domain name for your environment.
```

src/pepr/operator/controllers/istio/virtual-service.ts
```diff
@@ -38,8 +38,13 @@ export function generateIngressVirtualService(
 
   const name = generateVSName(pkgName, expose);
 
-  // Get the correct domain based on gateway
-  const domain = gateway === Gateway.Admin ? UDSConfig.adminDomain : UDSConfig.domain;
+  // Get the correct domain based on gateway or custom domain
+  let domain = UDSConfig.domain;
+  if (expose.domain) {
+    domain = expose.domain;
+  } else if (gateway === Gateway.Admin || gateway.includes("admin")) {
```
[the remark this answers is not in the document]
joelmccoy: bad bot: this is not the spec described

src/pepr/operator/controllers/istio/virtual-service.ts
```diff
@@ -89,8 +94,8 @@ export function generateIngressVirtualService(
     },
   };
 
-  // If the gateway is the passthrough gateway, apply the TLS match
-  if (gateway === Gateway.Passthrough) {
+  // If the gateway is the passthrough gateway or includes passthrough in the name, apply the TLS match
+  if (gateway === Gateway.Passthrough || gateway.includes("passthrough")) {
```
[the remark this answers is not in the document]
joelmccoy: bad bot: this is not the spec described

docs/reference/configuration/service-mesh/custom-gateways.md
```diff
@@ -0,0 +1,94 @@
...
+Then for your values file (`config-custom.yaml` above) you will want to need to setup your configuration. Reference the [default values file](https://github.com/defenseunicorns/uds-core/blob/main/src/istio/charts/uds-istio-config/values.yaml) for full configuration options, but you will need to at minimum provide:
+```yaml
+name: custom
+
+domain: mydomain.dev # Set domain if different from default tenant domain for this gateway
+
+tls:
+  servers:
+    custom:
+      mode: # One of `SIMPLE`, `MUTUAL`, 'OPTIONAL_MUTUAL', `PASSTHROUGH`
```
joelmccoy, line 66: should probably call out setting the TLS certs here. I think these are required values they would need to set for this to work right?
```suggestion
tls:
  # The TLS certificate for the gateway, if not in 'PASSTHROUGH' mode (base64 encoded)
  cert: ""
  # The TLS key for the gateway, if not in 'PASSTHROUGH' mode (base64 encoded)
  key: ""
  # The CA certificate for the gateway when using `MUTUAL' or 'OPTIONAL_MUTUAL' (base64 encoded)
  cacert: ""
 # The name of the secret containing the TLS certificate to use for this gateway, this will override cert, key and cacert
  credentialName: ""
  servers:
    custom:
      mode: # One of `SIMPLE`, `MUTUAL`, 'OPTIONAL_MUTUAL', `PASSTHROUGH`
```

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -39,15 +39,35 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+        );
+      }
+    }
+    if (expose.gateway && isStandardGateway(expose.gateway) && expose.domain) {
+      return req.Deny(
+        "domain cannot be set for the standard gateways (tenant, admin, or passthrough)",
+      );
+    }
+
+    if (expose.gateway === Gateway.Passthrough || expose.gateway?.includes("passthrough")) {
```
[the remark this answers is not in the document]
joelmccoy: bad bot. this is not the spec

docs/reference/configuration/service-mesh/custom-gateways.md
```diff
@@ -0,0 +1,94 @@
...
+  description: "Custom gateway for UDS Core"
+components:
+  - name: istio-custom-gateway
+    required: true
+    charts:
+      # This is the Ingress Gateway from upstream Istio
+      - name: gateway
+        url: https://istio-release.storage.googleapis.com/charts
+        version: 1.26.2 # This should match the Istio version currently in UDS Core
+        releaseName: custom-ingressgateway
```
joelmccoy: forgot to swap this when testing locally.  would be helpful to have a comment early on
```suggestion
        releaseName: custom-ingressgateway # This should be <gateway-name>-ingressgateway
```

docs/reference/configuration/service-mesh/custom-gateways.md
```diff
@@ -0,0 +1,94 @@
...
+        namespace: istio-custom-gateway
```
joelmccoy:
```suggestion
        namespace: istio-custom-gateway # This should be istio-<gateway-name>-gateway
```

docs/reference/configuration/service-mesh/custom-gateways.md
```diff
@@ -0,0 +1,94 @@
...
+      - name: uds-istio-config
+        version: v0.52.0 # This should match the version of UDS Core you are deploying
+        url: https://github.com/defenseunicorns/uds-core.git
+        gitPath: src/istio/charts/uds-istio-config
+        namespace: istio-custom-gateway
```
joelmccoy:
```suggestion
        namespace: istio-custom-gateway # This should be istio-<gateway-name>-gateway
```

src/pepr/operator/controllers/istio/virtual-service.ts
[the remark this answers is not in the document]
mjnagel: Just to respond to AI - this is intentional. For the admin side it's mostly a convenience to pre-select the right domain. I would expect that someone might have `custom-admin`, `public-admin`, etc. We can't guarantee that it's at the start or end, so includes is the easiest here.

src/pepr/operator/controllers/istio/virtual-service.ts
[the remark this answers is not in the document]
mjnagel: Just to respond to AI - this is intentional. For the passthrough side it's necessary to setup the VS correctly. I would expect that someone might have `custom-passthrough`, `public-passthrough`, etc. We can't guarantee that it's at the start or end, so includes is the easiest here.

joelmccoy (commented):
Generally LGTM! Tested the docs locally with my own custom gateway and domain with self signed certs.  A couple comments/suggestions for improvement

docs/reference/configuration/service-mesh/custom-gateways.md
```diff
@@ -0,0 +1,94 @@
...
+        valuesFiles:
+          - "config-custom.yaml"
+```
+
+Then for your values file (`config-custom.yaml` above) you will want to need to setup your configuration. Reference the [default values file](https://github.com/defenseunicorns/uds-core/blob/main/src/istio/charts/uds-istio-config/values.yaml) for full configuration options, but you will need to at minimum provide:
+```yaml
+name: custom
```
mjnagel:
```suggestion
name: custom # This should be i<gateway-name>
```

docs/reference/configuration/service-mesh/custom-gateways.md
mjnagel: Thanks - somehow stopped typing here before I

docs/reference/configuration/service-mesh/custom-gateways.md
```diff
@@ -0,0 +1,94 @@
...
+        url: https://github.com/defenseunicorns/uds-core.git
+        gitPath: src/istio/charts/uds-istio-config
+        namespace: istio-custom-gateway # This should be istio-<gateway-name>-gateway
+        valuesFiles:
+          - "config-custom.yaml"
+```
+
+Then for your values file (`config-custom.yaml` above) you will want to need to setup your configuration. Reference the [default values file](https://github.com/defenseunicorns/uds-core/blob/main/src/istio/charts/uds-istio-config/values.yaml) for full configuration options, but you will need to at minimum provide:
+```yaml
+name: custom # This should be i<gateway-name>
```
mjnagel:
```suggestion
name: custom # This should be <gateway-name>
```

docs/reference/configuration/service-mesh/custom-gateways.md
mjnagel: Yeah - initially left this off because PASSTHROUGH doesn't need certs but I like your example. It's also a bit weird because normally people wouldn't set these in the values file (they'd do it via bundle overrides later), but the same could be said for domain I guess.

mjnagel: Applied suggestion - but I can also/instead link a reference here to the domain and TLS config from the other page (https://uds.defenseunicorns.com/reference/configuration/service-mesh/ingress/#configure-domain-name-and-tls-for-istio-gateways)

joelmccoy: Ah yeah you are totally right. These should really be passed in as variables (as the key at least is a secret).  You can undo my suggestion and either link to to docs or maybe include just a small snippet for the overrides block for this example below. I think it might be helpful just to have a easy reference for the overrides here.  I personally always gets confused when doing these 😂

mjnagel: Updated - let me know if that hits the mark.

joelmccoy: excellent. love the rework.

joelmccoy (approved):

PR #1917, opened 2025-09-15
fix!: improve Keycloak stability

slaskawi:
## Description

This Pull Request contains essential stability improvements that were split out from https://github.com/defenseunicorns/uds-core/pull/1898:

* Aligned Kubernetes Probes with the upstream
* Removed `podManagementPolicy` as anything other than `OrderedReady` may lead to issues:
  * Scaling out rapidly will clash on database schema update transactions (only the first one will succeed)
  * Scaling in rapidly may cause data loss in Infinispan cluster (killed both primary and secondary owners of the data segment)

## Related Issue

Relates to https://github.com/defenseunicorns/uds-core/issues/1748

## Type of change

## Steps to Validate

Tested automatically

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
fix!: improve Keycloak stability (https://github.com/defenseunicorns/uds-core/pull/1917)

BREAKING CHANGE: Keycloak `podManagementPolicy` Helm Chart value has been removed as setting it to anything other than `OrderedReady` can cause clustering outages. During the upgrade procedure, if the previously used value was `Parallel`, Keycloak will temporarily scale in to 1 replica. Once the StatefulSet gets updated, it will scale out to previously set value. 

Release-As: 0.52.1
END_COMMIT_OVERRIDE

### DIFF
--- a/src/keycloak/common/zarf.yaml
+++ b/src/keycloak/common/zarf.yaml
@@ -27,7 +27,7 @@ components:
               set +e
 
               echo "Checking if Keycloak StatefulSet should be deleted"
-              KEYCLOAK_INSTALLED=$(uds zarf tools kubectl get sts -n keycloak keycloak > /dev/null 2>&1; echo $?)
+              KEYCLOAK_INSTALLED=$(./zarf tools kubectl get sts -n keycloak keycloak > /dev/null 2>&1; echo $?)
               if [ "${KEYCLOAK_INSTALLED}" -eq 0 ]; then
                 echo "Keycloak has been previously installed. Checking the podManagementPolicy..."
                 POD_MGMT_POLICY=$(uds zarf tools kubectl get sts -n keycloak keycloak -o jsonpath='{.spec.podManagementPolicy}' 2>/dev/null || true)
--- a/src/keycloak/common/zarf.yaml
+++ b/src/keycloak/common/zarf.yaml
@@ -34,7 +34,7 @@ components:
                 echo "Detected podManagementPolicy: ${POD_MGMT_POLICY:-<none>}"
                 if [ -n "${POD_MGMT_POLICY}" ] && [ "${POD_MGMT_POLICY}" != "OrderedReady" ]; then
                   echo "podManagementPolicy is '${POD_MGMT_POLICY}', deleting StatefulSet (cascade=orphan) to allow re-create with OrderedReady"
-                  uds zarf tools kubectl delete sts -n keycloak keycloak --cascade=orphan
+                 ./zarf tools kubectl delete sts -n keycloak keycloak --cascade=orphan
                 else
                   echo "podManagementPolicy is correct. Skipping..."
                 fi
--- a/src/keycloak/common/zarf.yaml
+++ b/src/keycloak/common/zarf.yaml
@@ -19,6 +19,28 @@ components:
           - ../chart/values.yaml
     actions:
       onDeploy:
+        before:
+          - description: Re-create Keycloak StatefulSet
+            # This can be switched to true in subsequent UDS Core versions
+            mute: false
+            cmd: |
+              set +e
+
+              echo "Checking if Keycloak StatefulSet should be deleted"
+              KEYCLOAK_INSTALLED=$(uds zarf tools kubectl get sts -n keycloak keycloak > /dev/null 2>&1; echo $?)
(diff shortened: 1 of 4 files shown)


mjnagel (dismissed):

joelmccoy (dismissed):

src/keycloak/common/zarf.yaml
```diff
@@ -19,6 +19,28 @@ components:
...
       onDeploy:
+        before:
+          - description: Re-create Keycloak StatefulSet
+            # This can be switched to true in subsequent UDS Core versions
+            mute: false
+            cmd: |
+              set +e
+
+              echo "Checking if Keycloak StatefulSet should be deleted"
+              KEYCLOAK_INSTALLED=$(uds zarf tools kubectl get sts -n keycloak keycloak > /dev/null 2>&1; echo $?)
```
mjnagel: These commands should all be `./zarf tools`... - that syntax will ensure that this works regardless of where zarf/uds zarf is on a user's machine (even if not on the path) - https://docs.zarf.dev/ref/actions/#action-transformations

mjnagel:
```suggestion
              KEYCLOAK_INSTALLED=$(./zarf tools kubectl get sts -n keycloak keycloak > /dev/null 2>&1; echo $?)
```

src/keycloak/common/zarf.yaml
```diff
@@ -19,6 +19,28 @@ components:
...
+              if [ "${KEYCLOAK_INSTALLED}" -eq 0 ]; then
+                echo "Keycloak has been previously installed. Checking the podManagementPolicy..."
+                POD_MGMT_POLICY=$(uds zarf tools kubectl get sts -n keycloak keycloak -o jsonpath='{.spec.podManagementPolicy}' 2>/dev/null || true)
```
mjnagel:
```suggestion
                POD_MGMT_POLICY=$(./zarf tools kubectl get sts -n keycloak keycloak -o jsonpath='{.spec.podManagementPolicy}' 2>/dev/null || true)
```

src/keycloak/common/zarf.yaml
```diff
@@ -19,6 +19,28 @@ components:
...
+                echo "Detected podManagementPolicy: ${POD_MGMT_POLICY:-<none>}"
+                if [ -n "${POD_MGMT_POLICY}" ] && [ "${POD_MGMT_POLICY}" != "OrderedReady" ]; then
+                  echo "podManagementPolicy is '${POD_MGMT_POLICY}', deleting StatefulSet (cascade=orphan) to allow re-create with OrderedReady"
+                  uds zarf tools kubectl delete sts -n keycloak keycloak --cascade=orphan
```
mjnagel:
```suggestion
                 ./zarf tools kubectl delete sts -n keycloak keycloak --cascade=orphan
```

mjnagel (commented):
LGTM overall, small nit to make sure this is portable.

briantwatson:
Nice changes! I hadn't caught we didn't have a startupProbe.  I'm wondering if this in itself would resolve the availability issues we're seeing.

joelmccoy (commented):
Generally LGTM pending the resolved comments on the zarf actions.

Wanted to call out one thing i noticed when testing the upgrade locally.  When you delete the statefulset and recreate it, only 1 keycloak pod stays up.  When you create the new statefulset it kills the other two pods and adds them back one by one.  So there might be a couple mins where keycloak is not running HA, but I think that is ok.

joelmccoy:
> Wanted to call out one thing i noticed when testing the upgrade locally. When you delete the statefulset and recreate it, only 1 keycloak pod stays up. When you create the new statefulset it kills the other two pods and adds them back one by one. So there might be a couple mins where keycloak is not running HA, but I think that is ok.

After some digging I think this behavior is due to use having the Statefulset replicas still being managed by the HPA.  The statefulset defaults to 1 replica and when recreated it scales down to 1 replica until the HPA scales it back up.

Unless we remove the HPA in this PR, I'm not sure there is a way to avoid this.  And I think for most environments a scale down to 1 replica for about a minute or so is generally ok

src/keycloak/common/zarf.yaml
slaskawi: Today I learned! Great trick!

mjnagel (approved):

joelmccoy (approved):

PR #1925, opened 2025-09-17
feat: x509 Authentication Flow control

slaskawi:
## Description

This Pull Request introduces settings required for controlling the x509 Authentication Sub Flow (see the technical explanation [here](https://github.com/defenseunicorns/uds-identity-config/issues/626#issuecomment-3301504023))

## Related Issue

Relates to https://github.com/defenseunicorns/uds-identity-config/issues/626

## Type of change

## Steps to Validate

Validated in UDS Identity Config

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/secret-kc-realm.yaml
+++ b/src/keycloak/chart/templates/secret-kc-realm.yaml
@@ -25,6 +25,7 @@ data:
   DENY_USERNAME_PASSWORD_ENABLED: {{ ternary "DISABLED" "REQUIRED" (.Values.realmAuthFlows.USERNAME_PASSWORD_AUTH_ENABLED) | b64enc }}
   RESET_CREDENTIAL_FLOW_ENABLED: {{ ternary "REQUIRED" "DISABLED" (.Values.realmAuthFlows.USERNAME_PASSWORD_AUTH_ENABLED) | b64enc }}
   REGISTRATION_FORM_ENABLED: {{ ternary "REQUIRED" "DISABLED" (or .Values.realmAuthFlows.USERNAME_PASSWORD_AUTH_ENABLED .Values.realmAuthFlows.X509_AUTH_ENABLED) | b64enc }}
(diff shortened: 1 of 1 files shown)


src/keycloak/chart/templates/secret-kc-realm.yaml
```diff
@@ -25,6 +25,7 @@ data:
   DENY_USERNAME_PASSWORD_ENABLED: {{ ternary "DISABLED" "REQUIRED" (.Values.realmAuthFlows.USERNAME_PASSWORD_AUTH_ENABLED) | b64enc }}
   RESET_CREDENTIAL_FLOW_ENABLED: {{ ternary "REQUIRED" "DISABLED" (.Values.realmAuthFlows.USERNAME_PASSWORD_AUTH_ENABLED) | b64enc }}
   REGISTRATION_FORM_ENABLED: {{ ternary "REQUIRED" "DISABLED" (or .Values.realmAuthFlows.USERNAME_PASSWORD_AUTH_ENABLED .Values.realmAuthFlows.X509_AUTH_ENABLED) | b64enc }}
+  X509_LOGIN_FLOW_ENABLED: {{ ternary "ALTERNATIVE" "DISABLED" (.Values.realmAuthFlows.X509_AUTH_ENABLED) | b64enc }}
```
mjnagel, line 28: Is there a reason this is different from `X509_LOGIN_ENABLED` in the secret? Could we just use that same one? They're both tied to the same helm value (`realmAuthFlows.X509_AUTH_ENABLED`) so I'd expect we could just us that one?

slaskawi: Yes, that's correct.

The reason we have two is that in the `X509_LOGIN_ENABLED` has a simple value `true` or `false`. It is used in theme to show/hide certain things. The `X509_LOGIN_FLOW_ENABLED` is used in the Authentication Flow and in our case has two values - `ALTERNATIVE` or `DISABLED`. 

The value needs to be put into Keycloak Pod environment variables to be effective and here's probably the most convenient place.

mjnagel: 🤦 I overlooked that this does the ternary - duh. Makes sense and LGTM.

mjnagel (approved):

Issue #1878, opened 2025-09-02
Establish/Document a UDS Core Observability Reference Architecture


joelmccoy:
### Is your feature request related to a problem? Please describe.
UDS Core lacks a single, documented “golden path” for observability across metrics, logs, dashboards, and alert routing.

Today:
- Metrics → Prometheus/Alertmanager alerts can be declared via Prometheus CRDs or written as Grafana alerts
- Logs → Loki alerting has multiple competing approaches (Loki ruler, Grafana alerts via configmaps/overrides/operator, manual UI) (see https://github.com/defenseunicorns/uds-core/issues/1845)

It’s unclear where users should declare rules, how those rules should be loaded, where the end user should view those alerts, how alerts should be routed externally, and which Kubernetes resources/CRDs are the source of truth.

This ambiguity makes it harder to optimize UDS Core around a consistent pipeline and increases the learning curve for end users.

### Describe the solution you'd like

- **Given** a UDS Core deployment
- **When** a user needs to: (a) create dashboards, (b) create metrics alerts, (c) create log alerts, (d) route alerts to external systems
- **Then** they follow the documented reference architecture for defining/managing these things

### Describe alternatives you've considered
- Leave it up to the end user on how they want to configure the observability stack

### Additional context
N/A

joelmccoy:
I think a diagram of the recommend flow for the observability stack would be helpful.  Here is what I would propose for the overall opinions:

Metrics alerts
- Declare with PrometheusRule (CRD, monitoring.coreos.com).
- Evaluate in Prometheus.

Logs alerts
- Declare Loki alerting rules as files delivered via ConfigMaps (labeled and reloaded by a Loki ruler sidecar).
- Evaluate in Loki Ruler.
- Route to Alertmanager (unified alert pipeline).

Dashboards
- Declare as ConfigMaps (labeled for Grafana config reloader) as the default GitOps path.

External alert routing
- Prefer Alertmanager receivers (Slack, PagerDuty, Email, Webhook) configured via AlertmanagerConfig in-cluster via helm overrides.

Governance & tenancy
- Namespace-scoped ownership of rules/dashboards.

joelmccoy:
Some good questions to help provide answers/documentations for from @ntwkninja around alerting:

> Will only critical alerts be configured by default? Do I have a way to opt into some things that are available / recommend but may not be critical? Are alert categories defined?
> 
[...]

PR #1930, opened 2025-09-18
feat!: enable loki ruler alerting/recording

### LINKED ISSUE #1845: Feature Request: Loki-based Alerting Support in UDS Core
UDS Core provides a clear workflow for alerting on metrics through Prometheus and Alertmanager. Users can declaratively configure rules via CRDs, which is consistent, extensible, and easy to adopt.  

However, there is currently no comparable story for alerting on logs with Loki. Logs are ingested into Grafana for visualization, but users do not have a straightforward, documented path to configure and manage alerts on log data. This creates an inconsistency between metrics-based and logs-based alerting.  

There are many ways to approach this and we should agree on the best approach for UDS Core.

---

- **Given** that users are ingesting logs into Loki with UDS Core,  
- **When** a user wants to define alerting conditions on those logs,  
- **Then** they should have a documented and supported way to configure Loki-based alerts in UDS Core (ideally declarative, and consistent with how metrics-based alerts are handled).  

---

### Possible solutions 
- **Loki Ruler Sidecar**: Use loki ruler to evaluate rules and enable the [sidecar](https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml#L3023) to pull in rules across the cluster via configmaps.  Additionally loki ruler will need to be setup to send alerts to alertmanager.
- **Grafana Config Reloader**: With grafana config reloader (which UDS Core has enabled), you can define alerts via configmaps and Grafana will load these as alerts dynamically in grafana (requires no additional tooling, and PBME does something like this today)
- **Grafana Alerts Helm Value Overrides**: Setup Grafana Alerts to send to Alertmanager via a contact point.  Document for users to extend the Grafana Helm chart with overrides to create Grafana Alert rules (example: [Grafana values rules section](https://github.com/grafana/helm-charts/blob/6acd5544482b1f417cd3e5d344516b97a2300307/charts/grafana/values.yaml#L694-L755)) (Requires no new tooling, but downside is that all alerts must be managed as overrides to UDS Core)
- **Grafana Operator**: Introduce the Grafana Operator to manage Grafana alerts declaratively via CRDs.  (Requires additional tool, but allows for declarative CRDS to manage alerts)

---

- **Manual Grafana Alerts Configuration**: Users can create alert rules and configure Alertmanager as a contact point via the Grafana UI. However, this is manual, not declarative, and does not fit well with GitOps workflows.  
---

### Additional context  
Prometheus CRDs already provide a great model for metrics alerts. A similar workflow or a well-documented pipeline for Loki-based alerting would make UDS Core more consistent and significantly improve the user experience.

slaskawi:
## Description

Enables Loki ruler alerting and recording with configmap/secret sidecar loading of rules.

- Introduce sidecar reloader to loki backend to load rule files from configmaps/secrets
- Force Loki ruler storage backend to local to load rules from configmaps/secrets (breaking change for people using object storage for loki ruler)
- Added docs for log based alerting

## Related Issue

Fixes #1845

## Type of change

## Steps to Validate
- Deploy k3d-standard bundle
- Deploy this configmap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: loki-test-rules
  namespace: loki
  labels:
    loki_rule: "1"
data:
  test-rules.yaml: |
    groups:
    - name: always-firing
      rules:
      - alert: LokiAlwaysFiring
        expr: vector(1)
        for: 0s
        labels:
          severity: test
        annotations:
          summary: "This is a test alert from Loki ruler"
          description: "It always fires to confirm Loki → Alertmanager wiring."
    - name: test-recording
      interval: 30s    # how often to evaluate this rule
      rules:
      - record: loki:test_constant
        expr: vector(1)
```
- Port forward to alertmanager UI or use Grafana Alerts UI to see if `LokiAlwaysFiring` is firing (make take a few minutes to trigger)
- Query prometheus metric `loki:test_constant` in Grafana and ensure the value is 1 (make take a few minutes to sync)
- View the recording and alerting rules under `Alert Rules` in Grafana (select Loki datasource)

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: enable loki ruler alerting/recording (#1930)

BREAKING CHANGE: Loki Ruler is now configured to use local storage backend for rules files by default to support loading rules as ConfigMaps.  If you created Loki Ruler rules with the Loki API with an object storage backend, those rules will be disabled and will need to be applied as ConfigMaps.
END_COMMIT_OVERRIDE


src/loki/values/values.yaml
```diff
@@ -26,7 +26,6 @@ loki:
   storage:
     bucketNames:
       chunks: uds
-      ruler: uds
```
joelmccoy, line 29: note: we are disabling s3 storage for the ruler with this PR. the ruler bucket literally only stores the rules files and no other data.  since we are using configmaps now this bucket is not necessary anymore.  This technically would be a breaking change but I doubt anybody has uploaded rules files to their s3 bucket for loki to then ingest them

docs/reference/configuration/observability/monitoring-metrics.md
joelmccoy, line 1: I am moving this into the Observability folder in order to group and prep for other docs introduced in https://github.com/defenseunicorns/uds-core/issues/1878

src/loki/values/values.yaml
```diff
@@ -134,23 +155,37 @@ minio:
...
     requests:
-      cpu: 100m
-      memory: 100Mi
-  securityContext:
-    allowPrivilegeEscalation: false
-    capabilities:
-      drop:
-        - ALL
-    seccompProfile:
-      type: RuntimeDefault
```
joelmccoy, line 150: setting resource requests to recommended upstream and removing securitycontext as this is already default upstreem and pepr will set this anyways.  not adding value having it in values here IMO

ref: https://github.com/grafana/loki/blob/v3.5.5/production/helm/loki/values.yaml#L3530-L3543

src/loki/values/values.yaml
```diff
@@ -134,23 +155,37 @@ minio:
...
   rules:
     # -- Whether or not to create a sidecar to ingest rule from specific ConfigMaps and/or Secrets.
-    enabled: false
+    enabled: true
+    # -- Label that the configmaps/secrets with rules will be marked with.
+    label: loki_rule
+    # -- Label value that the configmaps/secrets with rules will be set to.
+    labelValue: ""
+    # -- Folder into which the rules will be placed.
+    folder: /rules/fake
```
slaskawi: Probably it's worth to use a more descriptive directory

docs/reference/configuration/observability/monitoring-metrics.md
mjnagel: We've in the past maintained redirects for the old content pages - https://github.com/defenseunicorns/uds-docs/blob/main/astro.config.mjs#L13-L43 - may be good to do that here as well.

src/loki/values/values.yaml
```diff
@@ -134,23 +155,37 @@ minio:
...
+    # -- Method to use to detect ConfigMap changes. With WATCH the sidecar will do a WATCH request, with SLEEP it will list all ConfigMaps, then sleep for 60 seconds.
+    watchMethod: WATCH
+    # -- WatchServerTimeout: request to the server, asking it to cleanly close the connection after that.
+    # defaults to 60sec; much higher values like 3600 seconds (1h) are feasible for non-Azure K8S.
+    watchServerTimeout: 300
+    # -- WatchClientTimeout: is a client-side timeout, configuring your local socket.
+    # If you have a network outage dropping all packets with no RST/FIN,
+    # this is how long your client waits before realizing & dropping the connection.
+    # Defaults to 66sec.
+    watchClientTimeout: 120
```
mjnagel: Were these values informed from some testing, docs, etc? Just curious why the diff from upstream defaults here.

test/vitest/integration-loki-ruler.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  try {
+    // Try to delete existing ConfigMap first (in case of previous test failures)
+    await K8s(kind.ConfigMap)
+      .InNamespace("loki")
+      .Delete("loki-test-rules")
+      .catch(() => {
+        // Ignore errors if ConfigMap doesn't exist
+      });
```
mjnagel: Very nit, but I think this could just call `removeTestConfigMap`?

test/vitest/integration-loki-ruler.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  }, 110000);
+
+  test("Loki ruler should send alerts to Alertmanager", async () => {
+    await pollUntilSuccess(
+      () => checkAlertInAlertmanager(alertmanagerProxy.url, "LokiAlwaysFiring"),
+      isAlertFiring => isAlertFiring === true,
+      "Checking for LokiAlwaysFiring alert in Alertmanager",
+      180000,
+    );
+  }, 190000);
```
mjnagel: The timeouts on the tests here feel pretty high 🫣. Is that how long these take to work (upwards of 2-3 mins)? Would love to find ways to shorten this cycle if so - maybe changing timing around when the configmap is created for example so that things are already firing/present when we get here.

docs/reference/configuration/observability/monitoring-metrics.md
```diff
@@ -1,5 +1,7 @@
 ---
 title: Monitoring and Metrics
+sidebar:
+    order: 1
```
mjnagel: There's some funny behavior with this order where it affects the order of the other folders on the sidebar. To not disrupt the existing order we probably want a 5 (or 4.5 🙃) to put it after (or right before) backup and restore. We'd also want to delete the ordering on the logging page, or adjust to a higher number that won't conflict/cause issue with the rest of the folders.

docs/reference/configuration/observability/logging-alerting.md
```diff
@@ -0,0 +1,183 @@
+---
+title: Log Based Alerting
+sidebar:
+    order: 2
```
slaskawi: Could we create a corresponding Draft PR in the uds-docs repo for this branch? This way, we could use the "preview" feature and see how future docs would look like.

docs/reference/configuration/observability/logging-alerting.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Architecture Flow
+
+- Applications send logs to Loki
+- Loki Ruler (a part of the Loki backend pods) evaluates rules against log data
+- Alerting rules trigger notifications via Alertmanager
+  - Alerts are routed through notification channels
+- Recording rules create metrics stored in Prometheus
+  - Metrics are available for dashboards and further alerting
+
```
slaskawi, line 26: I'm OK with addressing this as a follow up but I would suggest adding some small diagram here.

docs/reference/configuration/observability/logging-alerting.md
```diff
@@ -0,0 +1,183 @@
...
+          - record: my_app:error_percentage
+            expr: |
+              (
+                sum(rate({namespace="my-app-namespace",app="my-app"} |= "ERROR" [1m]))
+                /
+                sum(rate({namespace="my-app-namespace",app="my-app"} [1m]))
+              ) * 100
+```
+
+You can then create Prometheus alerting rules based on these recorded metrics using a [PrometheusRule CRD](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PrometheusRule) (these rules will automatically be sent to Alertmanager):
```
slaskawi, line 130: Should we add a Grafana screenshot with these alerts?

docs/reference/configuration/observability/monitoring-metrics.md
slaskawi: I also requested a "draft" UDS Docs PR so that we could have a look if the everything rendered correctly: https://github.com/defenseunicorns/uds-core/pull/1930/files#r2371576154

src/loki/values/values.yaml
slaskawi: Related to this, I think these are present in the upstream Helm chart values: https://github.com/grafana/loki/blob/v3.5.5/production/helm/loki/values.yaml#L3552-L3581

Can we just use these instead of providing our owns?

slaskawi (commented):
I tested creating Alerts based on Loki and viewing them in Grafana. This use case worked really well - it's fantastic work @joelmccoy !

test/vitest/integration-loki-ruler.spec.ts
joelmccoy: Locally the tests have been taking anywhere from 30 secs to 1 min and 30 secs to complete.

I think these might have taken a bit longer in CI so i bumped them higher here just for some buffer.  

I'm not sure where we could put the CM deployment earlier in the pipeline to give it more time.  If you have some ideas I'm all ears.

Right now this test file should run in parallel with and I think the bottleneck is still the network tests: https://github.com/defenseunicorns/uds-core/actions/runs/17865942215/job/50808267131?pr=1930#step:5:1377

vs: loki integration: https://github.com/defenseunicorns/uds-core/actions/runs/17865942215/job/50808267131?pr=1930#step:5:1462

So maybe this isn't that bad in current state?

docs/reference/configuration/observability/logging-alerting.md
joelmccoy: you can actually see the preview locally too with the task `uds run dev-docs`

docs/reference/configuration/observability/logging-alerting.md
joelmccoy: yeah think we will cover the diagram in: https://github.com/defenseunicorns/uds-core/issues/1878

test/vitest/integration-loki-ruler.spec.ts
mjnagel: It looks like the network tests take a combined total of 16 seconds (16713ms) whereas these new Loki tests take a total of 134 seconds (134879ms)? So definitely significantly longer than our other tests - adding almost 2 mins to CI I think (134s - 16s = 118s / 1:58). Let me know if I'm misreading the CI output here though.

It might feel a bit strange but we could add the test configmaps to our `src/test` folder/zarf package. That's the approach we take with the network tests to ensure that all pods are already up, with the necessary network policies and other things in place, well before we get to the vitests.

src/loki/values/values.yaml
mjnagel: Resolving based on changes.

test/vitest/integration-loki-ruler.spec.ts
joelmccoy: I missed a digit there 😂 . that is the correct analysis.  definitely see the concern now.  I think adding the configmap to our test package makes sense.  will add that in

test/vitest/integration-loki-ruler.spec.ts
joelmccoy: this is removed now

src/loki/values/values.yaml
joelmccoy: I updated this to be `/rules/uds`.  the subdirectory usually indicates the loki `tenant`. since we are single tenant `fake` is usually the default tenant.  however, i don't think it matters what name we use here.  loki is going to pick up all rules under `/rules/**`.  So maybe we can make this prettier by calling it `/rules/uds`.  Thoughts or any better suggestion on other names?

test/vitest/integration-loki-ruler.spec.ts
joelmccoy: moved the CM to the `src/test` package and it seems that by the time the test come around it is already firing.  good call: https://github.com/defenseunicorns/uds-core/actions/runs/17955605184/job/51066090534?pr=1930#step:5:1373

docs/reference/configuration/observability/monitoring-metrics.md
joelmccoy: interesting... i made a bad assumption that the order was based on subdirectory. but it sounds like it is global.  i adjusted the observability to 4.5 to put it right before backup and restore and removed the ordering for the logging page. looks like this now: 
image
Let me know if this makes sense?

test/vitest/integration-loki-ruler.spec.ts
mjnagel: Perfect, yeah much better timing.

docs/reference/configuration/observability/monitoring-metrics.md
mjnagel: I think this makes sense - and yeah the ordering is weird the way it works...I believe it's global within `configuration`?

docs/reference/configuration/observability/logging-alerting.md
joelmccoy: not sure how useful a screenshot in grafana will be here. it would just list the rule in the UI like so
image

if you think adding in something like this is still useful, I can apply the example rules here and take a screenshot like the one above for these specific ones

docs/reference/configuration/observability/monitoring-metrics.md
joelmccoy: draft PR: https://github.com/defenseunicorns/uds-docs/pull/297

src/loki/values/values.yaml
joelmccoy: actually i had to bring this back.  having this set to `/rules/fake` is important for grafana to ingest and view the rules in the UI.  when loki is setup with no auth, the tenant defaults to `fake` and grafana will query the ruler api with `tenant=fake` to load the rules.  i tested this locally.  when i put uds the alerts/recording rules are missing from the UI and when i set `fake` they properly show up under grafana managed alerts.

docs/reference/configuration/observability/logging-alerting.md
slaskawi: Sounds good to me!

docs/reference/configuration/observability/logging-alerting.md
slaskawi: Ha! Today I learned!

src/loki/values/values.yaml
slaskawi: oooh my! Thanks for checking!

slaskawi (commented):
LGTM!

src/loki/values/values.yaml
mjnagel: I see this PR is noted as a breaking change - is this why? Would someone hit issues if they were overriding this value to some bucket, but not actually using the bucket? Or the rules would just be copied to the bucket but otherwise no problems?

Just want to understand what we may want to say in release notes/put into a release override for the breaking change note.

mjnagel (approved):
LGTM overall, just one question about the breaking change and impact.

src/loki/values/values.yaml
joelmccoy: Yeah this is the reason why I am classifying it as a breaking change (but also could be convinced that it might not be??)

Essentially what will happen is that if a user deployed any loki alerting or recording rules via the Loki Ruler API they would be stored in this bucket.  On this upgrade we point the ruler to local storage by default.  So any old rules in the bucket would not be active and they would have to migrate them to configmaps or set a bunch of overrides to revert the values we set here.  I honestly doubt anyone has actually done this.  But based on this we may want to still classify this as a breaking change?

Also to note.. I think we can remove the need to have the loki ruler bucket in any infra setups.  Was going to create a follow up issue on the EKS module to potentially phase out the loki ruler bucket at some point.

mjnagel: Thanks for explaining - I'd still agree on marking this as a breaking change just in case someone does have rules in their existing bucket.

In testing it seems like having the bucket set as an override doesn't cause any issues so hopefully this doesn't have much of an effect on end users from that perspective.

PR #1937, opened 2025-09-22
fix: switch Keycloak to JDBC_PING

### LINKED ISSUE #1929: [Keycloak 26.4.x] Switch to JDBC_PING stack
## Summary

Keycloak 26.4.0 will introduce JDBC_PING2 stack as default for Kubernetes with improved Split Brain handling (see https://github.com/keycloak/keycloak/discussions/42681). 

The goal of this ticket is to switch the UDS Core stack to JDBC_PING and remove the DNS_PING configuration settings.

slaskawi:
## Description

This Pull Request changes Keycloak JGroups Discovery stack to `jdbc-ping`. 

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1929

## Type of change

## Steps to Validate

- Only automated - check if CI builds stabilized

## Checklist before merging

### DIFF
--- a/.github/bundles/aks/uds-bundle.yaml
+++ b/.github/bundles/aks/uds-bundle.yaml
@@ -4,7 +4,7 @@
 kind: UDSBundle
 metadata:
   name: uds-core-aks-nightly
-  description: A UDS bundle for deploying UDS Core on AKS
+  description: A UDS bundle for deploying and testing UDS Core on AKS
   # x-release-please-start-version
   version: "0.52.1"
   # x-release-please-end
--- a/.github/bundles/aks/uds-bundle.yaml
+++ b/.github/bundles/aks/uds-bundle.yaml
@@ -4,7 +4,7 @@
 kind: UDSBundle
 metadata:
   name: uds-core-aks-nightly
-  description: A UDS bundle for deploying and testing UDS Core on AKS
+  description: A UDS bundle for deploying UDS Core on AKS
   # x-release-please-start-version
   version: "0.52.1"
   # x-release-please-end
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -167,13 +167,6 @@ spec:
             # Dumb value (not used in the nginx provider, but required by the SPI)
             - name: KC_SPI_X509CERT_LOOKUP_{{ .Values.x509LookupProvider | upper }}_SSL_CLIENT_CERT_CHAIN_PREFIX
               value: UNUSED
-          {{- if .Values.fips }}
-            # This is a workaround for https://github.com/keycloak/keycloak/issues/39454
-            # It silences the TLS Errors coming from DNS_PING out. Once the above issue is resolved and released,
-            # this can be removed.
-            - name: QUARKUS_LOG_CATEGORY__ORG_BOUNCYCASTLE_JSSE_PROVIDER__LEVEL
-              value: WARNING
-          {{- end }}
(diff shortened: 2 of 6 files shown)


src/keycloak/chart/values.yaml
```diff
@@ -339,6 +339,9 @@ autoscaling:
   # The scaling policy to use. This will scale up quickly but only scale down a single Pod per 5 minutes.
   # This is important because caches are usually only replicated to 2 Pods and if one of those Pods is terminated this will give the cluster time to recover.
   behavior:
+    scaleUp:
+      # Keycloak CPU spikes at boot time, this aims to prevent massive unwanted scale outs during the boot time or upgrades
+      stabilizationWindowSeconds: 600
```
slaskawi, line 344: This is probably the most crucial part to avoid HPA misbehaving. 

Keycloak uses extensive CPU when booting up and rebalancing the cluster. Booting a cluster of 3 takes roughly 5 minutes and can lead to scale out events (specifically at AKS where for some reason the CPU spikes are higher). Booting up 5 nodes usually causes Zarf to timeout.

.github/bundles/aks/uds-bundle.yaml
```diff
@@ -4,7 +4,7 @@
 kind: UDSBundle
 metadata:
   name: uds-core-aks-nightly
-  description: A UDS bundle for deploying UDS Core on AKS
+  description: A UDS bundle for deploying and testing UDS Core on AKS
```
mjnagel: more of a note for future reference - we should be able to use the `test-aks`, `test-eks`, and `test-rke2` labels to on-demand trigger these workflows on a PR (without needing to modify the bundles/code).

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -200,17 +193,17 @@ spec:
...
             - name: KC_CACHE
               value: ispn
-            - name: KC_CACHE_STACK
-              value: kubernetes
+            # Hints Infinispan to distribute data across machines in a Multi AZ environment
+            - name: KC_SPI_CACHE_EMBEDDED_DEFAULT_MACHINE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+                  apiVersion: v1
```
mjnagel, line 204: Is there a good reference for what this env is/does? Mostly for my own understanding to understand how this works/helps with multi-AZ.

mjnagel (commented):
LGTM overall with one small question.

.github/bundles/aks/uds-bundle.yaml
slaskawi: aaaaaaaaah I missed that!

src/keycloak/chart/templates/statefulset.yaml
slaskawi: The details will soon be released as part of the 26.4.0 and it's a more of a future improvement, see https://github.com/keycloak/keycloak/issues/41933

This setting will start being effective once we use 26.4.0 and will help Infinispan to better distribute the data. I added it here to avoid things getting flaky as we pick the new version up.

.github/bundles/aks/uds-bundle.yaml
slaskawi: Reverted

src/keycloak/chart/templates/statefulset.yaml
mjnagel: Thanks, that's helpful!

mjnagel (approved):

PR #1944, opened 2025-09-24
feat: falco extra rulesets

### LINKED ISSUE #1908: Ship Falco Incubating/Sandbox Rules with UDS Core
By default, the Falco deployment in UDS Core will only include the stable rules from [default rules](https://falco.org/docs/reference/rules/default-rules/) which is including the Falco container image.

Falco in UDS Core is designed to work in the Airgap, therefore we have automatic updates from the web disabled.  In order to include these rules, we will need to ship them as data in the Zarf Package.

- **Given** a deployment of UDS Core
- **When** a user toggles sandbox/incubating rules
- **Then** falco enables these rulesets without reaching out to the internet

- Allow updates from the web (not suitable for airgap environments and UDS Core does not ship the image needed to pull in automatic updates)
- Allow users to pull down these rules as data in their environment and attach it to their falco deployment (unideal user flow)

### Additional context
It would be relatively simple to have the UDS Core repo pull down the Falco rules from upstream prior to package creation and then store them as configmaps so that Falco can mount them.

chance-coleman:
## Description
Add Incubating and Sandbox rulesets. Task to update rules in CI via autogen. doc update for enabling rulesets and how to exclude specific rules.

## Related Issue

Fixes #1908

## Type of change

## Steps to Validate
- `uds run test-uds-core`
- exec into a pod and run `echo "test" >> /etc/passwd` i used the neuvector enforcer pod.
- verify that an alert for the rule `Write below etc` is logged
- try different variations of the bundle overrides

## Checklist before merging


chance-coleman:
this PR is currently set to merge into Joel's branch for implementing falco zarf package. Will probably wait till that is merged and then switch this to merge with main.

src/falco/tasks.yaml
```diff
@@ -25,3 +25,28 @@ tasks:
         cmd: |
           npm ci && npx vitest run "falco" --exclude "**/*integration*"
         dir: test/vitest
+
+  - name: update-rules
```
joelmccoy, line 29: I wonder since we are already downloading incubating/sandbox, maybe we should include the default rules here as well and wire them up to overwrite the rules shipped with the container?

That way we always release UDS Core with the most up to date rules.  

I think there is a scenario where a rule is promoted from sandbox to stable.  But falco hasn't had a release yet (they release only a couple times a year) and UDS Core wouldn't have this rule enabled anymore because the container doesn't include it in stable.

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -21,10 +21,50 @@ UDS Core ships Falco with [stable default rules](https://github.com/falcosecurit
 
 See more about default rules in the [Falco documentation](https://falco.org/docs/reference/rules/default-rules/).
 
-:::caution[Sandbox/Incubating Rules]
-UDS Core currently does not enable the Falco sandbox/incubating ruleset by default.
+#### Additional Rulesets
+
+:::note
+UDS Core ships with **sandbox** and **incubating** rulesets from the Falco community, but they are **disabled by default**. These rulesets contain experimental rules that may need tuning to reduce false positives and noise.
```
joelmccoy: I think we can remove the word experimental here. I think we could almost encourage people to enable at least the incubating rules. Incubating rules aren't considered experimental.  Maybe we can point people to the definitions here as well in this note: https://github.com/falcosecurity/rules/blob/main/proposals/20230605-rules-adoption-management-maturity-framework.md#rules-maturity-framework

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -21,10 +21,50 @@ UDS Core ships Falco with [stable default rules](https://github.com/falcosecurit
...
+            value: true
+          - path: "incubationRulesEnabled"
+            value: true
+          - path: "disabledRules"
+            value: ["Write below root"]
+```
+
+This configuration:
+
+1. Enables the sandbox ruleset while excluding the "Terminal shell in container" rule
```
joelmccoy:
```suggestion
1. Enables the sandbox ruleset while excluding the "Write below root" rule
```

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -21,10 +21,50 @@ UDS Core ships Falco with [stable default rules](https://github.com/falcosecurit
...
+2. Enables the incubating ruleset while excluding the "Read environment variable from /proc files" rule
```
joelmccoy: I think we are missing the configuration for this

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -21,10 +21,50 @@ UDS Core ships Falco with [stable default rules](https://github.com/falcosecurit
...
+
+#### Finding Rule Names for `disabledRules`
+
+The rule names used in the `disabledRules` array should match the `rule` field from the Falco rules files. You can find these rule names in the following locations:
```
joelmccoy: does the disabledRules also apply to the stable rules?

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -21,10 +21,50 @@ UDS Core ships Falco with [stable default rules](https://github.com/falcosecurit
...
+1. **In the rule files shipped with UDS Core**:
+   - Sandbox rules: [`src/falco/chart/rules/sandbox-rules.yaml`](https://github.com/defenseunicorns/uds-core/blob/main/src/falco/chart/rules/sandbox-rules.yaml)
+   - Incubating rules: [`src/falco/chart/rules/incubating-rules.yaml`](https://github.com/defenseunicorns/uds-core/blob/main/src/falco/chart/rules/incubating-rules.yaml)
+
+   Look for entries that start with `- rule:` to find the rule names.
+
+2. **From Falco logs**:
+   When Falco detects an event, it logs the rule name in the output. You can find these logs by querying Loki with:
+   ```txt
+   {component="falco"} |~ "rule"
```
joelmccoy: I don't think this query works as expected.  I'm not sure component is part of the schema at all here.

I think this could just be:
```suggestion
   {rule=~".+"}
```

docs/reference/configuration/runtime-security/alerting.md
joelmccoy: if so it might be good to note that in these docs

src/falco/chart/values.yaml
```diff
@@ -10,3 +10,13 @@ additionalNetworkAllow: []
...
+
+# Toggle inclusion of Falco's "sandbox" ruleset (experimental but more polished than incubating ruleset)
+sandboxRulesEnabled: false
+
+# Toggle inclusion of Falco's "incubating" ruleset (newer, less stable rules)
+incubationRulesEnabled: false
+
+# List of rules to explicitly disable by name
+# Example: ["Read environment variable from /proc files"]
+disabledRules: []
```
joelmccoy, line 22: love having this as a simple input here ❤️

joelmccoy (commented):
Generally looking good! Based on one of my comments, I'm curious if it is worth syncing the stable rules with this process as well

src/falco/tasks.yaml
chance-coleman: I think this makes a lot of sense, i've already got the scaffolding for incubating and sandbox so this would be a simple addition at this point.

docs/reference/configuration/runtime-security/alerting.md
chance-coleman: updated to add a link to the official docs

docs/reference/configuration/runtime-security/alerting.md
chance-coleman: updated

docs/reference/configuration/runtime-security/alerting.md
chance-coleman: good call, switched this to be more accurate

docs/reference/configuration/runtime-security/alerting.md
chance-coleman: good call, added call out to be more explicit about what the disabled rules applies too (which is everything)

docs/reference/configuration/runtime-security/alerting.md
chance-coleman: updated

src/falco/tasks.yaml
```diff
@@ -25,3 +25,28 @@ tasks:
...
+            echo "Error: Failed to download sandbox rules"
+            exit 1
+          fi
+
+          if ! curl -sSfL --connect-timeout 30 --max-time 60 \
+               https://raw.githubusercontent.com/falcosecurity/rules/main/rules/falco-incubating_rules.yaml \
+               -o "${RULES_DIR}/incubating-rules.yaml"; then
+            echo "Error: Failed to download incubating rules"
+            exit 1
+          fi
```
mjnagel, line 59: I think these should be pinned to specific git tags (`falco-sandbox-rules-5.0.1` and `falco-incubating-rules-5.0.1` are the latest I think) and then tracked with renovate. Otherwise we're at risk of breaking CI at random points (such as during a release PR) and pulling in code that hasn't gone through their release process (not sure if they do additional testing or not on it).

chance-coleman: added git tags and uds-common renovate rules should capture the github raw tags with semver-coercion.

src/falco/tasks.yaml
mjnagel: I think this makes sense to do. Hopefully any breaking changes (new falco version required) are obvious/properly reflected with a major version bump. The git tags/versioning on the stable rules are a bit weird but as long as they work 😄

chance-coleman: @joelmccoy @mjnagel I've added the stable rules, i verified the integration tests still work. As far as i could tell, by changing the `rules_files` value to not include the default rules that are packaged in the image, there is no overriding or duplicating of those images with configmap defined stable rules.

src/falco/chart/templates/rules-configmap.yaml
```diff
@@ -0,0 +1,35 @@
...
+{{ .Files.Get "rules/sandbox-rules.yaml" | nindent 4 }}
+  {{- end }}
+
+  # Include incubating rules if incubating ruleset is enabled in values.yaml
+  {{- if .Values.incubationRulesEnabled }}
+  incubating-rules.yaml: |-
+{{ .Files.Get "rules/incubating-rules.yaml" | nindent 4 }}
+  {{- end }}
+
+  # If either sandbox or incubating excludes are defined, generate a disable-rules file.
```
joelmccoy: nit: i think this comment is not accurate? 
```suggestion
  # If disabledRules are passed in, generate a disable-rules file
```

src/falco/chart/values.yaml
```diff
@@ -10,3 +10,13 @@ additionalNetworkAllow: []
 #     remoteGenerated: Anywhere
 #     description: "from falcosidekick to anywhere"
 #     port: 443
+
+# Toggle inclusion of Falco's "sandbox" ruleset (experimental but more polished than incubating ruleset)
```
joelmccoy: nit: I think this might be backwards.  Sandbox is not more polished than incubating. Pulling definitions from here: https://github.com/falcosecurity/rules?tab=readme-ov-file#falco-rules-2x
```suggestion
# Toggle inclusion of Falco's "sandbox" ruleset (experimental rules)
```

src/falco/chart/values.yaml
```diff
@@ -10,3 +10,13 @@ additionalNetworkAllow: []
...
+sandboxRulesEnabled: false
+
+# Toggle inclusion of Falco's "incubating" ruleset (newer, less stable rules)
```
joelmccoy: from here: https://github.com/falcosecurity/rules?tab=readme-ov-file#falco-rules-2x
```suggestion
# Toggle inclusion of Falco's "incubating" ruleset (provide a certain level of robustness guarantee but have been identified by experts as catering to more specific use cases, which may or may not be relevant for each adopter.)
```

joelmccoy: it also might be good to put incubating rules before the sandboxrules here just so people don't get confused.  Incubating are more mature than sandbox.

bundles/k3d-standard/uds-bundle.yaml
```diff
@@ -223,3 +223,14 @@ packages:
...
             - path: "falcosidekick.replicaCount"
               value: 1
+        uds-falco-config:
+          variables:
+            - name: FALCO_SANDBOX_RULES_ENABLED
+              description: Enable sandbox rules
+              path: sandboxRulesEnabled
+            - name: FALCO_INCUBATION_RULES_ENABLED
+              description: Enable incubation rules
+              path: incubationRulesEnabled
```
joelmccoy: nit: maybe keep the naming the same as the upstream name (incubating vs incubation)
```suggestion
            - name: FALCO_INCUBATING_RULES_ENABLED
              description: Enable incubating rules
              path: incubatingRulesEnabled
```

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -21,10 +21,53 @@ UDS Core ships Falco with [stable default rules](https://github.com/falcosecurit
...
+To enable the [sandbox and incubating](https://falco.org/docs/reference/rules/default-rules/) rulesets and exclude specific rules, override the `extraRules` value in your UDS Core bundle:
+
+```yaml
+  overrides:
+    falco:
+      uds-falco-config:
+        values:
+          - path: "sandboxRulesEnabled"
+            value: true
+          - path: "incubationRulesEnabled"
```
joelmccoy: nit: maybe keep the naming the same as the upstream name (incubating vs incubation)
```suggestion
          - path: "incubatingRulesEnabled"
```

src/falco/chart/templates/rules-configmap.yaml
```diff
@@ -0,0 +1,35 @@
...
+{{ .Files.Get "rules/stable-rules.yaml" | nindent 4 }}
+
+  # Include sandbox rules if sandbox ruleset is enabled in values.yaml
+  {{- if .Values.sandboxRulesEnabled }}
+  sandbox-rules.yaml: |-
+{{ .Files.Get "rules/sandbox-rules.yaml" | nindent 4 }}
+  {{- end }}
+
+  # Include incubating rules if incubating ruleset is enabled in values.yaml
+  {{- if .Values.incubationRulesEnabled }}
```
joelmccoy: nit: maybe keep the naming the same as the upstream name (incubating vs incubation)
```suggestion
  {{- if .Values.incubatingRulesEnabled }}
```

src/falco/chart/values.yaml
```diff
@@ -10,3 +10,13 @@ additionalNetworkAllow: []
...
+incubationRulesEnabled: false
```
joelmccoy: nit: maybe keep the naming the same as the upstream name (incubating vs incubation)
```suggestion
incubatingRulesEnabled: false
```

tasks/deploy.yaml
```diff
@@ -21,7 +21,7 @@ tasks:
         description: "Extra args for k3d"
     actions:
       - description: "Deploy the UDS Core Standard Bundle"
-        cmd: uds deploy bundles/k3d-standard/uds-bundle-k3d-core-demo-${UDS_ARCH}-${VERSION}.tar.zst --set INSECURE_ADMIN_PASSWORD_GENERATION=true --set K3D_EXTRA_ARGS="${{ .inputs.K3D_EXTRA_ARGS }}" --confirm --no-progress
+        cmd: uds deploy bundles/k3d-standard/uds-bundle-k3d-core-demo-${UDS_ARCH}-${VERSION}.tar.zst --set INSECURE_ADMIN_PASSWORD_GENERATION=true --set FALCO_SANDBOX_RULES_ENABLED=true --set FALCO_INCUBATION_RULES_ENABLED=true --set K3D_EXTRA_ARGS="${{ .inputs.K3D_EXTRA_ARGS }}" --confirm --no-progress
```
joelmccoy: nit: maybe keep the naming the same as the upstream name (incubating vs incubation)
```suggestion
        cmd: uds deploy bundles/k3d-standard/uds-bundle-k3d-core-demo-${UDS_ARCH}-${VERSION}.tar.zst --set INSECURE_ADMIN_PASSWORD_GENERATION=true --set FALCO_SANDBOX_RULES_ENABLED=true --set FALCO_INCUBATING_RULES_ENABLED=true --set K3D_EXTRA_ARGS="${{ .inputs.K3D_EXTRA_ARGS }}" --confirm --no-progress
```

tasks/test.yaml
```diff
@@ -195,7 +195,7 @@ tasks:
       - cmd: |
           # Set UDS_CONFIG for bundle deployment
           export UDS_CONFIG="${{ .inputs.configFile }}"
-          uds deploy bundles/k3d-standard/uds-bundle-k3d-core-demo-${UDS_ARCH}-${VERSION}.tar.zst --packages core --confirm --no-progress
+          uds deploy bundles/k3d-standard/uds-bundle-k3d-core-demo-${UDS_ARCH}-${VERSION}.tar.zst --set FALCO_SANDBOX_RULES_ENABLED=true --set FALCO_INCUBATION_RULES_ENABLED=true --packages core --confirm --no-progress
```
joelmccoy: nit: maybe keep the naming the same as the upstream name (incubating vs incubation)
```suggestion
          uds deploy bundles/k3d-standard/uds-bundle-k3d-core-demo-${UDS_ARCH}-${VERSION}.tar.zst --set FALCO_SANDBOX_RULES_ENABLED=true --set FALCO_INCUBATING_RULES_ENABLED=true --packages core --confirm --no-progress
```

joelmccoy (commented):
Couple nits on docs and naming things. Otherwise LGTM

src/falco/chart/templates/rules-configmap.yaml
chance-coleman: ahh yeah good catch, at one point i had individual `exclude` values before i consolidated into a single value.

src/falco/tasks.yaml
```diff
@@ -25,3 +25,35 @@ tasks:
...
+
+          if ! curl -sSfL --connect-timeout 30 --max-time 60 \
+               https://raw.githubusercontent.com/falcosecurity/rules/falco-sandbox-rules-5.0.1/rules/falco-sandbox_rules.yaml \
+               -o "${RULES_DIR}/sandbox-rules.yaml"; then
+            echo "Error: Failed to download sandbox rules"
+            exit 1
+          fi
+
+          if ! curl -sSfL --connect-timeout 30 --max-time 60 \
+               https://raw.githubusercontent.com/falcosecurity/rules/falco-incubating-rules-5.0.1/rules/falco-incubating_rules.yaml \
```
joelmccoy, line 55: do you know if these are renovated with the default config?  and will they keep the proper tag structure, for example: `falco-rules-4.0.0 -> `falco-rules-4.0.1 and `falco-sandbox-rules-5.0.1` -> `falco-sandbox-rules-5.0.2`.

joelmccoy (dismissed):
LGTM with a nonblocking question.

src/falco/tasks.yaml
chance-coleman: yeah good question. Short answer, yes they are. Long answer, this is a [regex101 link](https://regex101.com/r/JW0RkZ/1) to one of the uds-common renovate configs that matches.

src/falco/chart/templates/rules-configmap.yaml
```diff
@@ -0,0 +1,35 @@
...
+  {{- if .Values.sandboxRulesEnabled }}
+  sandbox-rules.yaml: |-
+{{ .Files.Get "rules/sandbox-rules.yaml" | nindent 4 }}
+  {{- end }}
+
+  # Include incubating rules if incubating ruleset is enabled in values.yaml
+  {{- if .Values.incubatingRulesEnabled }}
+  incubating-rules.yaml: |-
+{{ .Files.Get "rules/incubating-rules.yaml" | nindent 4 }}
+  {{- end }}
```
mjnagel, line 24: I guess we'll find out when this becomes an issue (if it does) but wondering if multiple configmaps might become a necessity at some point due to size of the rules and configmap size limitations?

mjnagel (approved):
LGTM with one comment for future consideration.

src/falco/values/values.yaml
```diff
@@ -16,6 +16,23 @@ falco:
...
+
+mounts:
+  volumes:
+    - name: extra-rules
+      configMap:
+        name: falco-extra-rules
+  volumeMounts:
+    - name: extra-rules
+      mountPath: /etc/falco/rules.d
+      readOnly: true
```
mjnagel, line 34: One other future consideration just to note it here - while this volume should update "in-place", I'm not sure Falco will automatically read the updated rules if/when they change (seems to just read at startup). We might want to figure out a good solve here other than zarf actions - one option would be to extend [this feature](https://uds.defenseunicorns.com/reference/deployment/secret-pod-reload/) to handle configmaps.

src/falco/chart/templates/rules-configmap.yaml
chance-coleman: for tracking this: https://github.com/defenseunicorns/uds-core/issues/1972

src/falco/values/values.yaml
chance-coleman: for tracking this: https://github.com/defenseunicorns/uds-core/issues/1972

PR #1953, opened 2025-09-25
chore: revert authservice 1.1.0, add slim-dev-ha task

mjnagel:
## Description

Reverts authservice version update due to upstream bug/breaking change (see [issue](https://www.github.com/istio-ecosystem/authservice/issues/304)).

Also adds a slim-dev-ha task for quicker iterations on keycloak/authservice HA.

## Type of change

## Checklist before merging

### DIFF
--- a/bundles/k3d-slim-dev/uds-ha-config.yaml
+++ b/bundles/k3d-slim-dev/uds-ha-config.yaml
@@ -1,0 +1,15 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+variables:
+  core-identity-authorization:
+    keycloak_ha: true
+    keycloak_pg_username: postgres
+    keycloak_pg_password: unicorn123!@#UN
+    keycloak_pg_database: keycloak
+    keycloak_pg_host: host.k3d.internal
+    keycloak_devmode: false
+
+  # Authservice config is managed by the operator in the base layer
+  core-base:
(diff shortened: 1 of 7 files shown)


bundles/k3d-slim-dev/uds-ha-config.yaml
```diff
@@ -0,0 +1,15 @@
...
+    keycloak_ha: true
+    keycloak_pg_username: postgres
+    keycloak_pg_password: unicorn123!@#UN
+    keycloak_pg_database: keycloak
+    keycloak_pg_host: host.k3d.internal
+    keycloak_devmode: false
+
+  # Authservice config is managed by the operator in the base layer
+  core-base:
+    AUTHSERVICE_REDIS_URI: redis://authservice:authservice@host.k3d.internal:6379
```
joelmccoy, line 15: digging into this... we don't actually deploy authservice HA? we just set the redis URI for our HA config?  is it worth bumping the replica count to at least 2 to truly test HA?  and if so, might be good to include this on our standard bundle.

joelmccoy: doesn't have to be done in this PR if so.  just wanted to call it out

joelmccoy (approved):
LGTM with a question

bundles/k3d-slim-dev/uds-ha-config.yaml
mjnagel: Yeah that's fair and a good catch. If it's okay I'll do a follow-on PR for that? I probably should've separated out the slim-dev-ha stuff already but I had started on it to test Authservice 1.1.0 😅

joelmccoy: totally sounds good

PR #1959, opened 2025-09-25
chore: split pepr and operator-deps into own renovate groups

joelmccoy:
## Description

All npm deps are being grouped into the pepr renovate group.  See: https://github.com/defenseunicorns/uds-core/pull/1954 as an example PR where this causes issues.

- Updates the renovate config to bump pepr as a completely separate group
- npm dev deps will get grouped to support-deps
- npm actual deps with get grouped to operator-deps

## Related Issue

N/A

## Type of change

## Checklist before merging

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -69,7 +69,14 @@
       "commitMessageTopic": "support dependencies"
     },
     {
+      "matchFileNames": ["package.json", "package-lock.json"],
+      "groupName": "operator-deps",
+      "excludeDepNames": ["pepr"],
+      "commitMessageTopic": "operator-deps"
+    },
+    {
       "matchFileNames": ["package.json", "package-lock.json", "tasks/create.yaml"],
+      "matchDepNames": ["pepr", "registry1.dso.mil/ironbank/opensource/defenseunicorns/pepr/controller", "defenseunicorns/pepr"],
       "groupName": "pepr",
       "commitMessageTopic": "pepr"
     },
--- a/renovate.json
+++ b/renovate.json
@@ -70,6 +70,12 @@
     },
     {
       "matchFileNames": ["package.json", "package-lock.json"],
+      "groupName": "support-deps",
+      "matchDepTypes": ["devDependencies"],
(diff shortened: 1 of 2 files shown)


slaskawi (dismissed):

renovate.json
```diff
@@ -68,8 +68,21 @@
...
+      "groupName": "support-deps",
+      "matchDepTypes": ["devDependencies"],
+      "commitMessageTopic": "support dependencies"
+    },
+    {
+      "matchFileNames": ["package.json", "package-lock.json"],
+      "groupName": "operator-deps",
+      "excludeDepNames": ["pepr"],
+      "commitMessageTopic": "operator-deps"
+    },
```
mjnagel: Did you happen to test this part locally/on a fork? I think it may need to be reordered - with this config the last match will win out. So any dev dependencies would actually be globbed up by the `operator-deps` match if I'm not mistaken?

renovate.json
```diff
@@ -68,8 +68,21 @@
...
     },
+    {
+      "matchFileNames": ["package.json", "package-lock.json"],
+      "groupName": "support-deps",
+      "matchDepTypes": ["devDependencies"],
+      "commitMessageTopic": "support dependencies"
+    },
+    {
+      "matchFileNames": ["package.json", "package-lock.json"],
+      "groupName": "operator-deps",
```
mjnagel, line 73: Also remembering now - if we are adding a new group and corresponding branch we may need to update the renovate-readiness action to ensure it processes that branch properly: https://github.com/defenseunicorns/uds-core/blob/split-pepr-renovate/.github/actions/renovate-readiness/action.yaml#L47-L58

renovate.json
joelmccoy: swapped the order here

renovate.json
joelmccoy: added a step for operator-deps.  and decided to not fail the job as it might be ok for CI to run on these updates (it shouldn't trigger IAC runs).  But let me know if you feel differently

mjnagel (approved):

PR #1973, opened 2025-09-29
chore: add keycloak default sso_session_idle_timeout value/add compliance mapping

joelmccoy:
## Description

Adds the default value for Keycloak `SSO_SESSION_IDLE_TIMEOUT` and adds a Lula compliance mapping.  Default value taken from identity config: https://github.com/defenseunicorns/uds-identity-config/blob/v0.17.0/src/realm.json#L12

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/values.yaml
+++ b/src/keycloak/chart/values.yaml
@@ -47,6 +47,10 @@ realmInitEnv:
   # Maximum number of concurrent active sessions per user. Set to 0 to disable the limit (unlimited).
   SSO_SESSION_MAX_PER_USER: "0"
   # @lulaEnd ef276ec4-7921-4287-a18c-80dec5f8e742
+  # @lulaStart 251f6498-05a3-416c-b6eb-b9fadd776704
+  # Specifies how long a session remains active without user activity
+  SSO_SESSION_IDLE_TIMEOUT: "600"
+  # @lulaEnd 251f6498-05a3-416c-b6eb-b9fadd776704
 
 # Realm-level configuration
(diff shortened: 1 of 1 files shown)


src/keycloak/chart/values.yaml
```diff
@@ -47,6 +47,10 @@ realmInitEnv:
   # Maximum number of concurrent active sessions per user. Set to 0 to disable the limit (unlimited).
   SSO_SESSION_MAX_PER_USER: "0"
   # @lulaEnd ef276ec4-7921-4287-a18c-80dec5f8e742
+  # @lulaStart 251f6498-05a3-416c-b6eb-b9fadd776704
+  # Specifies how long a session remains active without user activity
+  SSO_SESSION_IDLE_TIMEOUT: "600"
+  # @lulaEnd 251f6498-05a3-416c-b6eb-b9fadd776704
```
mjnagel, line 53: I think this is fine, but we did also add Lula to the identity-config repo CI. To get closer to the source it would be good to wrap [this value](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/realm.json#L12) with Lula comments. In theory the `lula crawl` command will support JSON, we'd just need to figure out a good way to make "comments" and validate it won't mess with the realm import if we have additional fields/values in the JSON (to simulate a comment).

Don't have to now, but might be good to figure out if there's anything viable there.

joelmccoy:
closing in favor of: https://github.com/defenseunicorns/uds-identity-config/pull/654

PR #1979, opened 2025-09-30
chore: disable noisy falco rules

joelmccoy:
## Description

Disables noisy falco rules:
  - "Change thread namespace"
  - "Contact EC2 Instance Metadata Service From Container"
  - "Contact cloud metadata service from container"
  - "Contact K8S API Server From Container"

## Type of change

## Checklist before merging

### DIFF
--- a/src/falco/chart/templates/rules-configmap.yaml
+++ b/src/falco/chart/templates/rules-configmap.yaml
@@ -23,8 +23,16 @@ data:
 {{ .Files.Get "rules/incubating-rules.yaml" | nindent 4 }}
   {{- end }}
 
-  # If disabledRules or udsDefaultDisabledRules are passed in, generate a disable-rules file
-  {{- $allDisabledRules := concat .Values.udsDefaultDisabledRules .Values.disabledRules | uniq }}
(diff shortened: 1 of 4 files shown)


src/falco/chart/templates/rules-configmap.yaml
```diff
@@ -23,10 +23,11 @@ data:
 {{ .Files.Get "rules/incubating-rules.yaml" | nindent 4 }}
   {{- end }}
 
-  # If disabledRules are passed in, generate a disable-rules file
-  {{- if (gt (len .Values.disabledRules) 0) }}
+  # If disabledRules or udsDefaultDisabledRules are passed in, generate a disable-rules file
+  {{- $allDisabledRules := concat .Values.udsDefaultDisabledRules .Values.disabledRules }}
```
mjnagel: I think this is a great place to add some helm unittests. Also curious if we need to/should do a [uniq](https://helm.sh/docs/chart_template_guide/function_list/#uniq-mustuniq) to dedupe this? Might not matter with the way Falco evaluates.

joelmccoy: good call. added this in.

mjnagel (approved):
LGTM 📈🐦

PR #1987, opened 2025-10-02
chore: add observability reference architecture docs

### LINKED ISSUE #1878: Establish/Document a UDS Core Observability Reference Architecture

joelmccoy:
## Description

Adds docs/best practices/architecture diagram for the UDS Core's Observability stack.

## Related Issue

Fixes #1878 

## Type of change

## Steps to Validate

- `uds run dev-docs` and check the docs site locally (including embedded likec4 diagram)

## Checklist before merging


joelmccoy:
Waiting until https://github.com/defenseunicorns/uds-core/pull/2000 is merged in to build the observability overview diagram

docs/reference/configuration/observability/metrics-alerting.md
```diff
@@ -0,0 +1,85 @@
...
+    repository: ghcr.io/defenseunicorns/packages/uds/core
+    ref: x.x.x
+    overrides:
+      kube-prometheus-stack:
+        kube-prometheus-stack:
+          values:
+            - path: defaultRules.disabled
+              value:
+                KubeControllerManagerDown: true
+                KubeSchedulerDown: true
```
mjnagel, line 32: I think this tracks, but it's interesting that there's a true/false for enabling individual rules too ([here](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L168-L204)). It seems like some rules might just have a toggle for the full rule CR ([example](https://github.com/prometheus-community/helm-charts/blob/026c5f47bda36e5d7ebf1c104df47cfe6d40d097/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/k8s.rules.container_memory_working_set_bytes.yaml#L7)) and others have specific "sub-rules" that are disable-able ([ref](https://github.com/prometheus-community/helm-charts/blob/026c5f47bda36e5d7ebf1c104df47cfe6d40d097/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/etcd.yaml#L27)). Not sure if it's worth calling out both here?

mjnagel (commented):
Looks great overall - lots of great information and a helpful overview + specific guides on each component. LGTM pending the diagram (or we can remove that TODO and ship as-is).

docs/reference/configuration/observability/metrics-alerting.md
joelmccoy: good call. updated this section to call out both options.

docs/.c4/_spec.c4
```diff
@@ -17,6 +17,14 @@ specification {
   color k8sNamespaces #d1bad3
   color coreWebhooks #d8bf7a
   color pods #bea6a6
+  color crds #9e9e9e
+  color configmaps #5a6b5d
```
joelmccoy, line 21: i let AI pick these colors 😂 . not sure if we have strong opinions for these ones

docs/.c4/diagrams/base-layer.png
joelmccoy, line 1: not sure why this changed. it looks functionally the same. something may be off a pixel or something here? all the other diagrams when regenerated did not change.

docs/reference/configuration/observability/overview.mdx
```diff
@@ -0,0 +1,30 @@
+---
+title: Observability Overview
+sidebar:
+    order: 4.1
+---
+import LikeC4View from '../../../../components/LikeC4View.astro';
```
joelmccoy: you can test this locally by changing this to:
`import LikeC4View from '../../../../uds-docs/src/components/LikeC4View.astro';`

docs/.c4/views.c4
```diff
@@ -161,4 +182,27 @@ views "uds-core" {
...
+    exclude
+      cluster.generic-customer.customer-waypoint,
+      cluster.loki.loki-read,
+      cluster.loki.loki-write,
+      cluster.monitoring.kube-state-metrics,
+      cluster.monitoring.node-exporter,
+      * -> *
+        where kind is not observability
+
+      autoLayout TopBottom 300 100
```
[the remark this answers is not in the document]
joelmccoy: fake news.  this is indented correctly

docs/reference/configuration/observability/monitoring-metrics.md
```diff
@@ -1,10 +1,10 @@
 ---
 title: Monitoring and Metrics
 sidebar:
-    order: 4.5
+    order: 4.2
```
chance-coleman, line 4: Do we feel strongly about the ordering of these docs? in the past we have only ordered the `Overview` and left everything else unordered, which was mainly just so we didn't have the headache of maintaining the orders and sidebar ordering of parent directory. This definitely works, and im fine with it as is.

docs/.c4/_spec.c4
chance-coleman: personally think they are fine

docs/reference/configuration/observability/monitoring-metrics.md
joelmccoy: i feel medium strong about the ordering here 😂 . I think it tells a better logical story to sequence like:
Overview -> Monitoring and Metrics (how do you monitor you applications) -> Alert Management (how do you manage alert routing) -> Metrics Alerting (how do you configure additional alerts) -> Log Based Alerting (how do you alert on logs) -> Dashboards (how do you visualize your data)

i think the 4.1/4.2/4.3/etc notation works kind of nice here.  if we assume all observability docs are 4.X, we can order them without affecting other parts

chance-coleman: Definitely on board with that. I was looking at the current docs and it looks like somewhere along the line it got screwed messed up again. service-mesh has both `order: 4` and `order: 5` and then observability goes back to `order: 4.1`... but might not be worth the effort when we're planning a docs rework anyway.

chance-coleman (dismissed):

docs/reference/configuration/observability/monitoring-metrics.md
mjnagel: I think as long as the order works out (which looks like it does) then this should be fine?

docs/.c4/diagrams/base-layer.png
mjnagel: This seems to happen occasionally and is strange...maybe something to track down if it keeps happening. For now probably okay - just something to be aware of.

docs/.c4/views.c4
```diff
@@ -41,6 +41,12 @@ views "uds-core" {
         istio-mutating-webhook,
         zarf-mutating-webhook,
         kubernetes-control-plane,
+
+      // exclude any observability relationships and tagged elements
+      exclude
+        * where tag is #observability,
+        * -> *
+          where kind is observability
```
mjnagel, line 49: This feels a bit annoying for scalability, but I guess it's because we include `*` up above?

docs/.c4/views.c4
```diff
@@ -161,4 +182,27 @@ views "uds-core" {
...
+  view observability-reference-architecture of infrastructure {
+    title "Reference Architecture / Observability Stack"
+
+    include
+      cluster,
+      cluster.monitoring.*,
+      cluster.grafana.*,
+      cluster.generic-customer.*,
+      cluster.loki.*,
+      external-apps.ext-alerts,
```
mjnagel, line 196: Is Vector intentionally left off? I saw some relationships defined in the model, unclear if that was intentional to not include in the view?

docs/.c4/model.c4
```diff
@@ -196,6 +196,11 @@ model {
           grafana -> keycloak-waypoint
           grafana -> monitoring.prometheus
           grafana -> loki-gateway
+
+          // observability relationships
+          grafana -[observability]-> monitoring.prometheus 'queries metrics data'
+          grafana -[observability]-> loki-gateway 'queries logs data'
+          grafana -[observability]-> monitoring.alertmanager 'queries and manages alerts'
```
mjnagel, line 203: Do we want to capture the user in here (accessing Grafana)? Okay if we want to just focus on system/components here.

mjnagel (commented):
Few small diagram comments.

docs/.c4/views.c4
joelmccoy: yea. we could specify the specific components in the overall view instead of *. but then you have the same scalability issue I think? anything you add to the overall view will need to be explicitly added to this diagram?  not opposed to the other approach. but I thought the exclusion here works well without modifying the other diagrams

docs/.c4/views.c4
joelmccoy: good catch. was not intentional to leave out. added this in and rearranged things

docs/.c4/model.c4
joelmccoy: good call! added this in

docs/.c4/views.c4
mjnagel: Yeah - I'm good with this. Agreed, exclusion problem is better than inclusion problem 😆

mjnagel (approved):

PR #1993, opened 2025-10-03
feat: keycloak relies on Istio for mTLS

### LINKED ISSUE #1581: Optimize Keycloak port configuration
## Summary

This ticket has been created in response of the following comment:

              @slaskawi @UnicornChance Instead of set the Port 7800 to PERMISSIVE you may consider to declare this port in the headless service. This will allow communication between all pods of the headless service but not from any other pods, which are not belongs to the cluster. see https://istio.io/v1.12/docs/ops/configuration/traffic-management/traffic-routing/#headless-services

_Originally posted by @johgoe in https://github.com/defenseunicorns/uds-core/issues/1529#issuecomment-2873786650_
            
This sounds like a perfect suggestion. Thanks @johgoe!

slaskawi:
## Description

This Pull Request disabled Infinispan Cache Encryption and relies on Istio for mTLS

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1581

## Type of change

## Steps to Validate

Tested automatically 

## Checklist before merging


mjnagel (dismissed):
LGTM - great to be cleaning up the peerauth exemption here :)

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -202,6 +202,10 @@ spec:
                 fieldRef:
                   fieldPath: spec.nodeName
                   apiVersion: v1
+            # Rely on Istio mTLS encryption
+            # See https://www.keycloak.org/server/caching#_running_inside_a_service_mesh
+            - name: KC_CACHE_EMBEDDED_MTLS_ENABLED
+              value: "false"
```
mjnagel, line 213: Just to confirm - this being nested inside the postgres conditional is the right place? This seems to match some of our other HA related things like ispn cache and the node name, but wanted to check.

slaskawi: You have a good point. Even though we use these settings in PostgreSQL mode, they are more related to HA. I'll move all of them to proper place.

slaskawi: Apart from moving this under HA Mode check, I added validation that PostgreSQL needs to be enabled in clustering mode. Otherwise, the configuration doesn't make any sense.

mjnagel (approved):

mjnagel:
Disabling auto-merge while we get the patch release running through CI - https://github.com/defenseunicorns/uds-core/pull/1981

PR #2012, opened 2025-10-10
chore: keycloak lula mappings

joelmccoy:
## Description

Adds more lula mappings for keycloak.

## Type of change

## Checklist before merging


src/keycloak/chart/values.yaml
```diff
@@ -57,7 +57,9 @@ realmConfig:
 realmAuthFlows:
   USERNAME_PASSWORD_AUTH_ENABLED: true
   X509_AUTH_ENABLED: true
+  # @lulaStart 96b7aa1b-b307-45c0-af40-8b57f3726693
   SOCIAL_AUTH_ENABLED: true
+  # @lulaEnd 96b7aa1b-b307-45c0-af40-8b57f3726693
```
andrewg-xyz: non-blocking: curiosity. 

When a control is mapped by an override config, how do we handle that?

i.e. I could change this with a zarf helm override.

joelmccoy: totally true and this question has been brought up a bit while working through this.  

Ultimately we are trying to do a best effort mapping of a default deploy of Core.  It's hard to account for every permutation of environment specific overrides.  I think these mappings act as supporting evidence (instead of absolute truth for compliance) in conjunction with the justification writeup we document in lula2.

cmwylie19 (dismissed):

mjnagel (approved):

PR #2013, opened 2025-10-10
fix: velero crd upgrade process

chance-coleman:
## Description
Registry1 updated their Velero onto a distroless base image, causing CRD upgrade jobs to fail. This PR introduces a fix to pull in the various CRDs as zarf manifests. This change also means we no longer need the kubectl image for Velero to do its upgrade process.

## Type of change

## Steps to Validate
- `uds run -f tasks/test.yaml uds-core-upgrade --set FLAVOR=registry1`

## Checklist before merging


src/velero/common/zarf.yaml
```diff
@@ -45,3 +45,49 @@ components:
         releaseName: velero
         valuesFiles:
           - ../values/values.yaml
+
+  - name: velero-crds
+    required: true
+    manifests:
+      - name: velero-crds
```
mjnagel, line 36: Shouldn't this go before velero itself? Otherwise we need to maintain the action for ownership perpetually I think, since installs would have the CRDs from the upstream chart.

src/velero/zarf.yaml
```diff
@@ -8,6 +8,11 @@ metadata:
   url: https://velero.io/
 
 components:
+  - name: velero-crds
+    required: true
+    import:
+      path: common
```
mjnagel, line 14: We'll need to import this component in the backup-restore layer and the standard package as well.

src/velero/common/zarf.yaml
mjnagel: I guess the import order a layer up is correct so it works, but probably makes sense to order right here.

brandtkeller:
Making a mental note on the utility of Zarf to orchestrate behaviors like these (CRD upgrades), as well as the constraints that require the action for "migrating" release resources.

src/velero/common/zarf.yaml
```diff
@@ -45,3 +45,49 @@ components:
...
+          - https://raw.githubusercontent.com/vmware-tanzu/helm-charts/ee1f0d5549f928bd27863b57b6263c79a3211c73/charts/velero/crds/serverstatusrequests.velero.io.yaml
+          - https://raw.githubusercontent.com/vmware-tanzu/helm-charts/ee1f0d5549f928bd27863b57b6263c79a3211c73/charts/velero/crds/volumesnapshotlocations.velero.io.yaml
+    actions:
+      onDeploy:
+        before:
+          - description: Handle Helm ownership metadata change on upgrade
+            cmd: |
+              echo "Patching Velero CRDs with Helm ownership metadata to prevent adoption errors..."
+              RELEASE_NAME="velero"
+              RELEASE_NAMESPACE="velero"
```
mjnagel: I'm surprised this is the expected name/namespace 🤔. Zarf generates a hash for manifests release-names based on the package, component, and manifest name so I would expect it to be random? And we aren't setting the namespace for the manifests as well from what I can see. It seems to work in CI on the upgrade, but I think that's just because of the missing imports a layer+ up?

mjnagel (changes requested):

mjnagel:
> Making a mental note on the utility of Zarf to orchestrate behaviors like these (CRD upgrades), as well as the constraints that require the action for "migrating" release resources.

@brandtkeller I had opened [this issue](https://github.com/zarf-dev/zarf/issues/2555) a while back, which I think is still valid/could be useful in some cases - added a few comments there hopefully explaining the behavior around CRD upgrades in Helm a bit better.

mjnagel (dismissed):

src/velero/common/zarf.yaml
```diff
@@ -30,6 +30,52 @@ variables:
...
+          - https://raw.githubusercontent.com/vmware-tanzu/helm-charts/refs/tags/velero-11.0.0/charts/velero/crds/restores.yaml
+          - https://raw.githubusercontent.com/vmware-tanzu/helm-charts/refs/tags/velero-11.0.0/charts/velero/crds/schedules.yaml
+          - https://raw.githubusercontent.com/vmware-tanzu/helm-charts/refs/tags/velero-11.0.0/charts/velero/crds/serverstatusrequests.yaml
+          - https://raw.githubusercontent.com/vmware-tanzu/helm-charts/refs/tags/velero-11.0.0/charts/velero/crds/volumesnapshotlocations.yaml
+    actions:
+      onDeploy:
+        before:
+          - description: Handle Helm ownership metadata change on upgrade
+            cmd: |
+              RELEASE_NAME="zarf-06ee059327936135c2152d44040c9f947f933034"
```
mjnagel: Adding a comment so we don't forget - this hash is dependent on the package name (see [zarf code](https://github.com/zarf-dev/zarf/blob/main/src/internal/packager/helm/common.go#L38-L42)) so it changes depending on whether Velero is deployed from the backup-restore layer or the standard package.

chance-coleman: decided to define zarf constants at the package level and pass that into the manifest annotation action to be used to avoid ownership issues.

src/velero/common/zarf.yaml
```diff
@@ -30,6 +30,52 @@ variables:
...
+                podvolumebackups.velero.io \
+                podvolumerestores.velero.io \
+                restores.velero.io \
+                schedules.velero.io \
+                serverstatusrequests.velero.io \
+                volumesnapshotlocations.velero.io; do
+
+                kubectl label crd "$crd" app.kubernetes.io/managed-by=Helm --overwrite || true
+                kubectl annotate crd "$crd" meta.helm.sh/release-name="$RELEASE_NAME" --overwrite || true
+                kubectl annotate crd "$crd" meta.helm.sh/release-namespace="$RELEASE_NAMESPACE" --overwrite || true
```
mjnagel: One small nit here - on an install this is pretty spammy in the logs. Might recommend suppressing stderr:

```suggestion
                kubectl label crd "$crd" app.kubernetes.io/managed-by=Helm --overwrite 2>/dev/null || true
                kubectl annotate crd "$crd" meta.helm.sh/release-name="$RELEASE_NAME" --overwrite 2>/dev/null || true
                kubectl annotate crd "$crd" meta.helm.sh/release-namespace="$RELEASE_NAMESPACE" --overwrite 2>/dev/null || true
```

mjnagel (dismissed):
LGTM, validated all 3 paths. While I don't love having to update helm ownership this is the cleanest approach for now and removes one image required (smaller core package 🎉).

mjnagel (approved):

src/velero/common/zarf.yaml
slaskawi: Yes, the Operator Lifecycle Manager (OLM for OpenShift) does it exactly the same way - CRDs first.

PR #2029, opened 2025-10-15
fix: velero crd helm ownership

joelmccoy:
## Description

Fixes the Velero CRD Helm ownership zarf action.  This was failing if you didn't have `kubectl` on your path.

This PR changes the logic so that:
1. `./zarf tools kubectl` is used
2. Fail if the label and annotate commands fail

## Type of change

## Steps to Validate
- Deploy UDS core 0.53.1 `uds deploy k3d-core-demo:0.53.1 --confirm`
- Deploy the standard bundle (core package only) on top of this
- Ensure CRDs are owned properly by the helm chart 

## Checklist before merging

### DIFF
--- a/src/velero/common/zarf.yaml
+++ b/src/velero/common/zarf.yaml
@@ -71,9 +71,15 @@ components:
                 serverstatusrequests.velero.io \
                 volumesnapshotlocations.velero.io; do
 
-                kubectl label crd "$crd" app.kubernetes.io/managed-by=Helm --overwrite 2>/dev/null || true
-                kubectl annotate crd "$crd" meta.helm.sh/release-name="$RELEASE_NAME" --overwrite 2>/dev/null || true
-                kubectl annotate crd "$crd" meta.helm.sh/release-namespace="$RELEASE_NAMESPACE" --overwrite 2>/dev/null || true
+                # Only apply the labels/annotations if the CRD exists and is not already managed by Helm
+                if ./zarf tools kubectl get crd "$crd" >/dev/null 2>&1; then
+                  MANAGED_BY=$(./zarf tools kubectl get crd "$crd" -o jsonpath='{.metadata.labels.app\.kubernetes\.io/managed-by}' 2>/dev/null)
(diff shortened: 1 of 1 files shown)


src/velero/common/zarf.yaml
```diff
@@ -71,9 +71,15 @@ components:
...
-                kubectl annotate crd "$crd" meta.helm.sh/release-name="$RELEASE_NAME" --overwrite 2>/dev/null || true
-                kubectl annotate crd "$crd" meta.helm.sh/release-namespace="$RELEASE_NAMESPACE" --overwrite 2>/dev/null || true
+                # Only apply the labels/annotations if the CRD exists and is not already managed by Helm
+                if ./zarf tools kubectl get crd "$crd" >/dev/null 2>&1; then
+                  MANAGED_BY=$(./zarf tools kubectl get crd "$crd" -o jsonpath='{.metadata.labels.app\.kubernetes\.io/managed-by}' 2>/dev/null)
+                  if [ "$MANAGED_BY" != "Helm" ]; then
+                    ./zarf tools kubectl label crd "$crd" app.kubernetes.io/managed-by=Helm --overwrite
+                    ./zarf tools kubectl annotate crd "$crd" meta.helm.sh/release-name="$RELEASE_NAME" --overwrite
+                    ./zarf tools kubectl annotate crd "$crd" meta.helm.sh/release-namespace="$RELEASE_NAMESPACE" --overwrite
+                  fi
```
mjnagel: I might advocate for unconditionally doing this part (conditional on existence but not on managed-by). Since the ownership (release name) could change depending on how you deploy (func layer vs full core) we might want to always update/"fix" these labels? At least until upstream adopts a better crd upgrade strategy.

mjnagel: To be more specific - if someone switched from deploying full core to deploying the func layers of core, this would be needed to "migrate" the release-name.

joelmccoy: makes sense. adjusted

mjnagel (approved):

cmwylie19 (approved):

PR #2030, opened 2025-10-15
fix: keycloak statefulset spi provider syntax

chance-coleman:
## Description
In a previous keycloak update they updated their spi statefulset reference syntax, it now needs to be double underscores. At the moment this is just a warning error that happens when not using the correct syntax.

`WARNING: The following SPI options are using the legacy format and are not being treated as build time options. Please use the new format with the appropriate -- separators to resolve this ambiguity: kc.spi-x509cert-lookup-provider`

## Type of change

## Steps to Validate
- `uds run test-single-layer --set LAYER=identity-authorization` from main and monitoring the logs of keycloak pod as it comes up to see the warning
- run the same command from my branch and see that the warning is no longer present.

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -166,13 +166,13 @@ spec:
               value: "true"
 
             ## Activate the nginx provider
-            - name: KC_SPI_X509CERT_LOOKUP_PROVIDER
+            - name: KC_SPI__X509CERT_LOOKUP__PROVIDER
               value: {{ .Values.x509LookupProvider }}
             # Set nginx provider header name
-            - name: KC_SPI_X509CERT_LOOKUP_{{ .Values.x509LookupProvider | upper }}_SSL_CLIENT_CERT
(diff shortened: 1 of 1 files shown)


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -166,13 +166,13 @@ spec:
               value: "true"
 
             ## Activate the nginx provider
-            - name: KC_SPI_X509CERT_LOOKUP_PROVIDER
+            - name: KC_SPI__X509CERT_LOOKUP__PROVIDER
```
slaskawi, line 169: According to the code, only this one causes troubles (see [here](https://github.com/keycloak/keycloak/blob/3724409c5ee57bbebf36ed745c82855a731942cd/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java#L100)). 

Even though other options should not trigger this warning, aligning them is beneficial from consistency perspective. 

Could we also open up a follow up PR and adjust other env. variable names?

slaskawi (approved):

PR #2041, opened 2025-10-17
fix: keycloak spi envs syntax

chance-coleman:
## Description
Updating the Keycloak spi envs to use the newer syntax. I was able to search the official docs and find various places these envs are referenced, unfortunately there isn't somewhere that references all of them in one place.

[Both the code](https://github.com/keycloak/keycloak/blob/3724409c5ee57bbebf36ed745c82855a731942cd/quarkus/runtime/src/main/java/org/keycloak/quarkus/runtime/configuration/mappers/PropertyMappers.java#L100) and the docs are very ambiguous with this change.

docs:
[provider spi envs](https://www.keycloak.org/server/all-provider-config)
[x509 lookup spi envs](https://www.keycloak.org/server/reverseproxy#_enabling_client_certificate_lookup)

## Type of change

## Steps to Validate
- Verified this by running locally both:
  - `uds run test-single-layer --set LAYER=identity-authorization --set FLAVOR=unicorn`
  - `uds run -f tasks/test.yaml uds-core-e2e --set FLAVOR=unicorn`
- Additionally verified in identity-config that i was able to run both:
  - `uds run uds-core-integration-tests --set FLAVOR=unicorn`
  - `uds run uds-core-smoke-test --set FLAVOR=unicorn`

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -154,11 +154,11 @@ spec:
               value: "true"
             # @lulaStart e4ea044c-75fc-4acc-a552-8bba2aab1b12
             {{- if .Values.detailedObservability.logging.enabled }}
-            - name: KC_SPI_EVENTS_LISTENER_JBOSS_LOGGING_SUCCESS_LEVEL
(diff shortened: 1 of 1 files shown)


mjnagel (approved):
Tested with identity-config, all looks good.

PR #2044, opened 2025-10-17
chore: add/update lula mapping keycloak observability

joelmccoy:
## Description

Adds a lula mapping and updates the position of an existing one.

## Type of change

## Checklist before merging


src/keycloak/chart/values.yaml
```diff
@@ -411,12 +411,12 @@ extraVolumeMounts: []
 truststorePaths: []
 
 # Detailed observability settings (Dashboards, Alerts, etc)
+# @lulaStart e4ea044c-75fc-4acc-a552-8bba2aab1b12
```
joelmccoy, line 414: note: moved this to include the whole detailed observability block. think any relevant changes here would make us want to look at the controls we map this to

cmwylie19 (approved):
yep!

PR #2047, opened 2025-10-20
fix: add script for checking unicorn multiarch images

chance-coleman:
## Description
Dynamically grab all unicorn images, based on `quay.io/rfcurated`, create a list of those images, then iterate over those images and do a crane manifest against each image. Keep track of each image that doesn't have both an `arm64` and `amd64` published architecture. If any images don't have both architectures, `exit 1` at the end of the script. Script automatically cleans up tmp files with the `trap` functionality.

Screenshot from 2025-10-20 10-16-45

## Related Issue

Fixes #

Relates to #

## Type of change

## Steps to Validate

## Checklist before merging


scripts/check-multiarch.sh
```diff
@@ -0,0 +1,71 @@
...
+
+# Create temporary files and ensure cleanup no matter script outcome
+TEMP_IMAGES=$(mktemp)
+TEMP_MISSING=$(mktemp)
+trap "rm -f $TEMP_IMAGES $TEMP_MISSING" EXIT
+
+echo "Scanning $SRC_DIR for unicorn images..."
+
+# Extract all quay.io/rfcurated/ images
+find "$SRC_DIR" -name "zarf.yaml" -type f -exec grep -h "quay.io/rfcurated/" {} \; | \
```
mjnagel: I know this isn't ready for review yet, but might be good to support registry1 in here also. While I don't think we've had them yank any arm images out from under us (yet), would be nice to be ready 😄

chance-coleman: added registry1 support 👍🏼

mjnagel (dismissed):

mjnagel (approved):

PR #2057, opened 2025-10-22
feat: pod reload on configmaps and falco separated configmaps

### LINKED ISSUE #1972: Falco Configmap Rework
At the moment all rulesets are combined into a single configmap. there is potential for that configmap to grow over time and eventually become to large. We should break each ruleset out into its own configmap.

At this time if rulesets update we have no way to cycle falco to make sure it uses the latest rulesets. We should address this, one solution is extending the existing functionality of the [secret-pod-reload functionality](https://uds.defenseunicorns.com/reference/deployment/secret-pod-reload/) to work with configmaps as well.

chance-coleman:
## Description
Previously all Falco configmaps for rules were combined into a single configmap, this raised concerns that over time as those rules change the configmap could become too large. This PR will separate each ruleset out into its own configmap.

Additionally, when Falco ruleset changes and we update a configmap we want our pods to reload so that Falco starts using that new configmap. Previously uds core only supported pod reloading on secret changes, this PR will expand that functionality to work with configmaps as well as secrets.

## Related Issue

Fixes #1972

## Type of change

## Checklist before merging


docs/reference/deployment/resource-pod-reload.md
```diff
@@ -1,14 +1,14 @@
 ---
-title: Secret Pod Reload
+title: Resource Pod Reload
```
mjnagel: Let's make sure to add a doc redirect to the docs site since this one has been linked in release notes, etc. I also wonder if it makes more sense to just call it "pod reloading"?

docs/reference/deployment/resource-pod-reload.md
```diff
@@ -1,14 +1,14 @@
...
 tableOfContents:
   maxHeadingLevel: 4
 ---
 
-UDS Core provides a Secret Pod Reload mechanism that automatically reloads pods when secrets are updated. This feature is particularly useful for credentials, certificates, and configuration data that applications need to refresh without manual intervention.
+UDS Core provides a Resource Pod Reload mechanism that automatically reloads pods when secrets or ConfigMaps are updated. This feature is particularly useful for credentials, certificates, and configuration data that applications need to refresh without manual intervention.
```
mjnagel: secrets/ConfigMaps have inconsistent capitalization here. This and my other suggestions will also take into account just changing to "pod reload" rather than "resource pod reload".


```suggestion
UDS Core provides a Pod Reload mechanism that automatically reloads pods when Secrets or ConfigMaps are updated. This feature is particularly useful for credentials, certificates, and configuration data that applications need to refresh without manual intervention.
```

docs/reference/deployment/resource-pod-reload.md
```diff
@@ -1,14 +1,14 @@
...
 
 ## How It Works
 
-The secret pod reload controller watches for changes to Kubernetes secrets labeled with `uds.dev/pod-reload: "true"`. When such a secret is updated, the controller will:
+The resource pod reload controller watches for changes to Kubernetes secrets and ConfigMaps labeled with `uds.dev/pod-reload: "true"`. When such a resource is updated, the controller will:
```
mjnagel:
```suggestion
The pod reload controller watches for changes to Kubernetes Secrets and ConfigMaps labeled with `uds.dev/pod-reload: "true"`. When such a resource is updated, the controller will:
```

docs/reference/deployment/resource-pod-reload.md
```diff
@@ -17,9 +17,9 @@ The secret pod reload controller watches for changes to Kubernetes secrets label
 
 ## Configuration
 
-### Enabling Secret Pod Reload
+### Enabling Resource Pod Reload
 
-To enable automatic pod reload when a secret changes, add the following label to your secret:
+To enable automatic pod reload when a secret or ConfigMap changes, add the following label to your resource:
```
mjnagel:
```suggestion
### Enabling Pod Reload

To enable automatic pod reload when a Secret or ConfigMap changes, add the following label to your resource:
```

docs/reference/deployment/resource-pod-reload.md
```diff
@@ -41,18 +41,20 @@ metadata:
...
 
-This will restart all pods matching the specified label selector when the secret changes.
+This will restart all pods matching the specified label selector when the resource changes.
 
 #### 2. Auto-Discovery (Default)
 
-If you don't specify a pod selector, UDS Core will automatically discover and restart pods that are consuming the secret through:
-- Volume mounts (secret volumes)
+If you don't specify a pod selector, UDS Core will automatically discover and restart pods that are consuming the resource through:
+- Volume mounts (secret or ConfigMap volumes)
```
mjnagel:
```suggestion
- Volume mounts (Secret or ConfigMap volumes)
```

docs/reference/deployment/resource-pod-reload.md
```diff
@@ -41,18 +41,20 @@ metadata:
...
 - Environment variables (env or envFrom)
-- Projected volumes using the secret
+- Projected volumes using the secret or ConfigMap
```
mjnagel:
```suggestion
- Projected volumes using the Secret or ConfigMap
```

docs/reference/deployment/resource-pod-reload.md
```diff
@@ -41,18 +41,20 @@ metadata:
...
 
-This auto-discovery approach ensures that only pods actually using the secret are restarted.
+This auto-discovery approach ensures that only pods actually using the resource are restarted.
 
-## Example
+## Examples
+
+### Secret Example
 
 Here's an example of a secret with pod reload enabled:
```
mjnagel:
```suggestion
Here's an example of a Secret with pod reload enabled:
```

docs/reference/deployment/resource-pod-reload.md
```diff
@@ -74,15 +76,38 @@ data:
...
+      output: Sensitive file opened for reading (user=%user.name file=%fd.name)
+      priority: WARNING
+```
+
+When this ConfigMap is updated (for example, when adding new security rules), all pods that mount this ConfigMap will be automatically reloaded to pick up the new configuration.
+
 ## Integration with SSO
 
-The Secret Pod Reload functionality can be used with SSO client secrets. You can enable this by adding the `uds.dev/pod-reload: "true"` label to your SSO client secrets through the `secretLabels` field in your Package CR.
+The Resource Pod Reload functionality can be used with SSO client secrets. You can enable this by adding the `uds.dev/pod-reload: "true"` label to your SSO client secrets through the `secretLabels` field in your Package CR.
```
mjnagel:
```suggestion
The Pod Reload functionality can be used with SSO client secrets. You can enable this by adding the `uds.dev/pod-reload: "true"` label to your SSO client secrets through the `secretLabels` field in your Package CR.
```

docs/reference/deployment/resource-pod-reload.md
```diff
@@ -74,15 +76,38 @@ data:
...
 
 For more details on configuring Secret Pod Reload for SSO clients, see the [Secret Templating documentation](/reference/configuration/single-sign-on/sso-templating#secret-pod-reload).
```
mjnagel:
```suggestion
For more details on configuring Pod Reload for SSO clients, see the [Secret Templating documentation](/reference/configuration/single-sign-on/sso-templating#secret-pod-reload).
```

docs/reference/deployment/pod-reload.md
```diff
@@ -74,15 +76,38 @@ data:
...
 
 For more details on configuring Secret Pod Reload for SSO clients, see the [Secret Templating documentation](/reference/configuration/single-sign-on/sso-templating#secret-pod-reload).
 
 ## Monitoring and Troubleshooting
 
-When a secret is updated and triggers a restart, the controller generates Kubernetes events with the reason `SecretChanged`. You can view these events using:
+When a resource is updated and triggers a restart, the controller generates Kubernetes events with the reason `SecretChanged` or `ConfigMapChanged`. You can view these events using:
 
 ```bash
 kubectl get events -n <namespace> --field-selector reason=SecretChanged
```
mjnagel, line 113:
```suggestion
kubectl get events -n <namespace> --field-selector reason=SecretChanged
kubectl get events -n <namespace> --field-selector reason=ConfigMapChanged
```

src/falco/chart/templates/rules-configmap.yaml
```diff
@@ -4,41 +4,73 @@
...
+  labels:
+    uds.dev/pod-reload: "true"
+data:
   incubating-rules.yaml: |-
+{{- if .Values.incubatingRulesEnabled }}
 {{ .Files.Get "rules/incubating-rules.yaml" | nindent 4 }}
-  {{- end }}
+{{- else }}
+    # Incubating rules are disabled by default
+    - required_engine_version: 0
```
mjnagel, line 29: For my own understanding - what does this flag do? I see this changed from being null/empty to including this value (and the tests changed accordingly).

mjnagel (commented):
Haven't gotten a chance to deploy locally but the changes to support configmaps look good from a review standpoint. Some other comments on docs mostly.

src/falco/chart/templates/rules-configmap.yaml
chance-coleman: It’s a  Falco "directive", `required_engine_version: 0` tells Falco this file requires at least engine version 0, which is always satisfied, so it parses cleanly but does nothing.

docs/reference/deployment/resource-pod-reload.md
chance-coleman: Yeah tracking the redirect, put [up a PR for that](https://github.com/defenseunicorns/uds-docs/pull/306). I like the idea of calling the doc Pod Reload, simpler, more straightforward.

src/falco/chart/templates/rules-configmap.yaml
mjnagel: Gotcha - thanks. That makes sense!

docs/reference/deployment/resource-pod-reload.md
mjnagel: Let's rename the file to match, which should make the docs PR correct 😄

docs/reference/deployment/resource-pod-reload.md
joelmccoy: I think you can rename this file to pod-reload too based on the other feedback?

joelmccoy: also it would be nice if we did a `git mv` on the file so we get cleaner diffs of the file in the PR review

src/pepr/operator/controllers/reload/pod-reload.ts
joelmccoy, line 1: are we able to get this file as a `git mv` so we can see the delta of the actual file change in this PR?

docs/reference/deployment/pod-reload.md
joelmccoy, line 1: are we able to get this file as a `git mv` so we can see the delta of the actual file change in this PR?

src/pepr/operator/index.ts
```diff
@@ -143,6 +148,18 @@ When(a.Secret)
...
+  .IsCreatedOrUpdated()
+  .WithLabel("uds.dev/pod-reload", "true")
+  .Reconcile(handleConfigMapUpdate);
+
+// Watch for deleted ConfigMaps to clean up the checksum cache
+When(a.ConfigMap)
+  .IsDeleted()
+  .WithLabel("uds.dev/pod-reload", "true")
+  .Reconcile(handleConfigMapDelete);
+
```
mjnagel, line 162: One other thing to consider - might be nice to have an e2e test of this (we have one for secrets in `test/vitest/pod-reload.spec.ts`)?

src/pepr/operator/controllers/reload/pod-reload.ts
```diff
@@ -0,0 +1,337 @@
...
+  checksumCache: Map<string, string>,
+  discoverResourceConsumers: (namespace: string, name: string) => Promise<kind.Pod[]>,
+  resourceType: string,
+) {
+  if (!resource.metadata?.name || !resource.metadata?.namespace) {
+    return;
+  }
+
+  const { name, namespace } = resource.metadata;
+  const cacheKey = `${namespace}/${name}`;
```
joelmccoy, line 167: could we have a collision with this cache key if we have a secret and a configmap with the same name in the same namespace? do we want to add whether it is a secret or configmap to this cache key?

src/pepr/operator/controllers/reload/pod-reload.ts
```diff
@@ -11,17 +11,21 @@ import { reloadPods } from "./reload-utils";
 
 const log = setupLogger(Component.OPERATOR_SECRETS);
 
-// Map to store secret checksums for change detection
+// Define resource types
+export type ResourceType = "Secret" | "ConfigMap";
```
joelmccoy, line 15: I don't think we actually use this type in the function signatures. will add comments to where this needs fixed.

src/pepr/operator/controllers/secrets/pod-reload.ts
```diff
@@ -103,51 +107,104 @@ export function parseSelectorString(value: string): Record<string, string> | nul
...
+ * @param resource The Kubernetes resource that was updated
+ * @param checksumCache The cache to use for this resource type
+ * @param discoverResourceConsumers Function to discover pods using this resource
+ * @param resourceType Type of resource ("Secret" or "ConfigMap")
+ */
+export async function handleResourceUpdate(
+  resource: kind.Secret | kind.ConfigMap,
+  checksumCache: Map<string, string>,
+  discoverResourceConsumers: (namespace: string, name: string) => Promise<kind.Pod[]>,
+  resourceType: string,
```
joelmccoy:
```suggestion
  resourceType: ResourceType,
```

src/pepr/operator/controllers/secrets/pod-reload.ts
```diff
@@ -165,61 +222,116 @@ export async function handleSecretUpdate(secret: kind.Secret) {
...
     return;
   }
 }
 
 /**
- * Handles a secret deletion event
+ * Generic function to handle resource deletion (Secret or ConfigMap)
  *
- * @param secret The Kubernetes secret that was deleted
+ * @param resourceType Type of resource ("Secret" or "ConfigMap")
```
joelmccoy: i don't think this documented param is actually there?

src/pepr/operator/controllers/reload/reload-utils.ts
```diff
@@ -22,8 +22,15 @@ import { createEvent, retryWithDelay } from "../utils";
  * @param pods List of pods to evict or restart
  * @param reason The reason for eviction/restart (for logging)
  * @param log Logger instance for logging
+ * @param resource Resource type responsible for eviction/restart (for logging)
```
joelmccoy: do we really need this extra parameter here?  Couldn't the parameter `reason` include the necessary data for logging?

src/pepr/operator/controllers/reload/reload-utils.ts
```diff
@@ -22,8 +22,15 @@ import { createEvent, retryWithDelay } from "../utils";
...
  */
-export async function reloadPods(namespace: string, pods: kind.Pod[], reason: string, log: Logger) {
+export async function reloadPods(
+  namespace: string,
+  pods: kind.Pod[],
+  reason: string,
+  log: Logger,
+  resource: string,
```
joelmccoy: If we do want to keep this parameter around (see above comment) it should be called `resourceType` and use the appropriate typing.  But this makes this function specific to only being used by ConfigMap and Secret reloads.

```suggestion
  resourceType: ResourceType,
```

joelmccoy (commented):
This is generally looking good to me and tested locally. A couple small comments and questions.

src/pepr/operator/controllers/reload/reload-utils.ts
joelmccoy: If we want to keep this function open to be used as a util for pod reloads (and not specific to just configmap and secret changes), I think we should remove this parameter.

src/pepr/operator/controllers/reload/pod-reload.ts
chance-coleman: the string value of cacheKey would be identical ("${namespace}/${name}") for a Secret and a ConfigMap with the same name in the same namespace. However, there wouldn't be a collision because they are stored in different caches, which are passed into the `handleResourceUpdate` and `handleResourceDelete` functions.

joelmccoy: ah. i see this now. thanks for the clarification.

src/pepr/operator/controllers/reload/reload-utils.ts
chance-coleman: Hmmm definitely see your point, i had passed this `resource` through so that the `restartController` function could accurate describe the [resource type](https://github.com/defenseunicorns/uds-core/blob/948a73a9ff679c5f5b9aa4e99b13c4e514e0069e/src/pepr/operator/controllers/secrets/reload-utils.ts#L205) that was changing.. I could just pass that type via hardcoded string since its [already in if statement logic](https://github.com/defenseunicorns/uds-core/blob/948a73a9ff679c5f5b9aa4e99b13c4e514e0069e/src/pepr/operator/controllers/secrets/reload-utils.ts#L69-L81) or i could just say generically that the `Resource Changed` because the [reason should include the resource type](https://github.com/defenseunicorns/uds-core/blob/falco-configmap-rework/src/pepr/operator/controllers/reload/pod-reload.ts#L266). I'd probably vote for the later solution.

mjnagel: Ultimately we want two pieces of information here (I think):
- The kind of resource that caused the change (secret v configmap)
- The actual name of the resource that caused it

And then we need to fill two fields in the event spec ([here](https://github.com/defenseunicorns/uds-core/blob/falco-configmap-rework/src/pepr/operator/controllers/reload/reload-utils.ts#L242-L243)):
- reason: why, when this was just secrets we had this as `SecretChanged` (this is limited to 128 chars)
- message: longer text description, in the past this is where we put the "reason" with the full resource name.

I could see some options here:
- Just change the event reason to `ResourceChanged`, and add the kind to the message (i.e. configmap/secret). This would be fine if we wanted to - less specific reason meaning you could find everything in one lookup I suppose. As long as the message (reason in this function) contains the kind this would OK, just don't want to lose that.
- Keep as-is, maintains all the info, but does mean we're passing around the kind a bit. Also doesn't keep the event logic generic.
- Pass an actual full "reason" to restartController and the event func - right now we're just passing `Secret` and `Configmap` (then appending `Changed`), but if we really want this to be generic we could actually pass in `SecretChanged` and `ConfigmapChanged`. This would also allow other uses to pass in `IstioInjection`, etc.

> If we want to keep this function open to be used as a util for pod reloads

For what its worth, this was the intent but the way events are created (previously with [SecretChanged](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/secrets/reload-utils.ts#L205)) I don't think it has been generic in the past either.

src/pepr/operator/index.ts
chance-coleman: added some tests that resemble the same tests the secrets have

src/pepr/operator/controllers/reload/reload-utils.ts
chance-coleman: okay reworked this and implemented Micah's third bullet to utilize `reason` and `message` according the event type keys. Moved the `resourceType + 'Changed'` at the event creation time out of that function to make it more dynamic/generic. Tests needed a bit of rework for syntax changes.

src/pepr/operator/controllers/reload/pod-reload.ts
```diff
@@ -278,7 +284,6 @@ export async function handleResourceUpdate(
  *
  * @param resourceType Type of resource ("Secret" or "ConfigMap")
  * @param resource The Kubernetes resource that was deleted
- * @param checksumCache The cache to use for this resource type
```
mjnagel: I think the wrong param comment got deleted here? Should still have `checksumCache`, but not `resourceType`.

src/pepr/operator/controllers/reload/reload-utils.ts
mjnagel: This LGTM.

mjnagel (dismissed):
LGTM with one small comment on a comment.

mjnagel (approved):

joelmccoy (approved):

PR #2060, opened 2025-10-22
fix: local velero crd manifests

chance-coleman:
## Description
Moves Velero CRDs into uds core manifests instead of pulling them from upstream. This allows us avoid helm permissions issues as well as a simplified deployment of velero.

Removed helm permission additions and velero-crds component.
Added `update-crds` task to `src/velero/tasks.yaml` that will pull velero crds directly from upstream and create manifest for them as well as license header. This task also has a renovate tracking the version of the  upstream charts. Autogenerated inclusion for making sure that updates are regenerated.

## Type of change

## Checklist before merging


src/velero/common/zarf.yaml
```diff
@@ -94,3 +45,16 @@ components:
         releaseName: velero
         valuesFiles:
           - ../values/values.yaml
+    actions:
+      onDeploy:
+        before:
+          - description: Handle Helm ownership metadata change on upgrade
+            cmd: |
+              # Update Helm ownership metadata for existing Velero CRDs
+              for crd in $(./zarf tools kubectl get crd -o name | grep velero.io); do
```
mjnagel: I think this is probably fine? But two considerations:
1. `grep`...should be on every machine?
2. Are there other `velero.io` CRDs we wouldn't have in this chart?

src/velero/tasks.yaml
```diff
@@ -104,3 +109,34 @@ tasks:
...
+          if [ -z "$CRD_FILES" ]; then
+            echo "Error: Could not fetch CRD list from upstream"
+            exit 1
+          fi
+
+          # Download each CRD
+          for crd_file in $CRD_FILES; do
+            echo "Downloading ${crd_file}"
+            curl -sSL "${BASE_URL}/${crd_file}" -o "${CHART_DIR}/crds/${crd_file}"
+            sed -i "1i# Copyright 2025 Defense Unicorns\n# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial\n" "${CHART_DIR}/crds/${crd_file}"
```
mjnagel: Should we just run the license task here? That way we aren't hardcoding the license/year separate from that task.

src/velero/tasks.yaml
```diff
@@ -104,3 +109,34 @@ tasks:
   - name: e2e-test
     actions:
       - description: "Run Velero E2E tests"
+
+  - name: update-crds
+    description: "Update Velero CRDs from upstream"
+    actions:
+      - description: "Download Velero CRDs from upstream"
+        cmd: |-
+          bash -c '
```
mjnagel: We should be able to set the shell to bash in the task to avoid this wrapper. I think we do that somewhere else?

src/velero/common/zarf.yaml
chance-coleman: updated to use kubectl label selector instead of grep. This will be removed in not so distant future release so for now not doing anything about getting all `velero.io` crds

src/velero/tasks.yaml
chance-coleman: yeah, switched over to use the fix-license uds-common task

src/velero/tasks.yaml
chance-coleman: changed this to use the `shell: linux: bash` syntax

src/velero/tasks.yaml
```diff
@@ -104,3 +112,34 @@ tasks:
...
+          linux: bash
+        cmd: |
+          CHART_DIR="src/velero/chart/templates"
+          BASE_URL="https://raw.githubusercontent.com/vmware-tanzu/helm-charts/refs/tags/velero-${VELERO_CRD_VERSION}/charts/velero/crds"
+
+          echo "Updating Velero CRDs to version ${VELERO_CRD_VERSION}"
+
+          # Fetch list of CRD files from GitHub API
+          echo "Fetching CRD list from upstream..."
+          CRD_FILES=$(curl -sSL "https://api.github.com/repos/vmware-tanzu/helm-charts/contents/charts/velero/crds?ref=velero-${VELERO_CRD_VERSION}" | grep -o "\"name\": \"[^\"]*\.yaml\"" | cut -d\" -f4)
```
mjnagel: This is probably fine - but since it's returning json we could do something like:
```
curl -sSL "https://api.github.com/repos/vmware-tanzu/helm-charts/contents/charts/velero/crds?ref=velero-11.1.0" | uds zarf tools yq '.[].name'
```

Instead of grep/cut. Feels a bit less clunky (could still grep for `.yaml` if you want after).

mjnagel (dismissed):

src/velero/tasks.yaml
chance-coleman: i like that. updated.

mjnagel (approved):

PR #2062, opened 2025-10-23
fix: renovate readiness check unicorn pepr image version

chance-coleman:
## Description
The current implementation only checks the REGISTRY1_PEPR_IMAGE version against package.json, only validates that registry1 matches package.json, but doesn't check if unicorn is also in sync.

I've updated the Pepr version checking logic in the renovate readiness action to validate that all three Pepr versions are in sync.

## Type of change

## Checklist before merging

### DIFF
--- a/.github/actions/renovate-readiness/action.yaml
+++ b/.github/actions/renovate-readiness/action.yaml
@@ -87,31 +87,62 @@ runs:
         echo "Unicorn image version: $UNICORN_IMAGE_VERSION"
 
         # Check if all three versions are in sync
-        VERSIONS_MATCH=true
+        IRONBANK_MISMATCH=false
+        UNICORN_MISMATCH=false
+        UPSTREAM_BEHIND=false
+
         if [[ "$PEPR_VERSION" != "$IRONBANK_IMAGE_VERSION" ]]; then
           echo "ERROR: package.json version ($PEPR_VERSION) does not match Ironbank image version ($IRONBANK_IMAGE_VERSION)"
-          VERSIONS_MATCH=false
+          # Check if package.json version is less than image version (upstream is behind)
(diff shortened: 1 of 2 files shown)


.github/actions/renovate-readiness/action.yaml
```diff
@@ -82,14 +82,35 @@ runs:
...
+        if [[ "$IRONBANK_IMAGE_VERSION" != "$UNICORN_IMAGE_VERSION" ]]; then
+          echo "ERROR: Ironbank image version ($IRONBANK_IMAGE_VERSION) does not match Unicorn image version ($UNICORN_IMAGE_VERSION)"
+          VERSIONS_MATCH=false
+        fi
+
+        # Apply labels based on version sync status
+        if [[ "$VERSIONS_MATCH" == "false" ]]; then
+          echo "Versions are not in sync. Waiting on image updates."
           gh pr edit ${{ github.event.pull_request.number }} --remove-label "needs-review" || true
           gh pr edit ${{ github.event.pull_request.number }} --add-label "waiting on ironbank"
```
mjnagel: Potentially this label should be something different depending on the comparison? Or maybe just a generic "waiting" label 😅. As we chatted about...in theory unicorn/upstream are always in sync unless there are publish issues (which do happen). Soo previously based on the check here `waiting on ironbank` was the only possible "wait". But now it could be waiting on unicorn or even upstream 🙃.

chance-coleman: added the logic for checking upstream version being behind unicorn/ironbank. adds a bit more complexity.

mjnagel (approved):

.github/actions/renovate-readiness/action.yaml
mjnagel: LGTM - it is more complex...assuming it doesn't explode I'm not too concerned. Should be mostly set and forget hopefully.

PR #2066, opened 2025-10-24
fix: add retries to all network tests

chance-coleman:
## Description
remove the network tests google curl's retries and bump the curl timeout down to 10 second. Add 2 retry attempts to all tests.

## Type of change

## Checklist before merging

### DIFF
--- a/test/vitest/network.spec.ts
+++ b/test/vitest/network.spec.ts
@@ -161,7 +161,7 @@ beforeAll(async () => {
   }
 });
 
-describe("Network Policy Validation", () => {
+describe("Network Policy Validation", { retry: 2 }, () => {
   const INTERNAL_CURL_COMMAND_1 = getCurlCommand("curl-pkg-deny-all-2", "curl-ns-deny-all-2");
   const INTERNAL_CURL_COMMAND_2 = getCurlCommand("curl-pkg-allow-all", "curl-ns-allow-all");
   const INTERNAL_CURL_COMMAND_5 = getCurlCommand(
@@ -174,11 +174,7 @@ describe("Network Policy Validation", () => {
     "curl",
     "-s",
     "-m",
-    "20",
-    "--retry",
-    "3",
-    "--retry-delay",
-    "1",
+    "10",
     "-o",
     "/dev/null",
     "-w",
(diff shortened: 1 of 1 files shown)


test/vitest/network.spec.ts
```diff
@@ -333,29 +284,6 @@ describe("Network Policy Validation", () => {
...
-      "curl-ns-remote-ns-2",
-      9999,
-    );
-    const denied_port_response = await execInPod(
-      "curl-ns-remote-ns-1",
-      curlPodName6,
-      "curl-pkg-remote-ns-egress",
-      blocked_port_curl,
-    );
-    expect(isResponseError(denied_port_response)).toBe(true);
```
mjnagel, line 358: Feel like we might want to keep one negative case?

test/vitest/network.spec.ts
```diff
@@ -373,38 +301,9 @@ describe("Network Policy Validation", () => {
...
-
-    // Default Deny for Blocked Port
-    const blocked_port_curl = getCurlCommand("curl-pkg-deny-all-2", "curl-ns-deny-all-2", 9999);
-    const denied_port_response = await execInPod(
-      "curl-ns-kube-api",
-      curlPodName8,
-      "curl-pkg-kube-api",
-      blocked_port_curl,
-    );
-    expect(isResponseError(denied_port_response)).toBe(true);
```
mjnagel, line 394: Same here maybe, at least keep 1 negative case?

test/vitest/network.spec.ts
```diff
@@ -373,38 +301,9 @@ describe("Network Policy Validation", () => {
...
-    // Validate successful request when using RemoteCidr
-    const success_response = await execInPod(
-      "test-admin-app",
-      testAdminApp,
-      "curl",
-      INTERNAL_CURL_COMMAND_7,
-    );
-    expect(success_response.stdout).toBe("200");
-
-    // Validate successful request to Google because of wide open remoteCidr
```
mjnagel, line 407: Between the two of these, hitting something internal to the cluster might be better? Avoid external dependencies? (i.e. keep the internal one, remove google?)

test/vitest/network.spec.ts
mjnagel, line 267: Of these two I'd probably keep the one with the header. That just ensures we don't let any random auth header through.

chance-coleman: reverted this change and used the header curl

test/vitest/network.spec.ts
chance-coleman: removed google request and kept internal

mjnagel (approved):

PR #2067, opened 2025-10-24
chore: bumb version references

andrewg-xyz:
## Description

I noticed various version references what were very out of date, this just updates those versions.

It'd be nice to be able to use `x-release-please...` but I didn't find an obvious way to include those annotations and keep them excluded from the published documentation.

## Type of change

## Steps to Validate
N/A

## Checklist before merging

### DIFF
--- a/docs/reference/configuration/service-mesh/ingress.md
+++ b/docs/reference/configuration/service-mesh/ingress.md
@@ -33,7 +33,7 @@ metadata:
 packages:
   - name: core
     repository: oci://ghcr.io/defenseunicorns/packages/uds/core
-    ref: 0.23.0-upstream
+    ref: 0.54.1-upstream
     # You must specify the istio-passthrough-gateway as an optionalComponent or else it will not be deployed
     optionalComponents:
       - istio-passthrough-gateway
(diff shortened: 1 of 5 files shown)


docs/reference/configuration/service-mesh/ingress.md
```diff
@@ -33,7 +33,7 @@ metadata:
 packages:
   - name: core
     repository: oci://ghcr.io/defenseunicorns/packages/uds/core
-    ref: 0.23.0-upstream
+    ref: 0.54.1-upstream
```
mjnagel, line 36: I think in some places we have used `x.x.x` as a version placeholder instead of an explicit version (maybe with a comment of "check available versions <here>". Okay with just bumping all for now as well.

chance-coleman: chatted in slack, going to kick this down the road to when we do the docs revamp. Would like to find a way to maintain the versions better in our docs across bundles and packages. Bundles have a latest tag which is easiest but packages do not.

Created [this issue](https://github.com/defenseunicorns/uds-core/issues/2068) for tracking this work during docs revamp.

chance-coleman (approved):

PR #2081, opened 2025-10-29
chore: make private-pki task macos compatible

joelmccoy:
## Description

This updates the private-pki task macos compatible.  Two things to call out:
* base64 on mac has different flags/syntax (no `-w` flag and requires `-i`) for input
* macos has a different way to trust CAs (keychain)

## Type of change

## Steps to Validate
- run `uds run -f tasks/test.yaml uds-core-private-pki` on macos

## Checklist before merging

### DIFF
--- a/.github/workflows/private-pki-test.yaml
+++ b/.github/workflows/private-pki-test.yaml
@@ -14,7 +14,7 @@ on:
     paths:
       - ".github/workflows/private-pki-test.yaml"
       - "bundles/k3d-standard/**"
-      - "test/private-pki/**"
+      - "test/playwright/private-pki/**"
 
 permissions:
   contents: read
--- a/.github/workflows/private-pki-test.yaml
+++ b/.github/workflows/private-pki-test.yaml
@@ -14,6 +14,7 @@ on:
     paths:
       - ".github/workflows/private-pki-test.yaml"
       - "bundles/k3d-standard/**"
+      - "test/private-pki/**"
 
 permissions:
   contents: read
--- a/src/grafana/tasks.yaml
+++ b/src/grafana/tasks.yaml
@@ -16,24 +16,17 @@ tasks:
             condition: Ready
       - description: Validate grafana virtual service
         cmd: |
-          # Note: this check forces to use /usr/bin/curl if it exists
-          # This is to force using the system curl on macos which is needed for our private-pki tests
-          CURL_CMD="/usr/bin/curl"
-          if [ ! -x "$CURL_CMD" ]; then
-            CURL_CMD="curl"
-          fi
-
-          if [ "$($CURL_CMD -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}')" = "302" ]; then
+          if [ "$(curl -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}')" = "302" ]; then
             echo "Grafana is up and running."
           else
-            echo "ERROR: Grafana returned a $($CURL_CMD -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}') code."
+            echo "ERROR: Grafana returned a $(curl -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}') code."
             exit 1
           fi
 
-          if $CURL_CMD -L -isS https://grafana.admin.uds.dev --output /dev/null -w '%{url_effective}' | grep "sso.uds.dev" 2>&1 1>/dev/null; then
+          if curl -L -isS https://grafana.admin.uds.dev --output /dev/null -w '%{url_effective}' | grep "sso.uds.dev" 2>&1 1>/dev/null; then
             echo "Grafana is redirecting to SSO as expected."
           else
-            echo "ERROR: Grafana is redirecting to $($CURL_CMD -L -isS https://grafana.admin.uds.dev --output /dev/null -w '%{url_effective}')."
+            echo "ERROR: Grafana is redirecting to $(curl -L -isS https://grafana.admin.uds.dev --output /dev/null -w '%{url_effective}')."
             exit 1
           fi
 
--- a/src/grafana/tasks.yaml
+++ b/src/grafana/tasks.yaml
@@ -16,17 +16,24 @@ tasks:
             condition: Ready
       - description: Validate grafana virtual service
         cmd: |
-          if [ "$(curl -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}')" = "302" ]; then
+          # Note: this check forces to use /usr/bin/curl if it exists
+          # This is to force using the system curl on macos which is needed for our private-pki tests
+          CURL_CMD="/usr/bin/curl"
+          if [ ! -x "$CURL_CMD" ]; then
+            CURL_CMD="curl"
+          fi
+
+          if [ "$($CURL_CMD -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}')" = "302" ]; then
             echo "Grafana is up and running."
           else
-            echo "ERROR: Grafana returned a $(curl -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}') code."
+            echo "ERROR: Grafana returned a $($CURL_CMD -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}') code."
             exit 1
           fi
 
-          if curl -L -isS https://grafana.admin.uds.dev --output /dev/null -w '%{url_effective}' | grep "sso.uds.dev" 2>&1 1>/dev/null; then
+          if $CURL_CMD -L -isS https://grafana.admin.uds.dev --output /dev/null -w '%{url_effective}' | grep "sso.uds.dev" 2>&1 1>/dev/null; then
             echo "Grafana is redirecting to SSO as expected."
           else
-            echo "ERROR: Grafana is redirecting to $(curl -L -isS https://grafana.admin.uds.dev --output /dev/null -w '%{url_effective}')."
+            echo "ERROR: Grafana is redirecting to $($CURL_CMD -L -isS https://grafana.admin.uds.dev --output /dev/null -w '%{url_effective}')."
             exit 1
           fi
 
--- a/test/playwright/private-pki/private-pki.test.ts
+++ b/test/playwright/private-pki/private-pki.test.ts
@@ -5,7 +5,7 @@
 
 import { expect, test } from "@playwright/test";
 
-test.describe("Grafana SSO with Private PKI", () => {
+test.describe("Private PKI tests", () => {
   test("Grafana SSO authentication flow", async ({ page }) => {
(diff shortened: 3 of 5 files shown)


chance-coleman (dismissed):
validated this worked locally with linux as well as macos

test/playwright/private-pki/tasks.yaml
mjnagel, line 29: Do we need an equivalent cleanup step here for macOS also?

.github/workflows/private-pki-test.yaml
```diff
@@ -14,6 +14,7 @@ on:
     paths:
       - ".github/workflows/private-pki-test.yaml"
       - "bundles/k3d-standard/**"
+      - "test/private-pki/**"
```
mjnagel:
```suggestion
      - "test/playwright/private-pki/**"
```

If I'm not mistaken?

test/playwright/private-pki/tasks.yaml
mjnagel: I did run through this a second time and it seemed to get past the `security add-trusted-cert` command but looking at my Keychain appears that the old ones are sticking around as well. In testing I think this would work for macOS?
```console
sudo security delete-certificate -c "UDS Test CA" /Library/Keychains/System.keychain
```

The one downside is that this command fails if multiple certs match (which would be the case if you've run this a few times already 😆):
```console
security: "UDS Test CA" is ambiguous, matches more than one certificate
Unable to delete certificate matching "UDS Test CA"
```

But we could `|| true` this with a warning about cleaning up old CAs? Or just fail it and it should work fine if people clean up before hand.

test/playwright/private-pki/tasks.yaml
```diff
@@ -44,16 +44,20 @@ tasks:
...
 
           # Base64 encode CA
-          base64 -w 0 build/private-pki/ca.crt > build/private-pki/ca.crt.b64
+          base64 < build/private-pki/ca.crt | tr -d '\n' > build/private-pki/ca.crt.b64
 
           # Add CA to system trust store
-          sudo cp build/private-pki/ca.crt /usr/local/share/ca-certificates/uds-test-ca.crt
-          sudo update-ca-certificates
+          if [[ "$OSTYPE" == "darwin"* ]]; then
+            sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain build/private-pki/ca.crt
```
mjnagel, line 63: This is probably okay, but poking around the Keychain got me thinking if we could avoid `sudo` by using the user's `login` keychain rather than the system one. I did a quick check and this command works for the login keychain:
```console
security add-trusted-cert -d -r trustRoot \
  -k "$HOME/Library/Keychains/login.keychain-db" build/private-pki/ca.crt
```

This does appear to work for zarf network wait actions, but maybe not on the browser? And still have the curl issue here regardless.

mjnagel (commented):
Not sure if I did something wrong in testing, but when I ran through this on my mac I hit this error:
```console
       •  Running "Validate grafana virtual service"                                                                                                                                                                                          
     curl: (60) SSL certificate problem: unable to get local issuer certificate                                                                                                                                                               
     More details here: https://curl.se/docs/sslcerts.html                                                                                                                                                                                    
                                                                                                                                                                                                                                              
     curl failed to verify the legitimacy of the server and therefore could not                                                                                                                                                               
     establish a secure connection to it. To learn more about this situation and                                                                                                                                                              
     how to fix it, please visit the webpage mentioned above.                                                                                                                                                                                 
     curl: (60) SSL certificate problem: unable to get local issuer certificate                                                                                                                                                               
     More details here: https://curl.se/docs/sslcerts.html                                                                                                                                                                                    
                                                                                                                                                                                                                                              
     curl failed to verify the legitimacy of the server and therefore could not                                                                                                                                                               
     establish a secure connection to it. To learn more about this situation and                                                                                                                                                              
     how to fix it, please visit the webpage mentioned above.                                                                                                                                                                                 
     ERROR: Grafana returned a 000 code.                                                                                                                                                                                                      
          ERROR:  Failed to run action: command "Validate grafana virtual service" timed out after 0 seconds                                                                                                                                  
     ERROR:  Failed to run action: command "for package in $(ls src); do; if [ ! $(echo ${EXCLUDED_..." timed out                                                                                                                             
             after 0 seconds
```

However in a browser the cert seems to be trusted. I did a bit of poking and it might be because I used homebrew-installed curl? It won't by default trust the system keychain but the env `CURL_CA_BUNDLE` would allow trusting it, or the `--cacert` command. I did test both routes and they seem to fix my issue, so maybe `CURL_CA_BUNDLE` as an input/env would work here.

test/playwright/private-pki/tasks.yaml
joelmccoy: found a way to list these certs and delete them individually. latest changes has that in

.github/workflows/private-pki-test.yaml
joelmccoy: fixed. good catch

test/playwright/private-pki/tasks.yaml
joelmccoy: as discussed in DMs, the user keychain does not work for chrome (only safari).  so we are going to keep using the system trust store even though it requires sudo

joelmccoy:
> Not sure if I did something wrong in testing, but when I ran through this on my mac I hit this error:
> 
> ```
>        •  Running "Validate grafana virtual service"                                                                                                                                                                                          
>      curl: (60) SSL certificate problem: unable to get local issuer certificate                                                                                                                                                               
>      More details here: https://curl.se/docs/sslcerts.html                                                                                                                                                                                    
>                                                                                                                                                                                                                                               
>      curl failed to verify the legitimacy of the server and therefore could not                                                                                                                                                               
>      establish a secure connection to it. To learn more about this situation and                                                                                                                                                              
>      how to fix it, please visit the webpage mentioned above.                                                                                                                                                                                 
>      curl: (60) SSL certificate problem: unable to get local issuer certificate                                                                                                                                                               
>      More details here: https://curl.se/docs/sslcerts.html                                                                                                                                                                                    
>                                                                                                                                                                                                                                               
>      curl failed to verify the legitimacy of the server and therefore could not                                                                                                                                                               
>      establish a secure connection to it. To learn more about this situation and                                                                                                                                                              
>      how to fix it, please visit the webpage mentioned above.                                                                                                                                                                                 
>      ERROR: Grafana returned a 000 code.                                                                                                                                                                                                      
>           ERROR:  Failed to run action: command "Validate grafana virtual service" timed out after 0 seconds                                                                                                                                  
>      ERROR:  Failed to run action: command "for package in $(ls src); do; if [ ! $(echo ${EXCLUDED_..." timed out                                                                                                                             
>              after 0 seconds
> ```
> 
> However in a browser the cert seems to be trusted. I did a bit of poking and it might be because I used homebrew-installed curl? It won't by default trust the system keychain but the env `CURL_CA_BUNDLE` would allow trusting it, or the `--cacert` command. I did test both routes and they seem to fix my issue, so maybe `CURL_CA_BUNDLE` as an input/env would work here.

took the approach of using `/usr/bin/curl` if present in the grafana task.  this should force using the system curl which should pull the keychain truststore

test/playwright/private-pki/private-pki.test.ts
```diff
@@ -5,7 +5,7 @@
 
 import { expect, test } from "@playwright/test";
 
-test.describe("Grafana SSO with Private PKI", () => {
+test.describe("Private PKI tests", () => {
```
joelmccoy, line 8: note: just noticed this top level test group name was not accurate for the tests here. changed it to reflect that this is not just grafana pki tests

test/playwright/private-pki/tasks.yaml
mjnagel: Worked great for me in testing!

src/grafana/tasks.yaml
```diff
@@ -16,17 +16,24 @@ tasks:
             condition: Ready
       - description: Validate grafana virtual service
         cmd: |
-          if [ "$(curl -isS https://grafana.admin.uds.dev --output /dev/null -w '%{http_code}')" = "302" ]; then
+          # Note: this check forces to use /usr/bin/curl if it exists
+          # This is to force using the system curl on macos which is needed for our private-pki tests
+          CURL_CMD="/usr/bin/curl"
+          if [ ! -x "$CURL_CMD" ]; then
+            CURL_CMD="curl"
+          fi
```
mjnagel: Commenting here because it's the closest spot to comment...I still see issues with homebrew curl during the uds-common `Creating the doug user in Keycloak` task. I think it might be okay to just say I'm on my own / add a small comment on the task about this gotcha. I did find that I was able to run:
```console
export CURL_CA_BUNDLE=path/to/uds-core/build/private-pki/ca.crt
uds run -f tasks/test.yaml uds-core-private-pki
```

And everything passed/worked (also reverted this curl change locally) - so maybe adding that in a comment might be nice. I _think_ inlining the env also works, although I didn't do a full e2e test with that (`CURL_CA_BUNDLE=path/to/uds-core/build/private-pki/ca.crt uds run -f tasks/test.yaml uds-core-private-pki`).

test/playwright/private-pki/tasks.yaml
```diff
@@ -44,16 +51,20 @@ tasks:
...
           # Base64 encode CA
-          base64 -w 0 build/private-pki/ca.crt > build/private-pki/ca.crt.b64
+          base64 < build/private-pki/ca.crt | tr -d '\n' > build/private-pki/ca.crt.b64
 
-          # Add CA to system trust store
-          sudo cp build/private-pki/ca.crt /usr/local/share/ca-certificates/uds-test-ca.crt
-          sudo update-ca-certificates
+          # Add CA to trust store
+          if [[ "$OSTYPE" == "darwin"* ]]; then
+            sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain build/private-pki/ca.crt
```
mjnagel, line 63: Not sure if it's unique to this new method for mac, but we have instruction output in `tasks/test.yaml` that says you need to load the certs for browser access separately:
```yaml
      - description: "Follow on Instructions"
        cmd: |
          echo ""
          echo "Load the build/private-pki/ca.crt into your browser if you want to use the Grafana, Keycloak, or httpbin UIs."
          echo ""
```

Is that wrong now/in general?

src/grafana/tasks.yaml
joelmccoy: i added this as a comment on the private pki task and choosing to ignore your special problem :P.

test/playwright/private-pki/tasks.yaml
joelmccoy: good catch. don't think it is relevant anymore.  removed this.

mjnagel (approved):

src/grafana/tasks.yaml
mjnagel: I also uninstalled homebrew curl now - although may be something to consider if private-pki is relevant for other testing/demo environments in the future (making sure that we know the limits/edge cases).

PR #2000, opened 2025-10-03
chore(docs): diagrams as code with likec4

chance-coleman:
## Description
We recently decided to use diagrams as code for creating and maintaining our diagrams. This should reduce maintenance burden and speed up creation of new diagrams.

Because this is diagrams as code, we lose some ability to customize how the diagram is generated. There is the ability with likec4 to manually edit and arrange diagrams, however doing this has some big pitfalls:

1. It causes headaches anytime we update the diagrams because the generator is not smart enough to take the customizations and add new things elegantly
2. When exporting the manually created diagrams, the export functionality is not smart enough to figure out where the center of the diagram is, resulting in a very large capture space which is mostly just white space.

Currently created diagrams:
- Overview Diagrams
  - Basic Overview
  - Namespace Overview
  - Ingress / Egress Overview
- Functional Layers (each layer is it's own diagram)
- Namespaces (each namespace is it's own diagram)
- Ports and Protocols

## Open Questions
- How do we like the colors?
- Do we want all these diagrams to be generated and stored in UDS Core right? Or do we only want a subset of them?
- Does anyone have any experience with likec4 and how to make manually customized diagrams export the right size? Or some automated way of trimming whitespace from a png?

## Related Issue

Fixes #1955

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md) followed

PR #2083, opened 2025-10-29
feat!: remove neuvector and require falco

### LINKED ISSUE #1950: Switch falco to required and remove NeuVector
This issue should not be worked until the iteration before the planned release on Nov 10.

Definition of done:
- Falco is included as a required component in core (any optional component overrides removed)
- NeuVector is fully removed from core (docs included)
- Clear plan for end users on what the upgrade looks like - is NeuVector automatically removed for them, conditionally removed, etc

chance-coleman:
## Description
Removal of NeuVector from UDS Core and requiring Falco as the new default runtime security application.

## Related Issue

Fixes #1950

Docs PRs
[NeuVector Package Guide](https://github.com/uds-packages/neuvector/pull/23)
[UDS Docs Tutorial Prerequisites updated and removal of NeuVector references](https://github.com/defenseunicorns/uds-docs/pull/308)

Also will need to get diagrams in from this [PR](https://github.com/defenseunicorns/uds-core/pull/2000) along with the docs updates for those diagrams to be up to date on the docsite.

## Type of change

## Steps to Validate
- 
## Checklist before merging


docs/reference/configuration/single-sign-on/overview.md
```diff
@@ -84,34 +84,6 @@ grafana:
 
 ```
 
-#### Neuvector
```
chance-coleman, line 87: Should I be moving all these docs to the NeuVector package docs?

chance-coleman: This is already documented, added external alerts section in [this PR](https://github.com/uds-packages/neuvector/pull/23).

docs/reference/configuration/uds-networking-configuration.md
```diff
@@ -146,30 +146,9 @@ The example above allows Grafana to query a remote Thanos instance in your clust
...
-                - direction: Egress
-                  selector:
-                    app: neuvector-manager-pod
-                  remoteGenerated: Anywhere
-                  description: "from neuvector to anywhere"
-                  port: 443
-```
-
-The example above allows NeuVector to send alerts to any external destination. Alternatively, you could use the remoteNamespace key to specify another namespace within the Kubernetes cluster (i.e. Mattermost).
+It may be desired send alerts from falco to locations in or outside of the cluster. To facilitate this, you can provide this [bundle override](https://uds.defenseunicorns.com/reference/configuration/runtime-security/alerting/#external-alert-forwarding) send alerts to any external destination. Alternatively, you could use the remoteNamespace key to specify another namespace within the Kubernetes cluster (i.e. Mattermost).
```
mjnagel:
```suggestion
It may be desired send alerts from falco to locations in or outside of the cluster. To facilitate this, you can provide this [bundle override](/reference/configuration/runtime-security/alerting/#external-alert-forwarding) send alerts to any external destination. Alternatively, you could use the remoteNamespace key to specify another namespace within the Kubernetes cluster (i.e. Mattermost).
```

Local refs generally preferred for local docs development (I think this is the right ref?)

packages/runtime-security/zarf.yaml
```diff
@@ -14,18 +14,42 @@ metadata:
...
+                echo "[cleanup] Removing legacy NeuVector CRDs if present..."
+                NV_CRDS="$(./zarf tools kubectl get crd -o name | grep -i neuvector || true)"
+                if [ -n "$NV_CRDS" ]; then
+                  echo "$NV_CRDS" | xargs -r ./zarf tools kubectl delete
+                else
+                  echo "[cleanup] No NeuVector CRDs found"
+                fi
+              else
+                echo "[cleanup] Skipping NeuVector CRD cleanup (CLEANUP_LEGACY_NEUVECTOR=false)"
+              fi
```
mjnagel, line 55: I _think_ this approach is sane but a few comments/thoughts:
- `zarf package remove core --components neuvector --confirm` is an interesting way to catch _everything_ (essentially a helm uninstall). We did this before for promtail -> vector...but I think the `core` name there might depend on how their bundle names the core package so it's a bit error-prone? I think something like `zarf p ls -o json 2>/dev/null | zarf tools yq '.[] | select(.components[] == "neuvector") | .package'` could grab the right package name, which is totally not insane right? 😛
- As a small improvement here on the CRD side we probably could grep for `neuvector.com` to be more specific, but I think either is safe

Regardless will want to call this out in release notes and/or document the manual equivalents also if people just want to one-time run some commands rather than setting the var.

mjnagel (commented):
LGTM overall! Small comments and we may want to consider merge timing given the docs removals/changes (or could split the docs removals off to a separate PR for closer to release day)?

packages/runtime-security/zarf.yaml
chance-coleman: if you have a preference here i'm happy to switch. I personally find just deleting the namespace cleaner and more straightforward. Agreed on the neuvector.com, i'll update that.

bundles/k3d-standard/uds-bundle.yaml
```diff
@@ -43,7 +43,6 @@ packages:
       - istio-passthrough-gateway
       - istio-egress-gateway
       - metrics-server
-      - falco
```
joelmccoy, line 46: just noting here. I was curious if people who have this in as an optional component in their bundle, would this break?  And I confirmed that it does not break.  So people who list falco as an optional component don't technically need to remove this from their bundle (which is nice).

docs/reference/configuration/runtime-security/overview.md
```diff
@@ -6,33 +6,17 @@ sidebar:
...
-    repository: oci://ghcr.io/defenseunicorns/packages/uds/core
-    ref: x.x.x-upstream
-    optionalComponents:
-      - falco # Deploys Falco as an optional component
+```bash
+# Example: deploy the runtime-security layer with cleanup enabled
+zarf package deploy packages/runtime-security --set CLEANUP_LEGACY_NEUVECTOR=true
 ```
+
+If you plan to deploy the standalone NeuVector package without Falco, do not enable this gate and skip the runtime-security layer entirely. See the [Runtime Security Migration Guide](./migration.md) and [Standalone NeuVector](./neuvector-standalone.md) for advanced scenarios.
```
mjnagel: These two docs are in the neuvector repo, not local to core I think? Links may need to change.

packages/runtime-security/zarf.yaml
mjnagel: I think this is fine - my "other option" was mostly thinking out loud and looking at what we did before. For NeuVector I think it's more clean and simple to just go with namespace/crd delete as well :+1:

docs/reference/configuration/runtime-security/overview.md
chance-coleman: oh yeah good call, i'll update that and run this branch through the docs to make sure everything is linked properly in case i missed any other ones

chance-coleman: looking at this again, and only the standalone doc is from the neuvector repo.

chance-coleman: broke the docs out into a separate branch, [link to PR](https://github.com/defenseunicorns/uds-core/pull/2095)

mjnagel (approved):
LGTM - cleanup flag worked as expected as well.

joelmccoy (approved):
LGTM - Tested the cleanup command locally (both the negative/positive)

PR #2088, opened 2025-10-30
fix: install test apps preupgrade and validate

### LINKED ISSUE #1995: Add sample package to upgrade CI pre-upgrade
In the current UDS Core (all package/demo bundle) upgrade CI, we deploy a number of test packages post-upgrade to validate all features within Core work as expected. However there is a small gap where these packages do not exist pre-upgrade, allowing for potential upgrade issues that we miss. This is somewhat mitigated by the fact that all Core packages go through the upgrade, but there may be some missing features that do not get validated across the upgrade.

One specific known gap is a Package CR with Authservice protection. While we validate this functionality post-upgrade, it is not installed prior to the upgrade in CI. This has lead to issues like https://github.com/defenseunicorns/uds-core/issues/1994 not being caught in CI.

Definition of done: Package (CR + full sample application) deployed in CI during the upgrade test, after initial install, before upgrade. Full validation of functionality is performed post-upgrade to ensure the upgrade does not disrupt app availability/reconciliation. At a minimum this Package CR should include authservice protection, as well as any other features/config that are not present in other Core Package CRs.

chance-coleman:
## Description

Previously we had a gap in our upgrade testing that meant we never actually tested a mission application being upgraded from one version of uds-core to the next in CI. This PR reworks the upgrade testing to make sure this is done.

### New Flow of the upgrade tests

1. install latest release uds-core
2. install latest release test-resources
3. build and deploy (upgrade) current branch
4. validate everything
5. e2e-tests (which is technically an upgrade of the test-resources and all the existing tests)

## Related Issue

Fixes #1995 

## Type of change

## Checklist before merging

### DIFF
--- a/src/test/tasks.yaml
+++ b/src/test/tasks.yaml
@@ -37,11 +37,11 @@ tasks:
     actions:
       - description: "Run Network Tests"
         cmd: |
-          npm ci && EGRESS_TESTS=${{ .inputs.validate_egress }} npx vitest run "network.spec.ts"
+          EGRESS_TESTS=${{ .inputs.validate_egress }} npx vitest run "network.spec.ts"
         dir: test/vitest
       - description: "Run Test Resources Tests"
         cmd: |
-          npm ci && npx vitest run "test-resources.spec.ts"
+          npx vitest run "test-resources.spec.ts"
         dir: test/vitest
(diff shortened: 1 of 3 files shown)


joelmccoy (approved):
LGTM

mjnagel (approved):

PR #2095, opened 2025-10-31
chore(docs): remove neuvector docs

chance-coleman:
## Description
PR to update docs for the removal of neuvector.
 
to be merged closer to release on Nov. 12th.

## Type of change

## Checklist before merging

### DIFF
--- a/docs/reference/configuration/runtime-security/migration.md
+++ b/docs/reference/configuration/runtime-security/migration.md
@@ -2,7 +2,7 @@
 title: Runtime Security Migration Guide
 ---
 
-This guide describes how to migrate to the new default runtime security posture in UDS Core, where Falco is the required solution and NeuVector is no longer managed by Core.
+This guide describes how to migrate to the new default runtime security posture in UDS Core, where Falco is the required solution and NeuVector is no longer managed by UDS Core.
 
 UDS Core now:
 - Includes Falco by default in the `core-runtime-security` layer.
@@ -12,13 +12,19 @@ UDS Core now:
 
 - **Falco only (remove legacy NeuVector on upgrade)**
   - Enable the cleanup gate during runtime-security deploy to remove legacy NeuVector resources (upgrade-only):
-    - Package level:
+    - Runtime Security Package:
       ```bash
       zarf package deploy packages/runtime-security --set CLEANUP_LEGACY_NEUVECTOR=true --confirm
       ```
-    - Bundle level:
+    - Standard package:
       ```bash
-      uds deploy bundles/neuvector-standalone/uds-bundle-neuvector-standalone-amd64-0.1.0.tar.zst --set CLEANUP_LEGACY_NEUVECTOR=true --confirm
+      zarf package deploy packages/standard --set CLEANUP_LEGACY_NEUVECTOR=true --confirm
+      ```
+    - Use a uds-config.yaml to set at the bundle level:
+      ```yaml
+      variables:
+        core:
+          CLEANUP_LEGACY_NEUVECTOR: "true"
       ```
   - This deletes the legacy `neuvector` namespace and any CRDs whose names contain `neuvector` if they exist.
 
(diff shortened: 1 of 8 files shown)


docs/reference/configuration/runtime-security/migration.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Runtime Security Migration Guide
+---
+
+This guide describes how to migrate to the new default runtime security posture in UDS Core, where Falco is the required solution and NeuVector is no longer managed by Core.
```
slaskawi: nit: Probably UDS Core instead of Core might sound better.

slaskawi (dismissed):
LGTM with 1 nit (up to you to implement or disregard it).

docs/reference/configuration/runtime-security/migration.md
```diff
@@ -0,0 +1,33 @@
...
+- Includes Falco by default in the `core-runtime-security` layer.
+- Does not manage NeuVector. If you still need NeuVector, deploy it as a standalone package.
+
+## Different Scenarios
+
+- **Falco only (remove legacy NeuVector on upgrade)**
+  - Enable the cleanup gate during runtime-security deploy to remove legacy NeuVector resources (upgrade-only):
+    - Package level:
+      ```bash
+      zarf package deploy packages/runtime-security --set CLEANUP_LEGACY_NEUVECTOR=true --confirm
```
joelmccoy, line 17: Won't the cleanup work for the standard core package as well since its a zarf var? I think many envs might just be using the full core package, so it might be good to clarify this can be done with the core standard package

docs/reference/configuration/runtime-security/migration.md
```diff
@@ -0,0 +1,33 @@
...
+      ```
+    - Bundle level:
+      ```bash
+      uds deploy bundles/neuvector-standalone/uds-bundle-neuvector-standalone-amd64-0.1.0.tar.zst --set CLEANUP_LEGACY_NEUVECTOR=true --confirm
```
joelmccoy: the way this is written it assumes you need a standalone bundle for this which I think is not the case?  I think many environments just have a single bundle with the core standard package.  For example the UDS Registry env bundle: https://github.com/defenseunicorns/infra-uds-registry/blob/main/bundles/core/uds-bundle.yaml

Couldn't these bundle users just set a variable in the actual bundle file to set `CLEANUP_LEGACY_NEUVECTOR` = `true`? If so, this would be a much easier suggestion for the cleanup and works for environments that need to go through specific CI.

joelmccoy: I just tested this locally. Should work with the a standard core bundle.  Additionally you can have someone set this var in their `uds-config.yaml` like so (easier for gitops instead of changing the commands):
```yaml
variables:
  core:
    CLEANUP_LEGACY_NEUVECTOR: "true"
```

docs/reference/configuration/runtime-security/overview.md
```diff
@@ -6,33 +6,17 @@ sidebar:
...
-
-packages:
-  - name: core
-    repository: oci://ghcr.io/defenseunicorns/packages/uds/core
-    ref: x.x.x-upstream
-    optionalComponents:
-      - falco # Deploys Falco as an optional component
+```bash
+# Example: deploy the runtime-security layer with cleanup enabled
+zarf package deploy packages/runtime-security --set CLEANUP_LEGACY_NEUVECTOR=true
```
joelmccoy, line 19: It might be worth calling out the bundle example here as an option as well (since most people use bundles to deploy):
you can have someone set this var in their `uds-config.yaml` like so (easier for gitops instead of changing the commands):
```yaml
variables:
  core:
    CLEANUP_LEGACY_NEUVECTOR: "true"
```

joelmccoy (dismissed):
LGTM, left some comments/suggestions, but non-blocking

slaskawi (approved):
LGTM but might be good to get also feedback from @joelmccoy

joelmccoy (approved):

PR #2107, opened 2025-11-04
feat: add experimental cluster health rules

jsevedge:
## Description

Add support for configuring UDS Cluster Health Score metric via recording rules, optionally enabled

## Related Issue

Not related to any existing uds-core issue.

## Type of change

## Steps to Validate

- Added helm unit tests to validate the conditional logic of all-or-nothing for the new `PrometheusRule`

## Checklist before merging


src/prometheus-stack/chart/templates/uds-rules.yaml
```diff
@@ -0,0 +1,55 @@
...
+  name: "{{ .Release.Name }}-uds-cluster-health"
+  namespace: {{ .Release.Namespace }}
+  labels:
+    release:  {{ .Release.Name }}
+spec:
+  groups:
+    - name: "{{ .Release.Name }}-uds-cluster-health.rules"
+      interval: 1m
+      rules:
+        - record: uds_cluster_health_score
```
joelmccoy, line 16: Just to spell out explicitly and think out loud here...  This metric essentially checks the health of the prometheus scrape targets for each of these services (not necessarily that the service is functional).  I think technically if prometheus is down, all these metrics will be down??  Does it make sense to even include prometheus in the check/weight if that is the case?

Still think this is a sane heuristic for now to determine if a service is up. I think the uptime metrics we introduce at a later point may be a better option to determine if a component is healthy.

src/prometheus-stack/chart/templates/uds-rules.yaml
```diff
@@ -0,0 +1,55 @@
...
+              (avg(up{job="apiserver"}) or vector(1)) * 25
+              +
+              # 25% weight for Kubelet health
+              (avg(up{job="kubelet"}) or vector(1)) * 25
+              +
+              # 15% weight for CoreDNS health
+              (avg(up{job="coredns"}) or vector(1)) * 15
+              +
+              # 15% weight for Pepr health
+              (avg(up{job="pepr-uds-core"}) or vector(1)) * 15
```
joelmccoy, line 29: pepr also has a scrape target for `job=pepr-uds-core-watcher`. This metric just watches the admission service as is. And you will likely want to include this job to grab the watcher service.

src/prometheus-stack/chart/templates/uds-rules.yaml
```diff
@@ -0,0 +1,55 @@
...
+              +
+              # 10% weight for Keycloak health
+              (avg(up{job="keycloak-http"}) or vector(1)) * 10
+              +
+              # 10% weight for Prometheus health
+              (avg(up{job="kube-prometheus-stack-prometheus"}) or vector(1)) * 10
+            )
```
joelmccoy, line 36: Any reason you didn't include metrics-server, Grafana, Loki, Falco, or Velero? Fine keeping it simple for now to get iterations in... But either way, before we make this metric not experimental anymore, I think we should have a published definition of what this metric captures and how we decided on the weight for each of the services.

src/prometheus-stack/chart/templates/uds-rules.yaml
```diff
@@ -0,0 +1,55 @@
...
+spec:
+  groups:
+    - name: "{{ .Release.Name }}-uds-cluster-health.rules"
+      interval: 1m
+      rules:
+        - record: uds_cluster_health_score
+          expr: |
+            (
+              # 25% weight for API Server health
+              (avg(up{job="apiserver"}) or vector(1)) * 25
```
joelmccoy, line 20: Should we default to vector(0) here instead? What happens if the scrape itself is broken and no metric exists? I think with vector(1) would read that everything is healthy. And same for all the others.

joelmccoy (commented):
Generally think this is a good approach. Left some comments/questions for consideration.  Before we decide to make this metric not experimental, I would like to have a published definition (maybe in the docs somewhere) of what this metric is and how we loosely determined the weight for each component in the score.

src/prometheus-stack/chart/templates/uds-rules.yaml
```diff
@@ -0,0 +1,55 @@
...
+              (avg(up{job="pepr-uds-core"}) or vector(1)) * 15
+              +
+              # 10% weight for Keycloak health
+              (avg(up{job="keycloak-http"}) or vector(1)) * 10
+              +
+              # 10% weight for Prometheus health
+              (avg(up{job="kube-prometheus-stack-prometheus"}) or vector(1)) * 10
+            )
+        - alert: UDSClusterHealthScoreWarning
+          expr: uds_cluster_health_score < 95
```
joelmccoy, line 38: I think this will also fire if it is below 90% creating duplicate alerts. consider making the expression: 
```suggestion
          expr: uds_cluster_health_score < 95 and uds_cluster_health_score >= 90
```

src/prometheus-stack/chart/templates/uds-rules.yaml
jsevedge: I erred towards fail open in this first pass with vector(1).  It is a valid point of consideration though... main blocker right now being with this simplistic expression we may have UDS Core deployed without some of the functional layers.

src/prometheus-stack/chart/templates/uds-rules.yaml
jsevedge: Just didn't invest the cycles thus far to test and include (our multi-cluster bundles only have a few functional layers included in my defense).  Imo before merging this should probably spend the cycles to include those so the score doesn't feel half baked.

src/prometheus-stack/chart/templates/uds-rules.yaml
jsevedge: Totally agreed, this is pretty rudimentary in general and looking at synthetic monitoring data would be better / in addition.  @soltysh suggested swapping grafana out for prometheus but of course if prometheus is partially or fully down none of this will work.  For partial prometheus failure scenarios (where metrics scraping and remote write capabilities are still up) doesn't hurt to include imo.

src/prometheus-stack/chart/templates/uds-rules.yaml
joelmccoy: totally fine for this as a first iteration, just wanted to ask the question

mjnagel:
@jsevedge is this still WIP that is valuable or are we okay to close this out for now? There may also be some overlapping/related work in https://github.com/defenseunicorns/uds-core/pull/2426.

mjnagel:
Tentatively closing - feel free to reopen if this is still applicable @jsevedge

PR #2121, opened 2025-11-06
chore: fix test task to check authservice protection

### LINKED ISSUE #2086: Check Authservice Protection Test Task Doesn't Run
The task here: https://github.com/defenseunicorns/uds-core/blob/d54c18b0a89a20d8c9fcdd6cd9880d75a5db297b/src/test/tasks.yaml#L158-L159 does not run in our CI/Task flows because of an invalid schema (we need an `actions` key).

Additionally when we enable this task it will most likely fail for the private-pki tests.  So we will need to update this task to work with the private-pki test workflow.

joelmccoy:
## Description

The check authservice protection task in our test zarf package wasn't running because of bad maru schema.  This fixes it.  Updates the task to use curl locally instead of launching pod on cluster. 

## Related Issue

Fixes #2086 

## Type of change

## Steps to Validate
- `uds run`
- `uds run -f tasks/test.yaml uds-core-private-pki`

## Checklist before merging

### DIFF
--- a/.github/workflows/private-pki-test.yaml
+++ b/.github/workflows/private-pki-test.yaml
@@ -16,6 +16,7 @@ on:
       - "bundles/k3d-standard/**"
       - "test/playwright/private-pki/**"
 
+
 permissions:
   contents: read
   id-token: write
--- a/src/test/tasks.yaml
+++ b/src/test/tasks.yaml
@@ -153,12 +153,23 @@ tasks:
             address: ${{ .inputs.address }}
             code: 200
 
-      - description: Verify the service at ${{ .inputs.address }} is protected by checking redirect
+      - description: Get tenant gateway IP
         maxRetries: 3
         task: utils:tenant-gw-ip
+
+      - description: Verify the service at ${{ .inputs.address }} is protected by checking redirect
+        maxRetries: 3
         cmd: |
           set -e
-          SSO_REDIRECT=$(uds zarf tools kubectl run curl-test --image=cgr.dev/chainguard/curl:latest -q --restart=Never --rm -i -- --resolve '${{ .inputs.address }}:$TENANT_GW_IP:443' -Ls -o /dev/null -w %{url_effective} "https://${{ .inputs.address }}")
+          # If running locally with k3d, access via the hostname directly
+          if curl --connect-timeout 2 -s "https://${{ .inputs.address }}" >/dev/null 2>&1; then
+            echo "Running in local environment, accessing via hostname"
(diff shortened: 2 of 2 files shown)


.github/workflows/private-pki-test.yaml
```diff
@@ -16,6 +16,7 @@ on:
       - "bundles/k3d-standard/**"
       - "test/playwright/private-pki/**"
 
+
```
joelmccoy, line 19: hack to trigger private pki tests in CI. can remove this if desired.

src/test/tasks.yaml
```diff
@@ -153,12 +153,23 @@ tasks:
...
-          SSO_REDIRECT=$(uds zarf tools kubectl run curl-test --image=cgr.dev/chainguard/curl:latest -q --restart=Never --rm -i -- --resolve '${{ .inputs.address }}:$TENANT_GW_IP:443' -Ls -o /dev/null -w %{url_effective} "https://${{ .inputs.address }}")
+          # If running locally with k3d, access via the hostname directly
+          if curl --connect-timeout 2 -s "https://${{ .inputs.address }}" >/dev/null 2>&1; then
+            echo "Running in local environment, accessing via hostname"
+            SSO_REDIRECT=$(curl -Ls -o /dev/null -w %{url_effective} "https://${{ .inputs.address }}")
+          # If running in CI or another environment, resolve via tenant gateway IP
+          else
+            echo "Running in CI/remote environment, resolving via tenant gateway IP"
+            SSO_REDIRECT=$(curl --resolve '${{ .inputs.address }}:443:$TENANT_GW_IP' -Ls -o /dev/null -w %{url_effective} "https://${{ .inputs.address }}")
+          fi
```
mjnagel: It's been a while since I looked at this task, but I'm trying to track why we need the resolution here when k3d (local or in CI) should resolve okay, and our IAC uses this [setup-hosts task](https://github.com/defenseunicorns/uds-core/blob/5dd5cf8f8cce5639e841158a97fd3e71faec8b32/tasks/utils.yaml#L101-L108). Are we missing some hosts in that list, or some other problem I'm overlooking?

If we do need separate paths for local/k3d vs IAC, I wonder if we could just use a more explicit input to detect this rather than relying on the curl to timeout?

joelmccoy: dohh. that makes sense. forgot we update our hosts here. there is no need for this then. I'll update the curl command

mjnagel: Oh I'm following it now - we had the resolution before because this was running in cluster, although coredns overrides should've handled that too. Possible this was before we ran full testing/had coredns overrides on IAC...hopefully just old history.

mjnagel (approved):

chance-coleman (approved):

PR #2127, opened 2025-11-07
fix: prevent network policies from selecting non-authservice SSO clients

### LINKED ISSUE #2120: Bug with networkpolicies when additional SSO client is deployed without authservice protection
Discovered while working with @TristanHoladay and @chance-coleman that there is a bug behavior where networkpolicies can incorrectly be attached to a waypoint that doesn't exist, for an SSO client that isn't "authservice enabled". This seems related to some incorrect filtering of clients in some of our netpol code: https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/policies.ts#L31-L37

It seems like this function may be treating an undefined `enableAuthserviceSelector` (key not present) as if it were an authservice client and should protect the whole namespace. We should fix this issue and make sure we align the ways we filter authservice clients from the general SSO list (maybe extracting [this](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/authservice/authservice.ts#L47-L50) into a util func). We should also add testing around this behavior (ensuring that netpols do/don't get adjusted to target the waypoint when necessary).

chance-coleman:
## Description

This PR fixes a bug where NetPols could be generated for SSO clients that were not "authservice-enabled" because a missing `enableAuthserviceSelector` was effectively treated as namespace-wide.

Pulled an existing authservice enabled sso client filter into a new utility to function that can be used to pre filter for the clients that we need. Implemented a bunch of new tests for the utility and the functions that use that utility to make sure i maintained existing expected behavior. Updated a few different spots that were filtering for authservice clients in different ways.
 
## Related Issue

Fixes #2120 

## Type of change

## Steps to Validate
- `uds run test:uds-core-e2e`
- `uds run test:unit-tests`

## Checklist before merging

### DIFF
--- a/src/pepr/operator/controllers/network/authorizationPolicies.spec.ts
+++ b/src/pepr/operator/controllers/network/authorizationPolicies.spec.ts
@@ -1170,7 +1170,7 @@ describe("findMatchingSsoClient", () => {
     expect(findMatchingSsoClient(pkg, { app: "x" })).toMatchObject({ clientId: "c1" });
   });
 
-  test("non-empty SSO selector uses any-label matching semantics", () => {
+  test("non-empty SSO selector uses all-label matching semantics", () => {
     const pkg: UDSPackage = {
       metadata: { name: "app", namespace: "ns" },
       spec: {
@@ -1179,10 +1179,12 @@ describe("findMatchingSsoClient", () => {
       },
     };
 
-    // Matches because 'tier' label matches (any-label semantics)
-    expect(findMatchingSsoClient(pkg, { tier: "prod" })).toMatchObject({ clientId: "c1" });
-    // Does not match when none of the labels match
-    expect(findMatchingSsoClient(pkg, { env: "prod" })).toBeUndefined();
+    // Matches only when all labels match
+    expect(findMatchingSsoClient(pkg, { app: "a", tier: "prod" })).toMatchObject({
+      clientId: "c1",
+    });
+    // Does not match when only a subset matches
+    expect(findMatchingSsoClient(pkg, { tier: "prod" })).toBeUndefined();
   });
 
   test("two enabled clients both match; first match wins", () => {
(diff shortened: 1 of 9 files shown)


src/pepr/operator/controllers/network/authorizationPolicies.spec.ts
```diff
@@ -1112,3 +1114,125 @@ describe("createDenyAllExceptWaypointPolicy", () => {
...
+    const pkg: UDSPackage = {
+      metadata: { name: "app", namespace: "ns" },
+      spec: {
+        sso: [{ clientId: "c1", name: "", enableAuthserviceSelector: { app: "a", tier: "prod" } }],
+        network: { serviceMesh: { mode: Mode.Ambient } },
+      },
+    };
+
+    // Matches because 'tier' label matches (any-label semantics)
+    expect(findMatchingSsoClient(pkg, { tier: "prod" })).toMatchObject({ clientId: "c1" });
```
mjnagel: This test is the one that made me suspicious of the logic I commented on (and this test would need to be changed if that tracks).

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -424,21 +424,17 @@ export function findMatchingSsoClient(
...
 
     // Empty object means namespace-wide protection
-    if (Object.keys(waypointSelector).length === 0) {
+    if (Object.keys(waypointSelector!).length === 0) {
       return true;
     }
 
     // For non-empty selectors, check if any label matches
-    return Object.entries(waypointSelector).some(([key, value]) => selector[key] === value);
+    return Object.entries(waypointSelector!).some(([key, value]) => selector[key] === value);
```
mjnagel: I think maybe this should be `.every` rather than `.some`? The equivalent label check for waypoints uses `.every` - https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/istio/waypoint-utils.ts#L74

mjnagel (commented):
One small bug I think - not new to this code but noticed since its in the same flow.

mjnagel (approved):

PR #2138, opened 2025-11-13
feat: adding custom pod annotations to keycloak statefulset

briantwatson:
## Description
- Adding new value `podAnnotations` to allow defining custom annotations
  - Added unittests for annotation scenarios
- Modified `lint-check` task to use `uv` and `uvx` from `pip3`, avoiding `externally-managed-environment` error with some pip installations
- Add setup-uv action to setup to accommodate switch to `uv`
- Updated CONTRIBUTING.md to reflect uv requirement

## Related Issue

N/A

## Type of change

## Steps to Validate
- Can test with with unit tests.  Additionally can render chart via `helm template test src/keycloak/chart`

## Checklist before merging

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,7 +39,7 @@ Before starting, ensure that you have the following installed:
 - **Go** (for pre-commit tooling): [Install Go](https://go.dev/doc/install)
 - **Helm** (for development and testing): [Install Helm](https://helm.sh/docs/intro/install/)
 - **Helm Unittest Plugin** (for development and testing): [Install Helm unittest](https://github.com/helm-unittest/helm-unittest?tab=readme-ov-file#install)
-- **uv** is required for running linting tooling:  [Install uv](https://docs.astral.sh/uv/getting-started/installation/#installing-uv).  The `lint` tasks will install latest if not already installed.
+- **uv** is required for running linting tooling:  [Install uv](https://docs.astral.sh/uv/getting-started/installation/#installing-uv).
 
 #### Setting Up Your Local Repository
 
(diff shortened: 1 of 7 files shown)


tasks/lint.yaml
```diff
@@ -20,12 +49,12 @@ tasks:
...
+      - description: Ensure python tools are installed
+        task: ensure-python-tools
       - description: install pepr deps
         cmd: npm ci
       - description: "Pepr Format check"
         cmd: npx pepr format --validate-only
-      - description: install yamllint and codespell deps
-        cmd: CMD=pip && which $CMD || CMD=pip3 && $CMD install yamllint codespell
       - description: yaml lint
         cmd: yamllint . -c .yamllint --no-warnings
```
mjnagel: As discussed in slack, let's switch to `uvx` for execution of `yamllint` and `codespell`. We may still want an "ensure-python-tools" type check to make sure `uvx` is present, but that should simplify this a bit and reduce the system installs we're doing.

CONTRIBUTING.md
```diff
@@ -39,6 +39,7 @@ Before starting, ensure that you have the following installed:
 - **Go** (for pre-commit tooling): [Install Go](https://go.dev/doc/install)
 - **Helm** (for development and testing): [Install Helm](https://helm.sh/docs/intro/install/)
 - **Helm Unittest Plugin** (for development and testing): [Install Helm unittest](https://github.com/helm-unittest/helm-unittest?tab=readme-ov-file#install)
+- **Python 3** is required for installing [pipx](https://pipx.pypa.io/) to install lint tooling.
```
mjnagel: Assuming we go the `uvx` route, make sure we update the contributing guide.

mjnagel (commented):
Keycloak changes LGTM.

tasks/lint.yaml
briantwatson: changed to uvx

CONTRIBUTING.md
briantwatson: Updated docs to reflect uvx change

mjnagel (dismissed):

mjnagel (approved):

PR #2145, opened 2025-11-14
chore: removing patch after upstream fix

briantwatson:
## Description

Removing patch no longer required

## Related Issue

N/A, observed temp patch for issue was resolved upstream

## Type of change

## Steps to Validate
Ran all the tests locally

## Checklist before merging

### DIFF
--- a/src/pepr/patches/index.ts
+++ b/src/pepr/patches/index.ts
@@ -3,40 +3,9 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
  */
 
-import { Capability, a } from "pepr";
-import { Component, setupLogger } from "../logger";
+import { Capability } from "pepr";
 
 export const patches = new Capability({
   name: "patches",
   description: "UDS Core Capability for patching miscellaneous things.",
 });
-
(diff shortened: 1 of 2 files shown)


src/pepr/patches/index.ts
```diff
@@ -3,40 +3,9 @@
...
 
 export const patches = new Capability({
   name: "patches",
   description: "UDS Core Capability for patching miscellaneous things.",
 });
-
-const { When } = patches;
-
-// configure subproject logger
-const log = setupLogger(Component.PATCHES);
```
mjnagel, line 17: Do these error as unused when not removed? Might be nice to leave them if not for the basic capability setup for reference? If it does error though, this delete seems fine.

briantwatson: I was getting eslint errors locally, yes.  I thought to maybe comment them out?  LMK preference, happy to adjust.

mjnagel: Yep that tracks - okay to just delete these here. I think we might have future use for this capability (why I think it's nice to keep around in general) but shouldn't be hard to setup the When and log again.

mjnagel (approved):

PR #2152, opened 2025-11-18
fix: decouple istio-cni and zarf cluster components

Racer159:
## Description

This PR excludes the `zarf` namespace from the Istio CNI pod so that zarf components don't depend on Istio to function creating a circular dependency (see related issue).

## Related Issue

Relates to https://github.com/defenseunicorns/uds-edge-os/issues/166#issuecomment-3411261788

## Type of change

## Steps to Validate

To test this - do the following:

1. Scale the `zarf-agent` pods to `0` replicas in the deployment
2. Delete / recycle the istio-cni pod (it will fail to come up due to the agent being missing)
3. Scale the `zarf-agent` pods back to `2` replicas (or your preferred value)

On this PR the pods will all come back - in the previous scenario the pods would get stuck and if you scaled down the zarf registry could get really really stuck.

## Checklist before merging

### DIFF
--- a/src/istio/values/base-cni.yaml
+++ b/src/istio/values/base-cni.yaml
@@ -6,6 +6,9 @@ profile: ambient
 cniConfDir: ###ZARF_VAR_CNI_CONF_DIR###
 cniBinDir: ###ZARF_VAR_CNI_BIN_DIR###
 
+podLabels:
+  zarf.dev/agent: ignore
+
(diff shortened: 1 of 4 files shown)


slaskawi (approved):
Approved but would be great to get a signoff from @mjnagel

mjnagel (approved):
This makes sense to me.

chance-coleman (approved):

PR #2157, opened 2025-11-19
feat: keycloak logout confirmation

### LINKED ISSUE #689: Display an explicit logoff message when users logoff from Keycloak
In order to ensure users are confident their session has been terminated by Keycloak, it would be useful if Keycloak explicitly displayed a logoff message/confirmation. The current behavior does terminate sessions properly, but typically users are redirected to a login form or the application itself after selecting to logout.

This is required by the [ASD STIG](https://stigviewer.com/stigs/application_security_and_development/2025-02-12/finding/V-222392) and would be great to solve at the platform (Keycloak) layer if possible.

A few places to consider this. If the implementation is in Keycloak hopefully this would cover all cases, but may be user flows to consider:
- Keycloak logoff (from account console)
- Applications in uds-core with a UI (mainly Grafana)
- Authservice protected applications (this will likely need to be docs on how to implement a "logoff button"?)

slaskawi:
## Description

This Pull Request introduces the logout confirmation dialog displayed by Keycloak when the session is destroyed.

⚠️ This is a stop-gap solution until we implement this feature directly in Keycloak, see https://github.com/keycloak/keycloak/discussions/44315

## Implementation details

Along with the to the [OpenID Connect RP-Initiated Logout 1.0](https://openid.net/specs/openid-connect-rpinitiated-1_0.html), Keycloak expect 3 parameters at the logout: `post_logout_redirect_uri`, `id_token_hint` and `client_id` to perform a frictionless logout with no additional confirmation dialogs. The solution removes  the`post_logout_redirect_uri` query param and rewrites it as `uds_post_logout_redirect_uri` (so it still can be read by the downstream services if needed). This forces Keycloak to render the confirmation dialog within the "info.ftl" page.

At the same time, the https://github.com/defenseunicorns/uds-identity-config/pull/726 provides a modified `info.ftl` template that adds a "Back to sign in" button that tries "to guess" what the URL should be based on (on this order):

1. Whether Keycloak is in the middle of a redirect process - based on `pageRedirectUri`
2. Whether it's a regular confirmation from an action (e.g. a user is configuring OTP) - based on `actionUri`
3. If a Client has set the Home URL - based on `client.baseUrl`
4. The JavaScript tries to guess it from either `uds_post_logout_redirect_uri` or `post_logout_redirect_uri`
5. If all fails - the last fallback is to the sso.uds.dev (or sso in a tenant gateway)

## Related Issue

Fixes https://github.com/defenseunicorns/uds-identity-config/issues/689

## Type of change

- [x] New feature (non-breaking changhttps://github.com/defenseunicorns/uds-identity-config/issues/689)

## Steps to Validate

- Install UDS Core as ususal
- Navigate to Grafana
- Hit the logout button
- Ensure you see the logout confirmation message

## Checklist before merging

- [x] Test, docs, adr added or updated as needed- Install UDS Core as ususal


test/playwright/grafana.test.ts
```diff
@@ -158,7 +158,7 @@ test("validate logout functionality", async ({ page }) => {
     // First to /login/generic_oauth, then to SSO
     await page.waitForURL(/.*sso\.uds\.dev.*/, { timeout: 15000 });
 
-    // Verify we're at the Keycloak login page
-    await expect(page.locator('h3:has-text("Sign in")')).toBeVisible({ timeout: 10000 });
+    // Verify we're at the Keycloak logout confirmation page
+    await expect(page.locator('p:has-text("You are logged out")')).toBeVisible({ timeout: 10000 });
```
slaskawi: ⚠️ Note for reviewers

We can't check the back link here as we'd need a new UDS Identity Config image. This can be added as a follow up (if needed).

src/keycloak/chart/templates/logout-confirmation-envoyfilter.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+                -- No query string => nothing to strip
+                if #query == 0 then
+                  return
+                end
+
+                -- Replace post_logout_redirect_uri with uds_post_logout_redirect_uri in the query parameters
+                local new_params = {}
+                local logout_redirect_value = nil
+                for param in string.gmatch(query, "[^&]+") do
```
chance-coleman: Edge case that i think is totally fine, but wanted to call it out
1. Multiple `post_logout_redirect_uri` params
   - Your code keeps the last one it sees, drops all previous ones.

src/keycloak/chart/templates/logout-confirmation-envoyfilter.yaml
```diff
@@ -0,0 +1,87 @@
...
+        operation: INSERT_BEFORE
+        value:
+          name: envoy.filters.http.lua
+          typed_config:
+            "@type": type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua
+            inlineCode: |
+              function envoy_on_request(request_handle)
+                local headers = request_handle:headers()
+                local path = headers:get(":path")
+                local host = headers:get(":authority")
```
chance-coleman, line 30: I'm trying to remember if `:authority` sometimes includes a port? if it does i think it might fail the equality check and this filter will no-op. Might be an over optimization, happy to accept the edge case.

chance-coleman (commented):
Tested this locally, everything is looking good. I'd be happy approving this as is. just a couple small comments.

src/keycloak/chart/values.yaml
```diff
@@ -27,6 +27,9 @@ additionalGatewayNamespaces: []
 # Should only be disabled if this is incorrectly blocking paths you expect to be allowed
 pathParameterProtection: true
 
+# Always enforce a logout confirmation screen by stripping post_logout_redirect_uri from logout URLs
+logoutConfirmation: true
```
mjnagel: I think enabling this by default makes sense since we want to default to security/compliance. Wondering if there's any risk to doing this now vs giving it a release+ before it's default? I guess worst case it adjusts the logout experience and someone could disable so probably safe? Just unsure if there might be more applications/edge cases we want to validate here.

src/keycloak/chart/templates/logout-confirmation-envoyfilter.yaml
mjnagel: Out of curiosity I tried to force a port in here and couldn't with our dev setup. I think it's possible, but unlikely? Would depend on a client (grafana, etc) having the port in their logout URL, and then anything in the middle (browser, external loadbalancer) not trimming that. Probably an acceptable behavior to ignore this case imo, or you could make the `is_keycloak_host` check match those too by just doing a string match. Feel like that's not necessary but 🤷

src/keycloak/chart/templates/logout-confirmation-envoyfilter.yaml
```diff
@@ -0,0 +1,87 @@
...
+                  -- the name "uds_post_logout_redirect_uri" so that downstream apps can still read it.
+                  if key == "post_logout_redirect_uri" then
+                    logout_redirect_value = string.match(param, "^[^=]+=(.*)$") or ""
+                  else
+                    table.insert(new_params, param)
+                  end
+                end
+                if logout_redirect_value ~= nil then
+                  table.insert(new_params, "uds_post_logout_redirect_uri=" .. logout_redirect_value)
+                end
```
mjnagel: I think this is okay and works, but reading it did take me some time to process. Currently we're doing a loop over all query params, then removing the one we don't want and adding our custom one.

I think we could more simply do something like:
```lua
local new_query = string.gsub(query, "post_logout_redirect_uri=", "uds_post_logout_redirect_uri=")
```

I don't really think performance is a huge change, but for code simplicity this feels way easier to follow? It might have some (acceptable) edge cases similar to what Chance called out about multiple logout uris (we'd maintain multiple), but I think it's straightforward?

Okay if we keep as-is too, this is ultimately just a readability comment and Lua is great fun to read either way 😛

mjnagel (commented):
Overall LGTM I think. Two small comments but neither are required changes/blocking.

src/keycloak/chart/templates/logout-confirmation-envoyfilter.yaml
slaskawi: I refactored this piece of code and changed this to a `string.find` check to ensure ports are supported.

src/keycloak/chart/templates/logout-confirmation-envoyfilter.yaml
slaskawi: This piece have been completely refactored and should now take all the duplicated query params into consideration.

src/keycloak/chart/values.yaml
slaskawi: Good point - I switched it back to false.

src/keycloak/chart/templates/logout-confirmation-envoyfilter.yaml
slaskawi: This is actually a very good suggestion! I update the code to use this approach.

slaskawi:
Closing in favor of https://github.com/keycloak/keycloak/pull/44463

Issue #2110, opened 2025-11-05
Provide experimental gateway api CRD installation option in uds-core-istio-common


andygodish:
### Is your feature request related to a problem? Please describe.

My team is working with a custom app integration that is utilizing the experimental set of CRDs provided by the [kubeernets-sigs/gateway-api project](https://github.com/kubernetes-sigs/gateway-api/tree/main/config/crd/experimental). We install the default kustomization provided in each release as an install script in the same way that the standard install takes place during the deployment of the [core-base package](https://github.com/defenseunicorns/uds-core/blob/main/src/istio/common/zarf.yaml#L102). 

Instead of targeting the standard file, I would like to download the CRDs defined in the `experimental-install.yaml`. 

```yaml
variables:
  - name: GATEWAY_CRD_INSTALLTION
    default: standard

- name: gateway-api-crds
    required: true
    manifests:
      - name: gateway-api-crds
        files:
          - https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.4.0/${GATEWAY_CRD_INSTALLTION}-install.yaml
```

### Describe the solution you'd like

- **Given** I provide the core base package an override configured in the uds-core-istio-common package as a zarf variable,
- **When** the package is created, zarf should read in the input and execute either the standard or experimental installation scripts provided by the upstream project
- **Then** my desired set of CRDs are installed.

### Describe alternatives you've considered

Currently, we have a helper script that reinstalls the experimental CRDs following each upgrade of UDS Core in our clusters.

mjnagel:
Given that this is to support a custom app, is there a reason you wouldn't just add these extra CRDs to the custom app zarf package? Are these entirely net-new CRDs that you need (like the xlistenersets, xmeshes, etc)? Or do you also need to replace some of the core provided ones with a newer/different version?

The challenge with supporting this is that this would be at create time - so you'd need to rebuild Core to take advantage of this since these manifests are pulled into the package at create time. While we could expose a variable for this (or really a [template](https://docs.zarf.dev/ref/create/#package-templates)), you'd still need to rebuild Core to get the experimental stream of CRDs. It feels like you gain very little from an exposed template - it slightly changes the UX in how the package gets modified, but you still aren't using the published/pre-created Core package.

I'm also unsure if/how using the experimental stream might affect Istio Waypoints that we use in Core, and would be hesitant to support this out of the box without testing.

mjnagel:
[...]

PR #2165, opened 2025-11-21
chore: making the experimental gateway-api-crds the default

andygodish:
## Description

Sets the experimental gateway api crds as the default for installations of the Istio "gateway-api-crds" component.

> [From the docs:](https://gateway-api.sigs.k8s.io/guides/#install-experimental-channel) The experimental release channel includes everything in the standard release channel plus some experimental resources and fields. This includes TCPRoute, TLSRoute, and UDPRoute.

Installation of the experimental CRDs via a separate package after the uds core deployment causes a helm ownership conflict on the CRDs that are common between the standard and experimental channels. 
...

## Related Issue

Relates to #2110 

## Type of change

## Checklist before merging


mjnagel (approved):
Istio actually documents the experimental CRDs as the [recommended install method](https://github.com/istio/istio.io/blob/48288cefd5dcc254318b06d1b9434e37967172c8/content/en/boilerplates/gateway-api-install-crds.md?plain=1#L8) now (seems to have coincided with 1.28.0) so this seems fine to me. As noted it's just additional fields/CRDs which come with the caveat of maybe being subject to future breaking changes. In Core I think we will _typically_ only leverage the standard feature set.

Kicked off our IAC testing to just confirm there's no weird behaviors across other cluster types.

chance-coleman:
Going to wait till after break to merge this in. Talks on the side with Istio to see if this was their expectation and/or if we should be aware of any side effects.

PR #2166, opened 2025-11-21
feat: add support for metrics collection on authservice protected services

### LINKED ISSUE #2105: Reconsider how to allow metrics scraping on authservice protected applications
This is related to/a follow-on to https://github.com/defenseunicorns/uds-core/issues/645

When developing a mission app, I would like offload auth to Authservice, but I also want to expose metrics from my app and have these picked up by Prometheus. Currently Authservice protection is absolute, and managed by the UDS Operator, meaning that Prometheus scraping will be blocked with no clear options to bypass this protection.

There are some alternative approaches to make this work:
- Migrate metrics to a separate pod/service without authservice protection
- Implement auth in the application layer (not using authservice)

However neither of these are great experiences.

We should consider automatically allowing Prometheus to bypass authservice protection, specifically and only for the metrics path/method (for example, GET `/metrics`). This could be done _only_ for monitors declared in `monitor` in the Package spec, and could be opt-in or opt-out if needed, however it may be simpler to to just automatically do it. Alternatively there may be a mechanism to have Prometheus scrape the endpoint authenticated, but I am not sure if this is possible.

Considerations:
- We want to avoid opening pandora's box in terms of "exclude this, exclude that" from Authservice - authservice should still be focused on simple protection only, and this issue/change would only focus on enabling metrics with that protection.
- There may be some requirements for this to work from the operator side - such as identical labels for protection and metrics - so that we can easily determine when "filtering out" prometheus traffic is possible and necessary.

mjnagel:
## Description

Implements modifications to authorizationpolicies to allow Prometheus to scrape metrics endpoints. This follows the [design](https://www.notion.so/defense-unicorns/Authservice-Metrics-Exemption-2b1e512f24fc80c28484ecf15fead1a9?source=copy_link) to (1) ignore metrics paths/ports on the CUSTOM authpol and (2) allow prometheus principal access to the metrics ports/paths only while leaving all other ports/paths protected as normal.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/2105

## Type of change

## Steps to Validate

Deploy slim-dev plus monitoring and the test resources and navigate to podinfo.uds.dev . Confirm any behavior that may be relevant such as protection on the main endpoint, metrics blocked by default, prometheus targets healthy.

```console
uds run test:single-layer --set layer=monitoring --set flavor=unicorn
uds run -f src/test/tasks.yaml create-deploy --with validate_egress=false
```

## Checklist before merging


src/pepr/operator/controllers/keycloak/authservice/authorization-policy.spec.ts
```diff
@@ -75,4 +205,94 @@ describe("authorization-policy.ts", () => {
...
+      labelSelector,
+      name,
+      namespace,
+      false,
+      undefined,
+      monitorExemptions,
+    );
+
+    const rules = (pol.spec!.rules ?? []) as IstioRule[];
+    expect(rules.length).toBeGreaterThanOrEqual(1);
```
chance-coleman: Should this be:
```
expect(rules.length).toBe(2);
```
You assert rules.length >= 1 but then unconditionally access rules[1]. If the implementation ever returns a single combined rule (or fewer rules than expected), this will silently access undefined and cause a confusing test failure.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -248,9 +374,35 @@ async function purgeOrphanPolicies(generation: string, namespace: string, pkgNam
...
+ */
+function computeMonitorExemptions(
+  pkg: UDSPackage,
+  labelSelector: Record<string, string>,
+): MonitorExemption[] {
+  const monitors: Monitor[] = pkg.spec?.monitor ?? [];
+  const out: MonitorExemption[] = [];
+
+  for (const m of monitors) {
+    const sel: Record<string, string> = m.podSelector ?? m.selector;
```
chance-coleman, line 392: im forgetting, is it possible to have neither podSelector or selector? this could potentially result in `undefined` at runtime.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -48,14 +50,117 @@ function setPolicyTarget(
...
+    rules.push({
+      to: nonMetricsOps,
+      when: [
+        {
+          key: "request.headers[authorization]",
+          notValues: ["*"],
+        },
+      ],
+    });
+  }
```
chance-coleman, line 156: nit: both of the rules arrays push in very similar objects, could remove some duplication with a helper function. I don't feel very strongly just noticed it.

chance-coleman (commented):
generally looks good to me, just a couple small comments

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.spec.ts
mjnagel: Looking at this - probably better to make this safer/clearer on failures, good catch.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
mjnagel: Our schema [requires selector](https://github.com/defenseunicorns/uds-core/blob/authservice-metrics/src/pepr/operator/crd/sources/package/v1alpha1.ts#L248) so we should be good here :)

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
mjnagel: Collapsing some of this, into shared `const`s for the `when` piece and the `source` pieces in the jwt policy. I think it makes it a bit clearer to read (hopefully 😅).

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.spec.ts
mjnagel: Changed in https://github.com/defenseunicorns/uds-core/pull/2166/commits/2335424827d10126738436576fce569140c9027f

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
mjnagel: Changed in https://github.com/defenseunicorns/uds-core/pull/2166/commits/2335424827d10126738436576fce569140c9027f - I also cleaned up some conditionals that were confusing/duplicative. Hopefully all makes sense still, I think the rule from/to is easier to read on the jwt policy with the vars in there.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
```diff
@@ -248,9 +368,35 @@ async function purgeOrphanPolicies(generation: string, namespace: string, pkgNam
...
+ */
+function computeMonitorExemptions(
+  pkg: UDSPackage,
+  labelSelector: Record<string, string>,
+): MonitorExemption[] {
+  const monitors: Monitor[] = pkg.spec?.monitor ?? [];
+  const out: MonitorExemption[] = [];
+
+  for (const m of monitors) {
+    const sel: Record<string, string> = m.podSelector ?? m.selector;
```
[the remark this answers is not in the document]
mjnagel: Selector is required in the schema so I'd lean towards not adding an unused fallback here?

docs/reference/configuration/single-sign-on/auth-service.md
```diff
@@ -86,6 +86,12 @@ When using `network.expose` with protected services:
...
 
+## Metrics Scraping
+
+If you enable Authservice for a workload (using `sso[].enableAuthserviceSelector`) and also configure Prometheus monitors for the same pods (using the `monitor` field on the `Package`) the operator will adjust protection on the workload to ensure that Prometheus can scrape the configured metrics endpoints. Due to limitations with Istio AuthorizationPolicy resources, other clients hitting the metrics endpoints will not receive a redirect for acquiring a token, and must already present a token with the initial request. All endpoints will remain protected behind SSO, with the exception of Prometheus being allowed to scrape metrics directly without an SSO flow.
+
+In order to ensure that metrics collection works, the `monitor[].podSelector` must be identical to the `sso[].enableAuthserviceSelector` for the protected workload. This allows the operator to identify the proper protection to allow Prometheus scraping to occur.
```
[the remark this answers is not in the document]
joelmccoy: I think this is an accurate statement and might be good to add in.  As discussed in a dm, the descriptions for the monitor `podSelector` vs `selector` can probably be updated to have a more accurate description of the difference between these two: https://github.com/defenseunicorns/uds-core/blob/411e22bb13b2b06c2fbfe3788be1dedcfc40378d/src/pepr/operator/crd/sources/package/v1alpha1.ts#L266-L281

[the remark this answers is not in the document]
mjnagel: This is fair enough - I need to clarify the descriptions around `selector` vs `podSelector` while I'm at it...

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.ts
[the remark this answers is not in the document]
joelmccoy: selector is a required field for monitor so i don't think this is necessary

docs/reference/configuration/single-sign-on/auth-service.md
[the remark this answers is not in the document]
mjnagel: Updated.

src/pepr/operator/controllers/keycloak/authservice/authorization-policy.spec.ts
```diff
@@ -29,6 +55,110 @@ describe("authorization-policy.ts", () => {
...
+      name,
+      namespace,
+      false,
+      undefined,
+      ex,
+    );
+    const sidecarRule = sidecarPol.spec!.rules?.[0] as IstioRule;
+    const toList = (sidecarRule.to ?? []) as IstioTo[];
+    expect(toList.length).toBe(0);
+  });
```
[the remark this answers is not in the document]
mjnagel: Added.

joelmccoy (approved):
LGTM - tested/verified locally

chance-coleman (approved):

PR #2167, opened 2025-11-21
feat: add centralized trust bundle cert management

### LINKED ISSUE #2126: Implement Centralized Trust Bundle Management in UDS Core Operator
Based on the design doc [here](https://www.notion.so/defense-unicorns/Central-Trust-Bundle-Management-299e512f24fc8025a40ff2559b2e571d?source=copy_link).

**Summary:**  
Add centralized trust bundle management to UDS Core to unify certificate trust across all namespaces containing UDS Packages.

**Details:**  
- Implement Operator Logic called out in design doc
- Add docs for usage of this feature
- Update Private-PKI tasks/workflows to use this feature

joelmccoy:
## Description

Implements the Centralized Trust Bundle management feature according to [design doc](https://www.notion.so/defense-unicorns/Central-Trust-Bundle-Management-299e512f24fc8025a40ff2559b2e571d?source=copy_link).

## Related Issue

Fixes #2126 

## Type of change

## Steps to Validate
- `uds run -f tasks/test.yaml uds-core-private-pki`
  -  can toggle/swap out values in the ClusterConfig manually to ensure expected reconciliation for configmaps in cluster

## Checklist before merging


src/pepr/operator/crd/sources/cluster-config/v1alpha1.ts
```diff
@@ -9,16 +9,36 @@ export const v1alpha1: V1CustomResourceDefinitionVersion = {
...
+    },
+  ],
+  subresources: {
+    status: {},
+  },
   schema: {
     openAPIV3Schema: {
       type: "object",
       properties: {
-        metadata: {
```
joelmccoy, line 16: note to self: add this back in when: https://github.com/defenseunicorns/kubernetes-fluent-client/issues/939 is resolved

src/pepr/operator/crd/validators/clusterconfig-validator.ts
```diff
@@ -17,10 +18,33 @@ export async function validateCfgUpdate(req: PeprValidateRequest<ClusterConfig>)
...
+    );
+
+    if (!certBlocks || certBlocks.length === 0) {
+      throw new Error("ClusterConfig: No valid certificates found in bundle");
+    }
+
+    // Validate each certificate in the bundle
+    certBlocks.forEach((certPem, index) => {
+      try {
+        new X509Certificate(certPem);
```
joelmccoy, line 44: note: add in a validation with the crypto library to ensure that provided values for CA_BUNDLE_CERTS are proper certs

docs/reference/configuration/trust-management/private-pki.md
joelmccoy, line 1: note to self: add a uds docs redirect link for this

tasks/test.yaml
```diff
@@ -232,3 +233,4 @@ tasks:
           uds zarf tools kubectl label namespace authservice-ambient-test-app zarf.dev/agent=ignore
           uds zarf tools kubectl apply -f src/test/app-ambient-authservice-tenant.yaml
       - task: private-pki:private-pki-tests
+      - task: trust-bundle:trust-bundle-tests
```
joelmccoy, line 236: note: I decided to keep these trust bundle e2e tests in the private-pki nightly workflow. if we feel that we want to run them as standard e2e tests that is fine too, just lmk.  We will just have to update our k3d-standard bundle deploy to enable public CA certs so we can test this.

test/vitest/network.spec.ts
```diff
@@ -40,76 +39,6 @@ async function getPodName(namespace: string, labelSelector: string): Promise<str
   }
 }
 
-// Execute commands inside a pod
-async function execInPod(
```
joelmccoy, line 44: note: moved this into a helper function folder so it can be reused

test/vitest/vitest.config.js
```diff
@@ -11,6 +11,7 @@ export default defineConfig({
     environment: 'node',
     globalSetup: ['./vitest.setup.js'],
     include: ['**/*.spec.ts'],
+    exclude: ['trust-bundle/**'],
```
joelmccoy, line 14: note: e2e vitest flow was running the trust-bundle ones as well, so had to explicitly exclude this

src/pepr/operator/controllers/config/config.ts
```diff
@@ -240,20 +409,30 @@ function areKubeNodeCidrsEqual(newCidrs: string[] = [], currentCidrs: string[] =
...
+ *
+ * @param reason Description of what triggered the authservice update
+ */
 async function performAuthserviceUpdate(reason: string) {
   const authserviceUpdate: AuthServiceEvent = {
     name: "global-config-update",
     action: Action.UpdateGlobalConfig,
     // Base64 decode the CA cert before passing to the update function
-    trustedCA: atob(UDSConfig.caCert),
+    trustedCA: atob(UDSConfig.caBundle.certs),
```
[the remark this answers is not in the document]
joelmccoy: we already validate this in the config validator

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (dodCerts) {
+      certs.push(dodCerts);
+    }
+  }
+
+  // Add public certs if included
+  if (UDSConfig.caBundle.includePublicCerts && UDSConfig.caBundle.publicCerts) {
+    const publicCerts = atob(UDSConfig.caBundle.publicCerts);
+    if (publicCerts) {
+      certs.push(publicCerts);
```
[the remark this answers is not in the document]
joelmccoy: the values of are guaranteed to be base64 from the validator, ignoring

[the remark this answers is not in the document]
joelmccoy: the values of are guaranteed to be base64 from the validator, ignoring

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -0,0 +1,189 @@
...
+      log.debug("No existing CA bundle ConfigMaps found to update");
+      return;
+    }
+
+    // Update each ConfigMap with the new CA bundle content
+    for (const configMap of caBundleConfigMaps.items) {
+      if (!configMap.metadata?.name || !configMap.metadata?.namespace) {
+        log.warn(`ConfigMap missing name or namespace, skipping update`);
+        continue;
+      }
```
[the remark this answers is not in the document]
joelmccoy: this check is just here for type safety. shouldn't be hit.  doesn't need to be tested.

schemas/package-v1alpha1.schema.json
```diff
@@ -42,6 +46,50 @@
...
+        "key": {
+          "type": "string",
+          "description": "The key name inside the ConfigMap (default: ca-bundle.pem)"
+        },
+        "labels": {
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          },
+          "description": "Additional annotations to apply to the generated ConfigMap (default: {})\nAdditional labels to apply to the generated ConfigMap (default: {})\nLabels to match pods in the namespace to apply the policy to. Leave empty to apply to all pods in the namespace\nThe labels to apply to the policy\nDeprecated: use selector\nDeprecated: use remoteSelector\nThe remote pod selector labels to allow traffic to/from\nSpecifies attributes for the client.\nLabels to match pods to automatically protect with authservice. Leave empty to disable authservice protection\nConfiguration options for the mapper.\nAdditional annotations to apply to the generated secret, can be used for pod reloading with a selector\nAdditional labels to apply to the generated secret, can be used for pod reloading\nA template for the generated secret"
```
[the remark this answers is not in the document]
joelmccoy: this is expected and auto generated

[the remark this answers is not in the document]
mjnagel: This one is unfortunately accurate, but also an upstream bug: https://github.com/defenseunicorns/kubernetes-fluent-client/issues/810

May be worth looking at whether we need to fix this in KFC or in Core with post-processing or using a different tool for the schema at this point. Not an issue for this specific PR though.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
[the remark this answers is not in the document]
joelmccoy: these values should be validated as proper base64 already. dismissing

src/pepr/operator/crd/sources/cluster-config/v1alpha1.ts
```diff
@@ -92,7 +140,7 @@ export const v1alpha1: V1CustomResourceDefinitionVersion = {
               required: ["allowAllNsExemptions"],
             },
           },
-          required: ["expose", "policy"],
+          required: ["expose", "policy", "caBundle"],
```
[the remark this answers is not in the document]
joelmccoy: fixed

src/pepr/operator/crd/validators/clusterconfig-validator.ts
```diff
@@ -17,10 +18,33 @@ export async function validateCfgUpdate(req: PeprValidateRequest<ClusterConfig>)
...
+      } catch (e) {
+        throw new Error(`ClusterConfig: Invalid certificate at index ${index}: ${e.message}`);
```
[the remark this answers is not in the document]
joelmccoy: fixed

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -0,0 +1,189 @@
...
+      log.debug(`Updating CA bundle ConfigMap ${configMapName} in namespace ${namespace}`);
+
+      // Find the first key for the CA bundle data
+      const dataKeys = Object.keys(configMap.data || {});
+      const caBundleKey = dataKeys[0];
+
+      if (!caBundleKey) {
+        log.warn(`No suitable key found in ConfigMap ${configMapName}, skipping update`);
+        continue;
+      }
```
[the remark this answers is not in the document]
joelmccoy: added test

src/pepr/operator/controllers/ca-bundles/ca-bundle.spec.ts
```diff
@@ -0,0 +1,501 @@
...
+      await caBundleConfigMap(mockPackage, "test-namespace");
+
+      expect(mockK8sApply).toHaveBeenCalledWith(
+        expect.objectContaining({
+          data: {
+            "custom-ca-bundle.pem": "",
+          },
+        }),
+      );
+    });
```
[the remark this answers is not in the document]
joelmccoy: added test

bundles/k3d-standard/uds-private-pki-config.yaml
```diff
@@ -3,8 +3,10 @@
 
 variables:
   core:
-    # CA certificate for Authservice
-    CA_CERT: "PLACEHOLDER_CA_CERT"
```
slaskawi, line 7: Shall we call this out as a breaking change?

hack/dev-manifests/uds-ca-certs.yaml
```diff
@@ -0,0 +1,11 @@
+# Copyright 2025 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: uds-ca-certs
+  namespace: pepr-system
+data:
+  dodCACerts: ""
```
slaskawi, line 10: ConfigMap has a limitation to 1 MB. I did some quick math and for a Base64 encoded certs that are quite large (2 KB, 2.72 KB after encoding), we can store around 370 of these. 

Even though it appears quite large, shall we mention somewhere in the docs that there is certain ceiling there and administrators should be aware of it?

src/pepr/zarf.yaml
```diff
@@ -73,3 +86,13 @@ components:
       - name: module
         valuesFiles:
           - values.yaml
+    actions:
```
slaskawi, line 89: Is this necessary? I though Zarf automatically checks the Ready field.

bundles/k3d-standard/uds-private-pki-config.yaml
```diff
@@ -30,6 +32,6 @@ variables:
     KEYCLOAK_EXTRA_VOLUMES:
       - name: ca-certs
         configMap:
-          name: private-ca
+          name: uds-trust-bundle
     KEYCLOAK_TRUSTSTORE_PATHS:
       - "/tmp/ca-certs"
```
mjnagel, line 37: This doesn't have to be part of this PR, but would love for us to think about adding `optional` mounts to all the core services by default to mount the CA configmap. This would be an out of the box value add for the time being while we figure out next steps for the operator (mutating mounts?).

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -0,0 +1,189 @@
...
+        // Set owner reference to the UDS Package for proper cleanup
+        ownerReferences: ownerRefs,
+      },
+      data: {
+        [configMapKey]: caBundleContent,
+      },
+    };
+
+    // Apply the ConfigMap
+    await K8s(kind.ConfigMap).Apply(configMapManifest);
```
mjnagel: nit: probably want to force apply here (this is pretty common with [other applies](https://github.com/defenseunicorns/uds-core/blob/2af1957e0ff5f5575bb8c30f94173f4e6f74a274/src/pepr/operator/controllers/network/policies.ts#L297-L299)). If we don't, we can end up with conflicts/issues when someone does manual edits (kubectl, etc). ([ref conflicts doc](https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts))

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -0,0 +1,189 @@
...
+        metadata: {
+          ...configMap.metadata,
+          managedFields: undefined,
+        },
+        data: {
+          [caBundleKey]: updatedCaBundleContent,
+        },
+      };
+
+      await K8s(kind.ConfigMap).Apply(updatedConfigMap);
```
mjnagel: same comment here, probably should force apply.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -0,0 +1,189 @@
...
+      // Find the first key for the CA bundle data
+      const dataKeys = Object.keys(configMap.data || {});
+      const caBundleKey = dataKeys[0];
+
+      if (!caBundleKey) {
+        log.warn(`No suitable key found in ConfigMap ${configMapName}, skipping update`);
+        continue;
+      }
+
+      // Update the ConfigMap data, removing managedFields to avoid conflicts
```
mjnagel: I think this is _okay_, but might want to diff `updatedCaBundleContent` against the current content to prevent excessive applies here.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -109,81 +172,183 @@ export async function handleCfgSecret(cfg: kind.Secret, action: ConfigAction) {
...
+  if (caBundle.certs === "###ZARF_VAR_CA_BUNDLE_CERTS###") {
+    caBundle.certs = "";
+  } else if (caBundle.certs === "###ZARF_VAR_CA_CERT###") {
+    caBundle.certs = "";
   }
 
-  if (UDSConfig.caCert !== expose.caCert) {
-    UDSConfig.caCert = expose.caCert || "";
+  // Load in the DoD and Public certs from the configmap
+  const caCertsConfigMap = await K8s(kind.ConfigMap).InNamespace("pepr-system").Get("uds-ca-certs");
```
mjnagel: Might want to try/catch this and re-throw a more clear error if it does fail. In my experience K8s get/apply failures can sometimes just timeout? and not return a helpful error (ref: https://github.com/defenseunicorns/kubernetes-fluent-client/issues/957). Probably would do something similar to [the handling in `loadUDSConfig`](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/config/config.ts#L193-L202).

src/pepr/operator/controllers/config/config.ts
```diff
@@ -109,81 +172,183 @@ export async function handleCfgSecret(cfg: kind.Secret, action: ConfigAction) {
...
-      UDSConfig.domain = "uds.dev";
+  try {
+    // Patch the status to Pending and set currentGeneration while we process the update
+    await K8s(ClusterConfig).PatchStatus({
+      metadata: {
+        name: cfg.metadata!.name,
+      },
+      status: {
+        phase: Phase.Pending,
+        observedGeneration: cfg.metadata!.generation,
```
mjnagel: We typically don't patch observedGeneration until [_after_ successfully reconciling](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/reconcilers/package-reconciler.ts#L130-L140). Was this intentional to do it before?

src/pepr/operator/controllers/config/config.ts
```diff
@@ -109,81 +172,183 @@ export async function handleCfgSecret(cfg: kind.Secret, action: ConfigAction) {
...
+    caBundle.certs = "";
   }
 
-  if (UDSConfig.caCert !== expose.caCert) {
-    UDSConfig.caCert = expose.caCert || "";
+  // Load in the DoD and Public certs from the configmap
+  const caCertsConfigMap = await K8s(kind.ConfigMap).InNamespace("pepr-system").Get("uds-ca-certs");
+
+  // Check if CA bundle ConfigMaps need updates based on configuration changes
+  const caBundleConfigMapsNeedUpdate =
```
chance-coleman: This is a bit confusing, having a hard time reading this. Wondering if it would make sense to break this out into a function of its own that uses a few if statements instead of a long ternary. For me it would a readability improvement, but mostly just preference.

src/pepr/uds-operator-config/values.yaml
```diff
@@ -12,11 +12,16 @@ cluster:
   attributes:
     clusterName: ""
     clusterTags: []
+  caBundle:
+    certs: "###ZARF_VAR_CA_BUNDLE_CERTS###"
+    includeDoDCerts: "###ZARF_VAR_CA_BUNDLE_INCLUDE_DOD_CERTS###"
+    includePublicCerts: "###ZARF_VAR_CA_BUNDLE_INCLUDE_PUBLIC_CERTS###"
   expose:
+    # Note caCert is deprecated, use caBundle.certs instead
+    caCert: "###ZARF_VAR_CA_CERT###"
```
mjnagel, line 21: I think we'll want/need to update the Istio CA Secret as well to handle this deprecation - https://github.com/defenseunicorns/uds-core/blob/ca-trust-bundle/src/pepr/uds-operator-config/templates/ca-secret.yaml

src/pepr/operator/controllers/config/config.ts
```diff
@@ -109,81 +172,183 @@ export async function handleCfgSecret(cfg: kind.Secret, action: ConfigAction) {
...
 }
 
-async function handleCAUpdate(expose: ConfigExpose, updateClusterResources?: boolean) {
+/**
+ * Handles updates to CA bundle configuration including DoD and public certificates
+ *
+ * @param caBundle The CA bundle configuration from the ClusterConfig
+ * @param updateClusterResources Whether to update cluster resources (ConfigMaps, etc.)
+ */
+async function handleCABundleUpdate(caBundle: ConfigCABundle, updateClusterResources?: boolean) {
```
mjnagel, line 229: Might be missing another mechanism where this happens, but I think we want/need a watch on the [uds-ca-certs configmap](https://github.com/defenseunicorns/uds-core/blob/ca-trust-bundle/src/pepr/uds-operator-config/templates/uds-ca-certs.yaml) for updates that change the public/dod certs, but not the clusterconfig (likely would call this same function)?

src/pepr/uds-operator-config/templates/clusterconfig.yaml
```diff
@@ -18,6 +18,12 @@ spec:
       {{- toYaml .Values.cluster.attributes.clusterTags | nindent 6 }}
     {{- end }}
   {{- end }}
+  caBundle:
+    {{- if or .Values.cluster.caBundle.certs .Values.cluster.expose.caCert }}
+    certs: {{ .Values.cluster.caBundle.certs | default .Values.cluster.expose.caCert | quote }}
+    {{- end }}
```
mjnagel, line 24: LGTM from a template perspective, but marking a note for myself to test the deprecated value e2e.

mjnagel (commented):
Initial review - need to deploy locally and validate the deprecated value at least but this looks pretty great overall.

chance-coleman (commented):
this is looking good to me, only really one place that tripped me up a bit.

src/pepr/operator/controllers/config/config.ts
joelmccoy: good callout. I think the way it is currently implemented if the CA certs for public/dod change it will not reconcile as it doesn't not change the cluster config. will work to get a watch added to the configmap as well

bundles/k3d-standard/uds-private-pki-config.yaml
joelmccoy: In order to not clutter this PR, I can cut a spin off fast follow issue and address it when this gets merged in.

Did you imagine the mounts happening by default?  I think I would need to confirm the behavior of providing an empty file mount for keycloak and grafana (which would be the default for the configmaps in default deploy of core).

Or did you imagine:
1. Having this behind a toggle helm value for keycloak and grafana?
2. Only auto mount if CA_BUNDLE_CERTS is set?

mjnagel: Hm, yeah I was thinking along the lines of 2 there - that this would only happen in cases where either custom certs or DoD/Public certs were set/configured. I actually might've had a poor assumption that I'm now reviewing in the code - I would've expected no configmap to be created if both DoD and Public are false, and no CA_BUNDLE_CERTS is set (that's why I mentioned optional mounts). I might make a separate comment on this in the relevant code spot - I'm not recalling if we discussed during the design process explicitly or not.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -0,0 +1,189 @@
...
+  const configMapKey = pkg.spec?.caBundle?.configMap?.key || DEFAULT_CONFIGMAP_KEY;
+  const configMapLabels = pkg.spec?.caBundle?.configMap?.labels || {};
+  const configMapAnnotations = pkg.spec?.caBundle?.configMap?.annotations || {};
+
+  // Always create ConfigMap - if no cert sources are available, create with empty content
+  try {
+    log.debug(`Reconciling CA Bundle ConfigMap for ${pkgName}`);
+
+    // Build the CA bundle content by combining available certs
+    const caBundleContent = buildCABundleContent();
```
mjnagel, line 42: Should we early return here if the caBundleContent is empty? Essentially a difference between making no configmap vs making an empty one.

If we do skip making empty configmaps we'll also want to account for deleting all configmaps if/when config changes to necessitate that.

src/pepr/uds-operator-config/values.yaml
joelmccoy: good catch. I also realized that i didn't add helm unit tests for the templating for this deprecation.  will add that in

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
joelmccoy: This wasn't called out in the design doc.... but I think I informally brought this up in a chat in gather when i was working through this.  

The problem with not creating the ConfigMaps by default is that when the clusterconfig is updated from not having `CA_BUNDLE_CERTS` set (or toggling public/dod) we need a way to reach out to all the namespaces and update the content (see [here](https://github.com/defenseunicorns/uds-core/blob/ca-trust-bundle/src/pepr/operator/controllers/config/config.ts#L232-L235)).  If the configmap doesn't exist (which is created from package reconciliation) we will need to trigger package reconciliation on all the packages on this update which felt like overkill?

I went down the route of just having all the configmaps created by default so we can list them and just update the contents with [this](https://github.com/defenseunicorns/uds-core/blob/ca-trust-bundle/src/pepr/operator/controllers/ca-bundles/ca-bundle.ts#L132) function.

Open to alternative approaches here if there is a better solution that I am not thinking of.

mjnagel: I don't think we'd need to re-reconcile all packages, we could just grab the list of packages in the cluster (single K8s Get) and then loop over the namespaces calling the CA configmap function? If you wanted to save that call could even in-memory keep a list of package namespaces, just have to make sure you handle add/delete to that list as needed.

joelmccoy: Ack. I think that approach makes sense. Will attempt to refactor this and take that approach.

src/pepr/operator/controllers/config/config.ts
joelmccoy: split out into a function

src/pepr/uds-operator-config/templates/clusterconfig.yaml
joelmccoy: added some helm tests for this as well. missed added those on initial pass

src/pepr/operator/controllers/config/config.ts
joelmccoy: added a new handler/watch for configmap updates

src/pepr/uds-operator-config/values.yaml
joelmccoy: added

src/pepr/operator/controllers/config/config.ts
joelmccoy: good catch.  removed patching the generation here

src/pepr/operator/controllers/config/config.ts
joelmccoy: added some error handling here

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
joelmccoy: good call. added that in

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
joelmccoy: added

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
joelmccoy: added

bundles/k3d-standard/uds-private-pki-config.yaml
joelmccoy: that totally makes sense given the discussion [here](https://github.com/defenseunicorns/uds-core/pull/2167#discussion_r2599964917). going to refactor the logic to not create configmaps unless those variables are set.

bundles/k3d-standard/uds-private-pki-config.yaml
mjnagel: We're leaving `CA_CERT` around, just deprecated so I don't think we need to. This removal is just in a CI only config file so not a downstream breaking change here.

src/pepr/zarf.yaml
mjnagel: Zarf will check for the Ready _condition_, but not the phase field by default. Think in the design doc the status stuff popped up towards the end of review/approval but we were leaning towards keeping status simple and not using conditions here. That would be a way to get rid of this action though if we wanted to.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
joelmccoy: Updated so that configmaps are not created if there is no content.  The refactor is in this [commit](https://github.com/defenseunicorns/uds-core/pull/2167/commits/c8ade8b32287df0432dbfb87538f2f94c67905ec).  Actually was way smoother than expected (Thanks for the suggestion!!). I just took the approach of listing all UDS Packages and calling `caBundleConfigMap` for the respective packages (was nicely reusable).  Also updated `caBundleConfigMap` to delete configmaps if no content is provided.

hack/dev-manifests/uds-ca-certs.yaml
joelmccoy: Yea, we did some math in the [design doc](https://www.notion.so/defense-unicorns/Central-Trust-Bundle-Management-299e512f24fc8025a40ff2559b2e571d?source=copy_link#299e512f24fc80bba8b9f6f86022d3a4) around this.  Realistically, we don't imagine users hitting the upper limit on the configmap.  We could put a callout in the docs about the upper limit, but I feel it is adding a caution to something that most likely won't happen?  I lean towards leaving it out but could be convinced to put it in

bundles/k3d-standard/uds-private-pki-config.yaml
joelmccoy: Have a fast follow issue [here](https://github.com/defenseunicorns/uds-core/issues/2203). Can pick this up once this feature is merged in.  The one thing to call out is we may need to think through upgrade paths for possible collisions on volume mounts for environments that already have extra volume mounts declared to solve this problem.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
mjnagel: LGTM in the code.

test/vitest/network.spec.ts
joelmccoy: https://github.com/defenseunicorns/uds-docs/pull/326

mjnagel (approved):

chance-coleman (approved):

docs/reference/configuration/trust-management/private-pki.md
joelmccoy: https://github.com/defenseunicorns/uds-docs/pull/326

bundles/k3d-standard/uds-private-pki-config.yaml
slaskawi: Perfect!

slaskawi (approved):

PR #2173, opened 2025-12-01
fix: install/use proper deps for check-ca-certs task

joelmccoy:
## Description

Fixes failed runs for check-ca-certs nightly workflow. See [example](https://github.com/defenseunicorns/uds-core/actions/runs/19809497884/job/56749064699#step:4:16).  The issue is that the deps aren't install in the github action workflow and that the task was using the top level package.json instead of the local script directory.

## Type of change

## Steps to Validate
- Tested here: https://github.com/defenseunicorns/uds-core/actions/runs/19828308383/job/56807418336?pr=2173

## Checklist before merging


mjnagel (approved):
Worked for me locally :)

Issue #2003, opened 2025-10-07
Keycloak scales rapidly down when modifying HPA to 1


slaskawi:
## Summary

@brianrexrode discovered that when changing HPA `minReplicas` and `maxReplicas` to 1, HPA ignores all safety barriers and scales down pretty much immediately. When Infinispan loses segments during the rapid scale down, Keycloak starts reporting errors. After certain timeout things stabilize. 

## Steps to reproduce

1. Connect to console via zarf connect keycloak.
2. Set HPA to 5
3. Wait until everything is ready
4. Set HPA to 1
5. Notice the rapid scale down (you can use `uds zarf connect` but most likely all Token Refreshes will fail due to a broken Infinispan cluster).

slaskawi:
This issue looks like a bug in Keycloak shutdown procedure - ignoring ongoing rebalance while closing the node off. 

I created https://github.com/keycloak/keycloak/issues/44620 and let's wait for reply before adding anything to the documentation.

slaskawi:
Putting this ticket back to backlog while waiting for https://github.com/keycloak/keycloak/issues/44620

slaskawi:
The primary tracking ticket is https://github.com/keycloak/keycloak/issues/43589. This functionality will be available in Keycloak  26.6 (end of March 2026).

slaskawi:
The primary Keycloak ticket (https://github.com/keycloak/keycloak/issues/43589) has been resolved and I successfully retested it with `26.6.1`. Closing as completed.

PR #2191, opened 2025-12-04
feat: clarify keycloak scaling down issues

slaskawi:
## Description

This Pull Request clarifies that scaling Keylcoak down rapidly is not recommended.

This is a stop-gap approach while waiting for https://github.com/keycloak/keycloak/issues/44620

## Related Issue

Relates to https://github.com/defenseunicorns/uds-core/issues/2003

## Type of change

## Steps to Validate

N/A

## Checklist before merging


docs/reference/configuration/resource-configuration-and-ha.md
```diff
@@ -144,6 +144,10 @@ packages:
               path: postgresql.password
 ```
 
+:::caution
+Scaling Keycloak down rapidly by modifying the replica count directly in the StatefulSet is not recommended and may result in data loss. This is a [known Keycloak limitation](https://github.com/keycloak/keycloak/issues/44620).
```
mjnagel, line 148: Is this also true if you _indirectly_ reduce the replica count by changing the HPA max replicas?

mjnagel (approved):

docs/reference/configuration/resource-configuration-and-ha.md
slaskawi: In our default HPA settings, we use custom Scale Down logic: https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/values.yaml#L357-L362

The only problem is when someone overrides it but I guess - that's a corner case.

PR #2194, opened 2025-12-04
feat: centralized ambient egress

### LINKED ISSUE #2064: Package spec.network.allow L7 (hostname) rule in one namespace breaks L4 (Anywhere) egress for pods in a different namespace
### Environment
Device and OS: Any
App version: Any that supports istio ambient mode
Kubernetes distro being used: EKS 1.33
Other:

### Steps to reproduce
I'm experiencing an issue where Package spec.network.allow rules in one namespace are negatively affected by spec.network.allow changes in a second, unrelated namespace. I encountered this with GitLab, but it appears to be a general issue relevant to any two applications in different namespaces.

I am running GitLab on top of uds-core. The GitLab Package CR is configured to allow all egress traffic for the registry pod using remoteGenerated: Anywhere. In this initial state, the registry pod can access external resources (like https.httpbin.org) without issue.
```
spec:
  network:
    allow:
    - description: Registry Storage
      direction: Egress
      remoteGenerated: Anywhere
      selector:
        app: registry
```

I then deployed an arbitrary application in a different namespace. This new application's Package CR includes a specific L7 egress rule, remoteHost: httpbin.org.
```
spec:
  network:
    allow:
    - direction: Ingress
      remoteGenerated: IntraNamespace
    - direction: Egress
      remoteGenerated: IntraNamespace
    - description: to httpbin.org
      direction: Egress
      port: 443
      remoteHost: httpbin.org
      selector:
        app: myapp
      serviceAccount: default
```

### Expected result
Both the new application's pod and the GitLab registry pod should be able to reach https://httpbin.org.

### Actual Result
The new application's pod can successfully reach https://httpbin.org, but the GitLab registry pod's access to the same URL is now broken and fails with an SSL_ERROR_ZERO_RETURN.

```
[git@gitlab-registry-xxxxxxxxxx-xxxxx /]$ curl -vkL https://httpbin.org
*   Trying 240.240.0.27:443...
* Connected to httpbin.org (240.240.0.27) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
*  CAfile: /etc/pki/tls/certs/ca-bundle.crt
* TLSv1.0 (OUT), TLS header, Certificate Status (22):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* OpenSSL SSL_connect: SSL_ERROR_ZERO_RETURN in connection to httpbin.org:443 
* Closing connection 0
curl: (35) OpenSSL SSL_connect: SSL_ERROR_ZERO_RETURN in connection to httpbin.org:443
```

### Severity/Priority
High. This suggests that the introduction of a single L7-based (remoteHost) rule anywhere in the cluster fundamentally changes the mesh's egress behavior by deploying a waypoint, [which is expected behavior](https://uds.defenseunicorns.com/reference/configuration/service-mesh/egress/#ambient-mode). It appears this new waypoint does not correctly honor the L4 remoteGenerated: Anywhere rule from the GitLab package, causing it to block the registry pod's traffic.

### Additional Context
From @chance-coleman, this is expected behavior from an istio perspective. [Source](https://www.github.com/istio/istio/issues/57661#issuecomment-3293464117)

chance-coleman:
## Description

- **Centralizes Ambient egress control per host.** For each external host:
  - Creates a shared `ServiceEntry` in `istio-egress-ambient`.
  - Creates a centralized `AuthorizationPolicy` in `istio-egress-ambient` targeting that ServiceEntry with from-only rules.
- **Identity resolution:** Unions owners (host-specific) and participants (`remoteGenerated: Anywhere`) with SA-first principals and namespace fallback.
- **Safety:** Skips SE/AP creation when no identities are resolved to avoid transient allow; uses generation-based purge to remove prior resources (Gateway, ServiceEntry, AuthorizationPolicy).
- **Quality:** Deterministic resource names via `sanitizeWithLimit()`. Ambient constants centralized in `istio-resources.ts`.

## New AP flow (Ambient)

1) Build per-host desired state from in-memory package map (union of ports/protocols + contributing packages).
2) Resolve identities (owners + Anywhere participants). If non-empty:
   - Apply shared per-host `ServiceEntry` (bound to egress waypoint).
   - Apply centralized `AuthorizationPolicy` with from-only rules and `targetRef` to the host’s `ServiceEntry`.
3) Increment generation and purge prior-gen ambient resources.

## Related Issue

Fixes #2064 

## Type of change

## Steps to Validate
- `uds run test:uds-core-e2e`
- `npm run test:unit`
- `npm run test-uds-core-upgrade`

### Manual steps
- apply [these manifests](https://gist.github.com/chance-coleman/268be7341f287e1aaf3439647379f640)
- verify central resources
  - `kubectl -n istio-egress-ambient get se,ap | grep httpbin`  # expect ambient-se-httpbin-org, ambient-ap-httpbin-org
- get pods
  - `R=$(kubectl -n ambient-egress-restricted get pod -l app=restricted-curl -o name)`
  - `A=$(kubectl -n ambient-egress-anywhere get pod -l app=anywhere-curl -o name)`
- curl checks
  - Restricted: `kubectl -n ambient-egress-restricted exec -it $R -- sh -c 'curl -s -w " HTTP_CODE:%{http_code}\n" https://httpbin.org'`  # ALLOW (retry if 503)
  - Restricted: `kubectl -n ambient-egress-restricted exec -it $R -- sh -c "curl -s -w ' HTTP_CODE:%{http_code}\n' http://httpbin.org"`   # DENY
  - Restricted: `kubectl -n ambient-egress-restricted exec -it $R -- sh -c 'curl -s -w " HTTP_CODE:%{http_code}\n" https://api.github.com'` # DENY
  - Anywhere:   `kubectl -n ambient-egress-anywhere exec -it $A -- sh -c 'curl -s -w " HTTP_CODE:%{http_code}\n" https://httpbin.org'`    # ALLOW
  - Anywhere:   `kubectl -n ambient-egress-anywhere exec -it $A -- sh -c "curl -s -w ' HTTP_CODE:%{http_code}\n' http://httpbin.org"`     # DENY
  - Anywhere:   `kubectl -n ambient-egress-anywhere exec -it $A -- sh -c 'curl -s -w " HTTP_CODE:%{http_code}\n" https://api.github.com'`  # ALLOW (L4 Anywhere)

## Checklist before merging


src/pepr/operator/controllers/istio/egress-orchestrator.ts
chance-coleman, line 1: This file came out of a refactor to avoid circular import dependency between egress.ts and istio-resources.ts due to ingress and egress orchestration having been intertwined. Previously resulted in unit-test failures due to the dependency issues.

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -121,9 +121,9 @@ spec:
...
-* An Istio ServiceEntry, in the package namespace, which is used to define the external service that the workload can access.
-* An Istio AuthorizationPolicy, in the package namespace, which is used to enforce that only traffic from workloads using the selected service account can egress. If no `serviceAccount` is specified, the `default` service account is used.
-* A shared Istio Waypoint, in the `istio-egress-ambient` namespace, which is used to route the egress traffic.
+* A shared Istio ServiceEntry, in the `istio-egress-ambient` namespace, one per external host across all Ambient packages. This registers the external service (host and union of ports/protocols) and binds it to the egress waypoint.
+* A centralized Istio AuthorizationPolicy, in the `istio-egress-ambient` namespace, that targets the per-host ServiceEntry (not the waypoint Gateway) and ALLOWs owners and Ambient "Anywhere" participants (ServiceAccount-first principal, else namespace). Rules use only `from:` sources; the destination host is implied by the ServiceEntry target.
+* The shared Istio Waypoint (`Gateway`), in the `istio-egress-ambient` namespace, used to route the egress traffic.
```
joelmccoy: nit:
```suggestion
* The shared Istio Waypoint (`Gateway`), in the `istio-egress-ambient` namespace, used to route ambient egress traffic.
```

src/pepr/operator/controllers/utils.ts
```diff
@@ -51,6 +51,24 @@ export function sanitizeResourceName(name: string) {
   );
 }
 
+/**
+ * Ensures a resource name stays within Kubernetes max length; if over, truncate and add short hash suffix
+ * @param base Base resource name
+ * @param maxLen Maximum allowed length (default: 253 for Kubernetes)
+ * @returns Safely truncated name with hash suffix if needed
+ */
+export function sanitizeWithLimit(base: string, maxLen = 253): string {
```
mjnagel: I'm a bit confused why this exists, since `sanitizeResourceName` should slice down to 250 characters? I don't see any places passing in a `maxLen` so the default of 253 would mean most of this function is never reached I think?

src/pepr/operator/controllers/istio/auth-policy.ts
```diff
@@ -3,64 +3,62 @@
...
+  const principalSources = identities.saPrincipals.length
+    ? [{ source: { principals: identities.saPrincipals } }]
+    : [];
+  const namespaceSources = identities.namespaces.length
+    ? [{ source: { namespaces: identities.namespaces } }]
+    : [];
+
+  // Build rules: only sources are needed; targetRef scopes to the specific host via ServiceEntry
+  type Source = { source: { principals?: string[]; namespaces?: string[] } };
+  type Rule = { from: Source[] };
```
mjnagel: These should be using the upstream istio authpol types ideally. We should end up with the same type checking since we use the upstream `IstioAuthorizationPolicy` below, but feels weird to be making our own types when we could import them?

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -2,20 +2,55 @@
...
+    const name = pkg.metadata?.name;
+    if (!ns || !name) {
+      continue;
+    }
+    const id = `${name}-${ns}`;
+    const mode = pkg.spec?.network?.serviceMesh?.mode || Mode.Sidecar;
+    if (!contributing.has(id) || mode !== Mode.Ambient) continue;
+    for (const allow of pkg.spec?.network?.allow ?? []) {
+      if (allow.direction !== Direction.Egress) continue;
+      if (allow.remoteHost !== host) continue;
```
mjnagel: nit: these inverted checks threw me off for a bit, feels like we could just do:

```ts
      if (allow.direction == Direction.Egress && allow.remoteHost == host) {
```

to make it more direct and explicit about what we're matching (then keeping the next if/else for serviceAccount nested in this one).

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -2,20 +2,55 @@
...
-import { generateLocalEgressSEName, generateLocalEgressServiceEntry } from "./service-entry";
-import { HostResourceMap } from "./types";
+import { generateCentralAmbientEgressAuthorizationPolicy } from "./auth-policy";
+import { inMemoryAmbientPackageMap, remapAmbientEgressResources } from "./egress";
+import { ambientEgressNamespace, log, sharedEgressPkgId } from "./istio-resources";
+import { generateSharedAmbientServiceEntry } from "./service-entry";
 
-export const ambientEgressNamespace = "istio-egress-ambient";
-export const sharedEgressPkgId = "shared-ambient-egress-resource";
+export { ambientEgressNamespace, sharedEgressPkgId };
```
joelmccoy: is there a reason for re-exporting these?

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -42,6 +49,8 @@ export let lastReconciliationPackages: Set<string> = new Set();
 let sidecarGeneration = 0;
 let ambientGeneration = 0;
 
+// (legacy) istioEgressResources removed in favor of orchestrator implementation in egress-orchestrator.ts
```
mjnagel: probably can cleanup this comment? `istioEgressResources` also doesn't appear to have been removed from this file (it was removed from istio-resources).

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -42,6 +49,8 @@ export let lastReconciliationPackages: Set<string> = new Set();
...
+
```
joelmccoy: rouge comment? unless I am missing something?

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -113,13 +114,22 @@ async function reconcilePackageFlow(pkg: UDSPackage): Promise<void> {
...
+  // Count unique hosts contributed by this package (central APs are per-host)
   let numEgressAuthPols = 0;
   if (istioMode === Mode.Ambient && pkg.spec?.network?.allow) {
-    numEgressAuthPols = egressRequestedFromNetwork(pkg.spec!.network!.allow!).length;
+    const uniqueHosts = new Set(
+      egressRequestedFromNetwork(pkg.spec.network.allow)
+        .map(a => a.remoteHost)
+        .filter((h): h is string => typeof h === "string" && h.length > 0),
+    );
+    numEgressAuthPols = uniqueHosts.size;
```
joelmccoy: nit: I don't have a strong opinion, but wanted to ask the question in case somebody does.... Do we want to count the central auth policy as a count for this package?

I almost lean towards not counting it as it's extra logic and doesn't reflect the namespaced scope # of auth policies we would see in the package namespace

test/vitest/network.spec.ts
```diff
@@ -421,6 +421,35 @@ describe("Network Policy Validation", { retry: 2 }, () => {
...
+        "egress-ambient-2",
+        curlPodNameEgressAmbient2,
+        "curl",
+        GITHUB_HTTP,
+      );
+      const msg = `Egress Ambient HTTP github denied for egress-ambient-2 SA: stdout=${resp.stdout}, stderr=${resp.stderr}`;
+      expect(isResponseError(resp), msg).toBe(true);
+    },
+  );
+
```
joelmccoy, line 407: do we maybe also want to add a test for the bug that kicked off this feature?

i.e. Test that a `remoteGenerated: Anywhere` ns pod can hit a host that is specified in another namespace as `remoteHost`?

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -174,9 +174,21 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+      const hostAllowed = isAmbient && !!policy.remoteHost;
+      const anywhereAllowed =
+        isAmbient &&
+        policy.remoteGenerated === RemoteGenerated.Anywhere &&
+        policy.direction === "Egress";
+      if (!hostAllowed && !anywhereAllowed) {
+        return req.Deny(
+          "serviceAccount is only valid for Ambient mode when using remoteHost or remoteGenerated: Anywhere on Egress rules",
+        );
+      }
```
joelmccoy, line 191: might be good to add a test for new `remoteGenerated` logic (should be allowed when remoteGenerated and egress rule)?

src/pepr/operator/controllers/utils.ts
joelmccoy: I'm trying to grok the need for this function. Did you add this in because you are expecting host names with something like:
- `superLongHostName................................1`
- `superLongHostName................................2`

Just using the original `sanitizeResourceName` could end up with the same resource name?  Or is there something else that makes this function necessary.

Either way, it might be good to add some simple tests for this function

src/pepr/operator/controllers/istio/istio-resources.spec.ts
```diff
@@ -10,8 +10,7 @@ import { Mode } from "../../crd/generated/package-v1alpha1";
 import * as utils from "../utils";
 import { defaultEgressMocks, pkgMock, updateEgressMocks } from "./defaultTestMocks";
 import * as egressMod from "./egress";
-import * as egressAmbientMod from "./egress-ambient";
-import { istioEgressResources } from "./istio-resources";
+import { istioEgressResources } from "./egress-orchestrator";
```
joelmccoy: does it make sense to move the istioEgressResources tests to a file called `egress-orchestrator.spec.ts` or are there issues with the mocking maybe based on this [comment](https://github.com/defenseunicorns/uds-core/pull/2194/files#r2598903416)?

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -34,72 +69,190 @@ export async function applyAmbientEgressResources(packageList: Set<string>, gene
...
+      saPrincipals: Set<string>;
+      namespaces: Set<string>;
+    };
+    byHost: Map<
+      string,
+      {
+        saPrincipals: Set<string>;
+        namespaces: Set<string>;
+      }
+    >;
```
joelmccoy: nit: it feels weird to define an interface in the middle of a function, any reason not to move this towards the top of the file?

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
```diff
@@ -63,6 +66,524 @@ describe("test applyAmbientEgressResources", () => {
...
+                remoteHost: "example.com",
+                port: 443,
+                remoteProtocol: RemoteProtocol.TLS,
+                serviceAccount: "sa1",
+              } as Allow,
+            ],
+          },
+        },
+      },
+      // Anywhere participant with same SA in another allow to create duplicates across owners/participants
```
joelmccoy, line 685: just want to callout that these little comments in these tests are SUPER HELPFUL for reviewing/trace expected behavior.  Thanks for adding these in 🚀

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
```diff
@@ -63,6 +66,524 @@ describe("test applyAmbientEgressResources", () => {
...
+    ];
+
+    defaultEgressMocks.getPkgListMock.mockResolvedValueOnce({ items: pkgItems });
+
+    await applyAmbientEgressResources(new Set(["pkg1-ns1", "pkg2-ns2", "pkg3-ns3"]), 1);
+
+    // Waypoint applied, plus one SE and one AP for example.com
+    expect(defaultEgressMocks.applyWaypointMock).toHaveBeenCalledTimes(1);
+    expect(defaultEgressMocks.applySeMock).toHaveBeenCalledTimes(1);
+    expect(defaultEgressMocks.applyApMock).toHaveBeenCalledTimes(1);
```
joelmccoy, line 822: should we maybe check the expected merged values for the SE and AP for this test?

joelmccoy (commented):
Generally LGTM! Tested some of the functionality locally with no issues.  A couple comments/questions for consideration

test/vitest/network.spec.ts
```diff
@@ -421,6 +421,35 @@ describe("Network Policy Validation", { retry: 2 }, () => {
...
+  });
+
+  test.concurrent(
+    "Egress Ambient HTTP to github denied for SA with Google TLS access",
+    async () => {
+      const GITHUB_HTTP = [
+        "sh",
+        "-c",
+        `curl -s -w " HTTP_CODE:%{http_code}" http://api.github.com`,
+      ];
```
mjnagel: Would probably avoid github api since this previously got us rate-limited 😭

mjnagel (commented):
Initial comments - need to go through some of the shared logic a bit more to comprehend the behavior and do testing.

test/vitest/network.spec.ts
chance-coleman: added an additional test that uses the admin-app (which needed to be switch from sidecar mode to ambient).

src/pepr/operator/crd/validators/package-validator.ts
chance-coleman: added some new tests here

src/pepr/operator/controllers/utils.ts
chance-coleman: removed this function altogether, it really wasn't necessary.

src/pepr/operator/controllers/istio/istio-resources.spec.ts
chance-coleman: moved tests to `egress-orchestrator.spec.ts` and implemented a few new ones for the `istio-resources`.

src/pepr/operator/controllers/istio/egress-ambient.ts
chance-coleman: moved out of function

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
chance-coleman: expanded the current test to also check the merged values

src/pepr/operator/controllers/utils.ts
chance-coleman: removed this function altogether

src/pepr/operator/controllers/istio/auth-policy.ts
chance-coleman: updated to use the correct types

src/pepr/operator/controllers/istio/egress-ambient.ts
chance-coleman: adopted this positive check

src/pepr/operator/reconcilers/package-reconciler.ts
chance-coleman: Chatted with @mjnagel about this and i think we're in agreement that this count doesn't really provide that much so we also lean towards ont counting it towards the policy count.

joelmccoy: Just to sanity check, we are in agreement that the remote host shared authpolicy entries should NOT be counted towards the package auth policy count? If so, I think we can still remove some of this logic here?

chance-coleman: this should be addressed in the lates commit 👍🏼

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -34,72 +81,203 @@ export async function applyAmbientEgressResources(packageList: Set<string>, gene
...
+
+    // Filter only egress rules for faster processing
+    const egressRules = (pkg.spec?.network?.allow ?? []).filter(
+      allow => allow.direction === Direction.Egress,
     );
 
-    log.debug(serviceEntry, `Applying Service Entry ${serviceEntry.metadata?.name}`);
+    for (const allow of egressRules) {
+      // Process Anywhere participants
+      if (allow.remoteGenerated === RemoteGenerated.Anywhere) {
```
mjnagel: I think we also need to check the port here to make sure that something allowed to "anywhere" on port 80 isn't allowed to hit a remote host on 443.

src/pepr/operator/controllers/istio/auth-policy.ts
```diff
@@ -3,64 +3,60 @@
...
+    },
+  ];
+
+  const ap: IstioAuthorizationPolicy = {
     metadata: {
-      name: generateAmbientEgressAuthorizationPolicyName(host, serviceAccount),
-      namespace,
+      name,
+      namespace: ambientEgressNamespace,
       labels: {
```
mjnagel, line 84: Functionality it doesn't seem like we need it, but might be good to add the `uds.dev/user-` annotations to link this back to the "owners" (the packages explicitly requesting it)?

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -34,72 +81,203 @@ export async function applyAmbientEgressResources(packageList: Set<string>, gene
...
+      const id = `${name}-${ns}`;
+      if (!contribIds.has(id)) continue;
 
-    log.debug(authPolicy, `Applying Authorization Policy ${authPolicy.metadata?.name}`);
+      await purgeOrphans(String(generation), ns, name, IstioServiceEntry, log, {
+        "istio.io/use-waypoint": "egress-waypoint",
+      });
+      await purgeOrphans(String(generation), ns, name, IstioAuthorizationPolicy, log, {
+        "uds/for": "egress",
+      });
```
mjnagel: I think there's a slim edge case here where this fails to delete things? Since `generation` is the new "shared" generation, if the Package CR happened to have been on that generation (for its "local" generation) then nothing would get purged. I think we could probably just directly get, then delete all of them (probably separate from `purgeOrphans` since that requires a generation for filtering).

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -34,72 +81,203 @@ export async function applyAmbientEgressResources(packageList: Set<string>, gene
...
+      IstioServiceEntry,
+      log,
+    );
+    await purgeOrphans(
+      generation,
+      ambientEgressNamespace,
+      sharedEgressPkgId,
+      IstioAuthorizationPolicy,
+      log,
+    );
```
mjnagel, line 357: I'm having trouble tracing down the "why" but this purge seems to have a race condition or other bug where it can result in most resources getting purged. I've been testing with a [large set of packages](https://gist.github.com/mjnagel/3cdf18684f32726fe48a680b643b3664) and when applying them all together I end up with no egress service entries or authpols in my cluster. Restarting the watcher typically works to create them, but there's also some strange behavior on delete at times where deleting all the packages together sometimes creates/leaves them.

src/pepr/operator/controllers/istio/egress-ambient.ts
mjnagel: In testing I think this will mean we need to split the authpols by port (to.port on the rules) in order to handle `remoteGenerated` allowed to one port but not another, where a `remoteHost` is set for multiple ports.

src/pepr/operator/controllers/istio/egress-ambient.ts
mjnagel: Trying to do more testing and debug, but definitely seems like the applies are ending up running out of order so I end up with an outdated generation "winning" for the resulting serviceentry/authpol, and then the purge functions remove them.

mjnagel (commented):
Did more e2e testing today and encountered some timing problems I believe as well as the port issue.

src/pepr/operator/controllers/utils.ts
```diff
@@ -100,7 +103,16 @@ export async function purgeOrphans<T extends GenericClass>(
...
+    const resourceGenLabel = resource.metadata?.labels?.["uds/generation"];
+    const resourceGen = Number(resourceGenLabel);
+
+    const shouldDelete =
+      resourceGenLabel == null ||
+      (currentGenIsNumber && Number.isFinite(resourceGen)
+        ? resourceGen < currentGen
+        : resourceGenLabel !== generation);
+
+    if (shouldDelete) {
```
mjnagel, line 139: I think this new purge behavior could cause some issues/delay in reconciliation/purging with our "shared" resources that have a made up generation. The generation would get reset on watcher restart (to 0) but then resource in cluster would have a newer generation and not get purged immediately.

chance-coleman: reverted the purge behavior to previous implementation of exact matching, added a loop to address the race condition that we were seeing when a package is modified during a reconcile.

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -2,104 +2,405 @@
...
+
+  // Apply per-host shared SE and centralized AP
+  for (const host of Object.keys(merged)) {
+    if (conflictedHosts.has(host)) {
+      log.warn(
+        { host },
+        "Skipping ambient egress resources for host due to protocol conflicts across packages",
+      );
+      continue;
+    }
```
mjnagel: I was able to hit this scenario and see port/protocol conflicts result in these logs, but the corresponding Packages were marked as "Ready". I might have to backtrack where this gets called, but can we throw an error here to ensure that the Package status properly represents these conflicts?

src/pepr/operator/controllers/istio/egress-ambient.ts
```diff
@@ -2,104 +2,405 @@
...
+      if (ownerSaPrincipals.size === 0 && ownerNamespaces.size === 0) {
+        const derived = deriveOwnersFromContributors(
+          host,
+          port,
+          resource.packages ?? [],
+          ambientMap,
+        );
+        for (const p of derived.ownerSaPrincipals) ownerSaPrincipals.add(p);
+        for (const n of derived.ownerNamespaces) ownerNamespaces.add(n);
+      }
```
mjnagel: I _think_ that this (and the corresponding `deriveOwnersFromContributors` func) are unreachable? Further up in this function we always add either a svcacct or a namespace:
```ts
        if (rule.serviceAccount) {
          portCache.saPrincipals.add(
            `cluster.local/ns/${entry.namespace}/sa/${rule.serviceAccount}`,
          );
        } else {
          portCache.namespaces.add(entry.namespace);
        }
```
Which is eventually the same cache we check here, meaning that while one could be empty it should be impossible for both to be empty?

mjnagel (dismissed):

docs/reference/configuration/service-mesh/egress.md
```diff
@@ -88,11 +88,11 @@ Wildcards in host names are NOT currently supported.
 :::
 
 :::caution
-Adding any `remoteHost` egress creates/uses a shared L7 waypoint in ambient. This can change egress for other namespaces and break L4-only allowances (e.g., `remoteGenerated: Anywhere`), sometimes causing TLS failures. This behavior is by design in Istio ambient today.
+Adding any `remoteHost` in Ambient creates/uses a shared L7 waypoint and a centralized, per‑host `ServiceEntry` + `AuthorizationPolicy` in `istio-egress-ambient`. Egress to that host is evaluated at L7. Namespaces without an AP for that host remain denied. This can change behavior for workloads that previously relied on L4‑only allowances (e.g., `remoteGenerated: Anywhere`) and may surface TLS/HTTP differences.
```
mjnagel: > Namespaces without an AP for that host remain denied. This can change behavior for workloads that previously relied on L4‑only allowances (e.g., `remoteGenerated: Anywhere`) and may surface TLS/HTTP differences.

This part of the doc is a bit confusing to me, the operator handles authpols for the end user now (making/modifying the ap to add L4 remote anywhere namespaces/service accounts). Might just not be reading it right?

src/pepr/operator/controllers/istio/auth-policy.ts
```diff
@@ -3,64 +3,100 @@
...
     metadata: {
-      name: generateAmbientEgressAuthorizationPolicyName(host, serviceAccount),
-      namespace,
+      name,
+      namespace: ambientEgressNamespace,
+      ...(Object.keys(annotations).length > 0 ? { annotations } : {}),
       labels: {
-        "uds/package": pkgName,
-        "uds/generation": generation,
-        "uds/for": "egress",
```
mjnagel, line 31: Was this label removal intentional? It doesn't actually seem to be used/relevant anymore in the purge functions (those target the legacy namespace-local authpols) but might not be bad to label still since we label the other authpols as being for "network".

src/pepr/operator/controllers/istio/istio-resources.ts
mjnagel: I think this can be removed since ambient authpols are now centralized (not in the package namespace) and purged separately as part of `purgeAmbientEgressResources`. Sidecar doesn't use authpols for host egress as well so this shouldn't be needed.

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -45,68 +91,130 @@ let ambientGeneration = 0;
...
     }
   }
 
-  // Start a new reconciliation
-  reconciliationMutex = performEgressReconciliation();
+  return { name, namespace, rules };
+}
+
+// Mutex-based reconciliation to prevent overwhelming the operator
+export async function performEgressReconciliationWithMutex(_pkgId: string): Promise<void> {
```
mjnagel: nit: can we just cleanup `_pkgId` since it is an unused param?

src/pepr/operator/index.ts
```diff
@@ -87,6 +93,31 @@ When(UDSPackage)
...
+
+      await performEgressReconciliationWithMutex("__uds_package_deleted__");
+    } catch (e) {
+      log.error(
+        { err: e, pkgId },
+        "Failed reconciling shared egress resources after UDSPackage deletion",
+      );
+    }
+  });
+
```
mjnagel: I think this should be part of the main finalizer function (`packageFinalizer`) rather than a separate deletion watch? It looks like we have some cleanup/deletion in the finalizer so maybe this can just be deleted although I'm unsure on the diffs between these,

src/pepr/operator/controllers/istio/egress.ts
```diff
@@ -160,98 +267,54 @@ export async function updateInMemoryPackageMap(
...
+
+  // Keep the queue healthy even if this update fails.
+  sidecarMapUpdateQueue = task.catch(() => undefined);
+  return task;
 }
 
 // Update the inMemoryAmbientPackages list with the latest package
 export async function updateInMemoryAmbientPackageMap(
-  hostResourceMap: HostResourceMap | undefined,
+  pkg: UDSPackage | undefined,
```
mjnagel: Maybe related to my comment on the deletion watch, could we allow only the `UDSPackage` type here and not have to account for `if (!pkg)`?

mjnagel (commented):
Functionally all seems good - just have a few final cleanup/clarification comments.

mjnagel (approved):

Issue #2196, opened 2025-12-04
AKS terraform improvements to testing


mjnagel:
A few items have come up surrounding our AKS testing that should be addressed:
- [ ] We currently create clusters with azapi rather than azurerm due to lack of VNET integration support, however that support has now been added upstream ([ref](https://github.com/defenseunicorns/uds-core/blob/main/.github/test-infra/azure/aks/main.tf#L54-L57))
- [ ] Our nodepool sizing could probably be optimized/improved given this is for CI testing only - right now we have 5 nodes across two pools
- [x] ~Currently we test against Ubuntu, but not FIPS - to better match our users we should switch to FIPS to validate full install (note that this is expected to cause some issues with Falco that we may need to work through)~ Removed from scope, spun off https://github.com/defenseunicorns/uds-core/issues/2217

mjnagel:
Linked PR (https://github.com/defenseunicorns/uds-core/pull/2198) resolves the first two bullets. In testing Ubuntu FIPS causes issues with Falco as expected:
- `modern_ebpf` is the driver kind we use, which is airgap friendly + the most modern option...but....
- if you try to run out of the box with modern_ebpf on AKS Ubuntu FIPS you'll get this error `libpman: ring buffer map type is not supported` due to the kernel not supporting bpf ring buffer
- assuming [AKS releases FIPS 22.04](https://www.github.com/Azure/AKS/issues/4857) at some point that should be resolved, but timeline is uncertain....

I went down the path of testing the other driver kinds (`ebpf`):
- alternative driver options (ebpf, kmod) require an additional image (driver-loader) that we do not include by default
- if you sideload that image, it runs as an init and reaches out to the internet to pull down the driver (so not airgap friendly)
- assuming you have connectivity and sideload that image, Falco starts up fine and seems to be doing the security things on my test install
- alternatively you can pre-load the driver on the nodes (however you want), and then not need that image/internet call - not sure how much of an option this is for AKS?

Assuming the linked PR is merged I will update the issue to check off the first two bullets and move this issue back to the backlog for prioritization.

PR #2198, opened 2025-12-05
chore(ci): update aks iac

mjnagel:
## Description

Modifies CI in a few ways for AKS:
- switch from azapi -> azurerm for AKS cluster creation: reduces extra provider dependency and should help fix some timing issues that intermittently cause permissions problems in CI
- changes node sizing: removes autoscaling, adjusts nodes to newer types (Dv5 series) - cost here will increase < $1 per month, with a significant memory gain and small CPU drop
   - Before: 40 cpu, 80gb memory
   - After: 32 cpu (savings mostly on system nodes), 128gb memory, newer hardware
   - Cost increase is $0.001 / hour
- switches to the standard SKU for better api stability
- fixes some deprecated items and cleans up (tflint) unused variables
- adds a few params explicitly in the init task to help with local usage

Additional CI changes (beyond AKS):
- Updated save-logs to save playwright outputs for IAC
- Updated cluster logs debug to save all pod logs

## Related Issue

Related to https://github.com/defenseunicorns/uds-core/issues/2196

## Type of change


mjnagel:
Going to recreate this with less changes to minimize effects.

PR #2200, opened 2025-12-08
feat!: keycloak HA guide and sizing

### LINKED ISSUE #1913: Investigate/Document High Availability AZ Considerations for Keycloak
## Affected Documentation Section

Which section of the documentation needs to be updated?  
https://uds.defenseunicorns.com/reference/configuration/resource-configuration-and-ha/#keycloak

---

## Update Details

Describe the changes you'd like to see.  
Investigate and provide guidance how UDS Core users should deploy their HA Keycloak setup.  Based on upstream docs the Keycloak deployment should live in a Single AZ.  Investigate and provide appropriate guidance based on this.

---

## Why Is This Update Needed?

Explain why this update is important.  
The [keycloak upstream docs](https://www.keycloak.org/getting-started/getting-started-scaling-and-tuning#_horizontal_scaling) mention recommend:

> This cache is not optimal for clusters that span multiple availability zones, which are also called stretch clusters. For embedded Infinispan cache, work to have all instances in one availability zone. The goal is to avoid unnecessary round-trips in the communication that would amplify in the response times. On Kubernetes, use Pod affinity to enforce this grouping of Pods.

---

## Additional Context (Optional)

Link to internal slack thread discussing this: https://defense-unicorns.slack.com/archives/C08KECJFRDF/p1757943367958649

slaskawi:
## Description

This Pull Request introduce some high level info about Keycloak HA configuration as well as decreases the HPA min/max configurations

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/1913

## Type of change

## Steps to Validate

N/A

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
feat!: keycloak HA guide and sizing (https://github.com/defenseunicorns/uds-core/pull/2200)

BREAKING CHANGE: Keycloak Horizontal Pod Autoscaler (HPA) configuration has been changed. The minimum number of replicas has been lowered to 2, whereas the maximum has been set to 5. The values have been calculated based on [Keycloak Performance Benchmarks](https://www.keycloak.org/2025/10/keycloak-benchmark).
END_COMMIT_OVERRIDE


chance-coleman (approved):

mjnagel (approved):

src/keycloak/chart/values.yaml
```diff
@@ -338,8 +338,8 @@ autoscaling:
   # Additional HorizontalPodAutoscaler labels
   labels: {}
   # The minimum and maximum number of replicas for the Keycloak StatefulSet
-  minReplicas: 3
-  maxReplicas: 10
+  minReplicas: 2
+  maxReplicas: 5
```
joelmccoy, line 342: Wouldn't we consider this a breaking change (which is fine, we would just want to highlight this in the release)?  We are changing the default scaling replicas and this will directly impact environments like UDS Registry who don't override these values.  This will possibly cause scale down behavior and the deploying user should probably be informed and have the option to override for their environment.

mjnagel: Based on our [versioning](https://github.com/defenseunicorns/uds-core/blob/main/VERSIONING.md) I could see it as changing a default value (under our breaking changes). Definitely a minor breaking change, but good to call out.

slaskawi: Ok, I'll apply commit oveeride for it in a few mins

Issue #2175, opened 2025-12-01
Authservice CI validation failures on IAC tests


mjnagel:
This error has popped up in CI relatively consistently lately:

```
•  Running "Verify the service at https://ambient-protected.uds.dev/ is accessible"
     ERROR:  Failed to run action: command "Verify the service at https://ambient-protected.uds.dev/ is
 accessible" timed out after 300 seconds
```

It's worth noting that this is not really a "timeout". The action that runs here uses a [zarf wait action](https://github.com/defenseunicorns/uds-core/blob/a6e74b28fd84f72427c04ef5e631ff5f81715a31/src/test/tasks.yaml#L149-L154) which is checking for a 200 response on requests to the specified address. If the initial request does not receive a 200, it has a [100ms delay](https://github.com/zarf-dev/zarf/blob/edf51e026aaf658d7c8c1d64deb68ba5bc4bc09d/src/pkg/utils/wait.go#L179-L183) before retrying, up to the total "timeout" limit (300 seconds in this case).

mjnagel:
May be worth looking at DNS setup task: https://github.com/defenseunicorns/uds-core/blob/429de085f128fce7b2048e9737fcea9a776b6d14/tasks/utils.yaml#L75-L108

Timebox to a 2 for identifying issue.

mjnagel:
From a run that hit this in CI I pulled some logs:

Istio tenant gateway:
```console
2025-12-08T03:14:44.6551321Z stdout F [2025-12-08T03:04:44.412Z] "GET / HTTP/2" 403 - via_upstream - "-" 0 0 600008 600007 "10.42.0.1" "Go-http-client/2.0" "c71a29d3-e0ac-476d-854a-7c04e386e3df" "ambient-protected.uds.dev" "envoy://connect_originate/10.42.0.103:15008" outbound|8000||httpbin.authservice-ambient-test-app.svc.cluster.local envoy://internal_client_address/ 10.42.0.109:443 10.42.0.1:33917 ambient-protected.uds.dev -
```

Waypoint (in front of `ambient-protected`):
```console
[...]

PR #2202, opened 2025-12-08
fix: ambient authservice validation use curl and retries

chance-coleman:
## Description
Switch Ambient Authservice validations to use curl instead of network wait. Conditionally displays error output if the curls fail, otherwise leave the terminal clean.

## Related Issue

Relates to #2175

## Type of change

## Steps to Validate
- `uds run test:uds-core-e2e`

## Checklist before merging

### DIFF
--- a/src/test/tasks.yaml
+++ b/src/test/tasks.yaml
@@ -151,18 +151,21 @@ tasks:
         cmd: |
           set -e
           URL="https://${{ .inputs.address }}"
+          echo "=== Begin authservice check: $URL ==="
           set +e
           OUT_ALL="$(curl -sS -L -o /dev/null --max-time 15 -v \
                     -w "HTTP_CODE=%{http_code}\nURL_EFFECTIVE=%{url_effective}\n" \
                     "$URL" 2>&1)"
-          RC=$?
           set -e
           if printf '%s\n' "$OUT_ALL" | grep -qx 'HTTP_CODE=200' && \
              printf '%s\n' "$OUT_ALL" | grep -qi '^URL_EFFECTIVE=https://sso\.uds\.dev'; then
+            echo "=== Authservice check passed: $URL ==="
+            echo
             true
           else
-            echo
             printf '%s\n' "$OUT_ALL"
+            echo "=== Authservice check FAILED: $URL ==="
+            echo
             exit 1
           fi
 
--- a/src/test/tasks.yaml
+++ b/src/test/tasks.yaml
@@ -147,28 +147,33 @@ tasks:
         required: true
     actions:
       - description: Verify the service at https://${{ .inputs.address }} is accessible
-        wait:
-          network:
-            protocol: https
-            address: ${{ .inputs.address }}
-            code: 200
-
-      - description: Verify the service at ${{ .inputs.address }} is protected by checking redirect
-        maxRetries: 3
+        maxRetries: 5
         cmd: |
           set -e
-          SSO_REDIRECT=$(curl -Ls -o /dev/null -w %{url_effective} "https://${{ .inputs.address }}")
+          URL="https://${{ .inputs.address }}"
+          OUT="$(curl -vkI --no-progress-meter --max-time 10 --url "$URL" 2>&1 || true)"
+          CODE="$(printf '%s\n' "$OUT" | awk '/^< HTTP/{print $3; exit}')"
+          if [ "$CODE" = "200" ] || [ "$CODE" = "302" ]; then
+            :
+          else
+            echo
+            printf '%s\n' "$OUT" | egrep -E '^(\* |< |> )' || printf '%s\n' "$OUT"
+            exit 1
+          fi
 
-          case "${SSO_REDIRECT}" in
-          "https://sso.uds.dev"*)
-              echo "Protected by authservice"
-              ;;
-          *)
-              echo "App is not protected by authservice"
-              echo $SSO_REDIRECT
-              exit 1
-              ;;
-          esac
+      - description: Verify the service at ${{ .inputs.address }} is protected by checking redirect
+        maxRetries: 5
+        cmd: |
+          set -e
+          URL="https://${{ .inputs.address }}"
+          OUT="$(curl -vkI --no-progress-meter --max-time 10 --url "$URL" 2>&1 || true)"
+          if printf '%s\n' "$OUT" | grep -qi '^location:\s*https://sso\.uds\.dev'; then
+            :
+          else
+            echo
+            printf '%s\n' "$OUT" | egrep -E '^(\* |< |> )' || printf '%s\n' "$OUT"
+            exit 1
+          fi
 
   - name: remove
     actions:
--- a/src/test/tasks.yaml
+++ b/src/test/tasks.yaml
@@ -154,10 +154,10 @@ tasks:
           OUT="$(curl -vkI --no-progress-meter --max-time 10 --url "$URL" 2>&1 || true)"
           CODE="$(printf '%s\n' "$OUT" | awk '/^< HTTP/{print $3; exit}')"
           if [ "$CODE" = "200" ] || [ "$CODE" = "302" ]; then
-            :
+            true
           else
             echo
-            printf '%s\n' "$OUT" | egrep -E '^(\* |< |> )' || printf '%s\n' "$OUT"
+            printf '%s\n' "$OUT" | grep -E '^(\* |< |> )' || printf '%s\n' "$OUT"
             exit 1
           fi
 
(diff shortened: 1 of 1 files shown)


src/test/tasks.yaml
```diff
@@ -147,28 +147,33 @@ tasks:
...
-            address: ${{ .inputs.address }}
-            code: 200
+        maxRetries: 5
+        cmd: |
+          set -e
+          URL="https://${{ .inputs.address }}"
+          OUT="$(curl -vkI --no-progress-meter --max-time 10 --url "$URL" 2>&1 || true)"
+          CODE="$(printf '%s\n' "$OUT" | awk '/^< HTTP/{print $3; exit}')"
+          if [ "$CODE" = "200" ] || [ "$CODE" = "302" ]; then
+            :
```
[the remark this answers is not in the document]
chance-coleman: addressed

src/test/tasks.yaml
```diff
@@ -147,28 +147,33 @@ tasks:
...
-          network:
-            protocol: https
-            address: ${{ .inputs.address }}
-            code: 200
+        maxRetries: 5
+        cmd: |
+          set -e
+          URL="https://${{ .inputs.address }}"
+          OUT="$(curl -vkI --no-progress-meter --max-time 10 --url "$URL" 2>&1 || true)"
+          CODE="$(printf '%s\n' "$OUT" | awk '/^< HTTP/{print $3; exit}')"
```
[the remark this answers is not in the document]
chance-coleman: i think this is a bit overkill and unrelated at the moment.

src/test/tasks.yaml
```diff
@@ -147,28 +147,33 @@ tasks:
...
-              exit 1
-              ;;
-          esac
+          URL="https://${{ .inputs.address }}"
+          OUT="$(curl -vkI --no-progress-meter --max-time 10 --url "$URL" 2>&1 || true)"
+          if printf '%s\n' "$OUT" | grep -qi '^location:\s*https://sso\.uds\.dev'; then
+            :
+          else
+            echo
+            printf '%s\n' "$OUT" | egrep -E '^(\* |< |> )' || printf '%s\n' "$OUT"
```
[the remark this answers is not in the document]
chance-coleman: addressed

src/test/tasks.yaml
```diff
@@ -147,28 +147,33 @@ tasks:
...
+        cmd: |
+          set -e
+          URL="https://${{ .inputs.address }}"
+          OUT="$(curl -vkI --no-progress-meter --max-time 10 --url "$URL" 2>&1 || true)"
+          CODE="$(printf '%s\n' "$OUT" | awk '/^< HTTP/{print $3; exit}')"
+          if [ "$CODE" = "200" ] || [ "$CODE" = "302" ]; then
+            :
+          else
+            echo
+            printf '%s\n' "$OUT" | egrep -E '^(\* |< |> )' || printf '%s\n' "$OUT"
```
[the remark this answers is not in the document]
chance-coleman: addressed

src/test/tasks.yaml
```diff
@@ -147,28 +147,33 @@ tasks:
...
-          *)
-              echo "App is not protected by authservice"
-              echo $SSO_REDIRECT
-              exit 1
-              ;;
-          esac
+          URL="https://${{ .inputs.address }}"
+          OUT="$(curl -vkI --no-progress-meter --max-time 10 --url "$URL" 2>&1 || true)"
+          if printf '%s\n' "$OUT" | grep -qi '^location:\s*https://sso\.uds\.dev'; then
+            :
```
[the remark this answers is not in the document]
chance-coleman: addressed

src/test/tasks.yaml
```diff
@@ -147,28 +147,33 @@ tasks:
...
       - description: Verify the service at https://${{ .inputs.address }} is accessible
-        wait:
-          network:
-            protocol: https
-            address: ${{ .inputs.address }}
-            code: 200
+        maxRetries: 5
+        cmd: |
+          set -e
+          URL="https://${{ .inputs.address }}"
```
slaskawi, line 153: Curl has a set of "fail" flags such as `--fail`, `--fail-with-body` and `--fail-early`. If I'm not mistaken, we could replace almost the entire code with `curl --fail-with-body`

chance-coleman: So i attempted this, and maybe i was doing something wrong but we can't use curl `--fail` or `--fail-with-body` here because they only change behavior for 4xx/5xx responses and don’t help with our real needs: explicitly accepting both 200 and 302 and inspecting headers ourselves (like checking the location in second curl).

joelmccoy (dismissed):

src/test/tasks.yaml
mjnagel: I don't have a strong opinion here, but I do tend to like simplicity where possible. We have two separate checks:
1. Check that the service is accessible (previously 200, but it followed redirects so really just not an error 400+). I think this actually could make use of the `--fail-with-body` flag? I'm not sure if that is easier or not, since you'd still have to handle "catching" the error and conditionally outputting the body?
2. Checking the redirect URL - this one probably doesn't benefit from the fail flags, and in theory we already know the service is accessible.

This does actually make me thing we could just collapse this into a single check though that does both? Check the effective URL and http code? Don't know if that needs to be in scope for this PR, but could help with stability as well if we only have one curl that handles both checks.

src/test/tasks.yaml
mjnagel: AI comment on this was a bit wild, but I still might suggest a different approach to checking the http code (and the effective URL) that would be a bit more foolproof/direct. Using the `-w` flag we can explicitly write out the code and effective URL with formatting, for example:
```console
❯ curl -L https://google.com -w "\nHTTP_CODE=%{http_code}\nURL_EFFECTIVE=%{url_effective}"
<!doctype html>...</html>
HTTP_CODE=200
URL_EFFECTIVE=https://www.google.com/
```

(if you use `-L` to follow the redirect the resulting http code should always be 200 so we don't need to check for 302)

Then we could store the output and just grep for exact line matches just with basic grep:
```console
❯ echo $GOOGLE_OUTPUT | grep "^HTTP_CODE="
HTTP_CODE=200
❯ echo $GOOGLE_OUTPUT | grep "^URL_EFFECTIVE="
URL_EFFECTIVE=https://www.google.com/
```

I think _if_ you want to collapse these two into a single action this could be a simpler approach

mjnagel (commented):
LGTM from a functionality perspective, some comments on approach and simplicity.

src/test/tasks.yaml
chance-coleman: Agreed on simplicity, now have a single check that validates both the effective URL and the HTTP code, so the two-action approach is OBE.

src/test/tasks.yaml
chance-coleman: Implemented. I collapsed the two steps into a single curl invocation that follows redirects and emits structured fields via `-w`. The task now checks both HTTP_CODE=200 and URL_EFFECTIVE=https://sso.uds.dev from one call.

To keep output quiet on success but rich on failure, I run curl once with `-v` and capture both stdout and stderr; on success we print nothing, on failure we print the captured verbose trace plus the `-w` fields.

src/test/tasks.yaml
```diff
@@ -146,29 +146,25 @@ tasks:
...
-              ;;
-          *)
-              echo "App is not protected by authservice"
-              echo $SSO_REDIRECT
-              exit 1
-              ;;
-          esac
+          URL="https://${{ .inputs.address }}"
+          set +e
+          OUT_ALL="$(curl -sS -k -L -o /dev/null --max-time 15 -v \
```
mjnagel: Would recommend dropping `-k` to not hide any weird cert errors, otherwise this LGTM.

chance-coleman: removed

mjnagel (dismissed):

src/test/tasks.yaml
```diff
@@ -152,7 +152,7 @@ tasks:
           set -e
           URL="https://${{ .inputs.address }}"
           set +e
-          OUT_ALL="$(curl -sS -k -L -o /dev/null --max-time 15 -v \
+          OUT_ALL="$(curl -sS -L -o /dev/null --max-time 15 -v \
                     -w "HTTP_CODE=%{http_code}\nURL_EFFECTIVE=%{url_effective}\n" \
                     "$URL" 2>&1)"
           RC=$?
```
mjnagel: nit: this is unused

mjnagel (approved):

PR #2205, opened 2025-12-10
chore(ci): add more debug on IAC, update AKS deprecations

mjnagel:
## Description

Changes:
- adds `kubectl top` output to debug
- enable playwright artifact collection on IAC runs
- adds collection of all pod logs to IAC save-logs, using parallel `kubectl logs`
- updates AKS terraform to fix deprecated values + add missing init values (only needed during re-init but helpful for local dev)

### DIFF
--- a/.github/actions/save-logs/action.yaml
+++ b/.github/actions/save-logs/action.yaml
@@ -60,8 +60,8 @@ runs:
           | xargs -P 8 -n 2 bash -c '
             ns="$1"; pod="$2"
             base="/tmp/pod-logs/${ns}-${pod}"
-            uds zarf tools kubectl logs -n "$ns" "$pod" --tail=-1 > "${base}.log" 2>&1 || true
-            uds zarf tools kubectl logs -n "$ns" "$pod" --previous --tail=-1 > "${base}-previous.log" 2>&1 || true
+            uds zarf tools kubectl logs --all-containers=true -n "$ns" "$pod" --tail=-1 > "${base}.log" 2>&1 || true
+            uds zarf tools kubectl logs --all-containers=true -n "$ns" "$pod" --previous --tail=-1 > "${base}-previous.log" 2>&1 || true
           ' _
         echo "::endgroup::"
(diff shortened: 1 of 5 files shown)


.github/actions/save-logs/action.yaml
```diff
@@ -44,37 +44,25 @@ runs:
...
-        uds zarf tools kubectl logs sts/keycloak -n keycloak --tail -1 --previous > /tmp/keycloak-previous-logs.log || true
+        echo "::group::Collecting logs for all pods"
+        mkdir -p /tmp/pod-logs
+        # Loop over all pods and grab current and previous logs (8x parallel processes for efficiency)
+        uds zarf tools kubectl get pods -A -o jsonpath='{range .items[*]}{.metadata.namespace}{" "}{.metadata.name}{"\n"}{end}' \
+          | xargs -P 8 -n 2 bash -c '
+            ns="$1"; pod="$2"
+            base="/tmp/pod-logs/${ns}-${pod}"
+            uds zarf tools kubectl logs -n "$ns" "$pod" --tail=-1 > "${base}.log" 2>&1 || true
+            uds zarf tools kubectl logs -n "$ns" "$pod" --previous --tail=-1 > "${base}-previous.log" 2>&1 || true
```
joelmccoy: grabbing logs for all containers might be helpful.  troubleshooting the loki sidecar configmap issue and that is a separate container that i wouldn't get with just this command
```suggestion
            uds zarf tools kubectl logs --all-containers=true -n "$ns" "$pod" --tail=-1 > "${base}.log" 2>&1 || true
            uds zarf tools kubectl logs --all-containers=true -n "$ns" "$pod" --previous --tail=-1 > "${base}-previous.log" 2>&1 || true
```

joelmccoy: or you could loop over the containers in each pod and create separate files if we don't want to merge all container logs into a single file

chance-coleman (approved):

PR #2207, opened 2025-12-11
fix: downgrade k8s-sidecar to 1.30.9

joelmccoy:
## Description

k8s-sidecar had releases for [1.30.10 ](https://github.com/kiwigrid/k8s-sidecar/releases/tag/1.30.10)and [1.30.11 ](https://github.com/kiwigrid/k8s-sidecar/releases/tag/1.30.11)yanked due to instability.

And there are multiple issues that prevent us from upgrade to 2.X.X (2.0.0 - 2.1.4) versions:
- [High memory usage issue](https://github.com/kiwigrid/k8s-sidecar/issues/462)
  - tested 2.1.4 locally and saw memory usage jump from 60MiB to ~270MiB which is not ideal
- [tmp directory mount issue affecting loki chart](https://github.com/kiwigrid/k8s-sidecar/issues/427)
  - This will be fixed in https://github.com/grafana/loki/pull/19759
  - We could use pepr to mutate this until that is merged in, but the high memory usage is a reason to not bump to any of the 2.0.0-2.1.4 versions

Updates the renovate config to skip these bad versions.

## Type of change

## Steps to Validate
- `uds run` (we hopefully shouldn't see flakiness in the loki ruler integration tests now)

## Checklist before merging

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -119,6 +119,14 @@
         "registry1.dso.mil/ironbank/opensource/falcosecurity/falcosidekick"
       ],
       "allowedVersions": "!/99.99.99/"
+    },
+    {
+      "matchPackageNames": [
+        "registry1.dso.mil/ironbank/kiwigrid/k8s-sidecar",
+        "quay.io/rfcurated/k8s-sidecar",
+        "ghcr.io/kiwigrid/k8s-sidecar"
+      ],
+      "allowedVersions": "!/^(1\\.30\\.(1[0-9]|[2-9][0-9])|2\\.[01]\\.[0-4])/"
     }
   ]
 }
--- a/src/grafana/values/registry1-values.yaml
+++ b/src/grafana/values/registry1-values.yaml
@@ -21,4 +21,4 @@ sidecar:
   image:
     registry: registry1.dso.mil
     repository: ironbank/kiwigrid/k8s-sidecar
-    tag: 2.1.4
+    tag: 1.30.9
--- a/src/grafana/values/unicorn-values.yaml
+++ b/src/grafana/values/unicorn-values.yaml
@@ -20,5 +20,5 @@ downloadDashboardsImage:
 sidecar:
   image:
     registry: quay.io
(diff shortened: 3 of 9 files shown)


renovate.json
```diff
@@ -119,6 +119,14 @@
...
       ],
       "allowedVersions": "!/99.99.99/"
+    },
+    {
+      "matchPackageNames": [
+        "registry1.dso.mil/ironbank/kiwigrid/k8s-sidecar",
+        "quay.io/rfcurated/k8s-sidecar",
+        "ghcr.io/kiwigrid/k8s-sidecar"
+      ],
+      "allowedVersions": "!/^(1\\.30\\.(1[0-9]|[2-9][0-9])|2\\.[01]\\.[0-4])/"
```
joelmccoy, line 129: tested with this regex: https://regex101.com/r/Yj7Rp2/1

joelmccoy: just remove the the leading `!/` and trail `/` and replace `//` with `/` as these are escaped in the renovate config

joelmccoy:
It appears the registry1 image for 1.30.9 is having this issue: 
```
2025-12-11T19:38:15.550398036Z stderr F {"time": "2025-12-11T19:38:15.550241+00:00", "level": "ERROR", "msg": "MaxRetryError when calling kubernetes: HTTPConnectionPool(host='localhost', port=80): Max retries exceeded with url: /api/v1/namespaces/grafana/configmaps?labelSelector=grafana_datasource&timeoutSeconds=60&watch=True (Caused by NewConnectionError(\"HTTPConnection(host='localhost', port=80): Failed to establish a new connection: [Errno 111] Connection refused\"))\n"}
```

Which points to this potentially this [issue](https://github.com/kiwigrid/k8s-sidecar/issues/431) (which doesn't make sense since this should only exist in 1.30.10+).  Need to further troubleshoot this.

joelmccoy:
no longer relevant. newer releases should have this fix included.

PR #2211, opened 2025-12-12
fix: gracefully handle missing uds-ca-certs configmap

joelmccoy:
## Description

Gracefully handle missing `uds-ca-certs` ConfigMap.  Although this shouldn't happen as we deploy this ConfigMap before pepr comes up in our config chart, this can happen in certain dev scenarios.  This was discovered and fixed [here](https://github.com/defenseunicorns/uds-cli/pull/1277) in UDS CLI CI.

However, for more resilient code it definitely makes sense to set default values and just warn on startup of Pepr instead of crashing if the ConfigMap is not there. 

This PR refactors to handle a missing `uds-ca-certs` ConfigMap.

## Type of change

## Steps to Validate
- delete the contents of the file `hack/dev-manifests/uds-ca-certs.yaml` to simulate missing configmap
- run `uds run dev-setup && npx pepr deploy --yes` to ensure pepr starts up correctly with missing configmap

## Checklist before merging

### DIFF
--- a/src/pepr/operator/controllers/config/config.spec.ts
+++ b/src/pepr/operator/controllers/config/config.spec.ts
@@ -792,11 +792,17 @@ describe("handleUDSConfig", () => {
     });
 
     it("handles ConfigMap not found gracefully with default empty values", async () => {
+      const warnSpy = vi.spyOn(configLog, "warn");
       mockCfg.spec!.caBundle!.includeDoDCerts = true;
       const notFoundError = Object.assign(new Error("ConfigMap not found"), { status: 404 });
       mockConfigMapGet.mockRejectedValue(notFoundError);
 
       await handleCfg(mockCfg, ConfigAction.UPDATE);
+
+      // Should log a warning about the missing ConfigMap
+      expect(warnSpy).toHaveBeenCalledWith(
+        "CA certs ConfigMap not found, using empty values for DoD and public certs",
+      );
 
       // Should use default empty values and continue processing
       expect(UDSConfig.caBundle.includeDoDCerts).toBe(true);
@@ -811,6 +817,8 @@ describe("handleUDSConfig", () => {
           observedGeneration: 2,
         },
       });
+
+      warnSpy.mockRestore();
     });
 
     it("handles ConfigMap with no data field", async () => {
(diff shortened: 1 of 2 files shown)


src/pepr/operator/controllers/config/config.spec.ts
```diff
@@ -791,19 +791,23 @@ describe("handleUDSConfig", () => {
       expect(UDSConfig.caBundle.dodCerts).toBe("");
     });
 
-    it("throws error when ConfigMap is missing", async () => {
+    it("handles ConfigMap not found gracefully with default empty values", async () => {
```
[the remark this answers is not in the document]
joelmccoy: added.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -266,16 +280,8 @@ async function handleCABundleUpdate(caBundle: ConfigCABundle, updateClusterResou
...
-    if (UDSConfig.caBundle.includeDoDCerts && caCertsConfigMap.data["dodCACerts"]) {
-      UDSConfig.caBundle.dodCerts = caCertsConfigMap.data["dodCACerts"];
-    }
-
-    if (UDSConfig.caBundle.includePublicCerts && caCertsConfigMap.data["publicCACerts"]) {
-      UDSConfig.caBundle.publicCerts = caCertsConfigMap.data["publicCACerts"];
-    }
-  }
+  UDSConfig.caBundle.dodCerts = dodCerts;
+  UDSConfig.caBundle.publicCerts = publicCerts;
```
mjnagel, line 284: Just checking my understanding, previously this was conditional on the `include*` toggles, but that was unnecessary because we check those values "later" when actually making/updating configmaps, is that correct?

mjnagel (approved):

src/pepr/operator/controllers/config/config.ts
joelmccoy: correct. caught this when i refactored.  we should probably still set the UDSConfig values for `dodCerts` and `publicCerts` because the actual building of the bundle is handled [here](https://github.com/defenseunicorns/uds-core/blob/99bdd4018be71c9615d07514b5e6ed9a7c0ba398/src/pepr/operator/controllers/ca-bundles/ca-bundle.ts#L121)  for toggling whether to include them or not.

PR #2219, opened 2025-12-16
fix: downgrade k8s-sidecar for Grafana to 1.30.10, pin helm unittest install

slaskawi:
## Description

This Pull Request fixes the `main` branch build. 

It root cause of the failures (like [this](https://github.com/defenseunicorns/uds-core/actions/runs/20259499091/job/58168267972)) was a combination of two things:

1. The `HELM_UNITTEST_VERSION=v1.0.3 helm plugin install https://github.com/helm-unittest/helm-unittest.git --version "$HELM_UNITTEST_VERSION"` incorrectly propagates the environment variable. At the time of invoking the `helm` command the `$HELM_UNITTEST_VERSION` is undefined. This can be easily illustrated with the following script:

```bash
$ bash
$ set -x
$ TEST=ok true "$TEST"
+ TEST=ok
+ true '' <-- Note that this is empty, it's not "ok"!!!
```

2. The https://github.com/helm-unittest/helm-unittest/issues/790 introduced an incompatible with Helm 3 change which was causing the plugin installation to fail. Originally a previous release (1.0.3) was retagged, but this has since been reverted so that 1.0.3 can be pulled down.

This also includes the helm chart bump for Grafana 10.3.1, resolving [this issue](https://www.github.com/grafana/helm-charts/issues/4039) (this replaces https://github.com/defenseunicorns/uds-core/pull/2206) and [enables the WAL](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#wal) to allow for more seamless upgrades on the internal DB. Finally it downgrades the k8s-sidecar image due to a number of upstream issues with 2.x images.

## Type of change

## Steps to Validate

Verified automatically

## Checklist before merging

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -49,7 +49,7 @@ runs:
     - name: Install helm unittest plugin
       shell: bash
       # renovate: datasource=github-tags depName=helm-unittest/helm-unittest versioning=semver
-      run: HELM_UNITTEST_VERSION=v1.0.3 helm plugin install https://github.com/helm-unittest/helm-unittest.git --version "$HELM_UNITTEST_VERSION"
+      run: export HELM_UNITTEST_VERSION="v1.0.2" && helm plugin install https://github.com/helm-unittest/helm-unittest.git --version "${HELM_UNITTEST_VERSION}"
 
     - name: Iron Bank Login
       if: ${{ inputs.registry1Username != '' }}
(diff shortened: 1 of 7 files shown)


slaskawi (commented):
LGTM!

chance-coleman (approved):

slaskawi (commented):
LGTM! Merging!

PR #2230, opened 2025-12-17
fix: backport task and github workflows

chance-coleman:
## Description

- update the backport task to include a force fetch snapshot tags
- update PR validate ci workflow to run on `release/**` branches
- update pull-request-conditionals workflow to pull out test-routing step into a new action and use that action
- use that test-routing action in slim-dev-test and checkpoint workflows

## Type of change

## Checklist before merging


.github/workflows/pull-request-conditionals.yaml
```diff
@@ -178,7 +165,7 @@ jobs:
   # Shim required checks for backport PRs to release branches (i.e., non-release-please PRs)
   run-package-test-shim:
     needs: check-paths
-    if: ${{ needs.check-paths.outputs.run_full_tests != 'true' && startsWith(github.base_ref, 'release/') }}
+    if: ${{ needs.check-paths.outputs.run_full_tests != 'true' && startsWith(github.base_ref, 'release/') && needs.check-paths.outputs.combined != '[]' }}
```
mjnagel: I think this should run regardless of `combined` output (still conditional on the other checks).

.github/workflows/slim-dev-test.yaml
```diff
@@ -42,7 +42,24 @@ concurrency:
...
+        id: test-routing
+        uses: ./.github/actions/test-routing
+        with:
+          event_name: ${{ github.event_name }}
+          base_ref: ${{ github.base_ref }}
+          head_ref: ${{ github.head_ref }}
+
   lint-check:
+    needs: routing
+    if: ${{ needs.routing.outputs.run_full_tests == 'true' }}
```
mjnagel: probably okay as-is but we could run lint on everything, and have this run in parallel with `routing`?

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -126,30 +126,17 @@ jobs:
...
+          event_name: ${{ github.event_name }}
+          base_ref: ${{ github.base_ref }}
+          head_ref: ${{ github.head_ref }}
 
   # @lulaStart 11948466-9230-4498-be44-dbac784d86d1
   # This job triggers a separate workflow for each changed source package, if any.
   run-package-test:
     needs: check-paths
-    if: ${{ needs.check-paths.outputs.run_full_tests == 'true' }}
+    if: ${{ needs.check-paths.outputs.run_full_tests == 'true' && needs.check-paths.outputs.combined != '[]' }}
```
mjnagel:
```suggestion
    if: ${{ needs.check-paths.outputs.run_full_tests == 'true' && needs.check-paths.outputs.combined != '[]' && needs.check-paths.outputs.combined != '' }}
```

In a previous PR I needed to do this to account for no packages changing (see https://github.com/defenseunicorns/uds-core/commit/8585a6fb6bfa6b820eddc417028bbdb3c9fbd832). I believe that's also why we had the `-n` check in the bash if. I'm not finding the specific issues that this was related to at the moment (upstream) but recall this being one option to make sure you caught 2 different ways the output might be formatted.

mjnagel (commented):
LGTM overall - small comments.

.github/workflows/pull-request-conditionals.yaml
chance-coleman: updated!

.github/workflows/slim-dev-test.yaml
chance-coleman: runs in parallel and linting runs on everything now

.github/workflows/pull-request-conditionals.yaml
chance-coleman: removed the combined check

mjnagel (approved):

PR #2234, opened 2025-12-18
docs: updating velero AWS policy with cluster tag

briantwatson:
- Small documentation update following some changes we've made to environment we're managing


joelmccoy (commented):
I think these changes look good but would like to match our CI to ensure this still works and there aren't any gotchas. Would request that we update the policy for [EKS](https://github.com/defenseunicorns/uds-core/blob/main/.github/test-infra/aws/eks/velero.tf)

mjnagel (approved):

joelmccoy (approved):

PR #2242, opened 2026-01-05
chore: switch to iso timestamps for pepr logs, redact config log

mjnagel:
## Description

This switches to the ISO timestamp format for pepr logs (https://docs.pepr.dev/user-guide/customization/#customizing-log-format) as well as redacting the CA cert values when printing the config log. While not sensitive these values are quite long and flood the pod logs on startup, so it seems wise to redact them. The values of read-in certs are pretty easy to identify elsewhere also (in resulting configmaps, source configmap, etc).

## Type of change

## Steps to Validate

Deploy at minimum the base layer and review log output.

## Checklist before merging

### DIFF
--- a/src/pepr/operator/controllers/config/config.ts
+++ b/src/pepr/operator/controllers/config/config.ts
@@ -451,7 +451,8 @@ function redactConfig() {
   const authserviceRedisUri = UDSConfig.authserviceRedisUri ? "****" : "";
   const dodCerts = UDSConfig.caBundle.dodCerts ? "****" : "";
   const publicCerts = UDSConfig.caBundle.publicCerts ? "****" : "";
-  const caBundle = { ...UDSConfig.caBundle, dodCerts, publicCerts };
+  const certs = UDSConfig.caBundle.certs ? "****" : "";
+  const caBundle = { ...UDSConfig.caBundle, dodCerts, publicCerts, certs };
   return { ...UDSConfig, authserviceRedisUri, caBundle };
 }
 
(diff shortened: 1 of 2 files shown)


src/pepr/operator/controllers/config/config.ts
```diff
@@ -443,13 +443,16 @@ export async function loadUDSConfig() {
...
+ * Creates a redacted copy of UDS config for logging (hides sensitive or large values)
  *
- * @returns UDS config with sensitive fields replaced with masked values
+ * @returns UDS config with sensitive or large fields replaced with masked values
  */
 function redactConfig() {
   const authserviceRedisUri = UDSConfig.authserviceRedisUri ? "****" : "";
-  return { ...UDSConfig, authserviceRedisUri };
+  const dodCerts = UDSConfig.caBundle.dodCerts ? "****" : "";
+  const publicCerts = UDSConfig.caBundle.publicCerts ? "****" : "";
```
mjnagel, line 453: Open to feedback on these, could also trim it to the starting 10 characters or so + `...` or something since they aren't "sensitive"? We also have the 3rd cert value (`UDSConfig.caBundle.certs`) which could also be redacted/shortened.

joelmccoy: I think redacting completely with a set short string is reasonable here. I don't think we need this information in the logs at all.

Would also probably agree that `UDSConfig.caBundle.certs` can be redacted as well as that can be pretty long and not necessary for logs.

joelmccoy (approved):

PR #2244, opened 2026-01-06
chore: add lula mapping

joelmccoy:
## Description

Adds lula mapping for istio.

## Type of change

## Checklist before merging

### DIFF
--- a/src/istio/zarf.yaml
+++ b/src/istio/zarf.yaml
@@ -3,9 +3,11 @@
 
 kind: ZarfPackageConfig
 metadata:
+  # @lulaStart 5e1176a3-b894-4e67-9ad0-96ec84c3fff8
   name: uds-core-istio
   description: "UDS Core Istio package deploys Istio with admin, tenant and passthrough gateways."
   url: https://istio.io/latest/
+  # @lulaEnd 5e1176a3-b894-4e67-9ad0-96ec84c3fff8
 
 variables:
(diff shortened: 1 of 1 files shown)


chance-coleman (approved):

mjnagel (approved):

cmwylie19:
> ## Lula Compliance Overview
> Please review the changes to ensure they meet compliance standards.
> 
> ### Reviewed Changes
> Lula reviewed 1 files changed that affect compliance.
> 
> File	Lines Changed
> `src/istio/zarf.yaml`	`147–150`
> **UUID:** `643060b2-0ddf-4728-9582-ef38dca7447a` **sha256:** `d2debad22d19ad9d6922a66e07f790d29fb7c444f72201a1fe933c9f9f84de70`
> 
> **Tip:** Customize your compliance reviews with [Lula](https://github.com/defenseunicorns/lula.git).

This is incorrect because in this file there are two instances with the same UUID, the crawl algorithm was not expecting the same UUID twice in the same file, once it found the first block of config encompassed by the UUID, it compared it to the new block's [config](https://github.com/defenseunicorns/lula/blob/main/cli/commands/crawl.ts#L208) but there was no guarantee that they were the correct blocks. The fix here is to ensure that if there are multiple blocks in the old and new files, that each block has a match or else it should be considered a change.

This way, you can move the mappings around in the file and it will not produce false positives.

PR #2258, opened 2026-01-12
chore: mappings for compliance 853

mjnagel:
## Description

Relates to https://github.com/defenseunicorns/uds-compliance/issues/853

## Type of change


src/istio/zarf.yaml
```diff
@@ -56,11 +56,13 @@ components:
       - name: ztunnel
         valuesFiles:
           - "values/registry1/ztunnel.yaml"
+    # @lulaStart b4367e52-bef0-4463-a906-e5af6b4aa015
```
joelmccoy, line 59: I think originally we avoided wrapping images because this would get flagged on all renovate updates?  But totally think this makes sense to wrap these.  It's the best way to prove compliance and if we ever lose the fips images we get informed correctly.  No action needed here, just wanted to say this out loud.

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -100,6 +100,7 @@ spec:
             # This will only import the realm if it does not exist
             - "--import-realm"
             # FIPS Mode
+            # @lulaStart e7a920f8-e6e2-4a97-80a0-edda1fab0f8b
```
joelmccoy: I assumed you put this mapping here based on the desire to maybe remove [this](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/values.yaml#L155-L156) in the future? Does it make sense to map that still until then?

mjnagel: Ah yeah good call, I can wrap the value as well or instead.

src/istio/zarf.yaml
mjnagel: Yeah - I tried to think of a better way to capture this but we (unfortunately) don't really have another clear/direct spot with this. I think it's an acceptable tradeoff to make Istio updates a bit noisy, but also ensure we are reminded of compliance impact of switching those.

src/keycloak/chart/templates/statefulset.yaml
joelmccoy: probably would stick to just mapping the value, and if it gets removed we should know to move the mapping if it is hardcoded in the statefulset

joelmccoy (approved):

PR #2260, opened 2026-01-13
feat: keycloak logout confirmation

### LINKED ISSUE #689: Display an explicit logoff message when users logoff from Keycloak

slaskawi:
## Description

This Pull Request leverages the new logout implementation from https://github.com/keycloak/keycloak/pull/44463 and sets logout confirmation to true by default.

## UDS Identity Config bug fix

This Pull Request doesn't strictly require any UDS Identity Config updates but https://github.com/defenseunicorns/uds-identity-config/pull/768 fixes the following bug:

image

## Related Issue

Fixes https://github.com/defenseunicorns/uds-identity-config/issues/689

## Type of change

## Steps to Validate

Validated automatically in Grarana tests.

The disabled confirmation code path is tested in UDS Identity Config

## Checklist before merging

### DIFF
--- a/docs/reference/configuration/single-sign-on/keycloak-session-management.md
+++ b/docs/reference/configuration/single-sign-on/keycloak-session-management.md
@@ -113,7 +113,7 @@ Keycloak 26.5.0+ (UDS Core 0.59.0+) will now prevent configuring an SSO client w
 
 ## Logout Confirmation
 
-By default, UDS Core will display a logout confirmation page when a user logs out from an application. This behavor can be configured on a per-client basis by setting the `logout.confirmation.enabled` attribute in the `sso` section of the Package CR, for example:
+By default, UDS Core will display a logout confirmation page when a user logs out from an application. This behavior can be configured on a per-client basis by setting the `logout.confirmation.enabled` attribute in the `sso` section of the Package CR, for example:
 
 ```yaml
 apiVersion: uds.dev/v1alpha1
--- a/docs/reference/configuration/single-sign-on/keycloak-session-management.md
+++ b/docs/reference/configuration/single-sign-on/keycloak-session-management.md
@@ -110,3 +110,23 @@ Keycloak 26.5.0+ (UDS Core 0.59.0+) will now prevent configuring an SSO client w
   - If a refresh token is valid, it can still obtain a new access token.
   - If the refresh token is also expired, the user must reauthenticate.
 - Realm-wide session settings take precedence over client settings, meaning once the realm session expires, the user is logged out from all clients.
+
+## Logout Confirmation
+
+By default, UDS Core will display a logout confirmation page when a user logs out from an application. This behavor can be configured on a per-client basis by setting the `logout.confirmation.enabled` attribute in the `sso` section of the Package CR, for example:
+
+```yaml
+apiVersion: uds.dev/v1alpha1
+kind: Package
+metadata:
+  name: my-package
+  namespace: my-namespace
+spec:
+  sso:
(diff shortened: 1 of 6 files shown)


joelmccoy (commented):
I have maybe a silly question... Is it even worth it to have a configuration flag to turn this off?  I think having this on all the time is a sane thing to have for our clients?

And if we want a flag, would it make sense to have this flag in the sso portion of the package CR instead of a global config flag?

mjnagel:
> I have maybe a silly question... Is it even worth it to have a configuration flag to turn this off? I think having this on all the time is a sane thing to have for our clients?
> 
> And if we want a flag, would it make sense to have this flag in the sso portion of the package CR instead of a global config flag?

I had similar thoughts reviewing this - I think it could be simpler just to default this client attribute to "true" and allow people to use normal attribute overrides in the Package CR sso spec if they want to turn it off? That would mean they have to disable it per-client (and we may want a helm value for Grafana and downstream UDS Packages for this), but given the compliance aspects and that it's really just a UX choice I think it's okay to add some friction for deviating from the baseline.

chance-coleman (commented):
with grafana the logout confirmation works as expected but not with sso.uds.dev. Probably not the biggest issue but curious if you know why that is?

slaskawi:
> I have maybe a silly question... Is it even worth it to have a configuration flag to turn this off? I think having this on all the time is a sane thing to have for our clients?

> And if we want a flag, would it make sense to have this flag in the sso portion of the package CR instead of a global config flag?

> I had similar thoughts reviewing this - I think it could be simpler just to default this client attribute to "true" and allow people to use normal attribute overrides in the Package CR sso spec if they want to turn it off? That would mean they have to disable it per-client (and we may want a helm value for Grafana and downstream UDS Packages for this), but given the compliance aspects and that it's really just a UX choice I think it's okay to add some friction for deviating from the baseline.

@mjnagel @joelmccoy I think it's beneficial to have both mechanisms. I agree on per-Client configuration via attributes (this is something I'll add in a moment - the UDS Operator shouldn't override this setting if user specified it), without a global switch it won't be possible to disable this in UDS-provided Packages - like Grafana. So unless you object, I would keep both.

> with grafana the logout confirmation works as expected but not with sso.uds.dev. Probably not the biggest issue but curious if you know why that is?

@chance-coleman Thanks for spotting this! I missed this in my testing. The Account Console uses a separate client that has this option turned off. I'll update the UDS Identity Config (and provide migrations steps) to address this.

slaskawi:
Comments addressed and ready for the next round.

slaskawi:
The code has been refactored and there's no global opt-out mechanism now.

chance-coleman (approved):
LGTM, tested manually updating clients via Admin console and package cr.

joelmccoy (approved):

mjnagel (approved):

Issue #2179, opened 2025-12-01
Determine and document CRD versioning/upgrade stance


mjnagel:
Currently the UDS Core CRDs are all versioned as v1alpha1 ([example](https://github.com/defenseunicorns/uds-core/blob/429de085f128fce7b2048e9737fcea9a776b6d14/src/pepr/operator/crd/sources/package/v1alpha1.ts#L507)). Going forward we should determine our plan for CRD versioning and management across upgrades, in particular:
- What is guaranteed based on the CRD version (will fields be added, removed, renamed)? Initial thoughts: I would look at the [K8s Deprecation Policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/) to help inform our thoughts here on v1, beta, and alpha
- How do CRD version bumps relate to UDS Core version? Initial thoughts: UDS Core semver is distinct from CRD versioning, but if we drop support for a specific CRD version that would be a breaking change. If an existing CRD changes in a backwards incompatible way, this _may_ be a breaking change?
- How long do we support specific versions of CRDs/do we support multiple versions of CRDs? Initial thoughts: heavily related to our [deprecation policy](https://github.com/defenseunicorns/uds-core/issues/2087) and UDS Core semver.
- What do we want to release with Core 1.0 in terms of CRD versions? Initial thoughts: I don't think we are required to "v1" our CRDs with a v1 of Core. We should however consider what our CRD version guarantees are (previous bullet) and how that relates to what we may want to do post-1.0. Bumping a version without support for older versions would also be a pain so this should be weighed.

Definition of done: Documentation/plan (ADR?) for how we version and change CRDs. The above questions should be considered and likely answered by the doc and any other relevant/related docs should be updated as well (versioning, upgrading, etc).

mjnagel:
I found the [k8s deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/) pretty useful for how k8s treats alpha/beta/v1. Summary of some of the key rules for k8s apis (that we might/should apply to our CRDs):
- only remove api elements with a version bump (v1alpha1 -> v1alpha2)
- objects must be "convertable" bi-directionally between versions in a given release
- lifecycle is determined by stability level:
  - GA ("v1"): cannot be removed
  - Beta: Deprecated for minimum of 9 months/3 minor release, and left deprecated for minimum of 9 months/3 minor releases (before reoval)
  - Alpha: Can be removed in any release without deprecation notice
[...]

Issue #1705, opened 2025-07-08
Remove deprecated fields/functionality


mjnagel:
This issue tracks removal of deprecated UDS Core fields and functionality ahead of a major version release.

Child issues:

- [ ] [CORE-403](https://linear.app/defense-unicorns/issue/CORE-403/delete-the-keycloak-fips-switch) — Keycloak FIPS switch
- [ ] [CORE-429](https://linear.app/defense-unicorns/issue/CORE-429/remove-deprecated-ca-cert-variable) — Keycloak values
- [ ] [CORE-430](https://linear.app/defense-unicorns/issue/CORE-430/remove-deprecated-keycloak-values) — Operator config chart values
- [ ] [CORE-431](https://linear.app/defense-unicorns/issue/CORE-431/remove-deprecated-operator-config-chart-values) — `CA_CERT` variable

Relevant code:

Package CR fields:

* [source 1](<https://github.com/defenseunicorns/uds-core/blob/0c01a66b0a295195f22d75dcba1e15335417d0ae/src/pepr/operator/crd/sources/package/v1alpha1.ts#L126-L140>)
* [source 2](<https://github.com/defenseunicorns/uds-core/blob/0c01a66b0a295195f22d75dcba1e15335417d0ae/src/pepr/operator/crd/sources/package/v1alpha1.ts#L204-L215>)
* [source 3](<https://github.com/defenseunicorns/uds-core/blob/1c7d4e14d3e8301a007bd6fa66b103e1b4efb2ef/src/pepr/operator/crd/sources/package/v1alpha1.ts#L366-L391>)

Keycloak FIPS switch:

* [source](<https://github.com/defenseunicorns/uds-core/blob/0c01a66b0a295195f22d75dcba1e15335417d0ae/src/keycloak/chart/values.yaml#L135>)

Keycloak values:

* [source](<https://github.com/defenseunicorns/uds-core/blob/825c1ed6ea58df6432080174408528915ea0059c/src/keycloak/chart/values.yaml#L399-L403>)

Operator config chart values:

* [source](<https://github.com/defenseunicorns/uds-core/blob/6547cad1efd71cddaacbc0ad7da3334c7a71f72a/src/pepr/uds-operator-config/values.yaml#L5-L8>)

`CA_CERT` variable:

* [source](<https://github.com/defenseunicorns/uds-core/blob/825c1ed6ea58df6432080174408528915ea0059c/src/pepr/zarf.yaml#L18-L21>)

[...]

PR #2261, opened 2026-01-13
docs(adr): add uds core deprecation policy

### LINKED ISSUE #2087: Create UDS Core Deprecation Policy
UDS Core currently does not have a formal deprecation policy. While we are tracking deprecated fields in [issue #1705](https://github.com/defenseunicorns/uds-core/issues/1705), there is no defined process for how and when deprecations are communicated, implemented, or removed.

Before the UDS Core 1.0 release, we should establish a clear and consistent deprecation policy. This policy should outline:

- How deprecations are announced (e.g., changelog, release notes, docs).
- Expected timelines or version ranges for removal.
- How deprecations interact with backporting and support policies (see [issue #1702](https://github.com/defenseunicorns/uds-core/issues/1702)).

The goal is to provide predictable behavior for users and maintainers, reduce surprise breakages, and align with broader UDS release and support practices.

joelmccoy:
## Description

Adds an ADR and docs around UDS Core's deprecation policy.

## Related Issue

Fixes #2087 

## Type of change

## Checklist before merging

### DIFF
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -16,6 +16,11 @@ When adding a deprecation, include:
 
 | Feature | PR | Deprecated In | Reason | Migration | Removal Target |
 |---------|-----|---------------|--------|-----------|----------------|
+| `allow.podLabels`, `allow.remotePodLabels`, `expose.podLabels`, `expose.match` | [#154](https://github.com/defenseunicorns/uds-core/pull/154) | 0.12.0 | Improve API naming and organization | Use `allow.selector`, `allow.remoteSelector`, `expose.selector`, `expose.advancedHTTP.match` instead | 1.0.0 |
+| Keycloak `fips` helm value | [#1518](https://github.com/defenseunicorns/uds-core/pull/1518) | 0.43.0 | FIPS mode is now enabled by default for all deployments | If you override `fips` to `false`, remove that override as soon as possible to enable FIPS mode before it becomes the only allowed method | 1.0.0 |
+| Keycloak `x509LookupProvider`, `mtlsClientCert` helm values | [#1670](https://github.com/defenseunicorns/uds-core/pull/1670) | 0.47.0 | Values will be hardcoded; EnvoyFilter manages headers for third-party integrations | No action needed; remove any overrides of these values | 1.0.0 |
+| `operator.KUBEAPI_CIDR`, `operator.KUBENODE_CIDRS` | [#1233](https://github.com/defenseunicorns/uds-core/pull/1233) | 0.48.0 | Moved to ClusterConfig CRD for centralized configuration | Use `cluster.networking.kubeApiCIDR` and `cluster.networking.kubeNodeCIDRs` instead | 1.0.0 |
+| `CA_CERT` Zarf variable | [#2167](https://github.com/defenseunicorns/uds-core/pull/2167) | 0.58.0 | Improved naming clarity for centralized trust bundle management | Use `CA_BUNDLE_CERTS` instead | 1.0.0 |
 | `sso.secretName`, `sso.secretLabels`, `sso.secretAnnotations`, `sso.secretTemplate` | [#2264](https://github.com/defenseunicorns/uds-core/pull/2264) | 0.60.0 | Simplify fields and make more coherent | Use `sso.secretConfig.name`, `.labels`, `.annotations`, `.template` instead | 1.0.0 |
 
 ## Recently Removed
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -1,0 +1,27 @@
+# UDS Core Deprecations
+
+This document tracks all currently deprecated features in UDS Core. Deprecated features remain functional but are scheduled for removal in a future major release.
+
+## Active Deprecations
+
+<!--
+When adding a deprecation, include:
+- Feature: What is being deprecated
+- PR: Link to the pull request that introduced the deprecation
+- Deprecated In: The version where deprecation was announced
+- Reason: Why it is being deprecated
+- Migration: The recommended replacement or migration path
+- Removal Target: The projected major version for removal
+-->
+
+| Feature | PR | Deprecated In | Reason | Migration | Removal Target |
+|---------|-----|---------------|--------|-----------|----------------|
+| `sso.secretName`, `sso.secretLabels`, `sso.secretAnnotations`, `sso.secretTemplate` | [#2264](https://github.com/defenseunicorns/uds-core/pull/2264) | 0.60.0 | Simplify fields and make more coherent | Use `sso.secretConfig.name`, `.labels`, `.annotations`, `.template` instead | 1.0.0 |
+
+## Recently Removed
+
+This section lists features that were removed in recent major releases for historical reference.
+
+| Feature | Deprecated In | Removed In | Migration |
+|---------|---------------|------------|-----------|
+| _None_ | - | - | - |
--- a/adrs/0006-deprecation-policy.md
+++ b/adrs/0006-deprecation-policy.md
@@ -1,4 +1,4 @@
-# 6. Deprcation Policy
+# 6. Deprecation Policy
 
 Date: 2026-01-13
 
@@ -20,15 +20,16 @@ UDS Core will adopt a **time-based deprecation model aligned with minor releases
 
 The policy is defined as follows:
 
-- Deprecated features will remain supported for **at least three minor releases** following their deprecation announcement.
+- Deprecated features will remain supported for **at least three subsequent minor releases** following their deprecation announcement.
 - Deprecated features **may be removed** in:
   - The next minor release after the three-minor deprecation window has elapsed, **or**
   - Any new major release.
 - Removal of deprecated functionality is considered a **breaking change**, even when it occurs in a minor release.
 - Breaking changes **must not** occur in minor releases unless they are the explicit removal of previously deprecated functionality.
 - Patch releases will never introduce deprecations or remove deprecated functionality.
+- **Users are expected to resolve all deprecation warnings before upgrading beyond the guaranteed support window** to avoid encountering breaking changes.
 
-**Example:** If a feature is deprecated in version `0.30.0`, it must remain supported through versions `0.31.0`, `0.32.0`, and `0.33.0`. It becomes eligible for removal starting in `0.34.0` or any subsequent release (including a hypothetical `1.0.0`).
+**Example:** If a feature is deprecated in version `0.30.0`, it must remain supported through the three subsequent minor releases: `0.31.0`, `0.32.0`, and `0.33.0`. It becomes eligible for removal starting in `0.34.0` or any subsequent release (including a hypothetical `1.0.0`).
 
 This approach aligns deprecation timelines with UDS Core’s existing backport and support policy, providing users with clear expectations and sufficient time to migrate.
 
--- a/adrs/0006-deprecation-policy.md
+++ b/adrs/0006-deprecation-policy.md
@@ -1,0 +1,67 @@
+# 6. Deprcation Policy
+
+Date: 2026-01-13
+
+## Status
+
+Accepted
+
+## Context
+
+UDS Core follows Semantic Versioning and provides regular minor releases on a two-week cadence, with patch support for the latest three minor versions. The platform exposes a broad, Kubernetes-native API surface (CRDs, configuration, defaults, and packaging) that is consumed by long-lived clusters and mission applications.
+
+Strict adherence to SemVer—requiring a major release for all breaking changes—would significantly slow iteration and create pressure to batch unrelated changes into infrequent major releases. This model does not align well with Kubernetes ecosystem norms, where breaking removals commonly occur in minor releases after a well-defined deprecation period.
+
+At the same time, removing deprecated functionality without clear guarantees would erode user trust and complicate upgrade planning. A clear, predictable deprecation policy is required to balance platform evolution, security posture improvements, and operator stability.
+
+## Decision
+
+UDS Core will adopt a **time-based deprecation model aligned with minor releases**, rather than requiring a major release for all breaking removals.
+
+The policy is defined as follows:
+
+- Deprecated features will remain supported for **at least three minor releases** following their deprecation announcement.
+- Deprecated features **may be removed** in:
+  - The next minor release after the three-minor deprecation window has elapsed, **or**
+  - Any new major release.
+- Removal of deprecated functionality is considered a **breaking change**, even when it occurs in a minor release.
+- Breaking changes **must not** occur in minor releases unless they are the explicit removal of previously deprecated functionality.
+- Patch releases will never introduce deprecations or remove deprecated functionality.
+
+**Example:** If a feature is deprecated in version `0.30.0`, it must remain supported through versions `0.31.0`, `0.32.0`, and `0.33.0`. It becomes eligible for removal starting in `0.34.0` or any subsequent release (including a hypothetical `1.0.0`).
+
+This approach aligns deprecation timelines with UDS Core’s existing backport and support policy, providing users with clear expectations and sufficient time to migrate.
+
+## Consequences
+
+### Positive
+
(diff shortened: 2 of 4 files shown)


adrs/0006-deprecation-policy.md
```diff
@@ -0,0 +1,67 @@
+# 6. Deprcation Policy
```
chance-coleman:
```suggestion
# 6. Deprecation Policy
```
typo

adrs/0006-deprecation-policy.md
```diff
@@ -0,0 +1,67 @@
...
+
+At the same time, removing deprecated functionality without clear guarantees would erode user trust and complicate upgrade planning. A clear, predictable deprecation policy is required to balance platform evolution, security posture improvements, and operator stability.
+
+## Decision
+
+UDS Core will adopt a **time-based deprecation model aligned with minor releases**, rather than requiring a major release for all breaking removals.
+
+The policy is defined as follows:
+
+- Deprecated features will remain supported for **at least three minor releases** following their deprecation announcement.
```
chance-coleman: Does “three” include the deprecation release or not? based on the example it's pretty clear but maybe tightening this up would make sense, something like:

> “three subsequent minor releases”

adrs/0006-deprecation-policy.md
```diff
@@ -0,0 +1,67 @@
...
+- Deprecated features **may be removed** in:
+  - The next minor release after the three-minor deprecation window has elapsed, **or**
+  - Any new major release.
+- Removal of deprecated functionality is considered a **breaking change**, even when it occurs in a minor release.
```
chance-coleman: i think this makes sense but wonder if we should call out more explicitly that users are expected to resolve deprecations before upgrading beyond the guaranteed window.

chance-coleman (commented):
some small comments, otherwise its looking good

adrs/0006-deprecation-policy.md
joelmccoy: added this in

adrs/0006-deprecation-policy.md
joelmccoy: added a caution callout to add this in.

docs/reference/uds-core/release-overview.md
```diff
@@ -45,6 +45,57 @@ While UDS Core has not yet reached version 1.0, it is considered production-read
...
+- UDS Core configuration values and defaults
+- Exposed Zarf variables
+- Default-enabled components or integrations
+- Default behaviors that are planned to change or be removed
+
+Internal implementation details are not subject to this policy.
+
+### How Deprecations Are Announced
+
+All deprecations are documented in the [CHANGELOG.md](https://github.com/defenseunicorns/uds-core/blob/main/CHANGELOG.md) under a `⚠ DEPRECATIONS` header and included in the GitHub release notes. Each deprecation announcement includes:
```
joelmccoy: note: I may have mad a bad assumption here that release-please could support this header out of the box.  Need to do some more digging here on what approach makes the most sense (and maybe this requires a custom release-please config if we want the deprecation header).

joelmccoy: it seems like it wouldn't be straight forward to add a Deprecations Section and link it to some sort of specific commit natively with release-please.

Do we think we can just call this out in the features section like so:
`feat: deprecate: field sso.secretName` and end users can grep for deprecate in the changelog?

Or maybe we don't worry about being specific in the changelog with deprecations and we can just add these to the github release notes in it's own section instead?

mjnagel: I don't think there'd be an easy way to support with release-please if we want a section since ultimately release-please mostly follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/). It does allow some [modification of changelog section titles](https://github.com/defenseunicorns/uds-core/blob/main/release-please-config.json#L7-L12) although we'd probably have to abuse a commit type to go that route.

I think some convention around `(deprecation)` in the PR title + more details in the github release itself is acceptable though.

adrs/0006-deprecation-policy.md
```diff
@@ -0,0 +1,68 @@
...
+
+UDS Core follows Semantic Versioning and provides regular minor releases on a two-week cadence, with patch support for the latest three minor versions. The platform exposes a broad, Kubernetes-native API surface (CRDs, configuration, defaults, and packaging) that is consumed by long-lived clusters and mission applications.
+
+Strict adherence to SemVer—requiring a major release for all breaking changes—would significantly slow iteration and create pressure to batch unrelated changes into infrequent major releases. This model does not align well with Kubernetes ecosystem norms, where breaking removals commonly occur in minor releases after a well-defined deprecation period.
+
+At the same time, removing deprecated functionality without clear guarantees would erode user trust and complicate upgrade planning. A clear, predictable deprecation policy is required to balance platform evolution, security posture improvements, and operator stability.
+
+## Decision
+
+UDS Core will adopt a **time-based deprecation model aligned with minor releases**, rather than requiring a major release for all breaking removals.
```
mjnagel: I like the premise of aligning our deprecation period with our support period (minimum of 3 minor releases), but I'm not sold on changing how we interpret versioning/semver (at least without more qualifications). I do recognize that not doing this could mean carrying deprecations around for longer, but I think that's the cost of providing a more stable product. I also think it will force a behavior change on our end around how we introduce/change fields.

If we are worried about potential for this "hamstringing" us then I'd recommend we think about changes to the [versioning/breaking change definitions](https://github.com/defenseunicorns/uds-core/blob/main/VERSIONING.md). One thing we could consider for example would be adding a section around alpha/beta/experimental features and defining different handling for those to allow more flexibility. I think this might better align with the callout below of:
> - Ecosystem Alignment: Matches common practices in Kubernetes and CNCF-adjacent projects.

I _think_ that Kubernetes and other projects only have flexibility in removing deprecated "APIs" for alpha/beta APIs. That might be some of the tension here is that Core could be 1.0, but that doesn't mean each API and feature is GA - and we likely want more flexibility with some of those to make "breaking changes".

docs/reference/uds-core/release-overview.md
mjnagel, line 34: If we do go forward with this proposal as-is we'll want to update this. Might even make sense to put the deprecation policy section under this versioning strategy header hierarchically?

adrs/0006-deprecation-policy.md
joelmccoy: These are good points.  And I totally be fine with sticking to strict semver if we have the ability to cut major releases more liberally as needed post 1.0.  Idk if we discussed this much. Is this going to be the case?

Tracking and maintaining experimental features could be useful as well, but feels like it would be very manual to track what features are experimental and the actual scope of what each experimental feature covers (could be very subjective?).

mjnagel: To date we have not committed to any cadence/"irregularity" around major updates, although I think the general sentiment has been that we want to avoid them being "too frequent".

Thinking about it there's a number of things at play here (and probably more I'm not thinking about)
- [CRD versioning](https://github.com/defenseunicorns/uds-core/issues/2179)
- Upstream chart/image versions
- Our own chart/image versions (think identity-config, our Keycloak chart, etc)
- "Unversioned" things like zarf variables, features, etc

Currently per our [versioning doc](https://github.com/defenseunicorns/uds-core/blob/main/VERSIONING.md) removals/breaking changes to any of these would reflect as a major version bump of Core, with the exception of upstream chart/image versions. I think the tension here is mainly that Core's version wraps up _a lot_ of stuff.

To maybe get a better sense of things I went back through the last ~10 releases to look at breaking changes:
- 0.58.0: Keycloak HPA default values changed (change of default helm values/behavior).
- 0.56.0: Removal of NeuVector (change of core components).
- 0.54.0: Change to opentofu keycloak client permissions (breaking change that came from upstream and required manual "fix").
- 0.53.0: Loki ruler storage backend changed to local storage (arguably _not_ a breaking change, but something we wanted to note for users).
- 0.52.1: Keycloak `podManagementPolicy` change (previous helm value removed).
- 0.50.0: Addition of new policies to our policy engine (previously allowed workloads might get denied).
- 0.49.0: Group auth protection added to Grafana/NeuVector (extra helm values needed to preserve prior behavior).
- 0.48.0: Clusterconfig added, some helm values removed

So out of the 8 we have:
- 5 changes to helm values (removed/changed/newly required)
- 1 component change: likely to be rare imo
- 1 policy/security change
- 1 change "caused by upstream" (keycloak tofu client) - also maybe could be classified as an experimental/beta feature
- 0 CRD changes

So looks like the bulk of our breaking changes (at least in a recent snapshot) are with helm values changing/adding/being removed. That does feel like a spot we'd want more flexibility so I could see that as something for us to reevaluate in our versioning interpretation (maybe at minimum default values changes aren't considered - we can deprecate and keep older values easily in most cases for renames/removals). A more drastic measure would be to treat our helm charts the same as "upstream" helm charts and version them appropriately, but not bubble those versions/breaking changes up to Core's version.

Another thing to maybe reevaluate is the "changes to security posture" - that feels like something we'd want more flexibility with too. Things like improvements to our policies, new default netpols/authpols, etc would probably reflect a risk we identified and while they would have downstream impacts could be looked at more as "fixes" with clear callouts but not the same semver impact.

I'd largely ignore the CRD conversation here (defer to separate CRD versioning convo), although I think we'd agree _removing_ an actual CRD version would be breaking for Core.

That's a lot of words...mostly to say I would still argue for re-evaluating our _interpretation_ of breaking changes _for Core_ rather than our adherence to semver. I think I'd agree with wanting more flexibility to not constantly be bumping our major version. I just want to be careful about broadly saying "we allow breaking changes in minor releases", and would prefer to more specifically state "we allow these types of changes in minors, which we don't consider breaking for Core's API". Might just be nuance in the words, but I think it changes the _scope_ of what can happen in a minor version.

DEPRECATIONS.md
```diff
@@ -0,0 +1,27 @@
...
+- PR: Link to the pull request that introduced the deprecation
+- Deprecated In: The version where deprecation was announced
+- Reason: Why it is being deprecated
+- Migration: The recommended replacement or migration path
+- Removal Target: The projected major version for removal
+-->
+
+| Feature | PR | Deprecated In | Reason | Migration | Removal Target |
+|---------|-----|---------------|--------|-----------|----------------|
+| `sso.secretName`, `sso.secretLabels`, `sso.secretAnnotations`, `sso.secretTemplate` | [#2264](https://github.com/defenseunicorns/uds-core/pull/2264) | 0.60.0 | Simplify fields and make more coherent | Use `sso.secretConfig.name`, `.labels`, `.annotations`, `.template` instead | 1.0.0 |
```
joelmccoy, line 24: Note: i will backfill the other deprecations in this PR if this format looks good to everyone

docs/reference/uds-core/release-overview.md
joelmccoy: moved this.

adrs/0006-deprecation-policy.md
joelmccoy: After synchronous discussions here is what actions have been taken:
1. Add a callout in our versioning that allows for a security exception to release a breaking change in minor version for security purposes
2. Move back to strict semver assumption in deprecation policy
3. Still keep true to deprecations will hang around for a minimum of 3 minor releases (but removal requires major release)
4. Added a process for tracking deprecations in a `DEPRECATIONS.md` file.
5. Added guidance that deprecations will follow the conventional commit style of `feat(deprecation): super cool feature description`.
6. Added a starter `DEPRECATIONS.md` file

adrs/0006-deprecation-policy.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Decision
+
+UDS Core will adopt a **deprecation model that requires major releases for all deprecation removals**, aligning with strict Semantic Versioning principles while ensuring users have adequate time to migrate.
+
+The policy is defined as follows:
+
+- Deprecated features will remain supported for **at least three subsequent minor releases** following their deprecation announcement.
+- Deprecated features **may only be removed** in major releases, and only after the three-minor-release window has elapsed.
+- Removal of deprecated functionality is considered a **breaking change** and will only occur in major releases.
```
slaskawi: nit: might be good to move it to the top. This sentence seems to be the "essence"

DEPRECATIONS.md
slaskawi: I love it!

adrs/0006-deprecation-policy.md
```diff
@@ -0,0 +1,108 @@
...
+- Patch releases will never introduce deprecations or remove deprecated functionality.
+- **Users are expected to resolve all deprecation warnings before upgrading to the next major version** to avoid encountering breaking changes.
+
+**Example:** If a feature is deprecated in version `0.30.0`, it must remain supported through at least the three subsequent minor releases: `0.31.0`, `0.32.0`, and `0.33.0`. It becomes eligible for removal starting in `1.0.0` (assuming `1.0.0` is released after `0.33.0`). If a major release were planned before the three-minor window elapsed, the deprecated feature would remain and could not be removed until the following major release.
+
+This approach provides users with clear, predictable upgrade paths, ensures adequate migration time, and aligns with strict Semantic Versioning expectations.
```
slaskawi, line 31: I think we should put a shortened version of the "Security Exception" paragraph (from the VERSIONING.md) here as well.

mjnagel (dismissed):

DEPRECATIONS.md
joelmccoy: added in rest of the deprecations here. lmk if i missed anything. also update our [issue](https://github.com/defenseunicorns/uds-core/issues/1705) for tracking.

adrs/0006-deprecation-policy.md
joelmccoy: moved up

adrs/0006-deprecation-policy.md
joelmccoy: added!

docs/reference/uds-core/release-overview.md
```diff
@@ -37,14 +37,65 @@ For a more detailed review of how UDS Core applies Semantic Versioning, see the
...
+#### Security Exception
+
+As a security-first platform, UDS Core reserves the right to release security-related breaking changes in minor versions when the security benefit to users outweighs the disruption of waiting for a major release. These changes will still be clearly advertised as breaking changes. We will always strive to minimize the impact on users and will only exercise this exception when the security improvement is necessary and urgent.
+
 ### Current Stability Status
 
 While UDS Core has not yet reached version 1.0, it is considered production-ready and stable. The pre-1.0 versioning reflects our commitment to maintaining flexibility as we continue to enhance our security posture.
 
-## Version Support
+### Version Support
```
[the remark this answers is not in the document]
joelmccoy: this is intentional

slaskawi:
After the recent changes, this doc is LGTM. Once comments are resolved, please ping me and I'll add my approval.

DEPRECATIONS.md
mjnagel: This looks great. Probably(?) a discussion separate from this ADR/PR, but "removal" of CRD fields might have some special nuances depending on where we go with CRD upgrade/management. Mostly thinking the "removal target" of 1.0.0 might be more nuanced but I don't think we need to spell that out here. Maybe something for @slaskawi on the CRD issue to consider how we best capture that here (and whether we ever directly remove fields from CRDs).

mjnagel (approved):

DEPRECATIONS.md
joelmccoy: yea totally agree. I think we can answer those questions in the CRD strategy.  I think for now having this as `Removal Target` does not guarantee we yank it in 1.0.0 and I suspect based on recent conversations we may have to keep this around much longer.  We can adjust these entries at a later point once we figure out the CRD stuff.

mjnagel: Yep - removal _could_ look more like an entire CRD version (v1alpha1 for example) getting deprecated/removed in a future version, rather than just a specific field.

chance-coleman (approved):
i think this is looking great. Happy with where this has landed

PR #2264, opened 2026-01-13
feat: add new package sso secret fields

### LINKED ISSUE #2111: Change `sso` spec to group secret fields together

In the `Package` CR we currently support a number of ways to customize the SSO secret via the spec. For example:
```yaml
spec:
  sso:
    - name: My App
...
      secretName: my-app-secret
      secretLabels:
        uds.dev/pod-reload: "true"
      secretAnnotations:
        uds.dev/pod-reload-selector: 'app=my-app' # Target a specific pod(s) to reload
      secretTemplate:
        config.json: |
          {
            "client_id": "clientField(clientId)",
            "client_secret": "clientField(secret)"
          }
```

This is very verbose for many of these secret keys and does not group them well. We should consider changing this to:
```yaml
spec:
  sso:
    - name: My App
...
      secret:
        name: my-app-secret
        labels:
          uds.dev/pod-reload: "true"
        annotations:
          uds.dev/pod-reload-selector: 'app=my-app' # Target a specific pod(s) to reload
        template:
          config.json: |
            {
              "client_id": "clientField(clientId)",
              "client_secret": "clientField(secret)"
            }
```

This would provide a better grouping/hierarchy and allow for a more scalable way of adding other secret controls in the future. 

Existing fields could/should be deprecated and noted in [this issue](https://github.com/defenseunicorns/uds-core/issues/1705) and we can use the [`migrate` function](https://github.com/defenseunicorns/uds-core/blob/23d0163b208f9dd238654ca3f4ead65ce1230c2a/src/pepr/operator/crd/migrate.ts#L15) to easily handle both paths the same in our reconcile code (by migrating to the new fields internally).

joelmccoy:
## Description

Adds new fields for SSO Secret Generation.  Deprecates old fields.

New fields:
```yaml
sso:
  secretConfig:
    name:
    labels:
    annotations:
    template:
```

Old deprecated (but still supported fields):
```yaml
sso:
  secretName: 
  secretLabels:
  secretAnnotations:
  secretTemplate:
```

## Related Issue

Fixes #2111 

## Type of change

## Checklist before merging


docs/reference/configuration/single-sign-on/sso-templating.md
```diff
@@ -2,7 +2,11 @@
...
+By default, UDS generates a secret for the Single Sign-On (SSO) client that encapsulates all client contents as an opaque secret. In this setup, each key within the secret corresponds to its own environment variable or file, based on the method used to mount the secret. If customization of the secret rendering is required, basic templating can be achieved using the `secretConfig.template` property. Below are examples showing this functionality. To see how templating works, please see the [Regex website](https://regex101.com/r/e41Dsk/3).
+
+:::caution Deprecated Fields
+The `secretName`, `secretLabels`, `secretAnnotations`, and `secretTemplate` fields are deprecated and will be removed in a future major release. Use `secretConfig.name`, `secretConfig.labels`, `secretConfig.annotations`, and `secretConfig.template` instead. The deprecated fields will be automatically migrated to the new structure.
```
mjnagel:
```suggestion
The `secretName`, `secretLabels`, `secretAnnotations`, and `secretTemplate` fields are deprecated and will be removed in a future release. Use `secretConfig.name`, `secretConfig.labels`, `secretConfig.annotations`, and `secretConfig.template` instead. The deprecated fields will be automatically migrated to the new structure.
```

Might just leave as future release since in 0.x we could remove them (although timing-wise 1.0 might be likely).

src/pepr/operator/README.md
```diff
@@ -106,7 +106,11 @@ spec:
 
 ### Example UDS Package CR with SSO Templating
```
mjnagel: Maybe not for this PR but we probably could just simplify this readme / point to the other docs so that we don't have duplication - just commenting because the duplication is more noticeable with the diffs in this PR 😅

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -206,7 +206,7 @@ export async function syncClient(
         // Use the CR as the owner ref for each VirtualService
         ownerReferences: getOwnerRef(pkg),
       },
-      data: generateSecretData(client, secretTemplate),
+      data: generateSecretData(client, secretConfig?.template),
     });
   }
```
mjnagel, line 211: Commenting here so this can be a 🧵 - I think there might be value in making one of our existing test packages use the deprecated SSO secret fields and adding some assertions around it, just to give some e2e confidence. The SSO client/secret doesn't even need to be used, just validate correctness of naming, labels, annotations, templating.

mjnagel (commented):
LGTM overall! One docs and testing comment.

docs/reference/configuration/single-sign-on/sso-templating.md
joelmccoy: removed

src/pepr/operator/README.md
joelmccoy: removed this readme per discussion in slack

src/pepr/operator/controllers/keycloak/client-sync.ts
joelmccoy: I think this is sane. I added an the deprecated sso specs to the test-tenant-app package and added an e2e test to make sure the secret still gets created correctly

mjnagel (approved):

slaskawi (approved):
Fantastic work! I really like the comment about the new format taking priority over deprecated fields ❤️

PR #2270, opened 2026-01-14
docs(adr): define cluster-less crd workflow

### LINKED ISSUE #2164: Consider defining UDS Core CRDs in YAML

The current workflow for creating and applying our UDS CRDs (`ClusterConfig`, `Package`, `Exemption`) is defined in https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/tasks.yaml.  

The workflow essential goes like this (using package as an example):
1. Define a JSON spec in ts (https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/crd/sources/package/v1alpha1.ts)
2. Apply the above spec with this function. This happens on startup of Pepr and also run this function in our tasks to apply the spec to a k8s cluster. (https://github.com/defenseunicorns/uds-core/tree/v0.56.0/src/pepr/operator/crd/register.ts#L16-L48)
3. Use kubernetes-fluent-client to export this CRD to generate ts types (https://github.com/defenseunicorns/uds-core/tree/v0.56.0/src/pepr/tasks.yaml#L15-L18)

This workflow has a couple issues:
1. We don't generate/check in the yaml CRD (for `Package` and `Exemption`)
2. This workflow requires a kubernetes cluster to generate the types

Related to https://github.com/defenseunicorns/uds-core/issues/272

### Additional context

Would be worth considering another workflow that doesn't require:
- the need for a Kubernetes cluster
- The conversion from TS -> Kubernetes YAML manifest -> TS Types

One option to consider is to:
1. Define our CRDs in YAML (maybe with a strong typing code to generate these yaml files)
2. Use kubernetes-fluent-client to convert these yaml files into proper TS types

This would get us defined YAML CRD specs checked into our source code (which is useful if you want to apply these CRDs before pepr) and also remove the need for a kubernetes cluster to generate the TS types we need in our operator code.

chance-coleman:
## Description
This PR introduces ADR 0006, which defines a new YAML-first, cluster-less workflow for authoring, generating, and registering UDS Core CRDs (ClusterConfig, Package, Exemption).

The ADR documents a shift away from the current cluster-dependent CRD generation flow toward a model where TypeScript remains the authoring source of truth, but static YAML CRD manifests are checked in and used as the primary artifact for registration and type generation.

## Related Issue

Fixes #2164 

## Type of change

## Checklist before merging


adrs/0006-cluster-less-crd-generation-workflow.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Workflow Architecture
+
+1. **Maintain TS Sources**: We will continue using `src/pepr/operator/crd/sources/*.ts` as the authoring source of truth for schema definitions, allowing us to leverage code reuse (e.g., shared Istio schema objects) and strong typing during design.
+2. **Static Manifest Generation**: A new generation step will "compile" these TS sources into static YAML manifests located in `src/pepr/operator/crd/manifests/*.yaml`. These files will be tracked in Git.
+3. **Local Type Generation**: The `gen-crds` task will be updated to point `kubernetes-fluent-client` at these local YAML files instead of a cluster.
+4. **Declarative Registration**: `src/pepr/operator/crd/register.ts` will be refactored to load these manifests and apply them via Pepr, ensuring the operator always uses the exact schema version it was compiled with.
```
joelmccoy: curious. is there a reason we would still want pepr to register the CRDs instead of applying them via helm chart since we have the manifests in code now?

chance-coleman: Good question. Ultimately, I think both options are possible with this approach, it mostly comes down to the user and the environment, and the pros/cons they care about.

Having Pepr register CRDs is nice because it makes things “just work,” especially in dev or throwaway clusters. You run the operator and it ensures the CRDs exist, so you don’t hit confusing errors or have to remember extra install steps.

In more locked-down or production environments, though, it’s usually better to install CRDs up front (via Helm, Zarf, etc.) and avoid giving the operator permission to create or modify cluster-scoped APIs at runtime. Some teams explicitly don’t want operators doing that, for security or change-control reasons.

adrs/0006-yaml-first-crd-workflow.md
```diff
@@ -0,0 +1,64 @@
...
+## Context
+
+The current workflow for managing Custom Resource Definitions (CRDs) in `uds-core` (ClusterConfig, Package, Exemption) is cluster-dependent. TypeScript definitions in `src/pepr/operator/crd/sources/` are applied to a running Kubernetes cluster, and `kubernetes-fluent-client` then fetches the applied CRDs from the cluster to generate TypeScript types.
+
+This workflow and the changes described in this ADR apply uniformly to all UDS Core CRDs, including ClusterConfig, Package, and Exemption.
+
+This workflow presents several challenges:
+
+1. **Cluster Dependency**: Generating TypeScript types requires a live Kubernetes cluster, preventing this task from running in standard CI/CD pipelines or for local development in cluster-less environments.
+2. **Missing Static Manifests**: Authoritative YAML manifests for `Package` and `Exemption` CRDs are not checked into the repository. This hinders manual inspection, pre-application in GitOps workflows, and local validation.
```
mjnagel: nit on `pre-application in GitOps workflows` I think we could more explicitly note something about "usage of CRs for resources before core" (thinking of use-cases for exemptions of pre-core items like MetalLB, etc).

adrs/0006-yaml-first-crd-workflow.md
```diff
@@ -0,0 +1,64 @@
...
+3. **Redundant Conversion Pipeline**: The current workflow effectively converts schemas from TypeScript to Kubernetes YAML (via the cluster) and then back into TypeScript types, adding complexity and increasing opportunities for drift.
+4. **Synchronization Overhead**: Keeping TS sources, the cluster's state, and the generated types in sync is a manual process that is prone to drift.
```
mjnagel: Not sure I follow this one - I think we're actually guaranteed these are in sync between the task and diff check we do? And I don't think that aspect (a task + diff check) change much with the updates here.

adrs/0007-cluster-less-crd-generation-workflow.md
```diff
@@ -0,0 +1,64 @@
...
+3. **Redundant Conversion Pipeline**: The current workflow effectively converts schemas from TypeScript to Kubernetes YAML (via the cluster) and then back into TypeScript types, adding complexity and increasing opportunities for drift.
+4. **Synchronization Overhead**: Keeping TS sources, the cluster's state, and the generated types in sync is a manual process that is prone to drift.
+
+## Decision
+
+We will transition to a YAML-first CRD management workflow that preserves the benefits of TypeScript-based schema composition while enabling cluster-less type generation and static manifest tracking. While TypeScript remains the authoring source of truth, the generated YAML manifests will be the primary checked-in artifact used for registration, inspection, and type generation.
+
+### Workflow Architecture
+
+1. **Maintain TS Sources**: We will continue using `src/pepr/operator/crd/sources/*.ts` as the authoring source of truth for schema definitions, allowing us to leverage code reuse (e.g., shared Istio schema objects) and strong typing during design.
```
mjnagel, line 32: It feels wrong to call this "yaml-first" when the typescript schema/source is the real source of truth?

adrs/0007-cluster-less-crd-generation-workflow.md
```diff
@@ -0,0 +1,64 @@
...
+2. **Static Manifest Generation**: A new generation step will "compile" these TS sources into static YAML manifests located in `src/pepr/operator/crd/manifests/*.yaml`. These files will be tracked in Git.
```
mjnagel, line 33: I recognize this isn't the full implementation - but is it safe to assume this would be a standalone util? I don't think this is a capability of KFC currently. I'd be curious if we think this is a pretty straightforward thing to build and/or if it's something we could push upstream to KFC too.

adrs/0006-cluster-less-crd-generation-workflow.md
mjnagel: I think there's a lot of hidden questions/complexity wrapped up in `load these manifests and apply them via Pepr`. If we stick with Pepr applying the CRDs (which I think is fine) then I'd argue we should keep the application method/source the same (using the types). Applying yaml manifests with Pepr doesn't make a lot of sense to me (over helm like Joel mentioned), even ignoring the "load" problem (where/how are those loaded).

adrs/0007-cluster-less-crd-generation-workflow.md
slaskawi: I echo what @mjnagel said here. YAML-first is a bit misleading as TS types are the authoritative source of truth. Maybe we should consider renaming this to "cluster-less CRD generation"?

adrs/0006-cluster-less-crd-generation-workflow.md
slaskawi: From the permission perspective, if we loaded the CRDs using Helm, we could trim a few Kube RBAC permissions (the ones that relate to CRDs) from the UDS Operator/Pepr. Technically, this would slightly (just a hair) strengthen our security posture.

adrs/0006-yaml-first-crd-workflow.md
chance-coleman: maybe i worded this weirdly, i was referring to the developer friction of having to spin up a cluster to satisfy the generation tool.

adrs/0007-cluster-less-crd-generation-workflow.md
chance-coleman: agreed, i like `cluster-less crd generation`. Updated different places that was referring to yaml-first.

adrs/0006-yaml-first-crd-workflow.md
mjnagel: Gotcha - yeah I might've just misread too 😅. I think the cluster dependency is covered in painpoint 1 so was thinking this was alluding to a different issue.

adrs/0007-cluster-less-crd-generation-workflow.md
chance-coleman: Yeah it would be a standalone TypeScript utility that we’d add to the repository. In my testing, I created a script that imports our current CRD definitions and uses `js-yaml` to serialize them into standard Kubernetes manifests. It isn’t a feature of KFC currently, but seems like a good canidate.

adrs/0006-cluster-less-crd-generation-workflow.md
chance-coleman: That’s a fair point. I originally proposed keeping the registration in Pepr just to make things 'self-heal' and keep that 'just works' feeling for developers.

But I agree that using Helm for the actual install makes more sense in the long run. I'm totally on board with updating the ADR to point to Helm as the way to go for registration.

adrs/0006-yaml-first-crd-workflow.md
chance-coleman: added more context for this

adrs/0006-cluster-less-crd-generation-workflow.md
```diff
@@ -0,0 +1,64 @@
+# ADR 0006: Cluster-less CRD Generation Workflow
```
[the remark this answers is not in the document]
joelmccoy: nah

adrs/0007-cluster-less-crd-generation-workflow.md
slaskawi: I'm good to resolve this one.

adrs/0006-cluster-less-crd-generation-workflow.md
slaskawi: I would definitely be in favour of it!

adrs/0007-cluster-less-crd-generation-workflow.md
mjnagel: I don't know if this should explicitly be in the ADR or not, but if we do go this route I'd definitely advocate for us contributing this upstream to KFC since I think it has broader benefits in the ecosystem.

adrs/0006-cluster-less-crd-generation-workflow.md
mjnagel: I do think if we switch to the helm based install approach there are some timing items to consider/evaluate since the CRDs will update "separately" from the operator. I largely think that is safe, but still something to consider/note.

The same is true if/when we publish these YAML CRDs as a separate package for install - end users might update the CRDs before our operator "understands" new fields (since it hasn't updated). Nothing we can't handle, but would benefit from some docs at a minimum and maybe some code handling too (that would be at implementation time, not this ADR, just thinking out loud).

adrs/0007-cluster-less-crd-generation-workflow.md
```diff
@@ -0,0 +1,64 @@
...
+
+   * **Linting**: The new manifests in `src/pepr/operator/crd/manifests/` will be automatically covered by the existing `yamllint` pre-commit hook (configured in `.husky/pre-commit` and `.yamllint`).
+   * **Licensing**: The generated YAML files will be processed by the existing `addlicense` logic in `tasks.yaml` to ensure compliance with project standards.
+   * **Formatting**: `prettier` (via `lint-staged`) will continue to handle formatting for both the TS sources and the generated TS types.
+4. **`register.ts` Refactor**: Update the registration logic to use standard Node.js `fs` and `yaml` libraries to fetch the CRD definitions from the `manifests/` directory.
+5. **CI Synchronization**: An enhancement to the existing validation suite will be added to ensure `manifests/` stay in sync with `sources/`.
+
+## Alternatives Considered
+
+1. **Pure YAML Definitions**: Rejected because it loses the benefits of programmatic schema composition, reuse of shared objects and constants, and type-safe authoring available in TypeScript, which are central to how UDS Core evolves CRDs.
```
mjnagel, line 68: This is a great point and one of the big reasons I like starting from the TS still - especially for pulling in other schemas as subsets of ours (like Istio, etc).

adrs/0007-cluster-less-crd-generation-workflow.md
```diff
@@ -0,0 +1,64 @@
...
+2. **`tasks.yaml` Refinement**: The existing `gen-crds` and `gen-manifests` tasks will be updated to remove cluster requirements. This maintains the current developer interface while changing the underlying implementation to be cluster-less.
+3. **Tooling Integration**:
+
+   * **Linting**: The new manifests in `src/pepr/operator/crd/manifests/` will be automatically covered by the existing `yamllint` pre-commit hook (configured in `.husky/pre-commit` and `.yamllint`).
+   * **Licensing**: The generated YAML files will be processed by the existing `addlicense` logic in `tasks.yaml` to ensure compliance with project standards.
+   * **Formatting**: `prettier` (via `lint-staged`) will continue to handle formatting for both the TS sources and the generated TS types.
+4. **`register.ts` Refactor**: Update the registration logic to use standard Node.js `fs` and `yaml` libraries to fetch the CRD definitions from the `manifests/` directory.
+5. **CI Synchronization**: An enhancement to the existing validation suite will be added to ensure `manifests/` stay in sync with `sources/`.
+
+## Alternatives Considered
```
mjnagel, line 66: One very interesting thing I attempted was using [kwok](https://kwok.sigs.k8s.io/) to spin up a "cluster" and then run the `gen-crds` task against it. It's _way_ faster than a k3d cluster creation/deletion (I think it takes around 1-2 seconds extra for cluster create/destroy after the initial image pulls) and otherwise requires no process changes. Might be worth testing that out and adding to alternatives depending on your thoughts.

The main benefit I see is we don't have create a new util/process for "cluster-less" CRD generation, but it adds a different dependency and a few seconds of time.

chance-coleman: Does the kwok approach help us with the 'missing manifests' issue? One of the main points in the ADR was that we wanted authoritative YAML checked into the repo for manual inspection and GitOps pre-application. If we use kwok, don't we still end up with a cluster-only source of truth?

I might be misunderstanding something here, it seems like an improvement to our current implementation of using K3d cluster generation.

mjnagel: With an extension of the [`gen-manifests` task](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/tasks.yaml#L81-L90) we could checkin the yaml for the CRDs as well (we already do this with clusterconfig). Our current process has no real limitation to us having the YAML/putting it in git, just something we'd have to add to the tasks. Kwok or a util script/command would basically just be "speeding up" the process to get from TS source -> YAML (kwok by still using a cluster but a "speedy one", and the current ADR by using a process without a cluster).

adrs/0007-cluster-less-crd-generation-workflow.md
slaskawi: I'd definitely echo @mjnagel here and would love to call this upstream contribution out. 

I would also be highly in favor of aligning the command invocation to the "controller-gen" (as described [here](https://book.kubebuilder.io/reference/generating-crd)). This would lower the transition barrier for folks more familiar with Kube-native tools.

adrs/0006-cluster-less-crd-generation-workflow.md
slaskawi: Conceptually, upgrading CRDs first (and then the Operator image) is what most of the Operator Frameworks do. The Operator Lifecycle Manager ([link1](https://docs.redhat.com/en/documentation/openshift_container_platform/4.2/html/operators/understanding-the-operator-lifecycle-manager-olm), [link2](https://docs.okd.io/latest/extensions/ce/crd-upgrade-safety.html)) performs additional CRD compatibility checks and then upgrades the CRDs very early stage of the process. 

Assuming we will follow Kubernetes versioning scheme (this will be documented as part of https://github.com/defenseunicorns/uds-core/issues/2179) and won't introduce breaking changes, we should be on the safe side.

adrs/0007-cluster-less-crd-generation-workflow.md
slaskawi: To me, using Kwok seems more like a workaround. While it will work for us, I think adding proper support for the upstream (as we mentioned in [this](https://github.com/defenseunicorns/uds-core/pull/2270/changes#r2692391381) thread) also brings benefits to the community. I would still vote to implement a dedicated task or util that does this. Seems much cleaner approach.

adrs/0006-cluster-less-crd-generation-workflow.md
mjnagel: Yep definitely - almost every operator I've used before (outside of UDS Operator 😛) has CRDs upgrade, then the controller/operator gets upgraded after. It should generally be safe imo, and that's a great callout on the CRD versioning being relevant to this too.

adrs/0007-cluster-less-crd-generation-workflow.md
mjnagel: Yeah I think I'd advocate for adding this as a rejected alternative. It was kind of neat to test out, but ultimately would require an extra tool whereas the current documented approach would just need KFC which we already use.

The upstream/community benefits is definitely a compelling reason to go the current route too.

joelmccoy (approved):
LGTM! I like the direction we are headed here.

adrs/0007-cluster-less-crd-generation-workflow.md
```diff
@@ -55,10 +59,13 @@ The implementation will focus on enhancing existing `uds-core` development patte
    * **Linting**: The new manifests in `src/pepr/operator/crd/manifests/` will be automatically covered by the existing `yamllint` pre-commit hook (configured in `.husky/pre-commit` and `.yamllint`).
    * **Licensing**: The generated YAML files will be processed by the existing `addlicense` logic in `tasks.yaml` to ensure compliance with project standards.
    * **Formatting**: `prettier` (via `lint-staged`) will continue to handle formatting for both the TS sources and the generated TS types.
-4. **`register.ts` Refactor**: Update the registration logic to use standard Node.js `fs` and `yaml` libraries to fetch the CRD definitions from the `manifests/` directory.
-5. **CI Synchronization**: An enhancement to the existing validation suite will be added to ensure `manifests/` stay in sync with `sources/`.
+4. **RBAC Reduction**: The Pepr operator's `register.ts` will be updated to remove CRD application logic. The operator's `ClusterRole` will be updated to remove `create`, `patch`, and `update` permissions for `customresourcedefinitions`, following the Principle of Least Privilege.
```
mjnagel, line 62: Just to note - we currently don't scope the operator's RBAC at all (https://github.com/defenseunicorns/uds-core/issues/2278) but this is definitely something we intend to do, and this would allow us to remove those permissions at that point.

mjnagel (approved):

adrs/0007-cluster-less-crd-generation-workflow.md
slaskawi: I'm 100% onboard! Great spot!

slaskawi (approved):

PR #2277, opened 2026-01-15
feat: add helm chart to deploy uds exemptions; add istio gateway nodeport default exemption

### LINKED ISSUE #2161: Conditional exemption in the core gateway chart
We need a way to add an exemption for a specific component such as istio in uds-core. Currently, on a fresh deployment of core, the deployment will fail because istio needs to run in nodeport mode and there is no way to add an exemption before deploying core.

- **Given** a conditional exemption in the core gateway chart is available
- **When** we toggle the given exemption in the chart
- **Then** we can perform a fresh deployment of uds-core with istio in nodeport mode

Alternative could be a separate CRD package that would allow us to add exemptions before uds-core.

joelmccoy:
## Description

Adds a helm chart that allows deploying UDS exemptions.  Additional adds a default exemption to allow istio gateways to be `NodePort` services.

## Related Issue

Fixes #2161 

## Type of change

## Steps to Validate
- Update the standard bundle to override tenant/admin gateway to a nodeport like so:
```
      istio-tenant-gateway:
        ...
        gateway:
          values:
            - path: "service.type"
              value: NodePort
      istio-admin-gateway:
        ...
        gateway:
          values:
            - path: "service.type"
              value: NodePort
```
- Update the exemption chart [value](https://github.com/defenseunicorns/uds-core/blob/istio-gateway-exemption/src/pepr/uds-exemptions/values.yaml#L8) to `true`
- `uds run` should pass successfully and you can check to make sure the istio gateway services are of type NodePort

## Checklist before merging


docs/reference/configuration/uds-operator/exemption.md
```diff
@@ -84,3 +84,57 @@ shared:
 variables:
  # package specific variables here
 ```
+
+### Deploying Exemptions in UDS Core Installation
+
+In some cases, you may want to deploy exemption CRs as part of your UDS Core installation. This can be done by overriding the `uds-exemptions` helm chart which gets applied after pepr is installed.
```
jasonwashburn: very much a nit, and may just be my personal preference
```suggestion
In some cases, you may want to deploy custom `Exemption` CRs as part of your UDS Core installation. This can be done by overriding the `uds-exemptions` helm chart which gets applied after pepr is installed.
```

jasonwashburn (dismissed):
just a small nit, otherwise looks good...seeing the unittests still warms my heart :)

docs/reference/configuration/uds-operator/exemption.md
```diff
@@ -84,3 +84,57 @@ shared:
...
+  - name: core
+    repository: ghcr.io/defenseunicorns/packages/uds/core
+    ref: x.x.x
+    overrides:
+      pepr:
+        uds-exemptions:
+          values:
+            - path: exemptions.istioGatewayNodeport.enabled
+              value: true
+```
```
mjnagel: _very_ nit but might make sense to indent this whole block to be "under" the bullet.

src/pepr/uds-exemptions/values.yaml
```diff
@@ -0,0 +1,38 @@
...
+    enabled: false
+    # Gateways to add exemption for
+    gateways:
+      - admin
+      - tenant
+      # - passthrough
+
+# Custom exemptions - allows you to define arbitrary UDS policy exemptions
+# Each entry creates a separate Exemption CR in the uds-policy-exemptions namespace
+custom: []
```
mjnagel: Feels a bit weird to have `custom` not nested under the `exemptions` key imo?

mjnagel (commented):
Default nodeport exemption worked great, everything else LGTM. One small docs comment and one comment on the values structure.

docs/reference/configuration/uds-operator/exemption.md
joelmccoy: added

docs/reference/configuration/uds-operator/exemption.md
joelmccoy: fixed

src/pepr/uds-exemptions/values.yaml
joelmccoy: good call. updated this.

mjnagel: I think may need updates to the unittests and schema? (edit: and docs)

joelmccoy: pushed the fix

mjnagel (approved):

PR #2280, opened 2026-01-15
feat: import custom image when running uds run

### LINKED ISSUE #2279: ability to import an image into k3d when running slim-dev

I am running tests against candidate Pepr images in UDS Core and I do not want to have to publish the candidate image to Dockerhub or any other container registry as this slows down the feedback loop.

Additional Feedback Collected:
- This probably does not belong in the `packages/checkpoint-dev/zarf.yaml. It would make more sense in the dev-docs and/or a task.

- **Given** I want to run a test with a Pepr candidate controller image in UDS Core
- **When** I call `uds run ...` and I pass in a `IMPORT_IMAGE`
- **Then** I want the ability to import that image into the cluster

Make the changes I need locally and not submit a PR to UDS Core.

cmwylie19:
## Description

I need to run quick and consecutive tests against UDS Core. This gives me the ability to import a testing image into the k3d cluster instead of needing to publish it to an external registry. This is for Pepr but I think a generic implementation so you can import any image into the cluster.

## Related Issue

Fixes #2279 

Relates to #

## Type of change

## Steps to Validate

## Checklist before merging

### DIFF
--- a/packages/checkpoint-dev/zarf.yaml
+++ b/packages/checkpoint-dev/zarf.yaml
@@ -11,6 +11,10 @@ metadata:
   # x-release-please-end
 
 variables:
+  - name: IMPORT_IMAGE
+    description: "Custom image to preload into the k3d cluster"
+    default: ""
+
   - name: CLUSTER_NAME
     description: "Name of the cluster"
     default: "uds"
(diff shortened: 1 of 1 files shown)


packages/checkpoint-dev/zarf.yaml
```diff
@@ -91,6 +95,11 @@ components:
                 -v "${DATA_DIR}/k3s_data:/var/lib/rancher/k3s@server:*" \
                 --image ghcr.io/defenseunicorns/uds-core/checkpoint:latest ${ZARF_VAR_K3D_EXTRA_ARGS} \
                 ${ZARF_VAR_CLUSTER_NAME}
+
+                if [[ -n "${ZARF_VAR_IMPORT_IMAGE:-}" ]]; then
+                  k3d image import "${ZARF_VAR_IMPORT_IMAGE}" -c "${ZARF_VAR_CLUSTER_NAME}"
```
mjnagel, line 100: Putting this into the checkpoint package (and maybe even in core) feels like the wrong place? Might be a better fit for [uds-k3d](https://github.com/defenseunicorns/uds-k3d), although there's an airgap aspect to consider there (for the airgap flavor). I also might just not follow the need/use-case well enough, could this just be a task rather than baked into the zarf package?

cmwylie19: I hear you. I am honestly not sure if it belongs, it is easy enough to reproduce without having it as an official part of the afterDeploy action. I think we can close this. I wanted to give you all the opportunity to have eyes on it and see if you all thought it was useful.

mjnagel: I have run the `k3d image import` before for some similar testing and definitely see a use-case for it. I think I'd lean more towards adding this in dev-docs and/or a task instead of directly in the package.

cmwylie19: makes perfect sense - i have an issue for tracking that i will update with this info

PR #2281, opened 2026-01-15
feat: unify trust bundle management across istio, authservice, and uds packages

### LINKED ISSUE #2235: Authservice and Istio should consume the UDS shared trust bundle (including DoD/Public CAs)
Authservice and Istio currently rely only on `CA_BUNDLE_CERTS` for their trust chain. When users enable DoD/Public CAs via UDS (`e.g. CA_BUNDLE_INCLUDE_DOD_CERTS`), those certs are correctly added to the generated UDS trust bundle, but Authservice and Istio do not automatically consume them. This leads to TLS verification failures unless users manually re-add DoD CAs to `CA_BUNDLE_CERTS`, which is confusing and error-prone.

This feels inconsistent with the spirit of the feature and creates unnecessary cognitive load for end users.

- **Given** DoD/Public CAs are enabled in UDS and included in the generated Cluster trust bundle
- **When** Authservice and Istio are deployed
- **Then** they should automatically trust the same combined CA bundle generated by the operator

- Requiring users to manually inline DoD/Public CAs into `CA_BUNDLE_CERTS`
- Adding additional, explicit opt-in flags just for `Authservice/Istio trust`

All alternatives increase complexity or user error compared to a unified trust bundle.

### Additional context
There are two places where this needs adjusted:
1. [JWKS Fetcher in Istio](https://github.com/defenseunicorns/uds-core/blob/b3190bf5df8273430b87fe31782091c2687fe6af/src/istio/values/base-istiod.yaml#L29-L41) - note: since Istio doesn't have a package CR, the operator code must be updated to create the configmap in this namespace without a Package
2. [Authservice trust chain](https://github.com/defenseunicorns/uds-core/blob/b3190bf5df8273430b87fe31782091c2687fe6af/src/pepr/operator/controllers/keycloak/authservice/authservice.ts#L240-L246) - this could be swapped to use an actual mounted file with [trusted_certificate_authority_file](https://github.com/istio-ecosystem/authservice/blob/36bece28d991a0f87010246505a59df4302ff4ad/config/README.md?plain=1#L89)

chance-coleman:
## Description
Implements a unified CA trust bundle in the Pepr operator, consolidating user-provided, DoD, and public certificates into a single source of truth.

- Centralized Trust: Reconciles a combined PEM bundle into UDSPackage ConfigMaps and the istio-system/sso-ca-cert secret.
- Improved Reliability: Replaces global manual reloads with native reloader mounting for packages to prevent cluster-wide thundering herd issues.
- Istio/Authservice Support: Ensures consistent trust propagation across Istio and Authservice for Private PKI environments.

## Related Issue

Fixes #2235 

## Type of change

## Checklist before merging


src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -163,31 +211,42 @@ export async function updateAllCaBundleConfigMaps(): Promise<void> {
     log.debug(`Found ${packages.items.length} UDS packages, processing CA bundle ConfigMaps`);
 
     // Process each package and update/delete its CA bundle ConfigMap as needed
-    for (const pkg of packages.items) {
+    const packageUpdates = packages.items.map(async pkg => {
```
joelmccoy: awesome refactor here 😎

src/pepr/operator/controllers/config/config.ts
```diff
@@ -475,8 +478,9 @@ async function performAuthserviceUpdate(reason: string) {
   const authserviceUpdate: AuthServiceEvent = {
     name: "global-config-update",
     action: Action.UpdateGlobalConfig,
-    // Base64 decode the CA cert before passing to the update function
-    trustedCA: atob(UDSConfig.caBundle.certs),
+    // Note: Use the combined CA bundle (User + DoD + Public) for Authservice trust.
+    // Authservice needs the raw PEM content, not base64. buildCABundleContent() gives us the combined, decoded string it expects.
+    trustedCA: buildCABundleContent(),
```
joelmccoy, line 491: I think this approach is fine for authservice.  But is there any strong reason why you went this way instead of mounting the configmap created by the UDS package to the authservice deployment and setting the [trust_certificate_authority_file](https://github.com/istio-ecosystem/authservice/blob/36bece28d991a0f87010246505a59df4302ff4ad/config/README.md?plain=1#L89)?  Then updates to the certs will happen on the configmap automatically and we can use the native pod reload logic instead of writing custom operator code to handle this specific scenario?

I think the approach you have in this PR now is fine.  But I _think_ we could remove some custom operator logic for handling cert changes for authservice and this would reflect the same pattern used by other mission apps.

src/pepr/operator/controllers/keycloak/authservice/authservice.spec.ts
joelmccoy, line 1: Commenting here because there is no other great place to put this- 
The private-pki workflow didn't trigger on this run and we might want to trigger it to show confidence in CI.  However, I'm also realizing we really don't have e2e tests for private-pki + authservice.  Instead of adding another e2e test, do you think it might be a good idea to run our standard [e2e tests](https://github.com/defenseunicorns/uds-core/blob/v0.59.1/tasks/test.yaml#L92-L124) that have authservice apps deployed as a part of the [private-pki workflow](https://github.com/defenseunicorns/uds-core/blob/v0.59.1/.github/workflows/private-pki-test.yaml#L56)?

joelmccoy (commented):
First code pass at this and generally looking good.  Going to play around with this locally later.  A couple small comments/questions.  Additionally, as a nit: Do you think it is worth updating the pr title/description to reflect that this PR integrates the centralized trust bundle management feature with Istio/Authservice? It might be more clear to end users that the existing trust bundle management feature now automatically works with Istio/Authservice.

src/pepr/operator/controllers/keycloak/authservice/authservice.spec.ts
mjnagel: Maybe missing what you're saying in the second half of this comment, but I think we do have testing?
- Deploy workload with authservice: https://github.com/defenseunicorns/uds-core/blob/v0.59.1/tasks/test.yaml#L230-L234
- Test of workload: https://github.com/defenseunicorns/uds-core/blob/v0.59.1/test/playwright/private-pki/private-pki.test.ts#L29-L46

Agreed on triggering private-pki for this PR.

joelmccoy: Ah disregard. We do have a test for authservice as you pointed out.

But still wondering if its worth doing the standard e2e tests we run in our typical test workflow in the private-pki workflow.  I'm just thinking as we have private-pki logic more integrated into core/other apps, we may want to make sure other things just don't break?  Right now we only smoke test grafana and authservice on our private-pki workflow.  But all the standard e2e tests should work as well? Definitely not needed for this PR, but thought I would call it out.

src/pepr/operator/controllers/config/config.ts
chance-coleman: That's a valid point for sure. My approach was mostly from a place of simplicity / least resistance. Unified secret meant no real changes to how authservice's configuration works with the config.json inside the secret and injecting the CA data kept the login in once place. Authservice also already required the rollout for any configuration change so reusing the checksum deployment was convenient. 

Benefits of what you're suggesting i think:
- Consistency: Matches the pattern used by tenant packages
- Simpler: We could potentially stop passing trustedCA through the AuthServiceEvent chain, as the pod would just look at its local disk
- Reduced Secret Size: Large CA bundles (like the DoD one) can make the AuthserviceConfig JSON quite large

docs/reference/configuration/trust-management/central-trust-bundle-management.md
```diff
@@ -14,7 +14,10 @@ The central trust bundle management feature addresses the challenge of maintaini
 - **Automatically distribute trust stores**: Generate and distribute trust bundles to all namespaces
```
mjnagel: Since it's top of mind - this is really `all namespaces with a Package CR in them` (could maybe update while we're editing this doc).

docs/reference/configuration/trust-management/private-pki.md
```diff
@@ -76,9 +76,16 @@ Many Go-based applications automatically check `/etc/ssl/certs/ca.pem` for addit
 
 ### Authservice
 
-Authservice can be configured with an additional trusted CA bundle when UDS Core ingress gateways are deployed with private PKI.
+Authservice automatically trusts the combined UDS CA bundle (Private PKI + DoD + Public CAs) when configured via the UDS Operator. This ensures it can verify the certificates of the tenant gateway without manual certificate injection.
 
-To configure, set `UDS_CA_BUNDLE_CERTS` as an environment variable with a Base64 encoded PEM formatted CA bundle that can be used to verify the certificates of the tenant gateway.
```
mjnagel, line 81: I think the doc rewrite lost the note about setting this via environment variable. Maybe okay to point folks more towards the config file path - just wanted to make sure that was intentional.

src/pepr/operator/controllers/config/config.ts
joelmccoy: Yep you nailed the benefits better than I could articulate.  Definitely think the current approach is still fine (and don't know if there would be quirks with the mounted file way), so I will defer to you (or is someone else wants to weigh in) on the best approach.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -98,6 +99,53 @@ export async function caBundleConfigMap(pkg: UDSPackage, namespace: string): Pro
   }
 }
 
+/**
+ * Updates the Istio sso-ca-cert secret with the combined CA bundle.
+ * Istio expects its root CA in the `extra.pem` key of this secret.
+ * We only update it if the secret already exists, otherwise, we let the Istio Helm chart handle the initial creation.
+ */
+export async function updateIstioCASecret(): Promise<void> {
```
mjnagel: Curious on your thoughts here - it feels a bit strange to me to manage the initial configmap creation [via the helm chart](https://github.com/defenseunicorns/uds-core/blob/2235-trust-bundle/src/pepr/uds-operator-config/templates/ca-secret.yaml) still, while managing the upgrades/changes with the operator (I'd think maybe we could just move creation into the operator too). As-is I think the flow is essentially:
- helm chart makes configmap, conditionally with the custom ca cert (but never with public/DoD)
- operator updates configmap (conditionally based on config) to add public/DoD CAs

The upgrade flow is also something that could get weird, although helm might just relinquish ownership? Regardless I think I would advocate for making the configmap in the operator as well unless there's a compelling reason to keep creation in helm.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -98,6 +99,53 @@ export async function caBundleConfigMap(pkg: UDSPackage, namespace: string): Pro
...
+  const namespace = "istio-system";
+  const secretName = "sso-ca-cert";
+
+  try {
+    // Build the combined CA bundle content
+    const caBundleContent = buildCABundleContent();
+
+    // Get existing secret
+    const secret = await K8s(kind.Secret).InNamespace(namespace).Get(secretName);
```
mjnagel: If we fully manage the secret (create -> update -> delete) we could probably remove this get and just always construct the secret, then apply. We could also switch this to a configmap if we want (don't think there was a good reason for making it a secret originally).

src/pepr/operator/controllers/config/config.ts
chance-coleman: I'm a bit torn. Mostly think the question is should we do this now or in a follow on PR. I believe the mounted approach to be a better long term approach but if i understand it correctly it comes with some baggage we would have to figure out. 
- If the mount is mandatory and the ConfigMap is missing, the pod won't start. If it's optional, the app might start but fail TLS handshakes.
- Kubernetes takes time to project ConfigMap changes into a pod. Even with a rollout, there's a tiny window where the file might not match the expected state.
- Since uds-ca-certs is in pepr-system, we need to ensure the operator logic reliably creates the uds-trust-bundle in the authservice namespace.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -276,6 +276,9 @@ async function handleCABundleUpdate(caBundle: ConfigCABundle, updateClusterResou
     if (updateClusterResources) {
       await performAuthserviceUpdate("change to CA Cert");
     }
+  } else if (caBundleConfigMapsNeedUpdate && updateClusterResources) {
+    // Even if user certs didn't change, DoD/Public flags or content might have, so update Authservice
+    await performAuthserviceUpdate("change to DoD/Public CA configuration");
```
mjnagel: This seems like it could be simplified to just the `else if` clause, or even just move the `performAuthserviceUpdate` call down inside the other conditional (line 290 currently)? I don't know if there's a ton of value in keeping the different "reasons" personally (would be more in favor of simplifying this to make it clear that we update both all configmaps and authservice based on the same conditional).

src/pepr/operator/controllers/config/config.ts
mjnagel: This is a great discussion here. Largely would agree that switching to `trusted_certificate_authority_file` would make more sense longer-term. I was initially concerned with complexity of the "reload" on the authservice side because I recalled `trusted_certificate_authority_refresh_interval`, but looks like that is actually [deprecated](https://github.com/istio-ecosystem/authservice/blob/36bece28d991a0f87010246505a59df4302ff4ad/config/README.md?plain=1#L90) and Authservice just auto-reloads when the file changes. This should save us from some pod restarts, and better align the management and control of that configmap with how we do it for other Packages (in fact that configmap should already exist for the authservice namespace).

My only real lingering question would be around the config and whether we need to add/remove `trusted_certificate_authority_file` from the secret when adding/removing a CA, or if we can just point to a file always in the config (even if it doesn't exist). If we can just have `trusted_certificate_authority_file` always set in the config, that will definitely simplify things. Otherwise we'll still need some management of cycling authservice for CA related things (albeit less frequently).

I'm okay with leaving this for a follow-on PR, but I do think we should make this change at some point.

chance-coleman: I've created [this issue](https://github.com/defenseunicorns/uds-core/issues/2288) to track moving to the mounted configmap approach. Feel free to update it with anything else

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
chance-coleman: refactored to use a configmap

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
chance-coleman: this got a little confusing with the switch to using a configmap instead of a secret for the ca-secret. but was able to consolidate the logic here a bit.

docs/reference/configuration/trust-management/private-pki.md
chance-coleman: added a note that its possible, although kept the focus on doing via the uds-config

docs/reference/configuration/trust-management/private-pki.md
```diff
@@ -76,9 +76,16 @@ Many Go-based applications automatically check `/etc/ssl/certs/ca.pem` for addit
...
+To configure, define the `caBundle` variables in your `uds-config.yaml` (or via Zarf variables like `UDS_CA_BUNDLE_CERTS` at deploy time). For example:
```
mjnagel:
```suggestion
To configure, define the `caBundle` variables in your `uds-config.yaml` (or via environment variables with the `UDS_` prefix like `UDS_CA_BUNDLE_CERTS` at deploy time). For example:
```

src/pepr/operator/controllers/config/config.ts
```diff
@@ -269,26 +269,19 @@ async function handleCABundleUpdate(caBundle: ConfigCABundle, updateClusterResou
...
-  // Handle global updates to Trust Bundle Configmaps
-  if (caBundleConfigMapsNeedUpdate && updateClusterResources) {
-    await updateAllCaBundleConfigMaps();
+  // Handle global updates to Trust Bundle Configmaps and Authservice
+  if (updateClusterResources) {
+    await performAuthserviceUpdate(
+      caBundleConfigMapsNeedUpdate ? "Global CA bundle change" : "Idempotent sync",
+    );
+    await updateAllCaBundleConfigMaps(!caBundleConfigMapsNeedUpdate);
   }
```
mjnagel, line 286: I'm not really following the change here to the conditional (removing `caBundleConfigMapsNeedUpdate`). It seems like that was what required the addition of the `skipReload` param, but not following why we couldn't keep the old conditional alternatively? If the cabundle hasn't changed there shouldn't be any need to re-run those applies?

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
mjnagel: Resolving this one, I think looks good with the latest push for the most part.

src/pepr/operator/controllers/keycloak/authservice/authservice.spec.ts
mjnagel: I think this still needs to be addressed - at minimum would like to see us trigger the private-pki workflow with these changes (could just make a comment change or something).

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -98,6 +99,57 @@ export async function caBundleConfigMap(pkg: UDSPackage, namespace: string): Pro
   }
 }
 
+/**
+ * Updates the Istio uds-trust-bundle ConfigMap with the combined CA bundle.
+ * Istio expects its root CA in the `extra.pem` key of this ConfigMap.
+ * This ConfigMap is managed entirely by the operator.
+ */
+export async function updateIstioCASecret(skipReload = false): Promise<void> {
```
joelmccoy:
```suggestion
export async function updateIstioCAConfigMap(skipReload = false): Promise<void> {
```

src/pepr/operator/controllers/config/config.ts
chance-coleman: By syncing every time, the operator ensures the desired state always exists. On fresh installs, this ensures the ConfigMap is populated even if the operator missed the initial ADD event or started after the ClusterConfig was already created. By always running the Apply but using skipReload when no change is detected, we guarantee the resource is present and correct without the performance hit or service churn of restarting pods (Istiod/Authservice). tldr: it moves from update only on change to ensure state is always correct

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -148,46 +200,38 @@ function buildCABundleContent(): string {
  *
  * @throws Error if the package listing or ConfigMap update operations fail
  */
-export async function updateAllCaBundleConfigMaps(): Promise<void> {
+export async function updateAllCaBundleConfigMaps(skipReload = false): Promise<void> {
```
joelmccoy: nit: might be good to update the tsdocs on this one.  reading this on first glance `skipReload` confused me a lot until i traced to the fact that it is only used on the istio functionality.  would it be better to name this `skipIstioReload` to be more clear?

src/pepr/operator/controllers/config/config.ts
mjnagel: Gotcha - definitely following better now. This change helps/supports ensuring that the configmap for istio-system is created "at startup" of Pepr (versus potentially getting skipped otherwise). I think this seems sane, just comprehending if there's any changes we could make to make it a bit clearer on the purpose/behavior.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -98,6 +99,57 @@ export async function caBundleConfigMap(pkg: UDSPackage, namespace: string): Pro
...
+ * Istio expects its root CA in the `extra.pem` key of this ConfigMap.
+ * This ConfigMap is managed entirely by the operator.
+ */
+export async function updateIstioCASecret(skipReload = false): Promise<void> {
+  const namespace = "istio-system";
+  const configMapName = "uds-trust-bundle";
+
+  try {
+    // Build the combined CA bundle content
+    const caBundleContent = buildCABundleContent();
```
joelmccoy: When i run the private pki tests locally: `uds run -f tasks/test.yaml uds-core-private-pki`

I am getting the `uds-trust-bundle` in `istio-system` to NOT have the dod/public certs even though it should. When i edit the clusterconfig and to trigger a reconciliation the dod/publics certs are added.  It seems there is some sort of timing issue that I haven't figured out exactly what is happening on initial creation of this configmap.

src/pepr/uds-operator-config/templates/uds-trust-bundle.yaml
```diff
@@ -2,10 +2,9 @@
...
-kind: Secret
+kind: ConfigMap
 metadata:
-  name: sso-ca-cert
+  name: uds-trust-bundle
   namespace: istio-system
-type: Opaque
 data:
-  extra.pem: ""
+  extra.pem: {{ default .Values.cluster.expose.caCert .Values.cluster.caBundle.certs | default "" | quote }}
```
joelmccoy: This is getting set to this
image and not being updated on my deploy when i run the private-pki tests.

As opposed to this other trust bundle created by the operator: 
image


A couple issues here:
1. I think this should be base64 decoded before putting in this manifest?
2. It doesn't include the dod/public certs if true/and it doesn't get updated unless i trigger a change to the cluster config

This essentially is not working for me locally and I get a JWT error with the authservice enabled application.
image

It seems we need to update our playwright test to catch this. I think by adding this at the end of this [private-pki-test](https://github.com/defenseunicorns/uds-core/blob/v0.59.1/test/playwright/private-pki/private-pki.test.ts#L44-L47) it will catch this:
```
    // Verify non-error response
    const response = await page.reload();
    expect(response?.status()).toBeLessThan(400);
```

joelmccoy (changes requested):
Having some things break for me locally.  Dropped some comments.  Still unsure of the root cause and will try to trace this and provide more context.  But wanted to drop my initial findings.

src/pepr/uds-operator-config/templates/uds-trust-bundle.yaml
mjnagel: Ah - I actually missed this the first time around. I don't think this template should even exist? Per my [previous thread](https://github.com/defenseunicorns/uds-core/pull/2281#discussion_r2709550617) we really should be entirely managing creation in the operator imo, which I thought the latest push enabled. I'm not 100% sure but this timing issue might disappear just by deleting this template/file?

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
chance-coleman: ahh yeah i think i know whats going on here. On initial creation, the uds-trust-bundle is missing the DoD/Public certs. When `pepr.ts` starts, it calls `loadUDSConfig()`, if the uds-ca-certs ConfigMap (which contains the DoD/Public raw data) hasn't been processed yet, or if the `handleUDSCACertsConfigMapUpdate` watcher hasn't completed, `UDSConfig.caBundle.dodCerts` will be empty. I'll look into the synchronization here and see what can be done

src/pepr/uds-operator-config/templates/uds-trust-bundle.yaml
joelmccoy: When i remove the template file istiod seems to hang because the configmap is not created:
```
MountVolume.SetUp failed for volume "sso-ca" : configmap "uds-trust-bundle" not found
```

The configmap should exist though? Pepr should create it on startup? I wonder if the `istio-system` namespace might not exist yet and would cause an issue?

mjnagel: I think maybe because of `if (updateClusterResources) {`? Which gates those functions to only run on updates. The istio configmap is a "unique" case where we need to run creation on startup (somewhat similar to how we create [other secrets on startup](https://github.com/defenseunicorns/uds-core/blob/main/pepr.ts#L36-L38)).

For the other configmaps we don't need to create/update them on startup because every package CR will get reconciled and created/updated later in the flow of the normal reconcile.

chance-coleman: Just to make sure i'm on the same page here, we are okay with Istiod requiring the `uds-trust-bundle` configmap to bootstrap itself? or was there something that we wanted to change, outside of the other issues with race conditions and base64 encoding

mjnagel: Istiod can/should still require that as a mount I think. We just need to make sure that the configmap always exists, even if it's empty (historically that's what the secret template did, just make a configmap with no data if there was no CA things configured).

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
chance-coleman: new commit ensures the loading of hte config is done before the handleCfg is called. also added a couple tests to make sure that pre-fetch logic is sound.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
chance-coleman: updated a couple tsdocs in this file

src/pepr/uds-operator-config/templates/uds-trust-bundle.yaml
mjnagel: I'm still not following why the template (`src/pepr/uds-operator-config/templates/uds-trust-bundle.yaml`) is being kept around? While istio does need that configmap for a mount, I think Pepr should be fully managing it (handling the initial create rather than that being in the chart). Timing-wise there shouldn't be any issue with this either since [Pepr starts up before istio](https://github.com/defenseunicorns/uds-core/blob/main/packages/base/zarf.yaml#L37-L53).

Unless we remove that we're still going to have the initial deploy only have `.Values.cluster.caBundle.certs`, and then get updated later with the DoD/Public certs which seems odd at best, potentially problematic at worst.

src/pepr/operator/controllers/keycloak/authservice/authservice.spec.ts
chance-coleman: this should be running now after comment in the private-pki tests.

src/pepr/uds-operator-config/templates/uds-trust-bundle.yaml
chance-coleman: removed template and added a check that the istio namespace is present and that the operator focus on the uds-trust-bundle and not all UDSPackages (which may not be present yet).

src/pepr/operator/controllers/config/config.ts
joelmccoy: This logic really has me tripped up. I think I'm fine with force syncing the configmaps everytime (I don't think there are any consequences other than maybe some extra unnecessary cycles?).

But now the function shouldUpdateCaBundleConfigMaps [here](https://github.com/defenseunicorns/uds-core/blob/2235-trust-bundle/src/pepr/operator/controllers/config/config.ts#L190) is VERY misleading as it now has nothing to do with determining whether CA Bundle Configmaps are updated. Essentially this function becomes just a way to determine if we need to restart istio. Can we rename this appropriately to something like caConfigChanged or something like that (with updated TSdocs?)

src/pepr/operator/controllers/config/config.spec.ts
```diff
@@ -834,20 +861,25 @@ describe("handleUDSConfig", () => {
       expect(UDSConfig.caBundle.publicCerts).toBe("");
     });
 
-    it("throws error for K8s API failures when fetching ConfigMap", async () => {
+    it("gracefully handles K8s API failures when fetching ConfigMap by using defaults", async () => {
```
joelmccoy: It looks like the refactor here doesn't throw any errors if we can't find the DoD/Public certs (which SHOULD always exist in a UDS Core deployment).  Is this an intended change? This may be fine? But would be concerning if an environment is using trust bundles and then somehow we hit a k8s api error here AND we essentially reset all their trust bundles to have null DoD/Public certs (potentially breaking their applications).  I think I lean towards throwing an error and stopping the reconciliation?

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -98,6 +99,66 @@ export async function caBundleConfigMap(pkg: UDSPackage, namespace: string): Pro
...
+ */
+export async function updateIstioCAConfigMap(skipIstioReload = false): Promise<void> {
+  const namespace = "istio-system";
+  const configMapName = "uds-trust-bundle";
+
+  try {
+    // Ensure the namespace exists
+    await K8s(kind.Namespace).Apply({
+      metadata: { name: namespace },
+    });
```
joelmccoy: note: on a new deploy of UDS Core, the istio-system namespace will now always be managed via the operator and not the zarf deploy.  It will now miss this annotation `app.kubernetes.io/managed-by=zarf`.  Which I _think_ is fine? just wanted to call this out unless there are other implications here that I am not thinking about?

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -98,6 +99,66 @@ export async function caBundleConfigMap(pkg: UDSPackage, namespace: string): Pro
...
+    // Skip reload if requested (e.g., during initial load or batch updates)
+    if (skipIstioReload) {
+      log.debug(`Skipping Istiod reload for ${configMapName} update`);
+      return;
+    }
+
+    // Reload Istiod to ensure it picks up the trust bundle change
+    try {
+      const istioPods = await K8s(kind.Pod).InNamespace(namespace).WithLabel("app", "istiod").Get();
+      await reloadPods(namespace, istioPods.items, "CA bundle update", log, "CA_BUNDLE");
```
joelmccoy: i'm now realizing.... can't we get rid of this custom operator code and just use the pod reload feature on the configmap?

I think we can just add this label to the configmap we create above and the reload logic will do the magic??
```
  labels:
    uds.dev/pod-reload: "true"
```

Then we can get rid the `skipIstioReload` parameter as a whole?

test/playwright/private-pki/private-pki.test.ts
joelmccoy, line 1: The Authservice private PKI test [here](https://github.com/defenseunicorns/uds-core/blob/2235-trust-bundle/test/playwright/private-pki/private-pki.test.ts#L30-L48) does not pick up failed cert issues.  Authservice will redirect but return a error status with `JWT Verification Fails`. See [this](https://github.com/defenseunicorns/uds-core/pull/2281#discussion_r2714041273) comment for more details.  Can we check the status on the redirected page to make sure we cover our basis here?

I think you could just add something like this at the end of the authservice test:
```
    // Verify non-error response
    const response = await page.reload();
    expect(response?.status()).toBeLessThan(400);
```

joelmccoy (commented):
Generally LGTM - tested locally.  But have a couple comments for consideration.  Being extra nit picky on this as the config logic is getting a bit confusing to follow and would want to simplify/be clear where possible.

test/playwright/private-pki/private-pki.test.ts
chance-coleman: I added a status code check to the Playwright tests. It now reloads the page and verifies a < 400 response, which confirms that Authservice successfully validated the JWT against the trust bundle.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
chance-coleman: I deleted the custom reload logic and added the uds.dev/pod-reload: "true" label to the uds-trust-bundle ConfigMap. Now, the generic UDS pod-reload controller handles the restarts automatically.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
chance-coleman: It's totally fine for the app to work, but it changes how Zarf cleanup works. Without that annotation, zarf package remove won't automatically delete the istio-system namespace because Zarf doesn't think it "owns" it anymore.

However, since the Operator is now the source of truth, it's actually safer, the Operator will proactively fix the namespace if it drifts, whereas Zarf only checks it during a deploy. If we want the best of both worlds, we could just slap that label on in our Apply call so Zarf still knows it's okay to clean up.

src/pepr/operator/controllers/config/config.spec.ts
chance-coleman: I updated fetchCACerts to explicitly throw an error on any non-404 failure. This stops the reconciliation and marks the ClusterConfig as Failed, preventing accidental data loss.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
mjnagel: We also do the same namespace creation [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/config.ts#L19-L23) for keycloak and [here](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/authservice/config.ts#L57-L61) for authservice so I don't have an issue with this.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
```diff
@@ -98,6 +98,55 @@ export async function caBundleConfigMap(pkg: UDSPackage, namespace: string): Pro
...
+ *
+ * @param skipIstioReload - If true, skips restarting Istiod pods after updating the ConfigMap.
+ *
+ * @throws Error if the ConfigMap cannot be applied to the cluster.
+ */
+export async function updateIstioCAConfigMap(): Promise<void> {
+  const namespace = "istio-system";
+  const configMapName = "uds-trust-bundle";
+
+  try {
```
mjnagel: Functionally I don't think it will ever cause conflicts, but this probably should be wrapped with a conditonal on watch or dev mode (like [this](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/config.ts#L17)). I believe for updates it will only get called on watch/dev, but for the initial load call (from `loadUDSConfig`) it would be called by all pods.

src/pepr/operator/controllers/keycloak/authservice/authservice.ts
```diff
@@ -238,7 +238,7 @@ export function buildConfig(config: AuthserviceConfig, event: AuthServiceEvent)
       };
     }
     if (!event.trustedCA) {
-      // Remove the trusted certificate authority if a CA is not provided
+      // Remove both trust configurations if a CA is not provided
```
mjnagel: super nit: what does "both" refer to here? Maybe more accurate to just note `trusted CAs` or something since it's no longer just a single CA, but also not necessarily strictly 2.

mjnagel (dismissed):
LGTM overall - 2 very small nits for your review. I'd also like to run a few upgrade/change scenarios and make sure things behave as expected but non-blocking since CI all worked.

src/pepr/operator/controllers/ca-bundles/ca-bundle.ts
chance-coleman: added this conditional and updated tests

mjnagel (dismissed):

src/pepr/operator/controllers/config/config.ts
chance-coleman: I updated the function from `shouldUpdateCaBundleConfigMaps` to `caConfigChanged`, updated tsdoc and references as well.

joelmccoy (approved):
LGTM

PR #2286, opened 2026-01-20
chore: add mappings

chance-coleman:
## Description
add mappings

## Type of change

## Checklist before merging

### DIFF
--- a/src/grafana/chart/templates/uds-package.yaml
+++ b/src/grafana/chart/templates/uds-package.yaml
@@ -1,7 +1,7 @@
 # Copyright 2024 Defense Unicorns
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
-# @lulaStart 506992d2-64cf-4fb3-95de-40a8ad127626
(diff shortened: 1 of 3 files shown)


src/grafana/chart/templates/uds-package.yaml
mjnagel, line 49: I'd probably advocate for just wrapping this expose section to reduce the scope of the mapping/uuid?

src/keycloak/chart/templates/uds-package.yaml
mjnagel, line 120: Similar comment on this one, I think we can scope to just the expose section.

chance-coleman: made the mapping a bit more fine grained to encapsulate the expose section only

src/grafana/chart/templates/uds-package.yaml
chance-coleman: made the mapping a bit more fine grained to encapsulate the expose section only

mjnagel (approved):

joelmccoy (approved):

PR #2287, opened 2026-01-20
feat!: istio service mesh ambient mode default

### LINKED ISSUE #2249: Switch to `ambient` as default Istio mode for Packages
Given the length of time that we have had ambient as an option in UDS Core for Packages (almost 9 months with the 0.40.0 release), we should consider switching to `ambient` as the default Istio mode for Packages.

This will be a breaking change and should be noted as such. We will also want to ensure all documentation is updated.

briantwatson:
## Description

- Sets ambient mode as default

## Related Issue

Fixes #2249 

## Type of change

## Steps to Validate
n/a

## Checklist before merging


docs/reference/configuration/uds-operator/package.md
```diff
@@ -39,9 +39,11 @@ metadata:
   name: grafana
   namespace: grafana
 spec:
-  network:
-    # Configure Istio Mode: sidecar or ambient
-    serviceMesh:
+    # Ambient mode is default; omit or set explicitly
+    serviceMesh: {}
```
joelmccoy:
```suggestion
    serviceMesh:
```
you have mode set below to ambient. I think you can remove the { } here

docs/reference/uds-core/istio-ambient-transition.md
```diff
@@ -35,7 +35,7 @@ These benefits are particularly important as UDS expands its **edge capabilities
 | Component                | Transition Path                                                                  |
 | ------------------------ | -------------------------------------------------------------------------------- |
 | **UDS Core Services**    | All deployments will use Istio Ambient Mode by default starting **0.43.0**.      |
-| **Mission Applications** | Both sidecar and ambient modes supported. Ambient is **opt-in** and recommended. |
+| **Mission Applications** | Both sidecar and ambient modes supported. Ambient is **default** and recommended. |
```
joelmccoy: nit: it might be good to call out the uds core version that ambient became default in here (like we did for the line above).  I'm assuming this would be 0.60.0 unless we decided to wait getting this feature in.

src/test/app-sidecar-authservice-tenant-package.yaml
joelmccoy, line 1: very nit: our [curl test packages](https://github.com/defenseunicorns/uds-core/blob/feat/istio-ambient-default/src/test/app-curl-packages.yaml) will now switch to ambient for our tests which I think is totally fine.

However, It just may be good to cleanup these sidecar [annotations](https://github.com/defenseunicorns/uds-core/blob/feat/istio-ambient-default/src/test/app-curl.yaml#L35-L36) in the deployment.

joelmccoy: I think we also need to switch this egress package to sidecar explicitly: https://github.com/defenseunicorns/uds-core/blob/feat/istio-ambient-default/src/test/app-egress-package.yaml#L17

This package is used to test egress when in sidecar mode (which is a different flow than ambient mode)

joelmccoy (commented):
Generally LGTM!  A couple small comments and make sure to mark this PR title as `feat!:` to indicate a breaking change

docs/reference/configuration/uds-operator/package.md
briantwatson: Thanks for catching this! Removed.

docs/reference/uds-core/istio-ambient-transition.md
briantwatson: Adjusted to add version number.  Chose some simple verbiage but welcome suggestions if we want the wording changed.

src/test/app-sidecar-authservice-tenant-package.yaml
briantwatson: Thanks for catching these.  Made adjustments you suggested.  I also removed annotations in the [app-egress-ambient.yaml](https://github.com/defenseunicorns/uds-core/blob/825c1ed6ea58df6432080174408528915ea0059c/src/test/app-egress-ambient.yaml#L36-L37), as those seem unnecessary in that test.

src/pepr/operator/controllers/istio/namespace.ts
```diff
@@ -49,12 +49,12 @@ export async function enableIstio(pkg: UDSPackage) {
   annotations[pkgKey] = "true";
 
   // Handle labels based on ambient opt-in or sidecar default
```
chance-coleman:
```suggestion
  // Handle labels based on sidecar opt-in or ambient default
```

chance-coleman (commented):
Should we be changing the operator to also be defaulting to Ambient mode? I might have missed a conversation somewhere but my understanding is that anywhere in the operator code that mentions `Mode.Sidecar` or `serviceMesh.mode` probably also needs to be evaluated. For example the [package validator has a check for istio mode](https://github.com/defenseunicorns/uds-core/blob/825c1ed6ea58df6432080174408528915ea0059c/src/pepr/operator/crd/validators/package-validator.ts#L25).

docs/reference/uds-core/istio-ambient-transition.md
```diff
@@ -32,10 +32,10 @@ These benefits are particularly important as UDS expands its **edge capabilities
...
 ## Transition Details
 
-| Component                | Transition Path                                                                  |
-| ------------------------ | -------------------------------------------------------------------------------- |
-| **UDS Core Services**    | All deployments will use Istio Ambient Mode by default starting **0.43.0**.      |
-| **Mission Applications** | Both sidecar and ambient modes supported. Ambient is **opt-in** and recommended. |
+| Component                | Transition Path                                                                   |
+| ------------------------ | --------------------------------------------------------------------------------- |
+| **UDS Core Services**    | All deployments will use Istio Ambient Mode by default starting **0.43.0**.       |
+| **Mission Applications** | Ambient is the **default** mode starting **0.60.0**.                              |
```
mjnagel:
```suggestion
| **UDS Core Services**    | All Core services use Istio Ambient Mode starting **0.43.0** (sidecar is not an option for Core services).       |
| **Mission Applications** | Ambient is the **default** mode starting **0.60.0**. Sidecar remains as a supported **opt-in** option.                              |
```

Might recommend some slight wording adjustment. First line is not something you changed, but I've heard comments on the "by default" being confusing. Second line I'd recommend explicitly putting something like that second sentence (noting that sidecar is still supported and an option).

src/test/app-egress-sidecar-package.yaml
```diff
@@ -15,6 +15,8 @@ metadata:
   namespace: egress-gw-1
 spec:
   network:
+    serviceMesh:
+      mode: sidecar
```
mjnagel, line 19: While you're in here...might be good to rename these files (app-egress-package and app-egress) to explicitly put `-sidecar-` in the name so that it's easier to spot in the future.

src/test/app-egress-package.yaml
mjnagel, line 43: I think this package may also need sidecar mode explicitly now?

mjnagel:
> Should we be changing the operator to also be defaulting to Ambient mode? I might have missed a conversation somewhere but my understanding is that anywhere in the operator code that mentions `Mode.Sidecar` or `serviceMesh.mode` probably also needs to be evaluated. For example the [package validator has a check for istio mode](https://github.com/defenseunicorns/uds-core/blob/825c1ed6ea58df6432080174408528915ea0059c/src/pepr/operator/crd/validators/package-validator.ts#L25).

Yep - there's a few like this that can be found semi-easily with a search for `|| Mode.Sidecar`. Definitely want to make sure that we have the same default everywhere that the operator is reading the current mode.

mjnagel (commented):
Looks great overall - small docs and test package comments only + the other mention from @chance-coleman about some lingering sidecar defaults in the operator.

src/pepr/operator/controllers/istio/namespace.ts
briantwatson: Thank you, made these adjustments

docs/reference/uds-core/istio-ambient-transition.md
briantwatson: Adjusted following recommendation, thanks.

src/test/app-egress-sidecar-package.yaml
briantwatson: renamed!

src/test/app-egress-package.yaml
briantwatson: set as sidecar

mjnagel (approved):

chance-coleman (approved):
🥳

PR #2298, opened 2026-01-22
chore: update k8s-sidecar to 2.4.0, switch unicorn back to hardened

### LINKED ISSUE #2272: Switch Unicorn k8s-sidecar image back to hardened variant

https://github.com/defenseunicorns/uds-core/pull/2266 migrated to the "less hardened" image build for the k8s-sidecar due to some missing dependencies. When issues are resolved, switch back to the hardened image.

Ensure that the image starts up as expected and functions across all usage.

### Additional context

Note that this image is present for both Grafana and Loki and their versions/variants should be kept in sync to ensure a smaller zarf package.

mjnagel:
## Description

Upstream (RF) noted that the image is now fixed so this switches back to the hardened variant.

Fixes https://github.com/defenseunicorns/uds-core/issues/2272


mjnagel:
This appears to have the same issue still - putting on hold for now.

joelmccoy (approved):

PR #2303, opened 2026-01-26
chore: add multiarch checks to release pipeline

chance-coleman:
## Description
Add multi arch script check to the release process for the release please jobs.

## Type of change

## Checklist before merging

### DIFF
--- a/.github/actions/check-multiarch/action.yaml
+++ b/.github/actions/check-multiarch/action.yaml
@@ -21,8 +21,8 @@ runs:
       shell: bash
       run: |
         CRANE_VERSION="${{ inputs.craneVersion }}"
-        OS="linux"
-        ARCH="amd64"
+        OS="Linux"
+        ARCH="x86_64"
         curl -fsSL "https://github.com/google/go-containerregistry/releases/download/${CRANE_VERSION}/go-containerregistry_${OS}_${ARCH}.tar.gz" | tar -xz crane
         sudo mv crane /usr/local/bin/crane
         crane version
--- a/.github/actions/check-multiarch/action.yaml
+++ b/.github/actions/check-multiarch/action.yaml
@@ -13,7 +13,7 @@ inputs:
     description: "Version of crane (go-containerregistry) to install"
     required: false
     # renovate: datasource=github-tags depName=google/go-containerregistry versioning=semver
-    default: "v0.20.6"
+    default: "v0.20.7"
 
 runs:
   using: "composite"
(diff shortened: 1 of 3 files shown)


.github/actions/check-multiarch/action.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+inputs:
+  srcDir:
+    description: "Directory to scan for zarf.yaml files"
+    required: false
+    default: "./src"
+  craneVersion:
+    description: "Version of crane (go-containerregistry) to install"
+    required: false
+    default: "v0.20.6"
```
mjnagel: Probably want renovate tracking on this.

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -115,6 +117,15 @@ jobs:
...
           filters: .github/filters.yaml
 
+      - name: Check paths - multiarch scan inputs
+        id: path-filter-multiarch
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3
+        with:
+          filters: |
+            multiarch:
+              - 'src/**/zarf.yaml'
+              - 'scripts/check-multiarch.sh'
```
mjnagel: Running on PRs that change zarf.yamls will mean a missing multi-arch manifest for _any_ layer will fail an unrelated PR (example: bumping a grafana version, if istio is missing an arm manifest it will still fail). I think this is probably okay though, especially if we don't make it a required check (would bring more visibility - could even seen us just always running it to get clear feedback if something is missing).

No real ask for any change here, but would be curious on other thoughts of when we should run this... can see a few options:
- Always (but not as a required check)
- Only on PRs changing `zarf.yaml` (maybe as a required check, or not?)
- On release-please PRs (as a required check)

Maybe just depends what we are targetting/have seen - I don't think that missing manifests _always_ come with a version bump, although that is the case sometimes.

chance-coleman: went with the always run check but no requirements to pass or change specific files.

.github/actions/check-multiarch/action.yaml
```diff
@@ -0,0 +1,33 @@
...
+inputs:
+  srcDir:
+    description: "Directory to scan for zarf.yaml files"
+    required: false
+    default: "./src"
+  craneVersion:
+    description: "Version of crane (go-containerregistry) to install"
+    required: false
+    # renovate: datasource=github-tags depName=google/go-containerregistry versioning=semver
+    default: "v0.20.6"
```
mjnagel: nit: 0.20.7 is the latest, we could avoid a support-deps PR 😆

mjnagel (dismissed):

mjnagel (approved):

PR #2307, opened 2026-01-26
feat: scope uds core operator pepr rbac

### LINKED ISSUE #2278: Scope UDS Operator RBAC
Currently the UDS Operator uses the default [RBAC Mode](https://docs.pepr.dev/user-guide/rbac/) of Pepr, `admin`. Admin mode provides cluster admin permissions which is overly permissive for what we need.

We should scope our access for the UDS Operator (and policy/admission pods) to only the required access. Note that this WILL require us to do more than just flip to `scoped` since the `scoped` method will only take into account Watches and Reconciles. We will also need to "audit" the other resources and actions we need for other `K8s` calls (ex: creating/deleting virtualservices). This should be clear through testing and the code where permissions are needed.

There may be some timing/"how" considerations since we currently use a dynamically generated chart for Pepr. We will likely want to "post-process" and replace the upstream roles/bindings before creating the zarf package.

joelmccoy:
## Description

Adds a custom k8s RBAC policy to the `ClusterRole` used by the UDS Operator.

## Related Issue

Fixes #2278 

## Type of change

## Steps to Validate
- You can also delete some of these permissions to ensure that a deployment of core fails (i tested this locally and confirmed RBAC issues in pepr logs)

## Checklist before merging

### DIFF
--- a/src/pepr/values.yaml
+++ b/src/pepr/values.yaml
@@ -3,23 +3,53 @@
 
 # Setup RBAC for Pepr
 rbac:
-  # Core API resources
+  # Configmaps and Secrets
   - apiGroups:
       - ''
     resources:
       - configmaps
-      - endpoints
-      - events
-      - namespaces
       - secrets
-      - services
     verbs:
       - create
       - delete
       - get
       - list
       - patch
-      - update
+      - watch
+
+  # Endpoints and events
+  - apiGroups:
+      - ''
+    resources:
+      - endpoints
+      - events
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+
+  # Namespaces
+  - apiGroups:
+      - ''
+    resources:
+      - namespaces
+    verbs:
+      - create
+      - get
+      - list
+      - patch
+
+  # Services
+  - apiGroups:
+      - ''
+    resources:
+      - services
+    verbs:
+      - get
+      - list
+      - patch
       - watch
 
   # Pods
@@ -32,7 +62,6 @@ rbac:
       - get
       - list
       - patch
-      - watch
 
   # Pod eviction subresource
   - apiGroups:
@@ -42,7 +71,7 @@ rbac:
     verbs:
       - create
 
-  # Nodes - read only
+  # Nodes
   - apiGroups:
       - ''
     resources:
@@ -72,7 +101,6 @@ rbac:
       - get
       - list
       - patch
-      - update
       - watch
 
   # Apps API resources
@@ -87,8 +115,6 @@ rbac:
       - get
       - list
       - patch
-      - update
-      - watch
 
   # Networking API resources
   - apiGroups:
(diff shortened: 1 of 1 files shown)


src/pepr/values.yaml
```diff
@@ -1,6 +1,202 @@
...
+      - secrets
+      - services
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
mjnagel, line 53: The only 2 permissions I couldn't trace back to any code here were `delete` permissions for services and namespaces. I'm okay with keeping them for consistency, but don't think we need them?

src/pepr/values.yaml
```diff
@@ -1,6 +1,202 @@
...
+    resources:
+      - networkpolicies
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
mjnagel: There's a number of resources (like netpols) that I don't _think_ we need watch on based on our k8s/pepr watch/reconciles. But I'm okay adding if we think it's helpful for future proofing (or if I'm just missing somewhere it's called).

src/pepr/values.yaml
```diff
@@ -1,6 +1,202 @@
...
+      - clusterconfig
+      - clusterconfig/status
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
mjnagel, line 198: I don't think we ever create or delete any of these resources, so might be able to able to remove those permissions?

mjnagel (dismissed):
LGTM overall and our test suite should well exercise code paths to ensure this covers everything. A few comments on some permissions that seem extra, but not overly concerned.

src/pepr/values.yaml
joelmccoy: good call. removed these

src/pepr/values.yaml
joelmccoy: yeah. good call. i generated the `scoped` rbac from pepr to compare and removed watch where necessary in all these resources.

src/pepr/values.yaml
joelmccoy: broke these out better. I don't think we create services either.  so put that in it's own category (same with `namespaces`).  This should be more specifically scoped now.

joelmccoy (commented):
scoped a bit more.  Also removed `update` everywhere as all our `update` calls are `patch` calls.

src/pepr/values.yaml
mjnagel: Great way to do it - I did a quick sanity check on the [KFC watch side](https://github.com/search?q=repo%3Adefenseunicorns%2Fuds-core%20registerWatchEventHandlers&type=code) as well, we only watch our own resources there so shouldn't be anything missing.

mjnagel (approved):

slaskawi (approved):

PR #2308, opened 2026-01-27
chore: crd versioning adr

### LINKED ISSUE #2179: Determine and document CRD versioning/upgrade stance

slaskawi:
## Description

This Pull Request introduces an ADR on CRD versioning. 

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/2179

## Type of change

## Steps to Validate

N/A

## Checklist before merging


adrs/0008-crd-versioning.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Decision
+
+UDS Core will follow the **Kubernetes-native CRD versioning strategy** as defined in the [Kubernetes API deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/) and [CRD versioning documentation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/). This approach provides the strongest backwards compatibility guarantees.
+
+The Kubernetes-native CRD versioning principles have been are outlined in the [Versioning Levels and Guarantees](#Kubernetes-native-CRD-versioning-principles) section below.
+
+UDS Core 1.0 will be released with all the CRDs promoted to v1beta1. This provides guaranteed backwards compatibility for production workloads and ensures that users can confidently build on UDS Core APIs knowing the stability contract.
```
slaskawi: I'm open for discussion about this but I think bumping the CRDs up to v1beta1 might be good from the optics perspective.

adrs/0008-crd-versioning.md
```diff
@@ -0,0 +1,82 @@
...
+The YouTube talk "[The Missing Talk About API Versioning & Evolution in Your Developer Platform](https://www.youtube.com/watch?v=pHRQpqCEvU8)" presents several alternative patterns for CRD versioning that prioritize faster iteration:
+
+* Using different `apiVersion` fields (e.g. `group.v1/v1`, `group.v2/v1`) for new CRD versions
+* Using embedded versions in the CRD `spec` (e.g. `spec.v1`, `spec.v2`) for the new API versions
+
+These approaches have been designed to enable rapid iteration and provide minimal backwards compatibility guarantees. Therefore, they are not suitable for UDS Core.
+
+The last approach considered is using `x-kubernetes-preserve-unknown-fields` annotation. This approach is strongly discouraged by the Kubernetes community and is not recommended for production use.
+
+## Kubernetes-native CRD versioning principles
```
slaskawi, line 46: This entire paragraph is purely informational. It could be removed but at the same time, it might be helpful to have it somewhere at hand. 

If you feel it's redundant, I'm open to removing it.

adrs/0008-crd-versioning.md
```diff
@@ -0,0 +1,81 @@
...
+- Fields may be added, removed, or renamed without notice
+- No schema compatibility guarantees between alpha versions
+- May be removed in any UDS Core release without prior deprecation
+- Suitable for experimental features and rapid iteration
+- Users should expect breaking changes
+
+**Beta (v1beta1, v1beta2, etc.)**
+- Schema is well-tested and considered stable
+- Fields may only be removed or have breaking changes with proper deprecation
+- Once marked as deprecated, beta versions must remain supported for a minimum of 9 months or 3 UDS Core minor releases (whichever is longer) before removal, aligning with the Kubernetes API deprecation policy
```
joelmccoy: just want to call out that 9 months will almost always be the longer period.  I wonder if this time frame still makes sense for UDS Core which releases every two weeks vs. k8s which does quarterly?  I think it could be fine for beta APIs, but just wanted to call this out in case someone has a strong opinion.

adrs/0008-crd-versioning.md
```diff
@@ -0,0 +1,81 @@
...
+### Positive
+
+- Strong backwards compatibility for production workloads
+- Predictable upgrade paths aligned with Kubernetes ecosystem best practices
+- Users can confidently build on UDS Core APIs knowing the stability contract
+- Familiarity for users already accustomed to Kubernetes API versioning
+
+### Negative
+
+- CRD schema evolution will require implementing a [Conversion Webhook](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#configure-customresourcedefinition-to-use-conversion-webhooks)
```
joelmccoy: Right now we have some custom [migrate](https://github.com/defenseunicorns/uds-core/blob/1d1a826ef54b68d6ee7697eb153f7a3707a44a93/src/pepr/operator/crd/migrate.ts#L15) code in Pepr for doing "_migrations_". But don't think it mutates the actual storage of the spec. Does this mean we shouldn't have this custom migrate code unless we introduce a way to save back to storage and instead we should only rely on the conversion webhooks?

adrs/0008-crd-versioning.md
joelmccoy: hmmmm maybe for packages and exemptions it makes sense as these are the more mature CRDs, but would be hesitant to bump the ClusterConfig as this CRD is relatively new and has had recent rapad changes

joelmccoy (commented):
Generally LGTM!  No qualms with this direction.  A couple small comments/open questions

adrs/0008-crd-versioning.md
mjnagel: I think this is fair from an optics perspective. What would the plan be for `v1alpha1` versions? Do they remain "supported" in 1.0, or would we try to phase them out before then?

adrs/0008-crd-versioning.md
mjnagel: I think this is something that really _could_ live with pepr, but it could be a tricky process since the CRD must have the webhook definition in it (including caBundle) and Pepr regenerates CAs on each build I believe?

Our current migrate I think is currently just functioning to simplify our operator logic (we do an in-memory migration so that the reconciliation doesn't have to care about all the paths). Conversion webhooks would solve this in a different way, by ensuring that the object hitting our watch is "already migrated". I think we could do this with a mutation too (maybe?), but then the end user wouldn't see their original kind/version. 

I'd also like to think through what other options we might have here...conversion webhooks seem uncommon across other projects (prometheus, istio, etc). I think in their case they just keep all versions identical which might not work for us. But I think worth capturing in alternatives?

adrs/0008-crd-versioning.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Beta (v1beta1, v1beta2, etc.)**
+- Schema is well-tested and considered stable
+- Fields may only be removed or have breaking changes with proper deprecation
+- Once marked as deprecated, beta versions must remain supported for a minimum of 9 months or 3 UDS Core minor releases (whichever is longer) before removal, aligning with the Kubernetes API deprecation policy
+- Breaking changes require incrementing the beta version (v1beta1 → v1beta2)
+- Beta versions must support conversion to/from other served versions
+
+**GA/Stable (v1, v2, etc.)**
+- Considered production-ready with strong stability guarantees
```
mjnagel: just a nit on this, I think we largely would already say our alpha APIs are production ready, I see this more as stable/minimal changes?

adrs/0008-crd-versioning.md
mjnagel: I'm _okay_ with a longer time period here, but maybe would lean towards a bit shorter than 9 months? 3-6 months feels like a better spot imo.

adrs/0008-crd-versioning.md
slaskawi: At the moment, we have the following versions of our CRDs:

```
clusterconfig uds.dev/v1alpha1
exemptions uds.dev/v1alpha1
packages uds.dev/v1alpha1
```

I would assume, we'd just promote them to `v1beta1` and use this opportunity as a fire-drill to see how hard it is to move from one version to another.

adrs/0008-crd-versioning.md
slaskawi: I'll switch this to 3 (the lower bound that you mentioned).

adrs/0008-crd-versioning.md
slaskawi: Change the wording to "fully-matured". I think this better describes the intention here.

adrs/0008-crd-versioning.md
slaskawi: We have a few options here. 

The most idiomatic one is using the Converting Webhooks that results in simplified UDS Operator code (it would need to care only about one version at a time).

We can also keep suing the migration logic from the UDS Operator but then, as the number of supported versions grows (`v1alpha1`, `v1alpha2`, `v1beta1` etc), this migration code inside the UDS Operator grows too. In addition to that, if we supported multiple versions of the CRD, our Operator our need to react on each one of them (which might be considered as a disadvantage to the Converting Webhook approach).

I don't think we need to make the decision now. I'll redact the text to highlight that some sort of migration logic will be necessary.

joelmccoy (dismissed):

adrs/0008-crd-versioning.md
mjnagel: So I think based on this ADR, we'd need to support both versions temporarily before 1.0 right?

> 4. **Storage version management**: The preferred/storage version cannot advance to a new version until after a release has been made that supports both the new version and the previous version. This ensures users can upgrade and rollback safely.

Just want to call out that effect of adopting this ADR, we will need to test/implement a conversion strategy pre-1.0 as well.

adrs/0008-crd-versioning.md
mjnagel: I don't think in-memory migrations actually solves this problem, and we will _need_ something like conversion webhooks if I'm not mistaken? Ultimately you can only have one stored version of a CRD, and our migration would just be in-memory for the operator. It may be worth some testing here, but I think basically we have 3 options:
1. Implement a conversion webhook (maybe with some pepr assistance): This is the cleanest path if we want/need to support multiple versions of the CRDs.
2. Keep schemas identical across served versions: This probably doesn't give us the flexibility we want?
3. Never provide overlap of CRD versions (hard cutovers each time): This is the easiest on our side, but hardest for end user (they have no "migration time" when switching between versions). I do think it's worth calling this out as an option though, if we are managing deprecating API fields in our older versions well.

adrs/0008-crd-versioning.md
slaskawi: That's correct. We'd need to go through a normal CRD upgrade cycle pre-1.0. 

Let me tag @jonnyvolpe here as well as this decision has a lot of impact on the Product.

adrs/0008-crd-versioning.md
slaskawi: You have a good point - only Converting Webhooks enable us to travel between the Storage Versions and ultimately cut old CRDs off. As you called out - in-memory migrations don't solve this problem and we'd need some 3rd party tool to perform the migration. 

I've rephrased this piece in the document, please check if it's OK now.

adrs/0008-crd-versioning.md
slaskawi: Resolving to clean up the doc.

adrs/0008-crd-versioning.md
slaskawi: Adding an information that we'll keep using `v1alpha1` as discussed internally

chance-coleman (dismissed):

adrs/0008-crd-versioning.md
mjnagel: I think we can/should drop this paragraph?

> UDS Core 1.0 will be released with all the CRDs promoted to v1beta1. This provides guaranteed backwards compatibility for production workloads and ensures that users can confidently build on UDS Core APIs knowing the stability contract.

It seems to conflict with the line before (UDS Core will keep using the `v1alpha1` CRD versions).

adrs/0008-crd-versioning.md
```diff
@@ -0,0 +1,86 @@
...
+
+UDS Core defines several Custom Resource Definitions (CRDs) that are currently all versioned as v1alpha1 (Package, Exemption, ClusterConfig). As UDS Core approaches its 1.0 release, there's need to establish a clear policy for CRD versioning, lifecycle management, and backwards compatibility guarantees. This is critical for providing stability to our users and ensuring predictable upgrade paths.
+
+## Decision
+
+UDS Core will follow the **Kubernetes-native CRD versioning strategy** as defined in the [Kubernetes API deprecation policy](https://kubernetes.io/docs/reference/using-api/deprecation-policy/) and [CRD versioning documentation](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/). This approach provides the strongest backwards compatibility guarantees.
+
+The Kubernetes-native CRD versioning principles are outlined in the [Kubernetes-native CRD versioning principles](#Kubernetes-native-CRD-versioning-principles) section below.
+
+Because of the time constraints, the UDS Core will keep using the `v1alpha1` CRD versions. Upgrading to `v1beta1` or `v1` will be considered for the future releases.
```
mjnagel:
```suggestion
For the 1.0 release, UDS Core will continue to include the `v1alpha1` CRD versions to reduce user impact. Upgrading to `v1beta1` or `v1` will be considered for the future releases.
```

Slight rewording to specifically call out 1.0 and focus on user impact rather than time constraints (both relevant, but user impact of removing v1alpha1 is more compelling imo).

mjnagel (commented):
LGTM overall - last comments on the 1.0 CRD versioning section.

adrs/0008-crd-versioning.md
slaskawi: Ohhh my! I forgot to delete this paragraph!

mjnagel (approved):

PR #2313, opened 2026-01-27
fix: broken dev setup because of missing trust bundle initialization

chance-coleman:
## Description
In the dev-setup task, istio is deployed before pepr, meaning the magic that pepr provides for istio doesnt happen causing a failed dev-setup because istiod never completely deploys. This bootstraps a blank trust-bundle configmap for the dev-setup task to successfully complete. Pepr will manage the configmap later if it needs updating.

## Type of change

## Checklist before merging

### DIFF
--- a/tasks.yaml
+++ b/tasks.yaml
@@ -47,10 +47,6 @@ tasks:
           uds zarf tools kubectl apply -f src/pepr/uds-cluster-crds/templates/clusterconfig.uds.dev.yaml
           uds zarf tools kubectl apply -f hack/dev-manifests/
 
-      - description: "Create uds-trust-bundle ConfigMap for Istio"
-        cmd: |
-          uds zarf tools kubectl create configmap uds-trust-bundle -n istio-system --from-literal=extra.pem="" --dry-run=client -o yaml | uds zarf tools kubectl apply -f -
-
       # Note: the `registry-url` flag used here requires uds 0.19.2+
       - description: "Deploy the Istio source package with Zarf Dev"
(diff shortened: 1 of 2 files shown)


tasks.yaml
```diff
@@ -47,6 +47,10 @@ tasks:
           uds zarf tools kubectl apply -f src/pepr/uds-cluster-crds/templates/clusterconfig.uds.dev.yaml
           uds zarf tools kubectl apply -f hack/dev-manifests/
 
+      - description: "Create uds-trust-bundle ConfigMap for Istio"
```
slaskawi: Do we need to call it somewhere during the deployment process?

tasks.yaml
```diff
@@ -47,6 +47,10 @@ tasks:
...
+        cmd: |
+          uds zarf tools kubectl create configmap uds-trust-bundle -n istio-system --from-literal=extra.pem="" --dry-run=client -o yaml | uds zarf tools kubectl apply -f -
```
joelmccoy: very nit: but you could also just include the yaml file in the `hack/dev-manifests` folder [here](https://github.com/defenseunicorns/uds-core/tree/main/hack/dev-manifests).  All these files are applied automatically in the step before this and are used to get around a similar problem.

slaskawi (dismissed):

tasks.yaml
chance-coleman: yea, good call! updated to use the manifest and removed the new cmd

joelmccoy (approved):

mjnagel (approved):

slaskawi (approved):

PR #2316, opened 2026-01-28
feat: cluster-less crd pipeline

### LINKED ISSUE #2292: feat: implement cluster-less CRD pipeline and move registration to Helm
UDS Core currently relies on a running cluster to sync CRD types and uses the Pepr operator to register those CRDs at runtime. This causes developer friction and requires the operator to have broad Cluster-wide RBAC permissions (`create/patch/update` on `customresourcedefinitions`).

- **Given** the CRD source definitions in [src/pepr/operator/crd/sources](cci:9://file:///home/chance/Unicorn/uds-core/src/pepr/operator/crd/sources:0:0-0:0)
- **When** the `uds run gen-crds` task is called
- **Then**:
  1. A `gen-manifests` script (using the new KFC logic or internal util) exports YAML to [src/pepr/operator/crd/manifests/](cci:9://file:///home/chance/Unicorn/uds-core/src/pepr/operator/crd/manifests:0:0-0:0).
  2. `kfc` generates types directly from these local manifests.
  3. Programmatic registration is removed from [register.ts](cci:7://file:///home/chance/Unicorn/uds-core/src/pepr/operator/crd/register.ts:0:0-0:0).
  4. Operator `ClusterRole` permissions for CRDs are removed.

**Acceptance Criteria:**
- `tasks.yaml` updated to remove the requirement for `KUBECONFIG`.
- CI check ensures [manifests/](cci:9://file:///home/chance/Unicorn/uds-core/src/pepr/operator/crd/manifests:0:0-0:0) are always in sync with [sources/](cci:9://file:///home/chance/Unicorn/uds-core/src/pepr/operator/crd/sources:0:0-0:0).
- [register.ts](cci:7://file:///home/chance/Unicorn/uds-core/src/pepr/operator/crd/register.ts:0:0-0:0) refactored to remove runtime `Apply()` calls for CRDs.

Keeping registration in Pepr while using manifests (rejected to achieve RBAC reduction).

### Additional context
This follows the "Cluster-less" transition documented in ADR 0007.

briantwatson:
## Description

- Adds new script to generate CRD yaml manifests and add to helm chart
- Removes register.ts
- Updates gen-crd to use this and local yaml files instead of requiring cluster
- Adds check for templates path to existing crd check

## Related Issue

Fixes #2292

## Type of change

## Steps to Validate

## Checklist before merging


src/pepr/values.yaml
```diff
@@ -140,6 +140,16 @@ rbac:
...
+  - apiGroups:
+      - apiextensions.k8s.io
+    resources:
+      - customresourcedefinitions
+    verbs:
+      - create
+      - get
+      - list
+      - patch
+
```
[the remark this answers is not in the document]
briantwatson: Confirmed they can't be removed as Pepr still needs permisssions for peprstore CRD

.github/actions/autogenerated-check/action.yaml
chance-coleman: I think this portion of our CI job can be cleaned up since we don't need the cluster for this to work anymore.

tasks/lint.yaml
```diff
@@ -57,7 +57,8 @@ tasks:
           failed=0
           for chart in $(find . -name Chart.yaml -type f -exec dirname {} \;); do
             if [ -d "$chart/tests" ]; then
-              if ! helm unittest --color=true "$chart"; then
+              echo "Testing $chart"
+              if ! helm unittest "$chart"; then
```
chance-coleman: i might be missing it, did we remove the `--color=true` for a reason?

src/pepr/uds-cluster-crds-v2/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2026 Defense Unicorns
```
chance-coleman: I might be missing something here, so maybe this is just a question. Should the legacy `src/pepr/uds-cluster-crds/Chart.yaml` be kept around? This would also affect the `packages/base/zarf.yaml` and `packages/standard/zarf.yaml` i think. Both of those packages have the legacy uds-crds included. We renamed the clusterconfig (the existing legacy crd), so theres no longer any templates for that crd.

chance-coleman (commented):
Looking good so far, couple comments and a question.

tasks/lint.yaml
mjnagel: Can we conditionally add `--color=true` if `helm version --short` returns v3? 😏

briantwatson: Opted to remove due to a bug with helm 4.  https://github.com/helm-unittest/helm-unittest/issues/777

src/pepr/zarf.yaml
```diff
@@ -57,6 +57,23 @@ components:
...
+    actions:
+      onDeploy:
+        after:
+          - description: Validate ClusterConfig CRD
+            maxTotalSeconds: 60
+            wait:
+              cluster:
+                kind: CustomResourceDefinition
+                name: clusterconfig.uds.dev
+
```
mjnagel, line 78: Discussed synchronously, but we should be able use the existing chart and just bump the version (0.2.0). That will require us using an `onDeploy.before` action to "adopt" the existing resources with the proper helm metadata, similar to [what we do here](https://github.com/defenseunicorns/uds-core/blob/main/src/velero/common/zarf.yaml#L53-L59).

On the `crds` folder vs `templates`, as discussed we want to ensure these are in `templates` so that they get upgraded over time.

src/pepr/uds-cluster-crds-v2/Chart.yaml
joelmccoy, line 1: nit: maybe related to Chance's comment, why did we create a new chart with v2 here?  IMO the `v2` in the name here is a bit confusing.  What is this v2 of?  And is there any reason why we can't just use the original chart?

src/pepr/zarf.yaml
mjnagel: Won't comment on other spots, but this should reduce the overall changes required since you won't need other `zarf.yaml` or chart changes I think.

src/pepr/uds-cluster-crds/templates/clusterconfig.uds.dev.yaml
joelmccoy, line 1: Just noting the delta on this file:
```
listKind: ClusterConfigList
```
is removed
and
```
  conversion:
    strategy: None
```
is removed
I think both of these fields are unimportant, but wanted to call out that there is a small delta. We could add these to the spec if we wanted to be explicit and consistent.

packages/base/zarf.yaml
```diff
@@ -23,6 +23,11 @@ components:
     import:
       path: ../../src/pepr
 
+  - name: uds-crds-v2
```
joelmccoy: Is there any reason why we can't just use the original `uds-crds` component?  `v2` on a directory name for this feels unnecessary IMO and kind of confusing.

src/pepr/operator/crd/register.ts
joelmccoy, line 1: we have a reference in our docs to this file: https://github.com/defenseunicorns/uds-core/blob/bda5384adce387707b43a0af3bc294e9f0b87894/docs/reference/configuration/uds-operator/overview.md?plain=1#L53

It doesn't look like this is getting updated.... And I would advocate for simplifying this section altogether?  We should have to update this doc everytime we add a file.  Maybe we can remove this section or at least only provide docs for the top level directory structure (controllers, crd, etc...).

src/pepr/uds-cluster-crds/templates/packages.uds.dev.yaml
joelmccoy, line 1: Just noting the delta on this file:

```
listKind: PackageList
``` 
is removed

and
```
  conversion:
    strategy: None
```
is removed

I think both of these fields are unimportant, but wanted to call out that there is a small delta. We could add these to the spec if we wanted to be explicit and consistent.

src/pepr/uds-cluster-crds/templates/exemptions.uds.dev.yaml
joelmccoy, line 1: Just noting the delta on this file:
```
listKind: ExemptionList
```
is removed
and
```
  conversion:
    strategy: None
```
is removed
I think both of these fields are unimportant, but wanted to call out that there is a small delta. We could add these to the spec if we wanted to be explicit and consistent.

src/pepr/values.yaml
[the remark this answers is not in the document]
joelmccoy: nit: might be worth a small comment here to clarify that. I was under the assumption that we could remove these, but the peprstore dep makes sense.  Might be useful to note this to future onlookers

src/pepr/zarf.yaml
```diff
@@ -57,6 +57,23 @@ components:
...
+    required: true
+    charts:
+      - name: uds-cluster-crds-v2
+        namespace: pepr-system
+        version: 0.1.0
+        localPath: uds-cluster-crds-v2
+    actions:
+      onDeploy:
+        after:
+          - description: Validate ClusterConfig CRD
```
joelmccoy: is there a reason you only validate the Clusterconfig here and not packages/exemptions as well?  Also I wonder if this is even needed, I think CRDs have a [kstatus](https://github.com/kubernetes-sigs/cli-utils/blob/5895ad6c17dd06b99c1e5c8af17e2b12a3504fad/pkg/kstatus/status/core.go#L611-L637) on them, so Zarf on deploy should automatically wait for CRDs to be `established`. See [docs](https://github.com/kubernetes-sigs/cli-utils/blob/5895ad6c17dd06b99c1e5c8af17e2b12a3504fad/pkg/kstatus/status/core.go#L611-L637)

So i think you can remove this check altogether?

tasks.yaml
```diff
@@ -31,20 +31,15 @@ tasks:
...
         cmd: |
           uds zarf tools kubectl create ns uds-policy-exemptions
           uds zarf tools kubectl create ns istio-system
           uds zarf tools kubectl create ns pepr-system
 
-      - description: "Deploy the UDS Cluster Config CRD and apply all hack dev manifests"
+      - description: "Deploy the UDS CRDs and apply all hack dev manifests"
         cmd: |
-          uds zarf tools kubectl apply -f src/pepr/uds-cluster-crds/templates/clusterconfig.uds.dev.yaml
+          uds zarf tools kubectl apply -f src/pepr/uds-cluster-crds-v2/crds/
```
joelmccoy: hmmm. I wonder if we even need this anymore since we deploy in a helm chart before we deploy pepr? I think we applied the clusterconfig because pepr needed clusterconfig CRD to read in it's config.  But now that we apply in chart beforehand I think we can remove this.

joelmccoy (commented):
Generally looking good! Left some comments for some consideration.

tasks.yaml
joelmccoy: actually I think we still need this as we don't deploy the CRDs/operator-config-chart in the dev-setup task.  closing this out.

src/pepr/values.yaml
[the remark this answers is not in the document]
briantwatson: Adding this to feedback commit

.github/actions/autogenerated-check/action.yaml
briantwatson: Totally missed the key piece of removal 😅 -- thanks!  Removing the cluster creation

src/pepr/uds-cluster-crds-v2/Chart.yaml
briantwatson: Discussed synchronously, but responding here for history. Given pepr was applying Exemption and Package CRs prior to this it created a scenario where we had two CRDs that weren't tracked by helm and helm would error on upgrade if I tried to put them in a templates folder of existing chart.  In my initial pass here to get everything cleanly installed by helm I moved them to a new chart in `/crds` and then removed the templates of the old chart so the old release was clean. 

In discussion, we landed on maintaining the original chart, using zarf action to annotate/label the pre-existing CRDs.  I'll make those changes so this new chart is no longer relevant.

src/pepr/uds-cluster-crds-v2/Chart.yaml
briantwatson: I'll link to explanation in response to Chance.  Appreciate the feedback on this one from all
https://github.com/defenseunicorns/uds-core/pull/2316#discussion_r2747223240

src/pepr/uds-cluster-crds/templates/clusterconfig.uds.dev.yaml
briantwatson: I did notice that as well, and from what I can tell and understand is those are being added by the Kubernetes API as defaults.  So the existing Package and Exemption CRDs that Pepr are applying do not have them on apply, and the original usage of Cluster to create ClusterConfig CRD yaml did not either.  Given that context, would you still prefer we add them here?

joelmccoy: I think we could at least add the listkind to script here: https://github.com/defenseunicorns/uds-core/blob/feat/clusterless-crd-pipeline/src/pepr/scripts/gen-crds.ts#L34 just to be explicit?

but the conversion one can be left off IMO

packages/base/zarf.yaml
briantwatson: This was a sequencing needed given my initial implementation of adding Package and Exemption and moving ClusterConfig to `/crds`.  The ClusterConfig CRD would be removed from cluster as the processing of `/templates` resources occurred after `/crds` and helm would see the ClusterConfig removed templates.

  New implementation discussed I'll implement will move back to original chart under `/templates` using a zarf action to annotate the pre-existing CRDs created by Pepr so helm can adopt them without erroring.

src/pepr/uds-cluster-crds/templates/packages.uds.dev.yaml
briantwatson: Following [comment above](https://github.com/defenseunicorns/uds-core/pull/2316#discussion_r2747240531) I believe ClusterConfig is actually the only delta of what we _apply_.  As Pepr is applying what's here via [register.ts](https://github.com/defenseunicorns/uds-core/blob/1a2515f0d201e13dc70180af503183048d84af31/src/pepr/operator/crd/register.ts#L66-L69) and we don't have `listKind` or `conversion` there?

src/pepr/uds-cluster-crds/templates/exemptions.uds.dev.yaml
briantwatson: Following [comment above](https://github.com/defenseunicorns/uds-core/pull/2316#discussion_r2747240531) I believe ClusterConfig is actually the only delta of what we _apply_.  As Pepr is applying what's here via [register.ts](https://github.com/defenseunicorns/uds-core/blob/1a2515f0d201e13dc70180af503183048d84af31/src/pepr/operator/crd/register.ts#L101-L104) and we don't have `listKind` or `conversion` there?

briantwatson: Will track decision on all three via thread above

src/pepr/uds-cluster-crds/templates/packages.uds.dev.yaml
briantwatson: Will track decision on all three via thread above

src/pepr/zarf.yaml
briantwatson: For context, I had hit an error that the ClusterConfig CRD was not present on the subsequent uds-operator-config component.  I _thought_, although not 100%, this was due to having moved to the `/crds` folder and from what I read helm doesn't wait to confirm install/established when installed from there unlike resources under `/templates`.  I'm not familiar with zarf implementation if that check occurs in all scenarios or just the templated resources--I'd be curious to know.  Certainly possible I had hit a different error misattributed.

That said, as noted above, moving to different approach to original helm chart in templates, so will remove this check along with other components for other helm chart.

joelmccoy: That would be interesting if the zarf package deploy completes before the CRD is ready. It shouldn't based on those docs.  It would be interesting to see if we can reproduce with that check removed.

src/pepr/scripts/gen-crds.ts
joelmccoy, line 1: Since KFC already [released](https://github.com/defenseunicorns/kubernetes-fluent-client/releases/tag/v3.11.0) the new conversion command `crd-manfiest`, can we just get rid of this and use the new command?

src/pepr/zarf.yaml
```diff
@@ -56,6 +56,25 @@ components:
         namespace: pepr-system
         version: 0.1.0
```
mjnagel: I think this should be 0.2.0 now? I guess Zarf doesn't care though since it's a local chart 🤔

tasks/lint.yaml
mjnagel:
```suggestion
              if ! helm unittest "$chart" --color; then
```
I believe moving color to after the chart seems to work for both helm 3 and 4?

mjnagel (commented):
Looking pretty good at this point - some minor things + I think we can switch to the new KFC command now 🎉

src/pepr/zarf.yaml
briantwatson: Revisited this out of curiosity and set up everything under /crds to see if a timing issue was present and the following step proceeded before CRD was established. I couldn't reproduce so I must have misattributed during other changes I made. Again, overcome by events, but thought I'd share the additional testing finding.  Thanks for calling this out.

src/pepr/tasks.yaml
```diff
@@ -78,33 +75,19 @@ tasks:
...
-          # Define CRD list as a space-separated string
           CRD_LIST="exemptions.uds.dev packages.uds.dev clusterconfig.uds.dev"
-          # For each CRD retrieve the CRD and create a Go struct file to then create a markdown file
           for CRD in $CRD_LIST; do
-              # Retrieve CRD Json from Cluster
-              kubectl get crd $CRD -o json | jq '.spec.versions[] | select(.name == "v1alpha1").schema.openAPIV3Schema' >> $CRD-crd.json
-              # Extract the version (e.g., v1alpha1)
-              version=$(kubectl get crd $CRD -o json | jq -r '.status.storedVersions[0]')
+              # Convert YAML to JSON and extract openAPIV3Schema for v1alpha1
+              yq -o=json '.spec.versions[] | select(.name == "v1alpha1").schema.openAPIV3Schema' src/pepr/uds-cluster-crds/templates/$CRD.yaml > $CRD-crd.json
```
mjnagel: super unimportant nit: `uds zarf tools yq` would ensure this is always on someone's machine.

chance-coleman (dismissed):

joelmccoy (dismissed):
LGTM!

chance-coleman (approved):

mjnagel (approved):

Issue #2289, opened 2026-01-21
Add Keycloak pre-Stop hook


slaskawi:
## Summary

When performing Keycloak load testing, it turned out the Keycloak Kubernetes Service emits occasional HTTP 503 even if there is enough Pods to serve the traffic. 

[This](https://stackoverflow.com/a/55871111) thread highlights that this might be a manifestation of the Kubernetes Proxy design choices and adding a pre-Stop hook might solve it. 

After performing some testing, indeed we can see the improvement.

## Implementation details:

1. Add the pre-stop hook to the Keycloak StatefulSet:

```
154         lifecycle:
155           preStop:
156             exec:
157               command:
158               - /usr/bin/bash
159               - -c
160               - sleep 1
```

2. Ensure that it works by invoking:

* [Terminal 1] `siege -l siege.log https://sso.uds.dev`
* [Terminal 2] `kubectl rollout restart statefulset -n keycloak keycloak`

Aim at 100% availability

slaskawi:
An upstream fix for HTTP 503s have been integrated, see https://github.com/quarkusio/quarkus/issues/50625

PR #2334, opened 2026-02-03
feat: improve Keycloak availability

slaskawi:
## Description

This Pull Request introduces a set of changes that aim to improve the availability of the SSO service within the UDS Registry.

## Related Issue

Relates to https://github.com/defenseunicorns/uds-core/issues/2289

## Type of change

## Steps to Validate

- Backwards compatible changes verified manually
- Deploy in UDS Registry to test the rest

## Checklist before merging


src/keycloak/chart/templates/destination-rule.yaml
```diff
@@ -1,7 +1,7 @@
 # Copyright 2024 Defense Unicorns
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
-{{- if not .Values.devMode }}
+{{- if .Values.autoscaling.sessionStickiness.enabled }}
```
mjnagel:
```suggestion
{{- if and (not .Values.devMode) .Values.autoscaling.sessionStickiness.enabled }}
```

I think to keep parity with previous behavior we'd want to `and` these conditions.

src/keycloak/chart/values.schema.json
```diff
@@ -143,6 +151,14 @@
...
       "$ref": "#/definitions/autoscalingSchema"
     },
+    "waypoint": {
+      "type": "object",
+      "properties": {
+        "deployment": {
+          "type": "boolean"
+        }
+      }
+    },
```
mjnagel: This looks off (and there's a duplicate waypoint below this I think)?

src/keycloak/chart/templates/waypoint-config.yaml
```diff
@@ -45,7 +45,11 @@ data:
   {{- with .Values.waypoint.horizontalPodAutoscaler }}
   horizontalPodAutoscaler: |
     spec:
+      {{- if $.Values.autoscaling.enabled }}
+      minReplicas: {{ $.Values.autoscaling.minReplicas }}
+      {{- else }}
       {{ with .minReplicas }}minReplicas: {{ . }}{{- end }}
+      {{- end }}
```
mjnagel: This feels odd imo to tie the min replicas together when we have a separate value for the waypoint HPA. Maybe instead this is something to document? Edit: after typing this I realized we do document this [here](https://github.com/defenseunicorns/uds-core/blob/main/docs/reference/configuration/resource-configuration-and-ha.md#:~:text=You%20can%20also,on%20your%20needs%3A) but maybe could be more explicit about the benefit of scaling up the min replicas?

One of the benefits of the waypoint over the older sidecar model is that it is independently scalable up/down, separate from the application (Keycloak). I would hesitate to tie these together like this values change does.

joelmccoy: agreed with this.  would recommend this be a documented override to `horizontalPodAutoscaler.minReplicas` instead of keying off the autoscaling... Technically we have "autoscaling" by default on the waypoint.  We just need to document that setting min replicas to 2 for failover.

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -187,6 +187,14 @@ spec:
...
             - name: KC_SPI_X509CERT_LOOKUP__{{ .Values.x509LookupProvider | upper }}__SSL_CLIENT_CERT_CHAIN_PREFIX
               value: UNUSED
+            # Below 3 settings are a workaround for https://github.com/keycloak/keycloak/issues/43589
+            # After Keycloak 26.6.0 release, they should be removed
+            - name: QUARKUS_SHUTDOWN_DELAY_ENABLED
+              value: "true"
+            - name: QUARKUS_SHUTDOWN_DELAY
+              value: "5s"
+            - name: QUARKUS_SHUTDOWN_TIMEOUT
+              value: "5s"
```
joelmccoy: mainly out of curiosity, is there a reason why 5s was picked for these values? is that recommended somewhere upstream? I couldn't pinpoint any specific guidance in the upstream issue.

slaskawi: The approach used is is based on https://github.com/keycloak/keycloak/pull/45833. The PR is likely to be closed in favor of the new Graceful Shutdown offered by Quarkus 4 (the HTTP server that Keycloak uses).

This till happens, this is just a workaround that gives extra 5s, so that Keycloak can serve the in-flight connections instead of dropping them. The idea is that within these 5s, Kube Proxy will plug the instance that is terminating out from the Load Balancer. 

In theory, 1s should be enough for all this but I used some extra safety buffer (so 5s), to ensure everything shuts down cleanly.

src/keycloak/chart/templates/destination-rule.yaml
slaskawi: After thinking about this for a moment, I think keying on `devMode` is not correct and we should key on HA Mode instead. Only HA Keycloak deployments need the stickiness guarantees and they do not make any sense for 1 instance. 

I'll update the PR in a second.

src/keycloak/chart/values.schema.json
slaskawi: Good find! Thanks!

src/keycloak/chart/templates/waypoint-config.yaml
slaskawi: I agree! Let me update the docs and revert changes from here.

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -225,8 +233,12 @@ spec:
                 fieldRef:
                   fieldPath: spec.nodeName
                   apiVersion: v1
+          # Based on https://www.keycloak.org/server/reverseproxy#_enable_sticky_sessions
+          # Infinispan session stickiness should be disabled when session stickiness based on cookies is enabled
+          {{- if not .Values.autoscaling.sessionStickiness.enabled }}
```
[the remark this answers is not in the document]
slaskawi: The exact quote is:

> If your proxy supports session affinity without processing cookies from backend nodes, you should set the spi-sticky-session-encoder--infinispan--should-attach-route option to false in order to avoid attaching the node to cookies and just rely on the reverse proxy capabilities.

So if we disable Session Stickiness using cookies (set the `.Values.autoscaling.sessionStickiness.enabled` to `false`), we should set the SPI opton to false. 

In other words, this suggestion is wrong.

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -225,8 +233,12 @@ spec:
...
             - name: KC_SPI_STICKY_SESSION_ENCODER__INFINISPAN__SHOULD_ATTACH_ROUTE
               value: "false"
```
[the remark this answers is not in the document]
slaskawi: Added tests

docs/reference/configuration/resource-configuration-and-ha.md
mjnagel:
```suggestion
You can also configure autoscaling for the Keycloak [waypoint](https://istio.io/latest/docs/ambient/usage/waypoint/) (its Istio Layer7 proxy). The independent scalability of the waypoint proxy ensures that you never encounter a bottleneck due to service mesh integration. An HPA is enabled by default for the waypoint, with access to the configuration parameters via values. The below provides an override example that can be changed based on your needs:
```

Since this no longer matches the defaults.

docs/reference/configuration/resource-configuration-and-ha.md
```diff
@@ -243,6 +243,10 @@ packages:
               value: 256Mi
 ```
 
+:::tip
+For HA Keycloak deployments running on multiple nodes, it is recommended to set `waypoint.horizontalPodAutoscaler.minReplicas` to at least 2. This ensures there is no downtime when waypoint pods are restarted or rescheduled.
```
mjnagel: Should we mention antiaffinity as well?

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -225,8 +233,12 @@ spec:
...
+          {{- end }}
```
mjnagel: I think the AI comments here read correct unless I'm misunderstanding? Currently the condition is based on "Istio" session stickiness being _DISABLED_, and then disabling infinispan sticky sessions as well. I think it should be inverted?

Reading that comment:
> If your proxy supports session affinity without processing cookies from backend nodes, you should set the spi-sticky-session-encoder--infinispan--should-attach-route option to false in order to avoid attaching the node to cookies and just rely on the reverse proxy capabilities.

In other words, if we are relying on Istio (our reverse proxy) for sticky sessions (`Values.autoscaling.sessionStickiness.enabled` == true), then we should disable Infinispan sticky sessions. Based on that I think we just need to remove the `not` in the helm condition?

docs/reference/configuration/resource-configuration-and-ha.md
mjnagel: This suggestion doesn't show properly on GitHub for me, but is meant to adjust the working of `The below override example includes the default values...`

src/keycloak/chart/templates/statefulset.yaml
slaskawi: It's written in such a way that I had to look at the code how it works. 

This setting is used by Infinispan to identify the Primary Owner of an Authentication Session (the session that is used during the login process) based on the `AUTH_SESSION_ID` cookie. Since we're not using it in Istio (we use own cookie, which is then managed by Istio), this option always has to be false. 

I'll update the code shortly.

docs/reference/configuration/resource-configuration-and-ha.md
slaskawi: Done!

mjnagel (approved):

joelmccoy (approved):

Issue #2288, opened 2026-01-20
Transition Authservice to ConfigMap-based Trust Management


chance-coleman:
## Description

Currently, UDS Core manages Authservice's trusted CAs by injecting the full PEM bundle directly into the `AuthserviceConfig` JSON stored in a Kubernetes Secret. While functional, this diverges from the pattern used by UDS Packages, which mount a centralized `uds-trust-bundle` ConfigMap.

This issue tracks the refactoring of Authservice to use a mounted ConfigMap and the `trusted_certificate_authority_file` configuration option.

## Goals

- **Standardization**: Align Authservice trust management with the pattern used by UDS Packages.
- **Efficiency**: Reduce the size of the Authservice configuration secret by offloading large CA bundles (e.g., DoD PKI) to a separate ConfigMap.
- **Maintainability**: Simplify the Pepr operator logic by removing the need to pass raw PEM strings through the Authservice event chain.

## Proposed Changes

1.  **Operator Update**: Modify the Pepr operator to ensure the `uds-trust-bundle` ConfigMap is mirrored into the `authservice` namespace during global trust updates.
2.  **Configuration Update**: Update `buildInitialSecret` and the OIDC configuration generator in the Pepr operator to use the `trusted_certificate_authority_file` path (e.g., `/etc/authservice/ca-bundle.pem`) instead of `trusted_certificate_authority`.
3.  **Deployment Update**: Update the Authservice Zarf package/Helm chart to include:
    *   A volume definition for the `uds-trust-bundle` ConfigMap.
    *   A `volumeMount` for the CA bundle file at the expected path.
4.  **Rollout Logic**: Ensure that updates to the ConfigMap trigger a rolling restart of the Authservice pods (reusing the existing `reloadPods` or checksum logic).

## Acceptance Criteria

- Authservice successfully validates OIDC tokens against CAs provided via the mounted ConfigMap.
- The `authservice-uds` secret no longer contains raw PEM certificate data.
- Global CA rotations (User, DoD, Public) are successfully propagated to Authservice.

PR #2337, opened 2026-02-03
feat: automount uds trust bundle to all applications

### LINKED ISSUE #2269: Automount CA bundle for all UDS Core Apps

Today, only some UDS Core applications (notably Keycloak and Grafana) automatically mount and trust the UDS Core CA bundle. Other UDS Core applications require manual CA mounting, even though the central trust bundle already exists.

This inconsistency is confusing for users, requires digging through per-module docs to understand behavior.

---

* **Given** a UDS Core deployment with a configured central trust bundle
* **When** any UDS Core application is deployed
* **Then** that application automatically mounts and trusts the UDS Core CA bundle by default (unless explicitly overridden)

Additionally:

* **Given** a user reading the UDS Core documentation
* **When** they reach the trust / PKI documentation
* **Then** it is clearly stated up front that all UDS Core applications automatically mount the trust bundle, instead of requiring users to infer this from individual module sections

---

* Improving documentation only while keeping per-application behavior
* Listing which applications auto-mount certs

Both approaches help, but do not resolve the underlying inconsistency.

---

### Additional context

Keycloak and Grafana already auto-mount the trust bundle, setting an expectation of consistent behavior across UDS Core. Aligning all UDS Core applications with this behavior would reduce confusion and improve the private PKI experience.

chance-coleman:
## Description
Add UDS Trust Bundle mount to all UDS Core applications by default. 

## Related Issue

Fixes #2269 

## Type of change

## Steps to Validate
- use a uds-config.yaml file like so in the k3d standard bundle directory:
```yaml
variables:
  core:
    CA_BUNDLE_CERTS: "PLACEHOLDER_CA_BUNDLE_CERTS"
    CA_BUNDLE_INCLUDE_DOD_CERTS: "true"
    CA_BUNDLE_INCLUDE_PUBLIC_CERTS: "true"
```
deploy uds-core standard ( i use `uds run test:uds-core`)
- validate mounts and things


src/authservice/chart/templates/deployment.yaml
```diff
@@ -49,6 +49,10 @@ spec:
           volumeMounts:
             - name: {{ include "authservice.name" . }}
               mountPath: /etc/authservice
+            - name: ca-certs
+              mountPath: /etc/ssl/certs/ca-bundle.pem
+              subPath: ca-bundle.pem
+              readOnly: true
```
mjnagel: I don't think this is necessary since we mount/handle the CA separately for Authservice (unless you want to work https://github.com/defenseunicorns/uds-core/issues/2288 as well into this).

mjnagel (commented):
Plan to spend a bit more time with this later, but would love for us to at least have some upstream docs/code links to reference where/how the mount path is honored by each app. My suspicion is that `/etc/ssl/certs/ca-bundle.pem` might _not_ actually be honored by all of these, but could be wrong there.

As we were discussing on the side it would probably also be beneficial for us to have testing around these but that's a bit tricky and will be app specific. A few thoughts:
- Loki: If we had a MinIO with TLS config, could store logs there and just do validation that connection works?
- Vector: Similar to ^ could have Vector directly send to a MinIO with TLS config
- Velero: Similar to ^, just backup to that MinIO
- Prometheus: Have it scrape an external metrics endpoint (something on a VS)
- Alertmanager: Not sure, need something "receiving alerts" that uses the CA?
- Falco Sidekick: Not sure, probably similar to ^
- Falco: Even more not sure...is there a use-case for this?
- Metrics-server: I can't think of a user-case here either?

joelmccoy:
> Plan to spend a bit more time with this later, but would love for us to at least have some upstream docs/code links to reference where/how the mount path is honored by each app. My suspicion is that /etc/ssl/certs/ca-bundle.pem might not actually be honored by all of these, but could be wrong there.

Just to give some context.  The `/etc/ssl/certs` path is a basically a hack if your application is using the go crypto library for it's TLS.  See [here](https://cs.opensource.google/go/go/+/master:src/crypto/x509/root_linux.go;l=21?q=%2Fetc%2Fssl%2Fcerts%2F&ss=go%2Fgo). Go just happens to check this directory.  For all go applications this path should work if using the underlying crypto module... otherwise you are going to have to use application specific dirs or override the system trust bundle (which differs based on image distro).

src/authservice/chart/templates/deployment.yaml
chance-coleman: reverted this change

src/falco/values/values.yaml
```diff
@@ -54,10 +58,25 @@ mounts:
       mountPath: /etc/falco/rules.d/disable-rules.yaml
       subPath: disable-rules.yaml
       readOnly: true
+    - name: ca-certs
+      mountPath: /etc/ssl/certs/ca-bundle.pem
+      subPath: ca-bundle.pem
+      readOnly: true
```
mjnagel: Falco itself is a C++ application, so I don't think this path is standard? Doing a bit of review of the chart and possible places a CA would be used I found [this](https://github.com/falcosecurity/charts/blob/master/charts/falco/values.yaml#L1228-L1231), for http_output, which might indicate `/etc/falco/certs/ca-bundle.pem` is a good path to put it in?

I also am still unsure if this one _needs_ to be configured, since most things should go through the sidekick for any outbound traffic?

src/falco/values/values.yaml
```diff
@@ -54,10 +58,25 @@ mounts:
...
+  extraVolumes:
+    - name: ca-certs
+      configMap:
+        name: uds-trust-bundle
+        optional: true
+  extraVolumeMounts:
+    - name: ca-certs
+      mountPath: /etc/ssl/certs/ca-bundle.pem
+      subPath: ca-bundle.pem
+      readOnly: true
```
mjnagel, line 72: For sidekick looks like we might want to set [`config.tlsclient.cacertfile`](https://github.com/falcosecurity/charts/blob/master/charts/falcosidekick/values.yaml#L167-L169)? Although the current path might works either way due to go crypto defaults/fallbacks? Maybe a philosophy question overall - we could explicitly configure using app-specific approach, or just use the "go generic". Not sure if there might be more obvious logging of the config if we use app approach?

For a test of this (doesn't have to be in this PR), I think we could configure to send to [grafana](https://github.com/falcosecurity/charts/blob/master/charts/falcosidekick/values.yaml#L749) (external endpoint) in our private-pki bundle, although not sure what that looks like on the grafana side/if there's additional config. Mostly just looking for something that wouldn't need extra things deployed in/out of the cluster.

src/metrics-server/values/values.yaml
```diff
@@ -1,6 +1,19 @@
...
+    configMap:
+      name: uds-trust-bundle
+      optional: true
+
+extraVolumeMounts:
+  - name: ca-certs
+    mountPath: /etc/ssl/certs/ca-bundle.pem
+    subPath: ca-bundle.pem
+    readOnly: true
+
```
mjnagel: I _don't think_ there's a use-case for metrics-server pinging anything external, outside of the nodes/kubelet which has [its own certificate requirement](https://github.com/kubernetes-sigs/metrics-server/tree/master?tab=readme-ov-file#requirements:~:text=Kubelet%20certificate%20needs%20to%20be%20signed%20by%20cluster%20Certificate%20Authority%20(or%20disable%20certificate%20validation%20by%20passing%20%2D%2Dkubelet%2Dinsecure%2Dtls%20to%20Metrics%20Server)). Might recommend dropping this one for simplicity?

src/vector/values/values.yaml
```diff
@@ -4,6 +4,19 @@
...
+    configMap:
+      name: uds-trust-bundle
+      optional: true
+
+extraVolumeMounts:
+  - name: ca-certs
+    mountPath: /etc/ssl/certs/ca-bundle.pem
+    subPath: ca-bundle.pem
+    readOnly: true
+
```
mjnagel, line 22: Vector is rust-based and has a different "probe" path for finding trusted certs: https://vector.dev/docs/reference/configuration/tls/#trusted-certificates

I think this path will still work though 🎉 based on [this](https://github.com/rustls/openssl-probe/blob/main/src/lib.rs#L138-L139) which adds the `/etc/ssl/certs` dir for searching on linux (and looks like it mirrors go crypto). Maybe worth a comment linking the Vector docs and/or rust openssl code.

mjnagel (commented):
A few comments. I think it would be worth us opening a follow-on issue/PR here to try and e2e test as many of these as we can in our private-pki workflow/tasks.

src/falco/values/values.yaml
chance-coleman: I was also unsure about this one.I added it just cause i was doing all the other ones but agree that it seems unnecessary because of sidekick. IMO we remove it and depend on sidekick.

src/falco/values/values.yaml
chance-coleman: > Maybe a philosophy question overall - we could explicitly configure using app-specific approach, or just use the "go generic"

IMO we go generic to keep this as minimal and straightforward as possible across apps and distros.

src/falco/values/values.yaml
mjnagel: Should be able to remove the volume too?

src/falco/values/values.yaml
mjnagel: This works for me - the main benefit I _might_ see to app-specific paths is that we could be a bit more "guaranteed" that they work + maybe they surface better to end users that it's configured this way. I wouldn't advocate for the distro specific paths since that's more spread out.

Okay rolling forward with the go generic path here tho.

mjnagel (commented):
One last comment - LGTM overall at this point.

mjnagel (approved):

PR #2343, opened 2026-02-05
fix: generate unique callback URIs to prevent authservice conflicts

### LINKED ISSUE #560: Compute authservice redirectUri
Currently the authservice implementation uses the first `redirectUri` specified in the Package `sso` block as the `callback_uri` in the authservice configuration for the mission app.  

Ideally we would compute an arbitrary `callback_uri` and include that URI as an acceptable redirectUri in the Keycloak client.  This would avoid scenarios where users can specify a URI that is incompatible with authservice (Keycloak allows  globs in the redirectUri). 

Steps to implement this fix:
* Replace the `callback_uri` in the authservice `buildChain` function with a computed value that is: `https://hostname/login` 
* Update the Keycloak client creation logic to ensure the computed URI is added as an additional value (`https://hostname/login` ) in the `client.redirectUris`

### LINKED ISSUE #2333: Pepr does not reject `redirectUris` for the root `/` URL path, but this crashes Authservice
### Environment
Unclear exact version, but current-ish UDS I think. Reported by another contractor we work with on a Delivery effort.

### Steps to reproduce
1. Deploy a Package CR with `enableAuthserviceSelector:` enabled and `redirectUris: ["https://app.example.com/"]` for an app hosted on `app.example.com`
2. Observe the Authservice Pod crashes with the following log message: 
   ```
   time="2026/01/28 22:59:34" level=error msg="unexpected exit" scope="default" error="1 errors occured:
   invalid OIDC override from chain[foo].filter[0]: must not be root path: invalid callback URL"
   ```

(I didn't test `https://app.example.com` without a trailing slash, but I expect it would have similar behavior)

### Expected result

This is really a misconfiguration (but shouldn't crash Authservice). In the case of authservice, the first `redirectUri` will be used as the `callback_uri` in the generated Authservice chain. The URL should be something like `https://app.example.com/login` and _not_ in use by the application. A wildcard like `https://app.example.com/*` should not be used either.

I think best case the Package CR is rejected by the validating webhook which allow us to reject this early and give the user a clear error message.

### Actual Result

Authservice crashes. One workload's Package CR is able to crash Authservice for all pods. It does look like Authservice doesn't become immediately unavailable, it does a rolling restart and the old version (with old config) stays running.

The Istio RequestAuthentication/AuthorizationPolicy works correctly and denies access to the (misconfigured) workload with a 403 error. This is UDS correctly being fail-closed in this situation.

See log message above

### Severity/Priority

Low? It's a misconfiguration causing the crash anyway.

### Additional Context

This is really a misconfiguration, see notes above. Clearer documentation about the recommended configuration of `redirectUris:` when using Authservice would help.

Also, avoiding the configuration of `redirectUris:` entirely might be an alternative solution, as discussed in https://github.com/defenseunicorns/uds-core/issues/560

chance-coleman:
## Description
Makes redirect URIs optional for authservice clients in ambient mode by auto-generating callback URIs from expose entries.

### Key changes:

- Ambient mode: Can omit redirectUris - uses expose host for callback URI
- Root paths: `/` or `/*` now replaced with generated callback URI
- Sidecar mode: Still requires redirectUris (no change)
- Normal OAuth: Still requires redirectUris (no change)

Simplifies configuration for ambient mode deployments while maintaining security and backward compatibility.

Implementation: Added utility modules for hostname extraction and redirect URI processing with proper fallbacks and validation.

## Related Issue

Fixes #560 #2333 

## Type of change

## Steps to Validate
- deploy test apps and view admin console to validate that two URIs are now populated, one for the package defined redirect and one that is computer based on the clientID

## Checklist before merging


docs/reference/configuration/single-sign-on/auth-service.md
```diff
@@ -26,7 +26,14 @@ spec:
 ```
 
 :::note
-The UDS Operator uses the first `redirectUris` to populate the `match.prefix` hostname and `callback_uri` in the authservice chain.
+The UDS Operator automatically generates a unique callback URI for Authservice in the format `https://{hostname}/.uds/auth/callback/{hash}` where the hash is derived from the clientId. This callback URI is automatically added to the Keycloak client's `redirectUris` and used in the authservice chain configuration. The hostname is extracted from the first `redirectUri` you specify.
```
slaskawi: We might want to clarify further how this hash is calculated. This will add a bit more transparency into how this feature is built.

docs/reference/configuration/single-sign-on/auth-service.md
```diff
@@ -26,7 +26,14 @@ spec:
...
+
+**Important:** When using `enableAuthserviceSelector`, your `redirectUris` must:
+- Include a specific path (e.g., `/login`, `/callback`)
+- Root paths (`/` or empty) are not allowed
```
slaskawi: Just to educate me - what will happen if a use specifies a slash at the end? Would UDS Operator fail the validation or would the authservice complain? If the latter, I'd encourage to add some small validation bit to the UDS Operator to show a clear error to the user.

slaskawi (commented):
Great work! Added two comments and once addressed, it's a LGTM!

Issue #2305, opened 2026-01-26
Implement UDS Core Uptime Monitoring


joelmccoy:
Implement the Operator Logic called out in: https://www.notion.so/defense-unicorns/UDS-Core-Uptime-Monitoring-25ce512f24fc80dea2e0f280704beaab?source=copy_link

## Scope

- Add **opt-in HTTP(S) uptime probes** via UDS Package CRD
- Support **SSO-aware probing** for Authservice-enabled applications
- Generate and manage **Keycloak clients** for probe access (least-privilege)
- Reconcile probes into **Prometheus Operator `Probe` CRs**
- Securely manage Blackbox config via Kubernetes Secret
- Trigger Blackbox config reloads on change
- Ship **default Grafana dashboard(s)** for uptime and latency

## Out of Scope

- Default alerting rules (future work)
- Non-2xx checks

PR #2344, opened 2026-02-05
feat: enable uds package uptime probes via blackbox exporter

joelmccoy:
## Description

Enables Blackbox Exporter uptime probes via UDS Package CR.

Note: This PR does not include the logic for supporting Authservice Enabled applications.  This will come as a follow up PR and pending final approval of the design doc.

Grafana dashboard screenshot:
image

## Related Issue

Relates to #2305 

## Type of change

## Steps to Validate
- Can modify uptime checks in the grafana and keycloak UDS package
  - set `expose[0].uptime.checks.enabled = false` - ensure probe goes away
  - set `expose[0].uptime.checks.enabled = true` - ensure probe comes back up
- query `probe_success` in grafana to see metrics
- check out the UDS / Monitoring / Uptime Grafana dashboard to see metrics visualized
 
## Checklist before merging


src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -0,0 +1,122 @@
...
+  pkgName: string,
+  generation: string,
+  ownerRefs: V1OwnerReference[],
+): PrometheusProbe {
+  const { uptime } = expose;
+  const fqdn = getFqdn(expose);
+
+  // Build the list of target URLs from paths (default to "/" if not specified)
+  const paths = uptime?.checks?.paths ?? ["/"];
+  const targets = paths.map(path => `https://${fqdn}${path}`);
```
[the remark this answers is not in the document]
joelmccoy: Actually going to block paths that don't start with `/`.  This is part of the design doc.  Will add this in the package validator.

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -223,6 +223,40 @@ const expose = {
...
+                },
+                default: ["/"],
+              },
+              interval: {
+                description: "How frequently to scrape the targets (e.g., 30s, 1m)",
+                type: "string",
+              },
+              scrapeTimeout: {
+                description: "Timeout for each scrape request (e.g., 10s)",
+                type: "string",
```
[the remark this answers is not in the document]
joelmccoy: actually, after seeing this, i realize we don't provide these values for podmonitor/servicemonitors. I think I am going to leave these off and let these be the default Prometheus CR values unless someone asks for them.

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -124,6 +125,9 @@ async function reconcilePackageFlow(pkg: UDSPackage): Promise<void> {
   monitors.push(...(await podMonitor(pkg, namespace!)));
   monitors.push(...(await serviceMonitor(pkg, namespace!)));
 
+  // Configure the Uptime Probes
+  await probe(pkg, namespace!);
```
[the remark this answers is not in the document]
joelmccoy: fixed

src/prometheus-stack/chart/dashboards/uptime-monitoring.json
```diff
@@ -0,0 +1,369 @@
...
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 0,
+  "id": 0,
+  "links": [],
+  "panels": [
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "PBFA97CFB590B2093"
```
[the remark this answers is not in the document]
joelmccoy: fixed

src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Get the FQDN for an expose entry based on host and gateway
+ *
+ * @param entry The expose entry
+ * @returns The fully qualified domain name
+ */
+export function getFqdn(entry: Expose): string {
+  const gateway = entry.gateway;
+  const domain = gateway === "admin" ? UDSConfig.adminDomain : UDSConfig.domain;
```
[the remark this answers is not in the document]
joelmccoy: fixed

src/pepr/operator/controllers/uptime/probe.ts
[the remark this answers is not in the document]
joelmccoy: added a validation that ensures paths start with `/`

docs/reference/configuration/observability/blackbox-exporter.md
```diff
@@ -1,66 +1,142 @@
...
+        host: myapp
+        gateway: tenant
+        port: 8080
+        uptime:
+          checks:
+            enabled: true
 ```
 
-## Usage
+This creates a Prometheus Probe that monitors `https://myapp.uds.dev/` using the default `http_2xx` module.
```
briantwatson: Might consider adding a little additional detail as to what the probe does here?  `...issues an HTTP GET requests at X interval and checks for 200 response`

src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -0,0 +1,120 @@
...
+): PrometheusProbe {
+  const { uptime } = expose;
+  const fqdn = getFqdn(expose);
+
+  // Build the list of target URLs from paths (default to "/" if not specified)
+  const paths = uptime?.checks?.paths ?? ["/"];
+  const targets = paths.map(path => `https://${fqdn}${path}`);
+
+  // Generate a sanitized name for the probe
+  const name = sanitizeResourceName(`uds-${fqdn}-uptime`);
```
briantwatson: If we're only using the fqdn for a resource name, will this conflict if a user has defined probes for multiple endpoints?

src/pepr/operator/crd/validators/package-validator.spec.ts
```diff
@@ -930,6 +930,96 @@ describe("Test Allowed SSO Client Attributes", () => {
...
+      ],
+      [],
+      [],
+      [],
+    );
+    await validator(mockReq);
+    expect(mockReq.Approve).toHaveBeenCalledTimes(1);
+  });
+
+  it("denies duplicate FQDNs with uptime enabled", async () => {
```
briantwatson: Maybe consider aligning naming on test descriptions by choosing `hosts` or `FQDNs`?  Could lead to confusion when revisiting these in the future.  When I read this and other tests that mention `hosts` I searched back through the PR to try to clarify what the limitation on 'one' per expose entry was and how that related. If we use FQDN, might consider using an FQDN in the test case(s) as well, e.g. `app.uds.dev`.

briantwatson (commented):
Nice addition!  I added a couple comments for consideration for clarity

src/prometheus-stack/values/blackbox-exporter-values.yaml
chance-coleman, line 15: Should we add the ca trust bundle for private pki? something like the [tls_config](https://github.com/prometheus/blackbox_exporter/blob/master/CONFIGURATION.md?plain=1#L152) configuration value.

src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -0,0 +1,120 @@
...
+      labels: {
+        "uds/package": pkgName,
+        "uds/generation": generation,
+      },
+      ownerReferences: ownerRefs,
+    },
+    spec: {
+      // Use default http_2xx module
+      module: "http_2xx",
+      prober: {
```
chance-coleman, line 100: I'm probably missing something, do we want to allow for changing default `interval` and `scrapeTimeout`? Might have larger impacts on the crd i suppose.

chance-coleman (commented):
🚀 🚀

src/prometheus-stack/values/blackbox-exporter-values.yaml
joelmccoy: hmmmm. this is a good callout. I think we can add this as a mount to the container instead of in the CR.  I will do that

src/pepr/operator/controllers/uptime/probe.ts
joelmccoy: I originally intend to do that (based on the design doc), but I also realized that we don't do this for the monitor CRs (podMonitor/serviceMonitor).  I am now leaning towards leaving these out of the CR until someone asks for them...  Would love a second opinion if we think that is the right move.

For now I am going to set this to 5s by default as that is what we defined in our Registry Uptime doc for polling metrics.  And if someone thinks that this should be configurable, I'd be open to adding that into the package CR.

docs/reference/configuration/observability/blackbox-exporter.md
joelmccoy: added

src/pepr/operator/crd/validators/package-validator.spec.ts
joelmccoy: switched to fqdn to standardize

docs/reference/configuration/observability/blackbox-exporter.md
```diff
@@ -1,66 +1,142 @@
...
+UDS Core includes Blackbox Exporter for HTTP/HTTPS probing of endpoints to monitor uptime and availability. Uptime checks can be configured through the UDS Package CR, and Prometheus [Probes](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.Probe) resources are automatically created based on your configuration.
 
-## Bundle Configuration
+## Uptime Monitoring
+
+To enable uptime monitoring for an exposed service, configure the `uptime.checks` section within your Package CR's `expose` entries.
 
 :::note
-Blackbox Exporter is deployed as an optional component within the core package. You must explicitly enable it in the `optionalComponents` section of your bundle configuration.
+Uptime checks do not currently support applications protected by Authservice. Support for probing Authservice-enabled applications is planned for a future release.
```
mjnagel: Maybe a nuance that we don't need to mention here - but they would be "probed" technically right? Just the probe wouldn't actually make it to the app (would just be probing the external interface for authservice)?

docs/reference/configuration/observability/blackbox-exporter.md
```diff
@@ -1,66 +1,142 @@
...
 
-# Check current configuration
-curl http://localhost:9115/config
+Monitor multiple services in a single package:
+
+```yaml
+spec:
+  network:
+    expose:
+      # monitors: https://app.uds.dev/healthz, https://api.uds.dev/health, https://api.uds.dev/ready, https://admin.uds.dev/
```
mjnagel: nit: I think the admin example here is on the admin gateway and would result in `admin.admin.uds.dev` for the host + admin domain?

docs/reference/configuration/observability/blackbox-exporter.md
```diff
@@ -1,66 +1,142 @@
...
-### Default Module
+The Blackbox Exporter provides some of the following key metrics (not exhaustive) that can be used for alerting and dashboarding:
 
-The default `http_2xx` module handles basic HTTP probing:
+| Metric | Description |
+|--------|-------------|
+| `probe_success` | Whether the probe succeeded (1) or failed (0) |
+| `probe_duration_seconds` | Total probe duration |
+| `probe_http_status_code` | HTTP response status code |
+| `probe_ssl_earliest_cert_expiry` | SSL certificate expiration timestamp |
```
mjnagel, line 132: Maybe this is straightforward but could we add an example promql query? Mostly the job naming, which I think would be driven off of the namespace and name of the probes?

packages/monitoring/zarf.yaml
```diff
@@ -26,7 +26,7 @@ components:
 
   # Blackbox Exporter
   - name: prometheus-blackbox-exporter
-    required: false
+    required: true
```
mjnagel, line 29: I think with this change we can drop `prometheus-blackbox-exporter` from the `optionalComponents` on all of our bundles (`.github/bundles` and `bundles/k3d-standard`). Looks like we have Falco also lingering in those 😅. Okay to do that last or even as a separate PR if we want since that would trigger all the IaC (don't want that to be run every commit here).

src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -0,0 +1,121 @@
...
+ * Get the FQDN for an expose entry based on host, domain, and gateway
+ *
+ * @param entry The expose entry
+ * @returns The fully qualified domain name
+ */
+export function getFqdn(entry: Expose): string {
+  const gateway = entry.gateway;
+  const domain = entry.domain ?? (gateway === "admin" ? UDSConfig.adminDomain : UDSConfig.domain);
+  return entry.host === "." ? domain : `${entry.host}.${domain}`;
+}
```
mjnagel: I think there's one missing condition here for `gateway.includes("admin")` comparing this to the VS function. Can we pull that out/ensure [this](https://github.com/defenseunicorns/uds-core/blob/black-box-uptime/src/pepr/operator/controllers/istio/virtual-service.ts#L42-L56) is in a shared function so these don't drift? You've already pulled this into a func, just want to make sure that both the VS and the probe use that func.

src/pepr/operator/controllers/uptime/probe.ts
mjnagel: I think for now I'd leave it static, but could see this being an ask in the future.

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -223,6 +223,32 @@ const expose = {
           type: "string",
         },
       },
+      uptime: {
+        description: "Uptime monitoring configuration for this exposed service.",
+        type: "object",
+        properties: {
+          checks: {
```
mjnagel, line 231: Thoughts on dropping the `enabled` key and making `paths` required without a default? I see merits to both paths, but I'm not seeing a strong argument for `checks.enabled = false`. Mostly feels like an extra step assuming you have custom paths to input for checks, although maybe it simplifies checks for the default root path (just need to input `true` rather than `['/']`).

Should've asked at design time - but thinking more now looking at the interface... The thing that feels off looking at this is that all of these values are optional. If we keep `enabled` we should make it required I think?

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -91,6 +94,28 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+
+      // Validate no duplicate FQDNs
+      const fqdn = getFqdn(expose);
+      if (uptimeFqdns.has(fqdn)) {
+        return req.Deny(
+          `Duplicate uptime probe for FQDN "${fqdn}". ` +
+            `Only one expose entry per FQDN can have uptime enabled. ` +
+            `Disable uptime on duplicate entries with "uptime.checks.enabled: false".`,
+        );
+      }
```
mjnagel, line 114: Is this something we could de-duplicate / coalesce at reconcile time rather than denying here? I have a foggy memory of discussing this _somewhere_ but can't find this in the design. Okay with denying too - just wanted to mention it (not sure how frequent multiple expose entries are for the same FQDN).

mjnagel (commented):
Think this tracks with the design - just a few comments.

packages/monitoring/zarf.yaml
joelmccoy: Will do this last. Good callout

src/pepr/operator/crd/validators/package-validator.ts
joelmccoy: We could technically do that... but there is some deep merge logic that would need to happen.  It's not a big deal now that we don't have interval/scrape timeout defined in the UDS Package CRD.... But if we have fields added to this in the future, you would need to have logic to deep merge (and decided what settings take precedent).  I think denying outright is the best option (and this shouldn't happen very often with mission apps I imagine).

src/pepr/operator/crd/sources/package/v1alpha1.ts
joelmccoy: Yeah that's a good callout.  I think I am fine with that approach and can update the design doc and this PR to reflect that.

src/pepr/operator/crd/validators/package-validator.ts
mjnagel: Good with this - going to resolve.

src/keycloak/chart/templates/uds-package.yaml
```diff
@@ -124,6 +124,9 @@ spec:
         selector:
           app.kubernetes.io/name: keycloak
         port: 8080
+        uptime:
+          checks:
+            enabled: true
```
mjnagel: For the purposes of our registry env we defined a few additional paths:
- `/`
- `/realms/uds/.well-known/openid-configuration`
- `/realms/uds/protocol/openid-connect/token`

I think these could have utility as default checks, or we could just expose helm templating to pass in paths?

joelmccoy: added them in as defaults for the keycloak package

src/pepr/operator/crd/sources/package/v1alpha1.ts
joelmccoy: updated.

src/pepr/operator/controllers/uptime/probe.ts
joelmccoy: refactored this out into its own util function

docs/reference/configuration/observability/blackbox-exporter.md
joelmccoy: added quick examples

docs/reference/configuration/observability/blackbox-exporter.md
joelmccoy: fixed

docs/reference/configuration/observability/blackbox-exporter.md
joelmccoy: added in some clarity here

src/pepr/operator/controllers/domain-utils.ts
joelmccoy, line 1: Note: I pulled this out to it's own file as there was a circular dep of imports of UDSConfig in `config.ts`.  I could also put this in `utils.ts` and have the function pass in the `UDSConfig.domain` & `UDSConfig.adminDomain` as parameters if we prefer that for cleanliness??

joelmccoy: But I also do like the idea of seprating utils based on function.  Maybe it is worth a refactor of having a `utils/<function>.ts`.

For example:
```
utils/
  domain.ts
  k8s.ts
  ...
```

src/keycloak/chart/templates/uds-package.yaml
joelmccoy: actually it appears that `/realms/uds/protocol/openid-connect/token` can only be called with a `POST` with proper parameters... are we sure this is something we want to monitor uptime for? it would require crafting a new POST module, and it feels weird that a monitoring solution could have a side effect

mjnagel: cc @slaskawi on this one. Making a POST feels strange I would agree - we added to the list since it's what other clients would be calling, but might just leave this off based on what you've seen.

src/pepr/operator/controllers/domain-utils.ts
mjnagel: I think this makes sense, good with deferring a "util refactor" until later too.

src/pepr/operator/controllers/domain-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Expose, Gateway } from "../crd";
+import { UDSConfig } from "./config/config";
+
+/**
+ * Get the FQDN for an expose entry based on host, domain, and gateway
+ *
+ * @param expose The expose entry
+ * @returns The fully qualified domain name
+ */
+export function getFqdn(expose: Expose): string {
```
mjnagel, line 15: I think we can also use this func in the [serviceentry generation](https://github.com/defenseunicorns/uds-core/blob/black-box-uptime/src/pepr/operator/controllers/istio/service-entry.ts#L67-L76) (forgot about this earlier).

src/pepr/operator/crd/sources/package/v1alpha1.ts
mjnagel: Resolving, appreciate the change here.

src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -0,0 +1,108 @@
...
+    // Apply the Probe and force overwrite any existing resource
+    await K8s(PrometheusProbe).Apply(payload, { force: true });
+
+    probeNames.push(payload.metadata!.name!);
+  }
+
+  // Purge any orphaned probes from previous generations
+  await purgeOrphans(generation, namespace, pkgName, PrometheusProbe, log);
+
+  return probeNames;
```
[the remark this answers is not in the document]
joelmccoy: fixed

src/pepr/operator/controllers/domain-utils.ts
joelmccoy: good catch. replaced.

src/keycloak/chart/templates/uds-package.yaml
joelmccoy: removed this endpoint for now until we get some clarity on if this is really needed

src/pepr/operator/controllers/domain-utils.ts
joelmccoy: keeping this as is for now

chance-coleman:
Probably should update the c4 diagrams both [in core](https://github.com/defenseunicorns/uds-core/tree/main/docs/.c4) and [here](https://github.com/defenseunicorns/uds-compliance/blob/main/uds-core/docs/.c4/ports-protocols.views.c4) for Blackbox (sorry i probably should have done this in the previous PR 🫣 ).

joelmccoy:
> Probably should update the c4 diagrams both [in core](https://github.com/defenseunicorns/uds-core/tree/main/docs/.c4) and [here](https://github.com/defenseunicorns/uds-compliance/blob/main/uds-core/docs/.c4/ports-protocols.views.c4) for Blackbox (sorry i probably should have done this in the previous PR 🫣 ).

updated the diagrams in this PR

joelmccoy:
I think most of the comments have been address here.  If all looks generally good, I will remove the optional component piece to trigger IAC tests.  Just want to make sure there aren't any other feedback points before I do this.

chance-coleman (commented):
i think this is LGTM, i'd approve it now pending your optional switch.

src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -0,0 +1,112 @@
...
+): PrometheusProbe {
+  const { uptime } = expose;
+  const fqdn = getFqdn(expose);
+
+  // Build the list of target URLs from paths
+  const paths = uptime!.checks!.paths!;
+  const targets = paths.map(path => `https://${fqdn}${path}`);
+
+  // Generate a sanitized name for the probe
+  const name = sanitizeResourceName(`uds-${fqdn}-uptime`);
```
mjnagel: Very very nit, but might use the host + gateway here rather than the full fqdn. Should still guarantee uniqueness, without making the names "environment specific" (and potentially quite long).

mjnagel (commented):
LGTM overall.

src/pepr/operator/controllers/uptime/probe.ts
mjnagel: Okay deferring this or sticking with the current naming - mostly was thinking through the FQDN having "known bad characters" (`.`) that would get sanitized, as well as potentially length domain names in some scenarios.

joelmccoy: good call. updated this.

mjnagel (approved):

chance-coleman (approved):

PR #2349, opened 2026-02-06
fix: validate authservice callback uri + redirect uri

### LINKED ISSUE #2333: Pepr does not reject `redirectUris` for the root `/` URL path, but this crashes Authservice

chance-coleman:
## Description
We cannot allow for root path URI's otherwise authservice will break. This PR add validation that doesn't allow authservice enabled apps to also have root path uri's defined in the redirect uri array field.

## Related Issue

Fixes #2333 

## Type of change

## Steps to Validate

## Checklist before merging

### DIFF
--- a/docs/reference/configuration/single-sign-on/auth-service.md
+++ b/docs/reference/configuration/single-sign-on/auth-service.md
@@ -29,6 +29,50 @@ spec:
 The UDS Operator uses the first `redirectUris` to populate the `match.prefix` hostname and `callback_uri` in the authservice chain.
 :::
 
+:::warning
+### Redirect URI Validation
+
+When using `enableAuthserviceSelector`, the UDS Operator validates the `redirectUris` array to ensure proper configuration:
+
+- **Root paths (`/` or `/*`) are not allowed** in `redirectUris` for authservice clients
+- **Valid redirect URIs must be fully qualified URLs** (e.g., `https://example.com/callback`)
+- **Non-authservice clients are not affected** by this validation
+
+#### Examples:
+
+✅ **Valid:**
+```yaml
+sso:
+  - name: "My App"
+    clientId: my-app
+    redirectUris: ["https://myapp.example.com/login"]
+    enableAuthserviceSelector:
+      app: myapp
+```
+
+❌ **Invalid:**
+```yaml
+sso:
+  - name: "My App"
+    clientId: my-app
+    redirectUris: ["/"]  # Root path not allowed
+    enableAuthserviceSelector:
+      app: myapp
+```
+
+❌ **Invalid:**
+```yaml
+sso:
+  - name: "My App"
+    clientId: my-app
+    redirectUris: ["https://myapp.example.com/login", "/"]  # Mixed with root path
+    enableAuthserviceSelector:
+      app: myapp
+```
+
+If invalid redirect URIs are detected, the Package will be denied with an error message indicating the issue.
+:::
+
 For complete examples, see [app-ambient-authservice-tenant.yaml](https://github.com/defenseunicorns/uds-core/blob/main/src/test/app-ambient-authservice-tenant.yaml) and [app-sidecar-authservice-tenant.yaml](https://github.com/defenseunicorns/uds-core/blob/main/src/test/app-sidecar-authservice-tenant.yaml)
 
 ## Multiple Services and Selectors
--- a/docs/reference/configuration/single-sign-on/auth-service.md
+++ b/docs/reference/configuration/single-sign-on/auth-service.md
@@ -55,7 +55,7 @@ sso:
 sso:
   - name: "My App"
     clientId: my-app
-    redirectUris: ["/"]  # Root path not allowed
+    redirectUris: ["https://myapp.example.com/"]  # Root path not allowed
     enableAuthserviceSelector:
       app: myapp
 ```
--- a/docs/reference/configuration/single-sign-on/auth-service.md
+++ b/docs/reference/configuration/single-sign-on/auth-service.md
@@ -65,7 +65,7 @@ sso:
 sso:
   - name: "My App"
     clientId: my-app
-    redirectUris: ["https://myapp.example.com/login", "/"]  # Mixed with root path
+    redirectUris: ["https://myapp.example.com/login", "https://myapp.example.com/"]  # Mixed with root path
     enableAuthserviceSelector:
       app: myapp
 ```
--- a/src/pepr/operator/crd/validators/package-validator.spec.ts
+++ b/src/pepr/operator/crd/validators/package-validator.spec.ts
@@ -843,6 +843,7 @@ describe("Test validation of Package CRs", () => {
     expect(mockReq.Deny).toHaveBeenCalledWith(
       'The client ID "test-client" has redirectUris containing root paths ("/" or "/*"). Authservice clients cannot have root path redirect URIs.',
     );
+    expect(mockReq.Approve).not.toHaveBeenCalled();
   });
 
   it("denies authservice clients with redirectUris containing URLs with wildcard root paths", async () => {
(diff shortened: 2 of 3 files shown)


docs/reference/configuration/single-sign-on/auth-service.md
```diff
@@ -29,6 +29,50 @@ spec:
...
+
+❌ **Invalid:**
+```yaml
+sso:
+  - name: "My App"
+    clientId: my-app
+    redirectUris: ["https://myapp.example.com/login", "/"]  # Mixed with root path
+    enableAuthserviceSelector:
+      app: myapp
+```
```
mjnagel, line 71: nit: for the invalid example can we use a `https://myapp.example.com/` (root path on the host)?

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -305,6 +305,31 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+          }
+
+          // Try to parse as URL to check the path component
+          const url = new URL(uri);
+          const path = url.pathname;
+          return path === "/" || path === "/*";
+        } catch {
+          // If URL parsing fails, fall back to checking if it's a root path
+          return uri === "/" || uri === "/*";
+        }
```
mjnagel, line 319: The initial `if` and the `catch` are duplicated (I don't think we need the initial `if`? I also think we'd need the redirectUri to always be a full URL, so maybe if it doesn't parse as a URL (catch block) we reject it?

docs/reference/configuration/single-sign-on/auth-service.md
chance-coleman: addressed

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -305,6 +305,28 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+            `The client ID "${client.clientId}" has an invalid redirect URI "${uri}". Redirect URIs must be valid URLs.`,
+          );
+        }
+      });
+      if (hasRootPaths) {
+        return req.Deny(
+          `The client ID "${client.clientId}" has redirectUris containing root paths ("/" or "/*"). Authservice clients cannot have root path redirect URIs.`,
+        );
+      }
+    }
```
joelmccoy: nit: This logic reads a bit weird.  The use of `.some()` here is may be a bit misleading? Functions passed to `.some()` should return booleans, but this callback sometimes returns `req.Deny()`, which is a side effect and not a boolean. Since we want to fail fast on invalid redirect URIs or disallowed root paths, a simple loop with early returns seems clearer:

```
if (client.enableAuthserviceSelector && client.redirectUris) {
  for (const uri of client.redirectUris) {
    let url: URL;

    try {
      url = new URL(uri);
    } catch {
      return req.Deny(
        `The client ID "${client.clientId}" has an invalid redirect URI "${uri}". Redirect URIs must be valid URLs.`,
      );
    }

    if (url.pathname === "/" || url.pathname === "/*") {
      return req.Deny(
        `The client ID "${client.clientId}" has redirectUris containing root paths ("/" or "/*"). Authservice clients cannot have root path redirect URIs.`,
      );
    }
  }
}
```

joelmccoy (dismissed):
Generally lgtm with 1 nit suggestion

src/pepr/operator/crd/validators/package-validator.spec.ts
```diff
@@ -821,6 +821,113 @@ describe("Test validation of Package CRs", () => {
...
+        },
+      ],
+      [],
+    );
+    await validator(mockReq);
+    expect(mockReq.Deny).toHaveBeenCalledTimes(1);
+    expect(mockReq.Deny).toHaveBeenCalledWith(
+      'The client ID "test-client" has an invalid redirect URI "/". Redirect URIs must be valid URLs.',
+    );
+  });
```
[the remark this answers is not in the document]
joelmccoy: req.Deny should return anyways. don't think this is necessary

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -305,6 +305,28 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+          // If URL parsing fails, reject the invalid URI
+          return req.Deny(
+            `The client ID "${client.clientId}" has an invalid redirect URI "${uri}". Redirect URIs must be valid URLs.`,
+          );
+        }
+      });
+      if (hasRootPaths) {
+        return req.Deny(
+          `The client ID "${client.clientId}" has redirectUris containing root paths ("/" or "/*"). Authservice clients cannot have root path redirect URIs.`,
+        );
```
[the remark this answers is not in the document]
joelmccoy: had a similar comment here: https://github.com/defenseunicorns/uds-core/pull/2349#discussion_r2783418458

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -305,6 +305,28 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+          const url = new URL(uri);
+          const path = url.pathname;
+          return path === "/" || path === "/*";
+        } catch {
+          // If URL parsing fails, reject the invalid URI
+          return req.Deny(
+            `The client ID "${client.clientId}" has an invalid redirect URI "${uri}". Redirect URIs must be valid URLs.`,
+          );
+        }
+      });
```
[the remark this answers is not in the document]
joelmccoy: I think this is an accurate comment as well?  I don't think we explicitly block empty `redirectUris` with authservice enabled and maybe we should with this logic?

src/pepr/operator/crd/validators/package-validator.ts
chance-coleman: definitely a good call out, this has been refactored to use a for loop instead. similar to what copilot was suggesting.

src/pepr/operator/crd/validators/package-validator.ts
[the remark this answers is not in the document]
mjnagel: I think technically this would be duplicative because we have a check above for:
```ts
    if (client.standardFlowEnabled !== false && !client.redirectUris) {
      return req.Deny(
        `The client ID "${client.clientId}" must specify redirectUris if standardFlowEnabled is turned on (it is enabled by default)`,
      );
    }
```
To ensure that redirects are always set for clients?

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -305,6 +305,33 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
       );
     }
+
+    // If this is an authservice client, deny redirectUris that contain any root paths
+    if (client.enableAuthserviceSelector) {
+      if (!client.redirectUris || client.redirectUris.length === 0) {
+        return req.Deny(
+          `The client ID "${client.clientId}" requires at least one redirect URI when enableAuthserviceSelector is true`,
+        );
+      }
```
mjnagel: Probably got lost in the resolved thread, but is there a specific need for this validation since we already "generically" validate for redirect URIs on all clients above?
```ts
    if (client.standardFlowEnabled !== false && !client.redirectUris) {
      return req.Deny(
        `The client ID "${client.clientId}" must specify redirectUris if standardFlowEnabled is turned on (it is enabled by default)`,
      );
    }
```

chance-coleman: the problem with `!client.redirectUris` is that it asserts as truthy for when redirectUris are defined as `[]` (meaning `![]` is `false`, so the condition wouldn't be met and it won't deny) where as with the new one we have to make sure that it has at least one entry. Maybe thats what we should have been doing all along with the original validation though.

mjnagel: Hm, yeah I think that's fair - and probably should be part of the earlier validation then?

chance-coleman: updated the previous check to validate there was at least 1 redirectUri defined and remove the new check that i added because its now duplication.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -305,6 +305,27 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
         `The client ID "${client.clientId}" is invalid as an Authservice client - Authservice does not support client IDs with the ":" character`,
       );
     }
+
+    // If this is an authservice client, deny redirectUris that contain any root paths
+    if (client.enableAuthserviceSelector && client.redirectUris?.length) {
```
mjnagel, line 310: nit: I think `&& client.redirectUris?.length` is duplicative here since we already would've denied if empty?

mjnagel (approved):
LGTM

src/pepr/operator/crd/validators/package-validator.ts
chance-coleman: is there a usecase where `enableAuthserviceSelector` and `standardFlow` don't overlap? at somepoint in this new if statement i will have to verify that the redirectUri is present because of downstream code depending on `redirectUri[0]` being present.

mjnagel: I think technically it's possible (as in: we don't deny if someone disables standard flow + enables authservice?), so okay leaving this one here. But I think functionally you need `standardFlow` for authservice to work right? Not 100% confident.

chance-coleman: Pretty sure that's correct. `standardFlowEnabled` is passed directly to Keycloak to enable/disable OAuth2 authorization code flow endpoints that authservice depends on. When `standardFlowEnabled` is false, Keycloak disables these endpoints but authservice still tries to use them, causing authentication failures.

PR #2358, opened 2026-02-10
feat: expose keycloak http client retry configurations

### LINKED ISSUE #791: Incorporate retries for Keycloak OCSP checks
https://github.com/defenseunicorns/uds-identity-config/issues/584 incorporated HTTP retries upstream in Keycloak (see [PR](https://github.com/keycloak/keycloak/pull/42535/changes)). Leveraging these options will require passing in new startup flags or environment variables.

We should configure core with sensible defaults (maybe even no retries?) but expose options to end users for modifying their retry behavior.

chance-coleman:
## Description
Expose the ability to configure Keycloak HTTP Client retries. Disabled by default. 

## Related Issue

[Fixes 791](https://github.com/defenseunicorns/uds-identity-config/issues/791)

## Type of change

## Steps to Validate
- Unfortunately this is really difficult to test and validate because it depends on faulty http requests and even when those happens there's little evidence that its happening in the background other than a request failing and the working.

## Checklist before merging

### DIFF
--- a/docs/reference/configuration/single-sign-on/keycloak-customization-guide.md
+++ b/docs/reference/configuration/single-sign-on/keycloak-customization-guide.md
@@ -138,7 +138,7 @@ Keycloak can automatically retry failed outgoing HTTP requests to handle transie
 
 ### Configuration
 
-HTTP retry behavior is disabled by default and must be explicitly enabled. Configure it via Helm chart values:
+HTTP retry behavior is disabled by default and must be explicitly enabled by setting `maxRetries` above `0`. Configure it via Helm chart values:
 
 ```yaml
 packages:
@@ -147,8 +147,6 @@ packages:
       keycloak:
         keycloak:
           values:
-            - path: httpRetry.enabled
-              value: true
             - path: httpRetry.maxRetries
               value: 2
             - path: httpRetry.initialBackoffMillis
@@ -165,8 +163,7 @@ packages:
 
 | Option                 | Description                           | Default |
 |------------------------|---------------------------------------|---------|
-| `enabled`              | Enable/disable HTTP retries           | `false` |
-| `maxRetries`           | Maximum retry attempts (0 = disabled) | `2`     |
+| `maxRetries`           | Maximum retry attempts (0 = disabled) | `0`     |
 | `initialBackoffMillis` | Initial backoff time in milliseconds  | `1000`  |
 | `backoffMultiplier`    | Exponential backoff multiplier        | `2.0`   |
 | `applyJitter`          | Apply jitter to prevent retry storms  | `true`  |
@@ -178,6 +175,7 @@ packages:
 - Keycloak communicates with external identity providers over unreliable networks
 - Database connections experience transient failures
 - External services (email, LDAP, etc.) have occasional availability issues
+- OCSP responders need multiple attempts to validate certificates
 - You need improved resilience for production deployments
 
 **Keep disabled when**:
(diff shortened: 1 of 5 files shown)


docs/reference/configuration/single-sign-on/keycloak-customization-guide.md
```diff
@@ -131,6 +132,60 @@ packages:
...
+| `backoffMultiplier`    | Exponential backoff multiplier        | `2.0`   |
+| `applyJitter`          | Apply jitter to prevent retry storms  | `true`  |
+| `jitterFactor`         | Jitter factor for backoff variation   | `0.5`   |
+
+### When to Enable HTTP Retries
+
+**Enable retries when**:
+- Keycloak communicates with external identity providers over unreliable networks
+- Database connections experience transient failures
+- External services (email, LDAP, etc.) have occasional availability issues
```
mjnagel, line 177: nit: do we explicitly want to call out OCSP here?

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -124,6 +124,18 @@ spec:
...
+            - "--spi-connections-http-client-default-max-retries={{ .Values.httpRetry.maxRetries }}"
+            - "--spi-connections-http-client-default-initial-backoff-millis={{ .Values.httpRetry.initialBackoffMillis }}"
+            - "--spi-connections-http-client-default-backoff-multiplier={{ .Values.httpRetry.backoffMultiplier }}"
+            {{- if not .Values.httpRetry.applyJitter }}
+            - "--spi-connections-http-client-default-apply-jitter=false"
+            {{- end }}
+            {{- if .Values.httpRetry.applyJitter }}
+            - "--spi-connections-http-client-default-jitter-factor={{ .Values.httpRetry.jitterFactor }}"
+            {{- end }}
+            {{- end }}
```
mjnagel: Can we add a based helm unittest or two for this templating?

src/keycloak/chart/values.yaml
```diff
@@ -446,6 +446,26 @@ truststorePaths:
...
   - "/tmp/ca-certs"
 
+# HTTP retry configuration for outgoing Keycloak requests
+# This configures retry behavior for transient network errors and temporary service unavailability
+httpRetry:
+  # Enable/disable HTTP retries for outgoing requests
+  # When disabled, no retries will be attempted for failed HTTP requests
+  enabled: false
+  # Maximum number of retry attempts for failed HTTP requests (0 = disabled)
+  maxRetries: 2
```
mjnagel: Would it make sense to drop `enabled` and just use `maxRetries: 0` as the default and the toggle?

chance-coleman: yeah good call, addressed!

mjnagel (approved):

src/keycloak/chart/values.yaml
```diff
@@ -446,6 +446,23 @@ truststorePaths:
   # Default path for UDS trust bundle certificates
   - "/tmp/ca-certs"
 
+# HTTP retry configuration for outgoing Keycloak requests
+# This configures retry behavior for transient network errors and temporary service unavailability
+httpRetry:
+  # Maximum number of retry attempts for failed HTTP requests (0 = disabled)
+  maxRetries: 0
```
slaskawi, line 453: hmmm would it make sense to enable it (with some sensible retries, like 1-2 maybe) by default?

PR #2367, opened 2026-02-12
chore: set explicit securityContext for Falco

mjnagel:
## Description

This PR explicitly sets the securityContext for Falco. By doing this we can prevent potential of race conditions where Pepr mutates the user/group for Falco and causes it to start incorrectly.

This is not an actual change to the security context, it just sets the implicit values explicitly:
- run as user/group: These are defaults in each of the images, `docker run -it --entrypoint id docker.io/falcosecurity/falco:0.42.1` is one way to check this
- privileged: This is a default in the chart that would get set if we did not provide our own security context ([upstream ref](https://github.com/falcosecurity/charts/blob/master/charts/falco/templates/pod-template.tpl#L399-L417))

### DIFF
--- a/src/falco/chart/templates/uds-exemption.yaml
+++ b/src/falco/chart/templates/uds-exemption.yaml
@@ -10,8 +10,9 @@ spec:
   exemptions:
     - policies:
         - DisallowHostNamespaces
+        - DisallowPrivileged
         - RequireNonRootUser
-        - RestrictCapabilities
+        - DropAllCapabilities
         - RestrictHostPathWrite
         - RestrictVolumeTypes
       matcher:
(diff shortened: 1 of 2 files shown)


joelmccoy (dismissed):
LGTM - no net changes here.  But did come across this value in the chart: https://github.com/falcosecurity/charts/blob/9bb50d88013d568cde85acecfa624f4626380d42/charts/falco/values.yaml#L350

I wonder if we could get away with running this in least privileged mode? https://falco.org/docs/setup/container/#docker-least-privileged

jasonwashburn (dismissed):

mjnagel:
> But did come across this value in the chart: https://github.com/falcosecurity/charts/blob/9bb50d88013d568cde85acecfa624f4626380d42/charts/falco/values.yaml#L350

I did notice that as well, but didn't explore it too far. Let me test that out before merging here to see if it works easy enough. The [template](https://github.com/falcosecurity/charts/blob/9bb50d88013d568cde85acecfa624f4626380d42/charts/falco/templates/pod-template.tpl#L397-L423) for security context upstream is a bit messy because you can't both use that value and customize the container security context (with user/group). So we'll just have to "manually" set the equivalents and keep that in sync. Probably not too bad/not much to keep up with. I don't know if we care either, but if someone did change the driver kind it would require different capabilities.

mjnagel:
Reverted back to the privileged setup based on inconsistent testing. Least privilege is likely worth exploring separate from this PR.

joelmccoy (approved):

PR #2376, opened 2026-02-12
chore(ci): switch azure account, improve test stability

mjnagel:
## Description

This PR includes a few changes primarily focused around CI stability:
- Azure/AKS:
  - Switches from Gov -> Public/Commercial (necessary restructure for org, not technical issues)
  - Adds optional controlplane/audit logging
  - Simplifies setup to a single nodepool
  - Adds [overlay](https://azure.microsoft.com/en-us/blog/announcing-the-general-availability-of-azure-cni-overlay-in-azure-kubernetes-service/) network plugin mode
- Adds a general `cluster-ready` task that waits on node and readyz endpoint health for all IaC tests
- Adjusts testing forward helper to use a host selected port rather than random selection by the test
- Increase pod reload test timeouts and de-dupe event checks
- Add a retryWithDelay around purge `LIST` calls


chance-coleman (approved):

src/pepr/operator/controllers/utils.ts
```diff
@@ -131,7 +131,8 @@ export async function purgeOrphans<T extends GenericClass>(
     }
   }
 
-  const resources = await query.Get();
+  const fetchResources = () => query.Get();
```
joelmccoy, line 134: Obviously this is the main failure we have been seeing, but there are also a [bunch of other Get() calls](https://github.com/search?q=repo%3Adefenseunicorns%2Fuds-core%20.get()&type=code).  Do you think it is worth wrapping those too?  Doesn't need to be this PR... Or we could just get KFC to implement retries on 429s?

joelmccoy (approved):
LGTM!

jasonwashburn (approved):

src/pepr/operator/controllers/utils.ts
mjnagel: I'd probably push for upstreaming this if we can - just chatted with @cmwylie19 and I think he's agreed on this being a KFC/Pepr thing.

PR #2378, opened 2026-02-13
chore: explicitly set runAsNonRoot false for vector/falco

mjnagel:
## Description

Similar/follow-on to https://github.com/defenseunicorns/uds-core/pull/2367

Reviewing our policy mutation logic I identified that this is another spot where being explicit will be helpful:
https://github.com/defenseunicorns/uds-core/blob/04c6c7b5927a80b8511ec45b206c75be54c90751/src/pepr/policies/security.ts#L179-L182

By explicitly defining `runAsNonRoot` we avoid any mutations, and ensure that the pod will be blocked (without being mutated) until the exemption is processed. This guarantees we don't create a poorly configured pod.

### DIFF
--- a/src/vector/values/values.yaml
+++ b/src/vector/values/values.yaml
@@ -85,13 +85,13 @@ customConfig:
       encoding:
         codec: "raw_message"
       labels:
-        namespace: '{{`{{ kubernetes.pod_namespace }}`}}'
-        app: '{{`{{ app }}`}}'
-        job: '{{`{{ kubernetes.pod_namespace }}`}}/{{`{{ app }}`}}'
-        container: '{{`{{ kubernetes.container_name }}`}}'
-        component: '{{`{{ component }}`}}'
-        host: '{{`{{ kubernetes.pod_node_name }}`}}'
-        filename: '{{`{{ file }}`}}'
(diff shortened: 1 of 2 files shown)


src/vector/values/values.yaml
```diff
@@ -85,13 +85,13 @@ customConfig:
...
-        component: '{{`{{ component }}`}}'
-        host: '{{`{{ kubernetes.pod_node_name }}`}}'
-        filename: '{{`{{ file }}`}}'
+        namespace: "{{`{{ kubernetes.pod_namespace }}`}}"
+        app: "{{`{{ app }}`}}"
+        job: "{{`{{ kubernetes.pod_namespace }}`}}/{{`{{ app }}`}}"
+        container: "{{`{{ kubernetes.container_name }}`}}"
+        component: "{{`{{ component }}`}}"
+        host: "{{`{{ kubernetes.pod_node_name }}`}}"
+        filename: "{{`{{ file }}`}}"
```
mjnagel, line 94: This was just auto formatting in my editor. Tested vector + loki to validate all labels are still there as expected and seems good.

chance-coleman (approved):

PR #2380, opened 2026-02-13
feat: falco rule overrides

### LINKED ISSUE #2254: Support bundle-level rule exceptions via ConfigMap overrides
## Problem

UDS Core ships Falco with curated stable / incubating / sandbox rulesets and supports disabling rules by name, but it does not support adding exceptions to enabled rules via bundle overrides.

As a result, users must either:

* disable a rule entirely (losing coverage), or
* fork/modify upstream rules (hard to maintain).

There is no way to keep a rule enabled while carving out known, expected behavior.

---

## Why this matters

Some Falco rules are correct but noisy in UDS Core due to expected behavior from system components.

Example:

* `ebs-csi-node` / `efs-csi-node` pods can legitimately trigger filesystem, mount, or syscall-based rules
* These are known-good workloads, but today the only mitigation is disabling the rule

We need a way to:

* keep rules enabled
* add **narrow, auditable exceptions** (pod, namespace, image, etc.)

This matches Falco’s upstream model of layering local overrides on top of base rules.

---

## Current State

* Falco rules are shipped via ConfigMaps:

  * `falco-stable-rules`
  * `falco-incubating-rules`
  * `falco-sandbox-rules`
* Bundle overrides support:

  * enabling/disabling rule tiers
  * disabling rules by name
* There is **no support** for:

  * overriding rule conditions
  * appending to macros or lists
  * pod-specific exceptions

---

### 1. Bundle-level Falco rule overrides

[Falco upstream exceptions that would be use for this configmap](https://falco.org/docs/concepts/rules/exceptions/#rule-exceptions)

Add support for an **optional ConfigMap** (e.g. `falco-local-overrides`) populated via bundle overrides that:

* is mounted into the Falco pod
* is loaded **after** stable/incubating/sandbox rules

This would allow:

* appending to lists
* appending/replacing macros
* overriding rule conditions when necessary

This mirrors Falco’s `falco_rules.local.yaml` pattern.

---

### 2. Investigate CSI pod exceptions

As part of this work, investigate:

* which Falco rules are commonly triggered by `ebs-csi-node` / `efs-csi-node`
* whether those rules already expose macros/lists suitable for exceptions
* what the **minimal safe exception** looks like

Outcome:

* documented exception guidance
* possible UDS-default exceptions where behavior is universally expected

briantwatson:
## Description

- Adds a new set of values for falco helm chart that allow for [rule overrides](https://falco.org/docs/concepts/rules/overriding/) including [rule exceptions](https://falco.org/docs/concepts/rules/exceptions/)
- Adds exception for privileged mount for csi pods for eks IaC bundle

Note Falco documentation:
> The keys that can be overridden vary by rules component and action being taken:
> - Lists (append or replace): items
> - Macros (append or replace): condition
> - Rules (append): condition, output, desc, tags, exceptions
> - Rules (replace): condition, output desc, priority, tags, exceptions, enabled, warn_evttypes, skip-if-unknown-filter

## Related Issue

Fixes #2254

## Type of change

## Steps to Validate

<details>
<summary>Testing Details</summary>

Added the following to falco bundle overrides for k3d-standard bundle to test:

```
      values:
                macros:
                  user_expected_terminal_shell_in_container_conditions:
                    action: replace
                    condition: '(k8s.ns.name="monitoring")'
                rules:
                  "Terminal shell in container":
                    exceptions:
                      action: append
                      items:
                        - name: allow_shell_in_ns
                          fields: ["k8s.ns.name"]
                          comps: ["="]
                          values:
                            - ["monitoring"]
                  "Unexpected UDP Traffic":
                    exceptions:
                      action: append
                      items:
                        - name: allow_udp_in_smoke_ns
                          fields: ["proc.name", "fd.l4proto"]
                          comps: ["=", "="]
                          values:
                            - ["iptables-restor", "udp"]
            - path: extraRules
              value:
                - rule: UDS Smoke Test - Open marker file
                  desc: Smoke test rule to verify extraRules is loaded (opens a known file)
                  condition: evt.type in (open,openat,openat2) and fd.name="/tmp/uds-falco-smoke.marker" and evt.is_open_read=true
                  output: "UDS SMOKE: opened marker file (proc=%proc.name user=%user.name container=%container.id)"
                  priority: NOTICE
                  tags: [local, uds_smoke]
```

Test extraRules:

```
kubectl run falco-smoke-marker \
  -n monitoring \
  --image=127.0.0.1:31999/library/busybox:1.37.0-zarf-2140033595 \
  --restart=Never \
  --command -- sh -c 'echo hello > /tmp/uds-falco-smoke.marker && cat /tmp/uds-falco-smoke.marker >/dev/null' \
  --overrides='
{
  "spec": {
    "securityContext": {
      "runAsNonRoot": true,
      "runAsUser": 1000
    },
    "containers": [{
      "name": "falco-smoke-marker",
      "securityContext": {
        "allowPrivilegeEscalation": false,
        "capabilities": { "drop": ["ALL"] }
      }
    }]
  }
}'
```
Observed "UDS SMOKE: opened marker file..." in logs

Test exception:

```
kubectl run falco-smoke-shell \
  -n monitoring \
  --image=127.0.0.1:31999/library/busybox:1.37.0-zarf-2140033595 \
  --restart=Never \
  --command -- sh -c 'while true; do sleep 3600; done' \
  --overrides='
{
  "spec": {
    "securityContext": {
      "runAsNonRoot": true,
      "runAsUser": 1000
    },
    "containers": [{
      "name": "falco-smoke-shell",
      "securityContext": {
        "allowPrivilegeEscalation": false,
        "capabilities": { "drop": ["ALL"] }
      }
    }]
  }
}
```

Opened shell in pod. Did not see "Terminal shell in container" in logs.  (Confirmed prior to rule exception I did see in logs)

Confirmed the following removed Unexpected UDP traffic in k3d demo deployments

```
"Unexpected UDP Traffic":
exceptions:
  action: append
  items:
	- name: allow_udp_in_smoke_ns
	  fields: ["proc.name", "fd.l4proto"]
	  comps: ["=", "="]
	  values:
		- ["iptables-restor", "udp"]
```

</details>

## Checklist before merging

### DIFF
--- a/.github/bundles/eks/uds-bundle.yaml
+++ b/.github/bundles/eks/uds-bundle.yaml
@@ -1,4 +1,4 @@
-# Copyright 2024 Defense Unicorns
+# Copyright 2024-2026 Defense Unicorns
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 kind: UDSBundle
--- a/.github/bundles/eks/uds-bundle.yaml
+++ b/.github/bundles/eks/uds-bundle.yaml
@@ -129,20 +129,21 @@ packages:
           values:
             - path: overrides
               value:
-                "Mount Launched in Privileged Container":
-                  exceptions:
-                    action: append
-                    items:
-                      - name: allow_csi_node_mounts
-                        fields:
-                          - k8s.ns.name
-                          - k8s.pod.name
-                          - proc.name
-                        comps:
-                          - =
-                          - startswith
-                          - =
-                        values:
-                          - ["kube-system"]
-                          - ["ebs-csi-node-", "efs-csi-node-"]
-                          - ["mount"]
+                rules:
+                  "Mount Launched in Privileged Container":
+                    exceptions:
+                      action: append
+                      items:
+                        - name: allow_csi_node_mounts
+                          fields:
+                            - k8s.ns.name
+                            - k8s.pod.name
+                            - proc.name
+                          comps:
+                            - =
+                            - startswith
+                            - =
+                          values:
+                            - ["kube-system"]
+                            - ["ebs-csi-node-", "efs-csi-node-"]
+                            - ["mount"]
--- a/.github/bundles/eks/uds-bundle.yaml
+++ b/.github/bundles/eks/uds-bundle.yaml
@@ -134,16 +134,13 @@ packages:
                     exceptions:
                       action: append
                       items:
-                        - name: allow_csi_node_mounts
-                          fields:
-                            - k8s.ns.name
-                            - k8s.pod.name
-                            - proc.name
-                          comps:
-                            - =
-                            - startswith
-                            - =
+                        - name: allow_csi_efs_node_mounts
+                          fields: [k8s.ns.name, k8s.pod.name, proc.name]
+                          comps: [in, startswith, in]
                           values:
-                            - ["kube-system"]
-                            - ["ebs-csi-node-", "efs-csi-node-"]
-                            - ["mount"]
+                            - [kube-system, efs-csi-node-, mount]
+                        - name: allow_csi_ebs_node_mounts
+                          fields: [k8s.ns.name, k8s.pod.name, proc.name]
+                          comps: [in, startswith, in]
+                          values:
+                            - [kube-system, ebs-csi-node, mount]
--- a/.github/bundles/eks/uds-bundle.yaml
+++ b/.github/bundles/eks/uds-bundle.yaml
@@ -136,11 +136,11 @@ packages:
                       items:
                         - name: allow_csi_efs_node_mounts
                           fields: [k8s.ns.name, k8s.pod.name, proc.name]
-                          comps: [in, startswith, in]
+                          comps: [=, startswith, =]
                           values:
                             - [kube-system, efs-csi-node-, mount]
                         - name: allow_csi_ebs_node_mounts
                           fields: [k8s.ns.name, k8s.pod.name, proc.name]
-                          comps: [in, startswith, in]
+                          comps: [=, startswith, =]
                           values:
                             - [kube-system, ebs-csi-node, mount]
--- a/.github/bundles/eks/uds-bundle.yaml
+++ b/.github/bundles/eks/uds-bundle.yaml
@@ -126,3 +126,23 @@ packages:
               description: Enable incubating rules
               path: incubatingRulesEnabled
               default: "true"
+          values:
+            - path: overrides
+              value:
+                "Mount Launched in Privileged Container":
+                  exceptions:
+                    action: append
+                    items:
+                      - name: allow_csi_node_mounts
+                        fields:
+                          - k8s.ns.name
+                          - k8s.pod.name
+                          - proc.name
+                        comps:
+                          - =
+                          - startswith
+                          - =
+                        values:
+                          - ["kube-system"]
+                          - ["ebs-csi-node-", "efs-csi-node-"]
+                          - ["mount"]
--- a/docs/reference/configuration/runtime-security/alerting.md
+++ b/docs/reference/configuration/runtime-security/alerting.md
@@ -57,7 +57,7 @@ To use this feature, provide an array of rule names under the `disabledRules` va
 
 **Falco Official Documentation:**
 
-  - [List of Falco Rules](https://falco.org/docs/reference/rules/default-rules/)
+- [List of Falco Rules](https://falco.org/docs/reference/rules/default-rules/)
 
 **UDS Core rule files:**
 
@@ -141,25 +141,25 @@ overrides:
 
 In AWS EKS environments, it is common to see Falco alerts triggered by the CSI (Container Storage Interface) drivers, such as EFS and EBS, because these drivers launch privileged containers to perform storage operations. These alerts are expected and do not indicate malicious activity. To reduce noise and avoid unnecessary investigation of these known benign events, it is recommended to add rule exceptions for the affected CSI driver pods. The following override demonstrates how to safely suppress these alerts while maintaining visibility into other privileged container activity.
 
-```
-  values:
-    - path: overrides
-      value:
-        rules:
-          "Mount Launched in Privileged Container":
-            exceptions:
-              action: append
-              items:
-                - name: allow_csi_efs_node_mounts
-                  fields: [k8s.ns.name, k8s.pod.name, proc.name]
-                  comps: [=, startswith, =]
-                  values:
-                    - [kube-system, efs-csi-node-, mount]
-                - name: allow_csi_ebs_node_mounts
-                  fields: [k8s.ns.name, k8s.pod.name, proc.name]
-                  comps: [=, startswith, =]
-                  values:
-                    - [kube-system, ebs-csi-node, mount]
+```yaml
+values:
+  - path: overrides
+    value:
+      rules:
+        "Mount Launched in Privileged Container":
+          exceptions:
+            action: append
+            items:
+              - name: allow_csi_efs_node_mounts
+                fields: [k8s.ns.name, k8s.pod.name, proc.name]
+                comps: [=, startswith, =]
+                values:
+                  - [kube-system, efs-csi-node-, mount]
+              - name: allow_csi_ebs_node_mounts
+                fields: [k8s.ns.name, k8s.pod.name, proc.name]
+                comps: [=, startswith, =]
+                values:
+                  - [kube-system, ebs-csi-node, mount]
 ```
 
 ### Querying Events with Loki
--- a/docs/reference/configuration/runtime-security/alerting.md
+++ b/docs/reference/configuration/runtime-security/alerting.md
@@ -27,7 +27,7 @@ UDS Core ships with **sandbox** and **incubating** rulesets from the Falco commu
 
 #### Enabling and Configuring Rulesets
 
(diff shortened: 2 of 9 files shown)


src/falco/chart/values.yaml
```diff
@@ -35,3 +35,41 @@ udsDefaultDisabledRulesIncubating:
...
+#       condition: "or evt.type=openat"
+#   rules:
+#     "Terminal shell in container":
+#       exceptions:
+#         action: append
+#         items:
+#           - name: allow_ci_shell
+#             fields: ["container.image.repository"]
+#             comps: ["in"]
+#             values: ["ghcr.io/my-org/ci-runner"]
```
briantwatson, line 67: This needs to be a list of list and will be adjusting

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -27,44 +27,115 @@ UDS Core ships with **sandbox** and **incubating** rulesets from the Falco commu
 
 #### Enabling and Configuring Rulesets
 
-To enable the [sandbox and incubating](https://falco.org/docs/reference/rules/default-rules/) rulesets and exclude specific rules, override the `extraRules` value in your UDS Core bundle:
+To enable the [sandbox and incubating](https://falco.org/docs/reference/rules/default-rules/) rulesets and exclude specific rules, override the `falco.uds-falco-config` value in your UDS Core bundle:
```
joelmccoy: nit: a more correct statement here would be
```suggestion
To enable the [sandbox and incubating](https://falco.org/docs/reference/rules/default-rules/) rulesets and exclude specific rules, override the `sandBoxRulesEnabled`, `incubatingRulesEnabled`, `disabledRules` values in the `uds-falco-config` helm chart in your UDS Core bundle:
```

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -27,44 +27,115 @@ UDS Core ships with **sandbox** and **incubating** rulesets from the Falco commu
...
-  - [List of Falco Rules](https://falco.org/docs/reference/rules/default-rules/)
+To use this feature, provide an array of rule names under the `disabledRules` value in your configuration. The names must match the `rule` field in the Falco rules files.
 
-2. **In the rule files shipped with UDS Core**:
-   - Sandbox rules: [`src/falco/chart/rules/sandbox-rules.yaml`](https://github.com/defenseunicorns/uds-core/blob/main/src/falco/chart/rules/sandbox-rules.yaml)
-   - Incubating rules: [`src/falco/chart/rules/incubating-rules.yaml`](https://github.com/defenseunicorns/uds-core/blob/main/src/falco/chart/rules/incubating-rules.yaml)
+**How to find rule names:**
 
-   Look for entries that start with `- rule:` to find the rule names.
+1. **Falco Official Documentation:**
```
joelmccoy: The formatting/numbering for this list doesn't look right

Image

src/falco/chart/values.schema.json
joelmccoy, line 1: nit: Adding a schema is a good idea, but I think this schema is missing a lot of our current values we set [here](https://github.com/defenseunicorns/uds-core/blob/4f0e0e6912c6e6898726e0678048cff8b1716344/src/falco/chart/values.yaml)?

I guess leaving them out won't cause issues because the top level `"additionalProperties": true,`.  It might be a good idea to add them in for completeness? But definitely not blocking here.

joelmccoy (commented):
Generally looks good to me! A couple very small comments

.github/bundles/eks/uds-bundle.yaml
```diff
@@ -124,3 +124,21 @@ packages:
...
+                        - name: allow_csi_efs_node_mounts
+                          fields: [k8s.ns.name, k8s.pod.name, proc.name]
+                          comps: [=, startswith, =]
+                          values:
+                            - [kube-system, efs-csi-node-, mount]
+                        - name: allow_csi_ebs_node_mounts
+                          fields: [k8s.ns.name, k8s.pod.name, proc.name]
+                          comps: [=, startswith, =]
+                          values:
+                            - [kube-system, ebs-csi-node, mount]
```
mjnagel, line 146: Do we also want to document this as an example in our runtime-security doc? Or somewhere else, essentially as a "if you're on EKS, you might want to do this...".

src/falco/chart/values.schema.json
briantwatson: I had considered that, one thought that had crossed my mind was all the values that might be out there 'out in the wild' that could then be breaking due to the schema addition. That's why I went that route with `additionalProperties` and just what was added here as least invasive. I hadn't thought through too deeply with the existing values though--Do you think that scenario might be a valid concern?

docs/reference/configuration/runtime-security/alerting.md
briantwatson: Thanks for catching this--will adjust!

.github/bundles/eks/uds-bundle.yaml
briantwatson: Following discussion will add this to a 'common overrides' section to alerting documentation.

src/falco/chart/values.schema.json
joelmccoy: Generally I think the schema should reflect all the values you can set.  It's an interface for interacting with your helm chart.  I would probably recommend adding our current values (don't need to do any validation on these things other than listing the type it should be). I think you can keep additionalProperties as true.  But wouldn't be opposed to flipping to false.  If someone is providing a value that isn't used in the helm chart they should get feedback (as they might think something is being set when in reality it isn't).

docs/reference/configuration/runtime-security/alerting.md
briantwatson: Adjusted.  If we prefer numbered here I can add back in

src/falco/chart/values.schema.json
briantwatson: Added to schema!

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -27,44 +27,140 @@ UDS Core ships with **sandbox** and **incubating** rulesets from the Falco commu
...
+            rules:
+              "Unexpected UDP Traffic":
+                exceptions:
+                  action: append
+                  items:
+                    - name: allow_udp_in_smoke_ns
+                      fields: ["proc.name", "fd.l4proto"]
+                      comps: ["=", "="]
+                      values:
+                        - ["iptables-restor", "udp"]
```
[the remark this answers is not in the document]
briantwatson: n/a

docs/reference/configuration/runtime-security/alerting.md
```diff
@@ -27,44 +27,140 @@ UDS Core ships with **sandbox** and **incubating** rulesets from the Falco commu
...
+**Exception Structure Rules**
+
+- `fields`, `comps` must have the same length.
+- When using multiple fields, each element in `values` must be an array (tuple) whose length matches the number of fields. When using a single field or omitting the `fields` specification, `values` can be a simple array of scalar values.
+
+#### Common Overrides
+
+In AWS EKS environments, it is common to see Falco alerts triggered by the CSI (Container Storage Interface) drivers, such as EFS and EBS, because these drivers launch privileged containers to perform storage operations. These alerts are expected and do not indicate malicious activity. To reduce noise and avoid unnecessary investigation of these known benign events, it is recommended to add rule exceptions for the affected CSI driver pods. The following override demonstrates how to safely suppress these alerts while maintaining visibility into other privileged container activity.
+
+```
```
mjnagel: nit: should have `yaml` formatting here.

mjnagel (dismissed):
LGTM overall - probably good to get a re-review from @joelmccoy as well before spinning CI on a small docs change 😅 (that could also be done as a very quick follow-on PR with just the docs CI).

src/falco/chart/values.schema.json
```diff
@@ -0,0 +1,306 @@
...
+        "port": {
+          "type": "integer",
+          "minimum": 1,
+          "maximum": 65535
+        },
+        "description": {
+          "type": "string"
+        }
+      }
+    },
```
joelmccoy, line 112: nit: This technically should match our `allow` field in our CRD: https://github.com/defenseunicorns/uds-core/blob/a9866964204e3ed88d196ded986255e6e69867ad/src/pepr/operator/crd/sources/package/v1alpha1.ts#L44-L149

I'd almost be fine with leaving this as is or an open object so that we don't have to keep this up.  No action needed here since we do allow additionalProperties.  But wanted to call this out.

joelmccoy: actually... i wonder if you could import this with json schema from here: https://github.com/defenseunicorns/uds-core/blob/a9866964204e3ed88d196ded986255e6e69867ad/schemas/package-v1alpha1.schema.json#L216-L307

That would ensure this is actually accurate and always up to date

joelmccoy (dismissed):
LGTM with 1 nit for consideration on the schema stuff

src/falco/chart/values.schema.json
briantwatson: Thanks for catching this and providing context.

I couldn't find any way to reference files locally and helm unittest to succeed how I think we'd want relative, reliably.  I checked the repo here and all refs in schemas are referencing local definitions from within the file.  If you have an idea or reference for specific implementation I'd welcome testing that out.

I'm going to push up making this just an object to reduce maintenance burden as you suggest.

docs/reference/configuration/runtime-security/alerting.md
briantwatson: added

joelmccoy (approved):

PR #2393, opened 2026-02-18
chore(docs): document airgapped x509 crl checking

### LINKED ISSUE #2335: Document airgapped certificate revocation check process
Currently UDS Core/Keycloak by default will query an OCSP server to check a certificate's status (whether it has been revoked). In an airgap this is often not possible (unless there is a mirror of the OCSP server running there). The preferred method for airgaps is to load CRLs for Keycloak.

In order to better support and understand this use-case, we should test and document the process for loading CRLs, especially for DoW CACs.

Documentation focus:
- Where to pull CRLs for CACs
- How to "airgap" CRLs
- How to load CRLs to Keycloak

chance-coleman:
## Description
Currently UDS Core/Keycloak by default will query an OCSP server to check a certificate's status (whether it has been revoked). In an airgap this is often not possible (unless there is a mirror of the OCSP server running there). The preferred method for airgaps is to load CRLs for Keycloak.

## Related Issue

Fixes #2335 

## Type of change

## Steps to Validate
- the dev doc is intended to be used to show how this is validated since it take a lot of steps and commands to do it. 

## Checklist before merging

### DIFF
--- a/docs/dev/keycloak-crl-airgap.md
+++ b/docs/dev/keycloak-crl-airgap.md
@@ -218,7 +218,7 @@ kubectl logs -n keycloak keycloak-0 | grep -i "x509\|crl\|revoc" | tail -n 120
 
 ## Negative tests (and how to make them reliably take effect)
 
-Keycloak caches CRLs in an internal Infinispan cache keyed by the configured CRL path. In practice, when iterating locally you may need to force Keycloak to re-read the CRL file.
+Keycloak caches CRLs in an internal Infinispan cache keyed by the configured CRL path. In practice, when iterating locally you may need to force Keycloak to re-read the CRL file. For debugging, prefer running a **single replica** so cached data is not preserved during Infinispan rebalances.
 
 If a negative test doesn’t change behavior immediately:
 
--- a/docs/dev/keycloak-crl-airgap.md
+++ b/docs/dev/keycloak-crl-airgap.md
@@ -1,7 +1,3 @@
----
-title: Local Testing Example (Fake CA + CRL) for Keycloak X.509
----
-
 # Local Testing Example (Fake CA + CRL) for Keycloak X.509
 
 This guide is for **local/dev testing only**. It validates that Keycloak:
--- a/docs/dev/keycloak-crl-airgap.md
+++ b/docs/dev/keycloak-crl-airgap.md
@@ -1,0 +1,329 @@
+---
+title: Local Testing Example (Fake CA + CRL) for Keycloak X.509
+---
+
+# Local Testing Example (Fake CA + CRL) for Keycloak X.509
+
+This guide is for **local/dev testing only**. It validates that Keycloak:
+
+- Accepts a client certificate presented through the UDS Istio gateway
+- Loads a local CRL file
+- Enforces revocation and freshness checks
+
+For production/airgapped implementation details (DoD CRLs, Zarf packaging, and platform overrides), use:
+
+- [Airgap CRL Configuration for Keycloak X.509 Authentication](/reference/configuration/single-sign-on/keycloak-crl-airgap/)
+
+---
+
+## Prerequisites
+
+- A local UDS Core cluster deployed (Keycloak + Istio gateways)
+- `kubectl`
+- `openssl`
+- Firefox (recommended for client certificate testing)
+
+---
+
+## 1) Deploy UDS Core locally
+
+Deploy your local dev bundle:
+
+```bash
+uds deploy k3d-core-slim-dev:latest
+```
+
+Confirm you can reach:
+
+- `https://sso.uds.dev`
+- `https://keycloak.admin.uds.dev`
+
+> Probably will need `zarf connect keycloak` to set up the admin user to access the Keycloak admin console.
+
+---
+
+## 2) Generate a fake CA, client certificate, and CRL
+
+Run on your workstation:
+
+```bash
+set -euo pipefail
+
+WORKDIR="${WORKDIR:-$(pwd)/pki-crl-test}"
+mkdir -p "$WORKDIR"
+cd "$WORKDIR"
+
+mkdir -p demoCA/newcerts
+touch demoCA/index.txt
+echo 1000 > demoCA/serial
+echo 1000 > demoCA/crlnumber
+
+cat > ca.conf <<'EOF'
+[ ca ]
+default_ca = CA_default
+
+[ CA_default ]
+dir = .
+database = $dir/demoCA/index.txt
+new_certs_dir = $dir/demoCA/newcerts
+certificate = $dir/ca.crt
+serial = $dir/demoCA/serial
+private_key = $dir/ca.key
+crlnumber = $dir/demoCA/crlnumber
+crl = $dir/crl.pem
+default_md = sha256
+default_days = 365
+default_crl_days = 30
+unique_subject = no
+policy = policy_any
+copy_extensions = copy
+
+[ policy_any ]
+commonName = supplied
+
+[ v3_client ]
+subjectAltName = @alt_names
+keyUsage = digitalSignature
+extendedKeyUsage = clientAuth, 2.16.840.1.101.3.6.8
+certificatePolicies = @policy_anything
+
+[ alt_names ]
+otherName.0 = 2.16.840.1.101.3.6.6;UTF8:test@mil
+email = unicorn.test@uds.dev
+
+[ policy_anything ]
+policyIdentifier = 2.16.840.1.101.3.2.1.3.12
+EOF
+
+# CA
+openssl genrsa -out ca.key 2048
+openssl req -x509 -new -nodes -key ca.key -sha256 -days 3650 -out ca.crt \
+  -subj "/CN=UDS CRL Demo CA" \
+  -addext "basicConstraints=critical,CA:TRUE" \
+  -addext "keyUsage=critical,keyCertSign,cRLSign"
+
+# Client cert
+openssl genrsa -out client.key 2048
+openssl req -new -key client.key -out client.csr -subj "/CN=UNICORN.TEST.USER.1234567890"
+openssl ca -batch -config ca.conf -in client.csr -out client.crt -extensions v3_client
+
+# PFX for browser import (empty export password)
+openssl pkcs12 -export -out client.pfx -inkey client.key -in client.crt -certfile ca.crt -passout pass:
+
+# CRL (PEM then convert to DER)
+openssl ca -batch -config ca.conf -gencrl -out crl.pem
+openssl crl -in crl.pem -inform PEM -out crl.der -outform DER
+```
+
+---
+
+## 3) Configure the Istio tenant gateway to request/accept your CA
+
+This makes the gateway:
+
+- Advertise your CA to the browser
+- Accept a client cert signed by that CA
+
+```bash
+kubectl -n istio-tenant-gateway patch secret gateway-tls \
+  --type merge \
+  -p '{"data": {"cacert": "'"$(base64 -w0 ca.crt)"'"}}'
+
+kubectl -n istio-tenant-gateway rollout restart deployment tenant-ingressgateway
+```
+
+---
+
+## 4) Ensure Keycloak trusts your CA
+
+Keycloak must trust the issuing CA:
+
+- To validate the client certificate chain
+- To verify the CRL signature
+
+```bash
+kubectl -n keycloak create configmap uds-trust-bundle \
+  --from-file=ca-bundle.pem=ca.crt \
+  --dry-run=client -o yaml | kubectl apply -f -
+
+kubectl -n keycloak rollout restart statefulset keycloak
+```
+
+---
+
+## 5) Deliver the CRL into the Keycloak pod
+
+`kubectl cp` requires `tar` in the container image, which Keycloak often does not include. Use streaming copy instead:
+
+```bash
+kubectl exec -n keycloak keycloak-0 -- mkdir -p /opt/keycloak/data/crls
+kubectl exec -i -n keycloak keycloak-0 -- sh -c 'cat > /opt/keycloak/data/crls/crl.der' < ./crl.der
+```
+
+Verify:
+
+```bash
+kubectl exec -n keycloak keycloak-0 -- ls -la /opt/keycloak/data/crls
+```
+
+---
+
+## 6) Import `client.pfx` into Browser
+
+Import `client.pfx` from the `pki-crl-test/` directory into browser’s certificate store.
+
+- Export password: empty
+
+---
+
+## 7) Configure the Keycloak X.509 authenticator (critical CRL Path gotcha)
+
+In Keycloak Admin Console:
+
+- Realm: `uds`
+- Authentication → `UDS Authentication`
+- `X509/Validate Username Form` → config (gear icon)
+
+Set:
+
+- **Check certificate validity:** enabled
+- **CRL Checking Enabled:** enabled
+- **CRL abort if non updated:** enabled
+- **OCSP Checking Enabled:** disabled
+- **OCSP Fail-Open Behavior:** disabled
+- **Enable CRL Distribution Point:** disabled
+
+### CRL Path must be relative to `/opt/keycloak/conf`
+
+Keycloak resolves the CRL path under `jboss.server.config.dir` (typically `/opt/keycloak/conf`).
+
+Since the CRL file is at `/opt/keycloak/data/crls/crl.der`, set:
+
+- **CRL Path:** `../data/crls/crl.der`
+
+Do **not** set `/opt/keycloak/data/crls/crl.der` or Keycloak will try to read:
+
+- `/opt/keycloak/conf//opt/keycloak/data/crls/crl.der`
+
+#### Make sure to save the config changes!
+---
+
+## 8) Validate success
+
+Open `https://sso.uds.dev` and select the client certificate.
+
+To watch logs during the attempt:
+
+```bash
+kubectl logs -n keycloak keycloak-0 | grep -i "x509\|crl\|revoc" | tail -n 120
+```
+
+---
+
+## Negative tests (and how to make them reliably take effect)
+
+Keycloak caches CRLs in an internal Infinispan cache keyed by the configured CRL path. In practice, when iterating locally you may need to force Keycloak to re-read the CRL file.
+
+If a negative test doesn’t change behavior immediately:
+
+- Ensure you refreshed the browser page and re-triggered a full auth attempt.
+- Confirm the CRL in the pod is actually updated (stream it back out and inspect locally).
+- **Restart the Keycloak StatefulSet** to force reloading cached CRL data:
+
+```bash
+kubectl -n keycloak rollout restart statefulset keycloak
+```
+
+> It can also be helpful to close and reopen the browser in between tests to ensure things are not cached in the sso.uds.dev session.
+
+### A) Revoked certificate
+
+Revoke and regenerate CRL:
+
+```bash
+cd "$WORKDIR"
+openssl ca -batch -config ca.conf -revoke client.crt
+openssl ca -batch -config ca.conf -gencrl -out crl.pem
+openssl crl -in crl.pem -inform PEM -out crl.der -outform DER
+```
+
+Update the CRL in the pod:
+
+```bash
+kubectl exec -i -n keycloak keycloak-0 -- sh -c 'cat > /opt/keycloak/data/crls/crl.der' < ./crl.der
+```
+
+Restart Keycloak:
+
+```bash
+kubectl -n keycloak rollout restart statefulset keycloak
+```
+
+Attempt login. Expect failure and log messages similar to:
+
+- `Certificate validation's failed. Certificate revoked or incorrect.`
+
+Further you can validate in the Keycloak pod logs that the CRL has been revoked and it failed the Authentication Flow:
+
+```bash
+kubectl logs -n keycloak keycloak-0 | grep -i "Certificate has been revoked,"
+```
+
+### B) Expired CRL (tests `CRL abort if non updated`)
+
+Generate an intentionally expired CRL:
+
+```bash
+cd "$WORKDIR"
+openssl ca -batch -config ca.conf -gencrl -out crl-expired.pem \
+  -crl_lastupdate 20000101000000Z \
+  -crl_nextupdate 20000102000000Z
+
+openssl crl -in crl-expired.pem -inform PEM -out crl-expired.der -outform DER
+```
+
+Update the CRL in the pod and restart Keycloak:
+
+```bash
+kubectl exec -i -n keycloak keycloak-0 -- sh -c 'cat > /opt/keycloak/data/crls/crl.der' < ./crl-expired.der
+kubectl -n keycloak rollout restart statefulset keycloak
+```
+
+Restart Keycloak:
+
+```bash
+kubectl -n keycloak rollout restart statefulset keycloak
+```
+
+Attempt login. Expect failure and log messages similar to:
+
+```bash
+kubectl logs -n keycloak keycloak-0 | grep -i "is not refreshed. Next update is"
+```
+
+- `CRL from '...' is not refreshed. Next update is ...`
+
+### C) Expired client certificate (tests `Check certificate validity`)
+
+Generate an expired client cert and PFX:
+
+```bash
+cd "$WORKDIR"
+openssl genrsa -out client-expired.key 2048
+openssl req -new -key client-expired.key -out client-expired.csr -subj "/CN=UNICORN.TEST.USER.EXPIRED"
+
+openssl ca -batch -config ca.conf -in client-expired.csr -out client-expired.crt -extensions v3_client \
+  -startdate 20000101000000Z \
+  -enddate 20000102000000Z
+
+openssl pkcs12 -export -out client-expired.pfx -inkey client-expired.key -in client-expired.crt -certfile ca.crt -passout pass:
+```
+
+Import `client-expired.pfx` into Firefox and explicitly select it on the login attempt.
+
+Attempt login, expect no x509 login prompts. However in the keycloak logs you should see a message similar to:
+`x509 client certificate is not available for mutual SSL.`
+
+```bash
+kubectl logs -n keycloak keycloak-0 | grep -i "x509 client certificate is not available for mutual SSL."
+```
--- a/docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
+++ b/docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
@@ -182,20 +182,6 @@ If any `Next Update` is in the past, that CRL is expired. Download a fresh copy 
 
 The CRL files need to reach the Kubernetes cluster. Package them as a Kubernetes **Secret** inside a Zarf package to transport them through your airgap pipeline alongside other images and packages.
 
-:::note[If you are not airgapped]
-For local/dev environments, you can skip Zarf packaging and apply the Secret directly:
-
-```bash
-kubectl create secret generic keycloak-crls \
-  --from-file=DODROOTCA3.crl=dod-crls/DODROOTCA3.crl \
-  --from-file=DODSWCA60.crl=dod-crls/DODSWCA60.crl \
-  --from-file=DODIDCA59.crl=dod-crls/DODIDCA59.crl \
-  --namespace keycloak \
-  --dry-run=client \
-  -o yaml | kubectl apply -f -
-```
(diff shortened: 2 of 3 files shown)


docs/dev/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,325 @@
...
+
+```bash
+kubectl logs -n keycloak keycloak-0 | grep -i "x509\|crl\|revoc" | tail -n 120
+```
+
+---
+
+## Negative tests (and how to make them reliably take effect)
+
+Keycloak caches CRLs in an internal Infinispan cache keyed by the configured CRL path. In practice, when iterating locally you may need to force Keycloak to re-read the CRL file.
```
slaskawi: Probably worth adding the detail that using a single replica here is a good idea. Otherwise, the data might be preserved with Infinispan rebalances.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+You do not need to rebuild or customize the `uds-identity-config` image to use CRLs in an airgapped environment.
+:::
+
+---
+
+## Quick Start
+
+1. Get the correct DoD CRL(s) for the certificate chain(s) you expect to see.
+2. Deliver those CRLs to the cluster (for example as a Kubernetes Secret).
+3. Mount them into the Keycloak pod (recommended: `/opt/keycloak/data/crls`).
```
slaskawi: We might want to add an example how to do perform this step.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
+---
+title: Airgap CRL Configuration for Keycloak X.509 Authentication
+---
+
+# Airgap CRL Configuration for Keycloak X.509 Authentication
+
+## Overview
```
slaskawi, line 7: I'm not sure if you agree with me, but I see the Certificate validation (in airgap) as a layered process. 

* The best case is to use OCSP backed by CRLs. With this setup, even if OCSP temporarily goes down, we can rely on CRLs.
* When OCSP is not accessible, we can rely on CRLs. But the real tradeoff here is that we'll always miss the latest revocations. 
* Finally - the worst case, no revocation at all. 

From this perspective, CRLs are complementary to OCSP rather than alternative. If you agree with this statement, we may want to slightly edit this paragraph to reflect that.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
slaskawi: It's actually referenced below. Could we add a cross link here?

slaskawi (commented):
Awesome work! I added a few comments (most of them are nits) and one question about the overview paragraph narrative.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: Updated the overview section to add some additional context.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+
+### DoD PKI CRL Sources
+
+If you do not yet have a certificate but know you need DoD CAC support, the authoritative CRL sources are:
+
+- **Primary:** [https://crl.disa.mil/](https://crl.disa.mil/)
+- **Alternate (GDS):** [https://crl.gds.disa.mil/](https://crl.gds.disa.mil/)
+- **Authoritative CA list:** [https://public.cyber.mil/pki-pke/](https://public.cyber.mil/pki-pke/)
+
+For typical DoD CAC deployments, you will need CRLs for the DoD Root CAs (e.g., `DODROOTCA2.crl` through `DODROOTCA6.crl`) and every SW CA or ID CA that issues end-entity CAC certs in your environment (e.g., `DODSWCA60.crl`, `DODIDCA59.crl`).
```
mjnagel, line 125: Is there any real downside to just using the `ALL CRL ZIP`? That seems like what we probably could/should point people towards?

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+
+---
+
+## Step 1: Identify Which CRLs You Need
+
+Not all CRL files are equal. Keycloak needs the CRL for **every CA in the certificate chain** that will be presented. For DoD CACs, this typically means the **DoD Root CA** CRL and every **DoD Intermediate/Issuing CA** CRL that signed the specific user certificates in your environment.
+
+### Inspect a CAC Certificate
+
+If you have a CAC certificate in PEM format, use `openssl` to list the CRL Distribution Points embedded in each certificate in the chain:
```
mjnagel, line 88: While this is probably useful information I might move this below the next section (DoD PKI CRL Sources) since generally end users would just want to grab all CRLs from those endpoints I would think? This might be more of an "advanced user config/check" if someone wants a subset of CRLs.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+
+```bash
+mkdir -p dod-crls
+
+# Repeat for every CRL URL identified in Step 1
+curl -sSL -o dod-crls/DODROOTCA3.crl  "http://crl.disa.mil/DODROOTCA3.crl"
+curl -sSL -o dod-crls/DODSWCA60.crl   "http://crl.disa.mil/DODSWCA60.crl"
+curl -sSL -o dod-crls/DODIDCA59.crl   "http://crl.disa.mil/DODIDCA59.crl"
+# ... add more as needed
+```
```
mjnagel, line 147: Might also show an example with the all CRL zip?

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+for crl in dod-crls/*.crl; do
+  echo "=== $crl ==="
+  openssl crl -in "$crl" -inform DER -noout -text \
+    | grep -E "Last Update|Next Update"
+done
+```
+
+:::caution
+If any `Next Update` is in the past, that CRL is expired. Download a fresh copy before continuing. Keycloak will reject an expired CRL at runtime and revocation checking will fail.
+:::
```
mjnagel, line 177: Just for my understanding - at what time does this check happen? Mostly trying to think through a couple of situations:
- If I downloaded a CRL, waited a week before transferring it to my airgap (next update date), then pulled it into Keycloak: Seems like this would get rejected?
- If I have a CRL already loaded in Keycloak and it hits the "next update" date: Does this get rejected/not used the next time someone signs in after that date/time?
- If I have a CRL in my cluster and Keycloak restarts after it hits the "next update" date: Does this get rejected?

mjnagel: Mostly just want to make sure we represent the effect and timing considerations well for end users. Looking at DoD CRLs they seem to have "next update" set to a week from today (for me right now), so maybe updates are spaced out a week?

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+  --from-file=DODSWCA60.crl=dod-crls/DODSWCA60.crl \
+  --from-file=DODIDCA59.crl=dod-crls/DODIDCA59.crl \
+  --namespace keycloak \
+  --dry-run=client \
+  -o yaml > keycloak-crls-secret.yaml
+```
+
+:::note[ConfigMap alternative:]
+CRL files are public information by definition, so using a ConfigMap instead of a Secret is also acceptable. Replace `secret` with `configmap` in the command above, and use `configMap:` instead of `secret:` in the volume definition in Step 4.
+:::
```
mjnagel, line 200: Any reason we wouldn't just document as configmap in this case?

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+
+:::note[How Keycloak resolves the CRL path]
+Keycloak's X.509 CRL file loader always resolves the configured `x509-cert-auth.crl-relative-path` as a file under `jboss.server.config.dir` (typically `/opt/keycloak/conf`).
+
+This means paths like `/opt/keycloak/data/crls/DODROOTCA3.crl` will **not** be treated as absolute filesystem paths.
+
+Recommended options:
+
+- Mount CRLs at `/opt/keycloak/data/crls` and set `CRL Path` to `../data/crls/<filename>`.
+- Or mount CRLs under `/opt/keycloak/conf/crls` and set `CRL Path` to `crls/<filename>`.
```
mjnagel, line 251: Less options is probably better here for end users 😄 - any reason to not just recommend `/opt/keycloak/conf/crls` (no `../` needed at config time)?

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+                - name: keycloak-crls
+                  secret:
+                    secretName: keycloak-crls
+                    defaultMode: 0444
+
+            - path: extraVolumeMounts
+              value:
+                - name: keycloak-crls
+                  mountPath: /opt/keycloak/data/crls
+                  readOnly: true
```
mjnagel, line 278: We may want to mention the [default CA volume/mount](https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/values.yaml#L428-L441) here that they would need to copy if they want to preserve that.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+
+---
+
+## Step 6: Verify and Troubleshoot
+
+It is recommended to restart Keycloak after making all these changes to make sure infinispan is reloaded with the new CRLs.
+
+```bash
+kubectl rollout restart statefulset keycloak -n keycloak
+```
```
mjnagel, line 341: No objection to this, just curious if it should be required/recommended? I would think this would be immediately effective similar to other realm settings, but might not know how this works - cc @slaskawi

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+kubectl exec -n keycloak keycloak-0 -- ls -lh /opt/keycloak/data/crls/
+```
+
+### Confirm CRL Files Are Not Expired
+
+The Keycloak container image typically does **not** include `openssl`. Use one of the following approaches:
+
+```bash
+# Stream one CRL locally and inspect it with openssl on your workstation
+kubectl exec -n keycloak keycloak-0 -- cat /opt/keycloak/data/crls/DODROOTCA3.crl > ./DODROOTCA3.crl
```
mjnagel, line 355: nit: [kubectl cp](https://kubernetes.io/docs/reference/kubectl/generated/kubectl_cp/) would be cleaner here

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+
+```bash
+kubectl logs -n keycloak -f keycloak-0 | grep -i "x509\|crl\|revoc"
+```
+
+---
+
+## CRL Renewal and Maintenance
+
+DoD CRLs have a validity window of roughly **24 hours**. After `nextUpdate`, Keycloak rejects the CRL entirely. Treat CRL renewal as a required operational process with a defined cadence, not optional maintenance.
```
mjnagel, line 377: From what I was seeing the DoD CRLs seemed to have a window of 7 days between last and next update?

> After `nextUpdate`, Keycloak rejects the CRL entirely

This is honestly kind of alarming 🫣 - is there any way to change this behavior? Wouldn't this mean you need to effectively update CRLs the instant they are updated "upstream" otherwise you risk failed logins (or fail open on your cert checks)?

mjnagel (commented):
Ended up with a number of comments 🫣. For after this doc, I'm seeing a few weak spots here that we could probably make changes in to improve this experience:
- Nightly? job that builds a DoD CRL Zarf package and publishes it. Might not cover all users, but would cover our main user-base (and the default istio/keycloak config with DoD certs only)
- Default optional mount for ^ CRL package/configmap
- Some templated realm values and/or Tofu for configuring CRL usage

All things to discuss later imo - but wanted to jot these down while this doc is fresh in my mind.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
mjnagel: Bleh - I should read other doc better:
> `kubectl cp` requires `tar` in the container image

resolving.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: Yeah this is my fault, its actually 7 day windows. I'm not seeing anything obvious outside of turning off failures if CRLs are stale, could document that option i suppose. 

implications are:
- You do not need to update the instant upstream publishes a new CRL to avoid outages.
- You need to update before your currently deployed CRL passes nextUpdate.
- However, if you want revocation freshness, you should update more frequently than the max window.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: I might have been using `cp` wrong but that wasn't working for me. kubectl cp relies on tar inside the container, and the Keycloak image often does not have tar.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: In my testing this was required, but i might be doing something wrong or not waiting long enough.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: No strong reasoning, if you prefer that i can do that.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: Keycloak checks nextUpdate during certificate validation (i.e., during an auth attempt), and it can also be affected by Keycloak's internal CRL caching.

- If you download a CRL, wait a week, then transfer and mount it
  - If nextUpdate is now in the past, Keycloak will reject it during auth.
  - Result: logins fail (for certs that require CRL checking) until CRLs are updated.

- If a CRL is already loaded/cached in Keycloak and passes nextUpdate
  - On the next authentication attempt after that time, Keycloak will treat it as stale.
  - With abort if non updated enabled, that becomes an auth failure.

- If the CRL in the cluster is expired and Keycloak restarts
  - After restart, Keycloak will still see an expired CRL at auth time.
  - Result: failures continue until CRLs are refreshed.
  
 You're right that the CRLs actually have a 7 day period. I definitely need to update portion

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: There are a couple real downsides, mostly operational:

1. Keycloak does not “scan a directory” for CRLs.
    - It loads CRLs from the explicit CRL Path list you configure (semicolon-separated list).
    - So this would turn into:
        - unzip
        - decide which files you actually want to reference
        - configure those exact filenames in Keycloak
     - Bigger artifact / more churn
         - More data to move through airgap steps and more files to validate on every refresh.

joelmccoy: hmmm, I wonder if you could make the CRL file a PEM file (instead of DER) and just concat all the things together to make a bundle.... Realistically in DoD environments you are going to want almost ALL the CRLs for each Root CA.  It would be nice if we could just give people a script that creates a single PEM file for the CRL.  I just don't know if Keycloak will iterate through the entries in the PEM file.  Would need to dig into the code there.  Maybe @slaskawi can confirm deny the behavior of CRL loading if we provide a PEM bundle.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
joelmccoy: I would strongly recommend just doing these all as Configmaps.  Creating them as a secret is unnecessary and confusing IMO.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,392 @@
...
+```bash
+kubectl exec -n keycloak -it keycloak-0 -- ls -la /opt/keycloak/data/crls/
+```
+
+## Step 5: Configure the Keycloak X.509 Authenticator
+
+The x509 authenticator must be configured to disable OCSP and enable local CRL file checking. This is done via the **Keycloak Admin Console** and persists in the Keycloak database across pod restarts.
+
+:::note[Why not `realmInitEnv`?]
+The `X509_OCSP_FAIL_OPEN` setting in `realmInitEnv` controls what happens when an OCSP check _fails_; it does not disable OCSP. More critically, `realmInitEnv` values are applied only during the **initial realm import**. If Keycloak is already running, changes there have no effect until the realm is re-imported from scratch. The Admin Console steps below take effect immediately.
```
joelmccoy, line 306: Just for my own understanding.... `X509_OCSP_FAIL_OPEN` is false by default.  And if you deploy in the Airgap CaC/x509 logins will fail automatically because the OSCP server can't be reached?  So anytime you are deploying in the airgap you should disable this?

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
joelmccoy: I'm not sure I'm following this entirely... Are we suggesting that CRLs need to be downloaded and updated every 7 days to be fresh?  Seems super unrealistic IMO based on the current manual process. I think the original goal was to document a manual process for this an maybe make this better in the future?  I would almost lean towards not documenting the manual process because it's an unrealistic ask.  And maybe we could find a way to provide a simple script for users to download CRL data and provide a single simple override/input to UDS core to update CRLs somehow?

We could propose a nightly CI job to show how this process could be automated in environments and bundled into an airgap ready artifact.

joelmccoy (commented):
A couple initial comments/questions on first pass

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: Happy to have @slaskawi confirm, when i looked through the upstream code for this my understanding is that each path gets its own CRLFileLoader, but the limitation is in [CertificateValidator.java](https://github.com/keycloak/keycloak/blob/main/services/src/main/java/org/keycloak/authentication/authenticators/x509/CertificateValidator.java), which calls cf.generateCRL() (singular), not cf.generateCRLs(), meaning a concatenated PEM bundle will only ever load the first CRL and silently ignore the rest. There is also a known bug ([#19918](https://github.com/keycloak/keycloak/issues/19918)) where using multiple ##-separated files from different CAs can cause auth failures entirely.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: If OCSP checking is enabled and OCSP fail-open is false (defualt), then in an airgap (OCSP responder unreachable), X.509 logins will fail because Keycloak can’t complete the revocation check.

If OCSP status can’t be checked:
- fail-open=true => warn + continue
- fail-open=false => throw error + fail auth

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: Yes that is the recommendation if someone is using CRLs, they need to be constantly updated for proper validation. [The original goal](https://github.com/defenseunicorns/uds-core/issues/2335?reload=1?reload=1) was to document how someone could provide keycloak the CRLs so that this validation could still be done in airgapped environments. Based on the outcome of this documentation we were going to decide if we should provide any tools or further improvements. I think you're saying the same thing as Micah said [here](https://github.com/defenseunicorns/uds-core/pull/2393#pullrequestreview-3826295589).

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
mjnagel: Hm, yeah this is all good context. Definitely would agree that ultimately most DoD environments will want need most/all of these CRLs, so I think that should be our target generally and what we advise.

I'm not too concerned about the unzip process and size of things, but obviously having to explicitly list each CRL file is not ideal - would be great if we could either point at a directory or combine them as @joelmccoy mentioned.

Assuming the current restrictions I'm okay documenting this way, but this seems like a good potential area for some upstream contributions to improve the CRL experience when you have multiple CRLs.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
mjnagel: Pending other conversation in https://github.com/defenseunicorns/uds-core/pull/2393/changes#r2828238769 - this probably doesn't make sense right now.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
mjnagel: 😬 ouch - that makes this process pretty painful, and you almost have to fail open in an airgap unless you have an automated process for doing this and follow the upstream update within minutes...

This also is something I don't fully understand - if a CRL has a new update, why does that make the existing one "invalid"/"unusable" (this is really a question of why Keycloak treats it this way). I would think it makes the CRL a bit stale, but not invalid? Would love to poke at this a bit more upstream as well and maybe some ways we could influence changes here too to better support airgap use-cases.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
joelmccoy: Definitely agree with the upstream contribution to support a bundled PEM file here if it doesn't currently work.  The UX on loading CRLs is not really ideal here :/.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
mjnagel: Yeah I think if they are public info (which they are since we download from a public site) we should just document as configmap only. RBAC around Secrets can be stricter, and there's really no need to do that here.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
mjnagel: Gotcha - in that case I would probably change the wording above from `recommended` to `required` or something similar.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
mjnagel: Related comment/thread - https://github.com/defenseunicorns/uds-core/pull/2393/changes#r2829323585 - I think the part that concerns me about this is that regardless of anything we do, we don't control the timing of updated CRLs getting transferred to and updated in their environment (although we could tie it into a Core release cycle). So There is basically guaranteed to be a window of time where the existing CRLs hit their "next update" but an airgapped environment hasn't updated yet.

Imo this is only barely better than just turning on fail open and not using CRLs. You'd likely need fail open to not block everyone logging in...and your CRL checking is only beneficial when you have the latest CRLs.

chance-coleman (commented):
### TL;DR of the review of this PR
CRL support in Keycloak is functional but deeply problematic for airgapped environments. The core issue is that Keycloak treats a CRL past its nextUpdate date as fully invalid, causing auth failures, rather than stale. DoD CRLs refresh daily with a 7-day window, meaning any airgapped environment that can't guarantee timely CRL transfers will inevitably hit outages or be forced to fail-open, which largely defeats the purpose of CRL checking. On top of that, Keycloak requires explicitly listing every CRL file individually, cf.generateCRL() silently drops everything after the first entry in a concatenated PEM bundle, and there's a known bug with multi-CA setups. The current experience is brittle enough that we're not comfortable recommending this as a supported path.
Paths forward:

1. Keep the doc but add a prominent warning, retain it as a public doc with a clear banner up top stating this is not a recommended or well-supported pattern in airgapped environments, and that we're actively working to improve it.
2. Demote to internal/dev-only doc, pull it from public docs entirely and keep it as a reference for our own understanding of current state while we work on a better solution.
3. Upstream contributions, open issues/PRs against Keycloak for: (a) a configurable grace window or "fail-open on stale CRL" toggle separate from the current all-or-nothing behavior, (b) cf.generateCRLs() support so concatenated PEM bundles work properly, (c) directory-based CRL loading so you can point at a folder instead of listing 60+ files, and (d) better logging that distinguishes "CRL stale" from other CRL errors.
4. Tooling/automation, explore a script or nightly CI job that bundles and refreshes CRLs into an airgap-ready artifact, potentially tied into a UDS Core release cycle, to reduce the manual burden until upstream improves.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
slaskawi: Thanks!

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
slaskawi: This needs to be check empirically but it seems there is a way to use multiple CRLs in a single go.

The `X509ClientCertificateAuthenticator` (the thing that validates user certificate) uses a singular path to the CRLs (controlled by `x509-cert-auth.crl-relative-path`) property. This path is later used to construct the `CRLListLoader` that splits the paths using a `##` separator (that why https://github.com/keycloak/keycloak/issues/19918 has been resolved). The `loadCRL` method at the end of its callstack delegates to `CertificateFactory#generateCRL`, which should delegate to BouncyCastle that automatically detects PEM/DER formats. 

So my point is - we may try to load multiple CRLs files with the separator. If this doesn't work, it's probably a bug. We should also be able to use PEM. 

All this is just theoretical - it needs to be checked in a running system with a debugger plugged in.

slaskawi: Actually, this test checks the multiple CRL scenario: https://github.com/keycloak/keycloak/blob/a6bf1944870aeafdc03bca64af10574fedf5fcb0/testsuite/integration-arquillian/tests/base/src/test/java/org/keycloak/testsuite/x509/X509BrowserCRLTest.java#L253

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
slaskawi: You can try to make this more graceful. In Master Realm, in server info there's a button to clear all caches. You can try that to avoid the restart (but on the other hand, this one is definitely faster than click ops).

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
slaskawi: @chance-coleman In the `AbstractX509ClientCertificateAuthenticator`, there's a flag called `x509-cert-auth-crl-abort-if-non-updated`. It seems it controls whether Keycloak rejects a CRLs if it's outdated. Maybe we could use it as 1w DoW update period doesn't fit our release cadence very well.

chance-coleman: I verified that this flag does work for ignoring out of date CRLs. The flag in the UI configuration is called `CRL abort if non updated`.

chance-coleman:
Based on lessons learned from this documentation and research. I'm going to close this PR and open a new one with a new approach geared towards providing scripts and processes for end users to use. This would avoid having to go through all these manual steps, simplifying the process. Will also simplify our lives with the amount of comments and concerns as currently documented. Will link this PR to the new one, once created.

Issue #2341, opened 2026-02-04
rke2 1.34 CI tests flakey


joelmccoy:
### Environment
Device and OS: AWS rke2 1.34
App version: 0.60.1
Kubernetes distro being used: rke2
Other:

### Steps to reproduce
1. Run RKE2 CI with 1.34: https://github.com/defenseunicorns/uds-core/pull/2340

### Expected result
1. All CI tests pass first try

### Actual Result
Experience some flakes: https://github.com/defenseunicorns/uds-core/actions/runs/21685865236/job/62538729739?pr=2340

metrics-server fails to come up

### Additional Context
This failure seems to only happen with rke2 when you bump to 1.34.

mjnagel:
This might require some changes on the image builder side: https://github.com/defenseunicorns/uds-rke2-image-builder

mjnagel:
One of the consistent errors we see on 1.34 is this (from kube-apiserver logs):
```
W0128 13:18:00.393443       1 logging.go:55] [core] [Channel #454 SubChannel #455]grpc: addrConn.createTransport failed to connect to {Addr: "127.0.0.1:2379", ServerName: "127.0.0.1:2379", BalancerAttributes: {"<%!p(pickfirstleaf.managedByPickfirstKeyType={})>": "<%!p(bool=true)>" }}. Err: connection error: desc = "transport: Error while dialing: dial tcp 127.0.0.1:2379: operation was canceled"
```

Which points to etcd connection issues. This has not been present on 1.33 in the past, so this seems to be the main issue, but unsure on cause.

briantwatson:
Confirmed https://github.com/defenseunicorns/uds-core/commit/162e06e28af8ea3a90d1a3edc2de8d02c4995a55 mitigates these flakey tests. Tentatively closing this issue.

PR #2397, opened 2026-02-19
chore: update rke2 get-kubeconfig

briantwatson:
## Description
- Some enhancements to make the this rke2 script more flexible with different version of sed and local troubleshooting
- Moved `**/uds-config.yaml` in nested gitignore to gitignore at root.  Pre-commit check for license will fail on it if file exists.  As it does when testing RKE2 from local machine
...

## Related Issue

Relates to #2341

## Type of change

## Steps to Validate

## Checklist before merging

### DIFF
--- a/.github/test-infra/aws/rke2/scripts/get-kubeconfig.sh
+++ b/.github/test-infra/aws/rke2/scripts/get-kubeconfig.sh
@@ -10,8 +10,9 @@ echo "tofu version: $(tofu --version)"
 # Get required outputs from tofu
 tofu output -raw private_key >key.pem
 chmod 600 key.pem
-trap 'rm -f key.pem' EXIT
-
+if [[ "${CI:-}" == "true" ]]; then # Don't delete the key if not running in CI, since it may be needed for debugging
+  trap 'rm -f key.pem' EXIT
+fi
 bootstrap_ip=$(tofu output -raw bootstrap_ip)
 node_user=$(tofu output -raw node_user)
 cluster_hostname=$(tofu output -raw cluster_hostname)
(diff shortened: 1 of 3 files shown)


.github/test-infra/aws/rke2/scripts/get-kubeconfig.sh
```diff
@@ -1,55 +1,80 @@
...
 
-# find existing host record in the host file and save the line numbers
-matches_in_hosts="$(grep -n $cluster_hostname /etc/hosts | cut -f1 -d:)"
+if [[ "${CI:-}" == "true" ]]; then
+  mkdir -p /home/runner/.kube
+  mv ./rke2-config /home/runner/.kube/config
+  # export KUBECONFIG=$(pwd)/rke2-config
+else
+  mkdir -p ~/.kube
+  mv ./rke2-config ~/.kube/config
```
mjnagel: nit: This will overwrite the entire local (default) kubeconfig. I think that's generally _fine_, but I wonder if this should be configurable and/or just be copied to something like `~/.kube/rke2-config` to not totally replace any existing content? Alternatively probably could combine/append them, but that might be a bit more annoying to do.

.github/test-infra/aws/rke2/scripts/get-kubeconfig.sh
```diff
@@ -1,55 +1,80 @@
...
+  sleep 15
 done
 
+if [[ "$ok" != true ]]; then
+  echo "ERROR: cloud-init did not finish after 20 attempts" >&2
+  exit 1
+fi
+
 # Make sure .kube directory exists
 mkdir -p ~/.kube
```
mjnagel: nit: this isn't necessary anymore I think since each branch for CI/local handles the dir separately?

.github/test-infra/aws/rke2/scripts/get-kubeconfig.sh
mjnagel: I guess technically it's already in `./rke2-config` as well, just need to export `KUBECONFIG` to use it. I think we could just have local instructions to export that env if we wanted to (before running anything to hit the cluster).

mjnagel (dismissed):
Two small nits that you can resolve as you see fit. LGTM.

.github/test-infra/aws/rke2/scripts/get-kubeconfig.sh
briantwatson: thanks for catching--removing

.github/test-infra/aws/rke2/scripts/get-kubeconfig.sh
briantwatson: Agreed, I'm generally not a fan of overwriting and on first pass was following some local troubleshooting docs.  I generally prefer the approach of merging/updating.  I'll test a little with `kubectl config view --merge --flatten` to see if we can achieve that.

briantwatson: Added handling. Tested both with existing kubeconfig with default context and no kubeconfig and worked well.

mjnagel (approved):

chance-coleman (approved):

PR #2398, opened 2026-02-19
feat: add uptime probe support for authservice enabled applications

### LINKED ISSUE #2305: Implement UDS Core Uptime Monitoring

joelmccoy:
## Description

Adds support for Uptime probes for Authservice-enabled applications according to [design doc](https://www.notion.so/defense-unicorns/UDS-Core-Uptime-Monitoring-25ce512f24fc80dea2e0f280704beaab?source=copy_link).

## Related Issue

Fixes #2305 

## Type of change

## Steps to Validate
- `uds run test:uds-core-e2e` - This will deploy and test the test packages which contain authservice enabled apps with uptime probes
  - You can play with removing the uptime probes for the test packages or remove authservice from the package to make sure everything switches back and forth correct.
  - You can verify uptime probes status in the grafana dashboard **UDS / Monitoring / Uptime**

## Checklist before merging

### DIFF
--- a/src/pepr/operator/controllers/keycloak/client-sync.ts
+++ b/src/pepr/operator/controllers/keycloak/client-sync.ts
@@ -51,13 +51,6 @@ export async function keycloak(pkg: UDSPackage) {
     clients.set(client.clientId, client);
   }
 
-  // Purge orphaned clients
-  try {
-    await purgeSSOClients(pkg, [...clients.keys()]);
-  } catch (e) {
-    log.error(e, `Failed to purge orphaned clients in for ${pkg.metadata!.name!}: ${e}`);
-  }
-
   // Purge orphaned SSO secrets
   try {
     await purgeOrphans(generation, pkg.metadata!.namespace!, pkg.metadata!.name!, kind.Secret, log);
--- a/src/pepr/operator/controllers/uptime/config.ts
+++ b/src/pepr/operator/controllers/uptime/config.ts
@@ -1,0 +1,156 @@
+/**
+ * Copyright 2026 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
+import { K8s, kind } from "pepr";
+import { Component, setupLogger } from "../../../logger";
+import { UDSConfig } from "../config/config";
+import { Mutex } from "../utils";
+
+const log = setupLogger(Component.OPERATOR_UPTIME);
+
+// Mutex for serializing reads/writes to the shared blackbox exporter config secret
+const blackboxConfigMutex = new Mutex();
+
+export const BLACKBOX_CONFIG_SECRET_NAME = "uds-prometheus-blackbox-config";
+export const BLACKBOX_CONFIG_NAMESPACE = "monitoring";
+export const BLACKBOX_BASE_CONFIG = {
+  modules: {
+    http_2xx: {
+      prober: "http",
+      timeout: "5s",
+      http: {
+        valid_http_versions: ["HTTP/1.1", "HTTP/2.0"],
+        follow_redirects: true,
+        preferred_ip_protocol: "ip4",
+      },
+    },
+  },
+};
+
+/**
+ * Updates the shared blackbox exporter config secret for a given namespace.
+ * Removes all existing SSO modules scoped to the namespace, regenerates them
+ * from the provided probe client credentials, and sorts all modules by name.
+ * Module names follow the convention: http_200x_sso_<namespace>_<probe_client_id>
+ *
+ * @param namespace The k8s namespace of the UDS Package, used to scope module ownership
+ * @param probeClients Probe client credentials to generate SSO modules from (pass in [] to remove all SSO modules for the namespace)
+ */
+export async function updateBlackboxConfig(
+  namespace: string,
+  probeClients: { clientId: string; secret: string }[],
+): Promise<void> {
+  // Acquire the mutex to ensure atomic read/modify/write of the shared config secret
+  const release = await blackboxConfigMutex.acquire();
+
+  try {
+    // Read and parse the current config secret
+    const secret = await K8s(kind.Secret)
+      .InNamespace(BLACKBOX_CONFIG_NAMESPACE)
+      .Get(BLACKBOX_CONFIG_SECRET_NAME);
+
+    const currentConfig = JSON.parse(atob(secret.data!["blackbox.yaml"])) as {
+      modules: Record<string, unknown>;
+    };
+
+    // Remove all existing SSO modules owned by this namespace
+    const namespacePrefix = `http_200x_sso_${namespace}_`;
+    const modules: Record<string, unknown> = Object.fromEntries(
+      Object.entries(currentConfig.modules).filter(([name]) => !name.startsWith(namespacePrefix)),
+    );
+
+    const tokenUrl = `https://sso.${UDSConfig.domain}/realms/uds/protocol/openid-connect/token`;
+
+    // Regenerate SSO modules for each probe client in this namespace
+    for (const { clientId, secret: clientSecret } of probeClients) {
+      modules[`http_200x_sso_${namespace}_${clientId}`] = {
+        prober: "http",
+        timeout: "5s",
+        http: {
+          valid_http_versions: ["HTTP/1.1", "HTTP/2.0"],
+          preferred_ip_protocol: "ip4",
+          oauth2: {
+            client_id: clientId,
+            client_secret: clientSecret,
+            token_url: tokenUrl,
+            endpoint_params: {
+              grant_type: "client_credentials",
+            },
+          },
+        },
+      };
+    }
+
+    // Sort all modules by name for deterministic output
+    const sortedModules = Object.fromEntries(
+      Object.entries(modules).sort(([a], [b]) => a.localeCompare(b)),
+    );
+
+    // Write the updated config back to the secret
+    await K8s(kind.Secret).Apply(
+      {
+        metadata: {
+          namespace: BLACKBOX_CONFIG_NAMESPACE,
+          name: BLACKBOX_CONFIG_SECRET_NAME,
+        },
+        data: {
+          "blackbox.yaml": btoa(JSON.stringify({ modules: sortedModules })),
+        },
+      },
+      { force: true },
+    );
+
+    log.debug(
+      `Updated blackbox config for namespace ${namespace} with ${probeClients.length} SSO module(s)`,
+    );
+  } finally {
+    release();
+  }
+}
+
+/**
+ * Sets up the blackbox exporter config secret used for uptime monitoring.
+ * If in dev/watch mode, ensures the monitoring namespace exists and creates
+ * the secret if it does not already exist.
+ */
+export async function setupUptimeConfig() {
+  if (process.env.PEPR_WATCH_MODE === "true" || process.env.PEPR_MODE === "dev") {
+    log.info("One-time blackbox exporter config secret initialization");
+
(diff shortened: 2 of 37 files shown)


src/pepr/operator/controllers/uptime/config.ts
```diff
@@ -0,0 +1,156 @@
...
+          preferred_ip_protocol: "ip4",
+          oauth2: {
+            client_id: clientId,
+            client_secret: clientSecret,
+            token_url: tokenUrl,
+            endpoint_params: {
+              grant_type: "client_credentials",
+            },
+          },
+        },
```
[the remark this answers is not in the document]
joelmccoy: no. we want to explicitly NOT allow redirects for Authservice because it can present false positives.

src/pepr/operator/controllers/uptime/config.ts
```diff
@@ -0,0 +1,156 @@
...
+export const BLACKBOX_CONFIG_NAMESPACE = "monitoring";
+export const BLACKBOX_BASE_CONFIG = {
+  modules: {
+    http_2xx: {
+      prober: "http",
+      timeout: "5s",
+      http: {
+        valid_http_versions: ["HTTP/1.1", "HTTP/2.0"],
+        follow_redirects: true,
+        preferred_ip_protocol: "ip4",
```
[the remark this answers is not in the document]
joelmccoy: nah. not necessary.

src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -49,13 +75,17 @@ export async function probe(pkg: UDSPackage, namespace: string): Promise<string[
       probeNames.push(payload.metadata!.name!);
     }
 
+    // Update the shared blackbox config with all SSO modules for this namespace
+    // Note: this will also remove any SSO modules for probes that no longer exist in this namespace
+    await updateBlackboxConfig(namespace, probeClients);
```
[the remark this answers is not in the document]
joelmccoy: Yea this is a good callout (although isn't really causing issues).  The reason i went this way is that we need to update the config all at once (to clear any lingering entries for the whole package/namespace) and we were applying the probes individually in the loops.  But I can collect the probe specs and apply them all at once after the config is applied.  Will push an update for this.

src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -110,3 +141,67 @@ export function generateProbe(
...
+          "introspection.token.claim": "true",
+          "id.token.claim": "false",
+          "lightweight.claim": "false",
+          "userinfo.token.claim": "false",
+        },
+      },
+    ],
+  });
+
+  return { clientId: (client as ClientWithId).clientId, secret: (client as ClientWithId).secret };
```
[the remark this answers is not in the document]
joelmccoy: added

src/pepr/operator/controllers/uptime/probe.ts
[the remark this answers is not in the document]
joelmccoy: added

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -51,13 +51,6 @@ export async function keycloak(pkg: UDSPackage) {
     clients.set(client.clientId, client);
   }
 
-  // Purge orphaned clients
```
joelmccoy, line 54: note: had to move this out to the main package reconciliation as we are now creating keycloak clients in two places

test/vitest/uptime.spec.ts
```diff
@@ -6,9 +6,10 @@
 import * as net from "net";
 import { afterAll, beforeAll, describe, expect, test } from "vitest";
 import { closeForward, getForward } from "./helpers/forward";
+import { pollUntilSuccess } from "./helpers/polling";
 import { queryPrometheusMetric } from "./helpers/prometheus";
 
-describe("Uptime Probes", { retry: 3 }, () => {
+describe("Uptime Probes", { timeout: 180000 }, () => {
```
joelmccoy: Note: its taking a good bit of time before Probes register and metrics become available.  It requires a large timeout even though we deploy the resources in our test package and there is some delay before we hit these e2e tests (deploy test package, playwright tests, vitest e2e).  If we feel this added e2e test time is significant, I'm open to other ideas here.  I do feel it is very important that we test our uptime probes on a Authservice package e2e.

src/prometheus-stack/values/blackbox-exporter-values.yaml
```diff
@@ -1,18 +1,8 @@
...
-      prober: http
-      timeout: 5s
-      http:
-        valid_http_versions:
-          - "HTTP/1.1"
-          - "HTTP/2.0"
-        method: GET
-        follow_redirects: true
+# This secret is created/managed by the UDS Core Operator
+configExistingSecretName: uds-prometheus-blackbox-config
```
mjnagel, line 5: Is there value in calling this out in the docs somewhere? Basically that the operator fully manages the config so end users won't be able to specify custom modules and shouldn't override this configmap?

test/vitest/uptime.spec.ts
mjnagel: I think it's definitely important to test this. I can spin up locally and play with a bit, but did it seem like this was just delays with Prometheus operator setting up the scrape config based on the `Probe` CRs?

Mostly curious if it's the number of probes we're configuring/checking here, or something else. If we were to slim this down to 3-4 checks configured instead of the 10 we have, would that help at all? I think we'd still have coverage of things we care about it we just check Keycloak + Grafana and a single authservice protected endpoint? I don't know that there's functional differences _for the probes_ in checking other non-authservice endpoints or even sidecar vs ambient authservice endpoints.

mjnagel (commented):
Looking good to me overall - would like to install and mess with it a bit locally as well but wanted to leave this very small feedback.

src/prometheus-stack/values/blackbox-exporter-values.yaml
joelmccoy: added a callout in the docs

test/vitest/uptime.spec.ts
joelmccoy: yeah this seems to be the delay from prometheus loading the probes and setting up the scrape config (plus generating the metrics in prometheus).  All the test probes in this test package start showing metrics at the same time.  So I don't think it would be an optimization to reduce the number of checks in this e2e test.

src/pepr/operator/controllers/uptime/config.ts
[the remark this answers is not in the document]
mjnagel: Reading the upstream [config doc](https://github.com/prometheus/blackbox_exporter/blob/master/CONFIGURATION.md) I think `follow_redirects` defaults to true? Does this need an adjustment based on that?

mjnagel, line 82: _Okay_ with this, but it might be nice to have some basic type checking around this config, similar to [what we do with authservice](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/authservice/types.ts). Definitely just a nice to have, but since we have the benefits of a typed language here may as well?

src/pepr/operator/controllers/uptime/probe.ts
```diff
@@ -110,3 +142,73 @@ export function generateProbe(
...
+  const probeClientId = `${sso.clientId}-probe`;
+
+  const client = await credentialsCreateOrUpdate({
+    clientId: probeClientId,
+    name: `${sso.name} Uptime Probe`,
+    serviceAccountsEnabled: true,
+    standardFlowEnabled: false,
+    publicClient: false,
+    redirectUris: [],
+    webOrigins: [],
```
mjnagel: Are these 3 defaults? Could we just leave them off?

test/vitest/uptime.spec.ts
```diff
@@ -19,35 +20,63 @@ describe("Uptime Probes", { retry: 3 }, () => {
...
+      result => result === 1,
+      `probe_success for ${instance}`,
+      180000,
     );
-    expect(metric).toBe(1);
+
+  test("probe_success metric should be 1 for sso.uds.dev", async () => {
+    const result = await expectProbeSuccess("https://sso.uds.dev/");
+    expect(result).toBe(1);
   });
```
mjnagel: A bit of a nit but one cleanup thing for the overall test structure might be making these tests concurrent and use `each`:
```ts
  const blackboxTargets = [
    "https://sso.uds.dev/",
    "https://sso.uds.dev/realms/uds/.well-known/openid-configuration",
    "https://keycloak.admin.uds.dev/",
    "https://grafana.admin.uds.dev/healthz",
    "https://ambient-protected.uds.dev/",
    "https://ambient2-protected.uds.dev/",
    "https://protected.uds.dev/",
    "https://demo.admin.uds.dev/status/200",
    "https://demo-8080.uds.dev/",
    "https://demo-8081.uds.dev/",
  ] as const;

  test.concurrent.each(blackboxTargets)("probe_success metric should be 1 for %s", async url => {
    const result = await expectProbeSuccess(url);
    expect(result).toBe(1);
  });
```

This way we don't have to explicitly add a test case for each endpoint, since they all run the same check. If we did expand what we check here (maybe alerts on cert things, etc) - would make it easier to just modify the one templated test and have it apply for all cases. The `concurrent` doesn't really improve the speed much, but might save a few seconds 🤷

test/vitest/uptime.spec.ts
mjnagel: I tried a couple different things but ultimately couldn't find anything that seemed to work. One thing I noticed is the config reload sidecar seems to operate on a 3 min interval, which is likely why we're seeing this delay. I tried to manually trigger the reload with a reload POST request but that didn't appear to make a difference.

I was having trouble following the code here, but I don't think the [watchInterval](https://github.com/prometheus-operator/prometheus-operator/blob/11a0910a5c1d03835451d5ee8c24ed53a7f7cd7c/cmd/prometheus-config-reloader/main.go#L67) is exposed in a way we can modify through the operator/prometheus CR?

joelmccoy: ditto. couldn't find a way to set this. for now i bumped the test timeout to 3m30s to give some buffer

src/pepr/operator/controllers/uptime/config.ts
joelmccoy: added some types here

test/vitest/uptime.spec.ts
joelmccoy: good call. refactored.

src/pepr/operator/controllers/uptime/config.ts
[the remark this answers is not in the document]
joelmccoy: good catch. explicitly set this to false

src/pepr/operator/controllers/uptime/probe.ts
joelmccoy: removed these

mjnagel (approved):

PR #2404, opened 2026-02-23
chore(docs): document airgapped certificate revocation check process

### LINKED ISSUE #2235: Authservice and Istio should consume the UDS shared trust bundle (including DoD/Public CAs)

chance-coleman:
## Description

Document and script the path for end users to use CRL checking and revocation in airgapped environments.

- Dev doc is meant for internal use mostly due to the complexity of testing these things.
- Main doc for end users
- script for generating an OCI with the CRLs
- small updates to the keycloak values.yaml for documenting the new CRL values

[Corresponding Identity Config PR](https://github.com/defenseunicorns/uds-identity-config/pull/802)

## Related Issue

Fixes #2235 

[Related PR ](https://github.com/defenseunicorns/uds-core/pull/2393)that I closed due to change in scope and large amounts of comments.

## Type of change

## Steps to Validate
- utilize the dev doc to test this. it covers the happy path and also upgrade path.

## Checklist before merging

### DIFF
--- a/docs/dev/keycloak-crl-airgap.md
+++ b/docs/dev/keycloak-crl-airgap.md
@@ -137,7 +137,7 @@ variables:
 
 Reference that in the deployment of slim-dev bundle:
 ```bash
-
+UDS_CONFIG=bundles/k3d-slim-dev/uds-config.yaml uds deploy bundles/k3d-slim-dev/uds-bundle-k3d-core-slim-dev-amd64-*.tar.zst --confirm --no-progress
 ```
 
 ### 1.3) Add keycloak-crls package to bundle
@@ -178,7 +178,7 @@ UDS_CONFIG=bundles/k3d-slim-dev/uds-config.yaml uds deploy bundles/k3d-slim-dev/
 # Configure tenant gateway to trust our demo CA and advertise to browsers
 kubectl -n istio-tenant-gateway patch secret gateway-tls \
   --type merge \
-  -p '{"data": {"cacert": "'"$(base64 -w 0 pki-crl-test/ca.crt)"'"}}'
+  -p '{"data": {"cacert": "'"$(base64 < pki-crl-test/ca.crt | tr -d '\n')"'"}}'
 
 # Restart tenant gateway to apply CA trust changes
 kubectl -n istio-tenant-gateway rollout restart deployment tenant-ingressgateway
@@ -234,7 +234,7 @@ zip -j demo-crls.zip demoCA/demo-ca.crl
 ```bash
 cd <path-to-uds-core-repo>
 
-bash scripts/keycloak-crl-airgap/update-keycloak-crl-oci-volume-package.sh \
+bash scripts/keycloak-crl-airgap/create-keycloak-crl-oci-volume-package.sh \
   --crl-zip "$WORKDIR/demo-crls.zip"
 
 uds zarf package deploy keycloak-crls/zarf-package-keycloak-crls-*.tar.zst --confirm
--- a/docs/dev/keycloak-crl-airgap.md
+++ b/docs/dev/keycloak-crl-airgap.md
@@ -1,0 +1,266 @@
+# Local Testing (Fake CA + CRL) using the CRL OCI Volume Scripts
+
+This guide is for **local/dev testing only**.
+
+---
+
+## 1) Create Fake CA and CRL
+
+> **Note:** Copy and paste the following snippet into a bash shell
+
+```bash
+# Setup working directory for PKI materials
+WORKDIR="${WORKDIR:-$(pwd)/pki-crl-test}"
+mkdir -p "$WORKDIR"
+cd "$WORKDIR"
+
+# Initialize OpenSSL CA database structure
+mkdir -p demoCA/newcerts
+: > demoCA/index.txt
+echo 1000 > demoCA/serial
+echo 1000 > demoCA/crlnumber
+
+# Create OpenSSL configuration for CA and certificate generation
+cat > ca.conf <<'EOF'
+[ ca ]
+default_ca = CA_default
+
+[ CA_default ]
+dir = .
+database = $dir/demoCA/index.txt
+new_certs_dir = $dir/demoCA/newcerts
+certificate = $dir/ca.crt
+serial = $dir/demoCA/serial
+private_key = $dir/ca.key
+crlnumber = $dir/demoCA/crlnumber
+crl = $dir/demoCA/demo-ca.crl
+default_md = sha256
+default_days = 365
+default_crl_days = 30
+unique_subject = no
+policy = policy_any
+copy_extensions = copy
+
+[ policy_any ]
+commonName = supplied
+
+[ v3_client ]
+subjectAltName = @alt_names
+keyUsage = digitalSignature
+extendedKeyUsage = clientAuth, 2.16.840.1.101.3.6.8
+certificatePolicies = @policy_anything
+
+[ alt_names ]
+# UPN (otherName) used by the UDS X.509 flow
+otherName.0 = 2.16.840.1.101.3.6.6;UTF8:test@mil
+email = unicorn.test@uds.dev
+
+# Include a policy OID that UDS accepts (must match Common.REQUIRED_CERT_POLICIES)
+[ policy_anything ]
+policyIdentifier = 2.16.840.1.101.3.2.1.3.12
+EOF
+
+# Generate Certificate Authority (CA) key and certificate
+openssl genrsa -out ca.key 2048
+openssl req -x509 -new -nodes -key ca.key -sha256 -days 3650 -out ca.crt \
+  -subj "/CN=UDS Scripted CRL Demo CA" \
+  -addext "basicConstraints=critical,CA:TRUE" \
+  -addext "keyUsage=critical,keyCertSign,cRLSign"
+
+# Generate client certificate key and signing request
+openssl genrsa -out client.key 2048
+openssl req -new -key client.key -out client.csr -subj "/CN=UNICORN.TEST.USER.1234567890"
+openssl ca -batch -config ca.conf -in client.csr -out client.crt -extensions v3_client
+
+# Create PKCS12 bundle for browser import (empty password)
+openssl pkcs12 -export -out client.pfx -inkey client.key -in client.crt -certfile ca.crt -passout pass:
+
+# Generate Certificate Revocation List (CRL) in PEM (keep as .crl)
+openssl ca -batch -config ca.conf -gencrl -out demoCA/demo-ca.crl
+
+# Verify locally before packaging (this is critical)
+openssl crl -in demoCA/demo-ca.crl -noout -text >/dev/null
+
+# Package CRL into ZIP for OCI volume creation
+zip -j demo-crls.zip demoCA/demo-ca.crl
+
+# Create Zarf OCI volume package containing the CRL
+cd ../
+bash scripts/keycloak-crl-airgap/create-keycloak-crl-oci-volume-package.sh \
+  --crl-zip "pki-crl-test/demo-crls.zip"
+```
+
+### 1.1) Configure Slim Dev Bundle
+
+Copy the Keycloak CRL Path from the generated crl paths text file into the uds-bundle.yaml:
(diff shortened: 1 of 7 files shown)


scripts/keycloak-crl-airgap/create-keycloak-crl-oci-volume-package.sh
```diff
@@ -0,0 +1,108 @@
...
+    CRL_PATH_LIST="${CRL_PATH_LIST}##${entry}"
+  fi
+done <<< "$crl_list"
+printf '%s\n' "$CRL_PATH_LIST" > "$paths_file"
+
+# Build data image
+cat > "$tmp/Dockerfile" <<'EOF'
+FROM scratch
+COPY stage/ /
+EOF
```
mjnagel: nit: this could go either way, but seems like we could just commit this to the same folder here?

scripts/keycloak-crl-airgap/create-keycloak-crl-oci-volume-package.sh
```diff
@@ -0,0 +1,108 @@
...
+metadata:
+  name: keycloak-crls
+  description: "DoD CRLs for Keycloak X.509 (OCI volume)"
+  version: "${PACKAGE_VERSION}"
+components:
+  - name: keycloak-crls
+    required: true
+    images:
+      - ${IMAGE_REF}
+EOF
```
mjnagel: similar comment here - any reason to not just commit this?

docs/dev/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,266 @@
...
+### 1.2) Create UDS Slim Dev Bundle Config
+
+Create uds-config.yaml:
+```yaml
+variables:
+  uds-k3d-dev:
+    k3d_extra_args: >-
+      --k3s-arg --kube-apiserver-arg=feature-gates=ImageVolume=true@server:0
+      --k3s-arg --kubelet-arg=feature-gates=ImageVolume=true@server:0
+```
```
mjnagel: Alternative here I think would be to export `UDS_K3D_EXTRA_ARGS` with those flags or `UDS_K3D_IMAGE` set to a 1.35+ image?

(no need to change necessarily, just might be slightly simpler?)

docs/dev/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,266 @@
...
+
+Reference that in the deployment of slim-dev bundle:
+```bash
+UDS_CONFIG=bundles/k3d-slim-dev/uds-config.yaml uds deploy bundles/k3d-slim-dev/uds-bundle-k3d-core-slim-dev-amd64-*.tar.zst --confirm --no-progress
+```
```
mjnagel: I think this is out of place? Need 1.3/1.4 first and then the create in 2 before we deploy the bundle?

docs/dev/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,266 @@
...
+kubectl -n istio-tenant-gateway rollout status deployment tenant-ingressgateway
+
+# Configure Keycloak to trust our demo CA for client cert validation
+kubectl -n keycloak create configmap uds-trust-bundle \
+  --from-file=ca-bundle.pem=pki-crl-test/ca.crt \
+  --dry-run=client -o yaml | kubectl apply -f -
+
+# Restart Keycloak to apply trust bundle changes
+kubectl -n keycloak rollout restart statefulset keycloak
+kubectl -n keycloak rollout status statefulset keycloak
```
mjnagel: We should be able to pass all of this into the bundle I would think? Since it's just a dev doc this is fine, but thinking about simplification we probably could have this in the bundle at deploy time.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,286 @@
+---
+title: Keycloak Airgap CRLs
+---
+
+## Overview
+
+In connected environments, Keycloak can use **OCSP (Online Certificate Status Protocol)** to check whether a client certificate has been revoked. In a true air-gap, OCSP responders are unreachable, so you must either:
+
+* disable revocation checks (not recommended), or
+* rely on **CRLs (Certificate Revocation Lists)** that you download *before* entering the air-gap and load locally into Keycloak.
```
mjnagel: Just going to comment on these two but we should be using `airgap` rather than `air-gap`.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,286 @@
...
+  - name: keycloak-crls
+    path: ./keycloak-crls/zarf-package-keycloak-crls-<arch>-<tag>.tar.zst
+    ref: x.x.x
+  - name: core-identity-authorization
+    ref: x.x.x
+```
+
+### Configure UDS policy exemptions (if required)
+
+If ImageVolume is denied by policy, add an exemption targeting Keycloak pods:
```
mjnagel: I think the `If` here is a bit strange since Core will deny it by default. I also opened [CORE-389](https://linear.app/defense-unicorns/issue/CORE-389/support-kubernetes-imagevolume-in-pepr-pod-policies) to address this for the future.


```suggestion
ImageVolumes are currently blocked by UDS Policies (future support will be added), so add an exemption targeting Keycloak pods:
```

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,286 @@
...
+
+If using k3s/k3d on a Kubernetes version that requires explicit enablement, create a `uds-config.yaml`:
+
+```yaml
+variables:
+  uds-k3d-dev:
+    k3d_extra_args: >-
+      --k3s-arg --kube-apiserver-arg=feature-gates=ImageVolume=true@server:0
+      --k3s-arg --kubelet-arg=feature-gates=ImageVolume=true@server:0
+```
```
mjnagel: This is definitely broader than just k3s/k3d. I might even suggest bringing this up to the top of the doc (right under the overview). We need to call out that you need to be on 1.31+ with the feature gate enabled or 1.35+ (where the feature gate is enabled by default).

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,286 @@
...
+            exemptions:
+              - policies:
+                  - RestrictVolumeTypes
+                matcher:
+                  namespace: keycloak
+                  name: "^keycloak-.*"
+                  kind: pod
+                title: "Allow Keycloak ImageVolume for CRLs"
+                description: "Allow Keycloak pods to mount CRLs via Kubernetes ImageVolume (OCI-backed)."
+```
```
mjnagel, line 214: Any reason to not just merge this with the other bundle config above?

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,286 @@
...
+          X509_CRL_RELATIVE_PATH: "<paste keycloak-crl-paths.txt contents here>"
+      - path: extraVolumes
+        value:
+          - name: ca-certs
+            configMap:
+              name: uds-trust-bundle
+              optional: true
+          - name: keycloak-crls
+            image:
+              reference: 127.0.0.1:31999/library/keycloak-crls:local # Common Zarf registry address; adjust for your environment
```
mjnagel: This should be able to point to the "upstream" image name and rely on Zarf to mutate it to the proper image (in the zarf registry).

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: My intention here was to separate them out so it was clearer what each section of the bundle config was doing. I have slight preference for that clarity but if you prefer i can combine those together and just have one giant copy paste section. That could include a few sections and result in just one very large yaml block.

docs/dev/keycloak-crl-airgap.md
chance-coleman: removed this, intended to be an example of how to do it but not necessary when to do it, but its a duplicate of the actual deploy command in step 2.

docs/dev/keycloak-crl-airgap.md
chance-coleman: i left this as is from an implementation perspective, although how its done is changed because of trying to move away from all the kubectl commands and implement everything in the bundle. so now there is a quick bash command to generate the config file with the base64 enc cacert things.

docs/dev/keycloak-crl-airgap.md
chance-coleman: utilizing the bundle now rather than individual commands, agreed its just a quicker and easier way to do this.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
chance-coleman: add a call out at the top of the doc to make sure this is clear.

docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md
```diff
@@ -0,0 +1,297 @@
...
+                  namespace: keycloak
+                  name: "^keycloak-.*"
+                  kind: pod
+                title: "Allow Keycloak ImageVolume for CRLs"
+                description: "Allow Keycloak pods to mount CRLs via Kubernetes ImageVolume (OCI-backed)."
+```
+
+### Enable ImageVolume on k3s/k3d clusters
+
+If running on k3s/k3d with Kubernetes < 1.35, enable the feature gate via `uds-config.yaml`:
```
mjnagel:
```suggestion
### Enable ImageVolume on uds-k3d

If running on the dev/demo bundles (uds-k3d) with Kubernetes < 1.35, enable the feature gate via `uds-config.yaml`:
```

I think this is a bit more specific/accurate since these values are only going to be picked up by uds-k3d (not generically k3s/k3d).

mjnagel (dismissed):
LGTM overall - with the ongoing work on docs I'm not sure if we might want to make some other adjustments here (move to how-to guides and use mdx + Steps)? But okay deferring that to the other tickets where we plan to address those sections.

mjnagel (approved):

PR #2412, opened 2026-02-25
fix: pull advancedHttp schema from istio upstream

### LINKED ISSUE #2354: Synchronize CRD sources with upstream

Several CRDs in UDS Core are sourced or parts are sourced from upstream CRDs and currently are not directly linked or synchronized. 

To avoid drift we should establish a process to synchronize them.  This could involve a script that sources the relevant pieces from upstream YAML and transforms to match our schemas.

Notably, we need to sync the components of `advancedHttp` in the [istio/virtualservice-v1beta.ts](https://github.com/defenseunicorns/uds-core/blob/a48c45a41663bbb919cb6dc026a493cb9817ab54/src/pepr/operator/crd/sources/istio/virtualservice-v1beta1.ts#L33)

Which is a subset of the [schema here](https://github.com/istio/istio/blob/e08bc95516b344ba3b7d887d66378015691326e3/manifests/charts/base/files/crd-all.gen.yaml#L10598-L11352)

Currently some fields are out place and this work should address these, e.g. `weight`:

```
13s         Warning   ReconciliationFailed   package/grafana   failed to create typed patch object (grafana/grafana-admin-grafana-80-grafana; networking.istio.io/v1beta1, Kind=VirtualService): .spec.http[0].weight: field not declared in schema
```

### Current source files

- https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/crd/sources/istio/virtualservice-v1beta1.ts
- https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/crd/sources/package/v1alpha1.ts
- https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/crd/sources/exemption/v1alpha1.ts
- https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/crd/sources/cluster-config/v1alpha1.ts

### Triggering mechanism
Consider a UDS task that runs custom script and then renovate the istio source file + add the "generated" file to our [autogen checks](https://github.com/defenseunicorns/uds-core/blob/main/.github/actions/autogenerated-check/action.yaml) in CI.

### References to upstream components
- corsPolicy: [https://istio.io/latest/docs/reference/config/networking/virtual-service/#CorsPolicy](https://istio.io/latest/docs/reference/config/networking/virtual-service/#CorsPolicy)
- directResponse: [https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPDirectResponse](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPDirectResponse)
- headers: [https://istio.io/latest/docs/reference/config/networking/virtual-service/#Headers](https://istio.io/latest/docs/reference/config/networking/virtual-service/#Headers)
- match: [https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPMatchRequest](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPMatchRequest)
- rewrite: [https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRewrite](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRewrite)
- redirect: [https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRedirect](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRedirect)
- retries: [https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRetry](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRetry)
- weight: sub-field on [https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRouteDestination](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRouteDestination)
- timeout: [https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRoute-timeout](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRoute-timeout)

mjnagel:
## Description

This PR adds a new script to generate the istio `source` CRD sub-schema from the upstream CRD manifest. That script is called before other CRD generation scripts to ensure we re-generate our sub-schema first.

There are a number of diffs on the actual schema that mostly amount to:
1. More strict typing (and validations) around sub-properties
2. Additional/new fields that hadn't been added
3. Fixes for fields that were out of place (`weight`)
4. Updates to descriptions

The script currently strips just the "top level" properties that we had before, but does bring them in with their entire sub-schema. The one exception is `match` where we additional strip only a subset of the fields (to keep parity with what we had done before).

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/2354

## Type of change

## Steps to Validate

You can run the new task with `uds run -f src/pepr/tasks.yaml gen-crds`.

Reviewing this PR is probably easiest approached by reviewing a subset of the files:
- The script: `src/pepr/scripts/gen-istio-crd-sources.ts`
- The task changes: `src/pepr/tasks.yaml`
- The Package CRD: `src/pepr/uds-cluster-crds/templates/packages.uds.dev.yaml` (YAML) and `src/pepr/operator/crd/generated/package-v1alpha1.ts` (generated types)

This will help to get the best sense of the diffs while ignoring the many generated files that are connected to the above.

## Checklist before merging


src/pepr/operator/crd/generated/package-v1alpha1.ts
```diff
@@ -366,10 +366,6 @@ export interface AdvancedHTTP {
    * Timeout for HTTP requests, default is disabled.
    */
   timeout?: string;
-  /**
-   * Weight specifies the relative proportion of traffic to be forwarded to the destination.
-   */
-  weight?: number;
```
mjnagel, line 372: This is a field that was misplaced previously (and using it would've caused errors during reconcile).

src/pepr/scripts/gen-istio-crd-sources.ts
```diff
@@ -0,0 +1,156 @@
...
+const EXCLUDED_MATCH_FIELDS = [
+  "gateways",
+  "headers",
+  "port",
+  "scheme",
+  "sourceLabels",
+  "sourceNamespace",
+  "statPrefix",
+  "withoutHeaders",
+];
```
mjnagel, line 43: The previous istio VS source excluded all of these fields from the `match` schema. While I think there could be some utility in exposing a few of these, generally we want to have a simpler interface in our Package CR and for now I wanted to take an approach of trying to aim for minimal diffs where possible. This also can help us stay a bit less tied to Istio VS schema if (in some distant future) we move to the gateway api or a different ingress approach.

Alternatively we could include all or some of these if there are strong arguments for them, but I might advocate for evaluating those on a case-by-case basis later.

renovate.json
```diff
@@ -91,12 +96,6 @@
       "groupName": "iac-support-deps",
       "commitMessageTopic": "iac support dependencies"
     },
-    {
-      "matchFileNames": ["package.json", "package-lock.json"],
-      "groupName": "operator-deps",
-      "commitMessageTopic": "operator-deps",
-      "matchDepNames": ["!pepr"]
-    },
```
mjnagel, line 99: Deleted while I was in here because this is an "unreachable" match - the following match for `support-deps` always matches the same files and takes precedence.

src/pepr/scripts/gen-istio-crd-sources.ts
```diff
@@ -0,0 +1,157 @@
...
+    d =>
+      d &&
+      typeof d === "object" &&
+      d.kind === "CustomResourceDefinition" &&
+      d.metadata?.name === "virtualservices.networking.istio.io",
+  );
+  if (!vs) throw new Error("VirtualService CRD not found in Istio bundle");
+
+  // Grab the VirtualService v1beta1 OpenAPI schema
+  const vsVersion = vs.spec?.versions?.find(v => v?.name === "v1beta1");
```
joelmccoy: very nit: if we ever wanted to update this version of the CRD it would be nice to have this declared at the top as a const variable and reused everywhere in this script to keep consistent

```ts
const VS_CRD_VERSION = "v1beta" 
```

joelmccoy (commented):
LGTM! Tested locally and resulted in no diff.

src/pepr/scripts/gen-istio-crd-sources.ts
mjnagel: Good call - adjusted this.

joelmccoy (approved):

PR #2413, opened 2026-02-26
chore: add gen-crds task for istio

mjnagel:
## Description

While working on https://github.com/defenseunicorns/uds-core/pull/2412, I noticed/remembered that we don't have a task tracking the origin of our istio CRDs. This is a bit more messy than the prometheus one since Istio stores all their CRDs together in a single file and we only need a subset of the types.

As far as I could tell `DestinationRule` is not used in our code so I also deleted that one.

## Related Issue

Related to https://github.com/defenseunicorns/uds-core/issues/2354

## Type of change

## Steps to Validate

Should just be primarily new fields/stricter schema/changed descriptions. Assuming all testing passes should be pretty straightforward.

## Checklist before merging


joelmccoy (approved):
LGTM!

Issue #1569, opened 2025-05-15
Add common labels for resources


noahpb:
### Description
For resources that are deployed as part of uds-core, supply a set of common labels that denote their belonging to uds-core. Additional labels should be included as well, such as the version of the particular release of core. 

These labels should be included on all namespaces that are deployed with uds-core at a minimum. Resources from helm charts managed directly in uds-core and from upstream in our zarf packages should also be considered.

### Additional context
BigBang common labels approach for [reference](https://repo1.dso.mil/big-bang/bigbang/-/blob/master/chart/templates/_helpers.tpl?ref_type=heads#L152-158).

PR #2426, opened 2026-03-03
feat: core uptime recording rules and dashboard

briantwatson:
## Description
Adds boolean "up" recording rules for Core components based on recent uptime design discussions 

### Shared helpers (prometheus-stack)

Added three generic recording rules that centralize health logic:

- **`daemonset:up`** — 1 if healthy, 1 if degraded due to node scaling event (2m window), 0 otherwise
- **`deployment:up`** — 1 if at least one replica available, 0 otherwise
- **`statefulset:up`** — 1 if at least one replica available, 0 otherwise
- **`daemonset:availability_ratio`** — ready/desired ratio for dashboards

All use `max by(type, namespace)` to strip extra labels (uid, instance, job) and guarantee one series per resource.

### Consumer pattern

Every consumer now uses the same unified pattern:

```promql
(max(TYPE:up{...}) or on() vector(0))
and on()
(kube_state_metrics:up == 1)
```

- **`max()`** collapses to `{}` labels — ensures a single time series per rule
- **`or on() vector(0)`** — when the resource doesn't exist, the shared rule produces no matching series, `max()` returns nothing, and `vector(0)` fills in `0` (down)
- **`and on() (kube_state_metrics:up == 1)`** — gates all results on KSM health; produces **no data** when KSM is down instead of a false `0`

### Behavior

| KSM | Resource | Result |
|-----|----------|--------|
| down/absent | any | no data |
| up | healthy | 1 (up) |
| up | degraded + scaling (daemonsets only) | 1 (forgiven) |
| up | unhealthy | 0 (down) |
| up | absent | 0 (down) |

### Removed after PR review

- All `absent()` fallback patterns — replaced by `vector(0)` to prevent split series for cleaner dashboard

...

## Related Issue

Related to INFRA-27 and INFRA-28

## Type of change

## Steps to Validate
Built locally and confirmed new recording rules present and dashboards load

## Checklist before merging


src/prometheus-stack/values/kube-prometheus-stack-values.yaml
```diff
@@ -121,6 +121,146 @@ alertmanager:
...
+          - record: uds:keycloak_endpoint:up
+            labels:
+              uds_layer: identity-authorization
+              uds_component: keycloak_endpoint
+            expr: min(probe_success{job="probe/keycloak/uds-sso-tenant-uptime"})
+          - record: uds:keycloak_admin_endpoint:up
+            labels:
+              uds_layer: identity-authorization
+              uds_component: keycloak_admin_endpoint
+            expr: min(probe_success{job="probe/keycloak/uds-keycloak-admin-uptime"})
```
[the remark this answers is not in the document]
briantwatson: Nah, while these and others are valid concerns I think should be handled in the alerts rather than recording rules.  No data is accurate here and otherwise would lead to false alarms. As an example,
 - Fire when `absent(uds:core:up)` persists beyond an expected scrape window
  (catching genuinely missing data)
  - Fire when `uds:core:up == 0` (catching actual downtime)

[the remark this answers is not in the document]
briantwatson: See https://github.com/defenseunicorns/uds-core/pull/2426#discussion_r2880620635

src/prometheus-stack/values/kube-prometheus-stack-values.yaml
```diff
@@ -121,6 +121,146 @@ alertmanager:
...
+            expr: clamp_max(kube_deployment_status_replicas_available{deployment="velero", namespace="velero"}, 1)
+  uds-core-uptime:
+    groups:
+      - name: uds-core-uptime
+        rules:
+          - record: uds:core:up
+            labels:
+              uds_layer: core
+              uds_component: core
+            expr: uds:keycloak_endpoint:up * on() uds:keycloak:up * on() uds:keycloak_waypoint:up * on() uds:tenant_ingressgateway:up * on() uds:ztunnel:up
```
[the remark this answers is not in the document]
briantwatson: See https://github.com/defenseunicorns/uds-core/pull/2426#discussion_r2880620635

src/prometheus-stack/chart/dashboards/uptime-core.json
```diff
@@ -0,0 +1,560 @@
...
+  "templating": {
+    "list": []
+  },
+  "time": {
+    "from": "now-3h",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "browser",
+  "title": "UDS / Monitoring / UDS Core Uptime",
```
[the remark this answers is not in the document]
briantwatson: I think it's alright here, although don't feel strongly either way

src/prometheus-stack/values/kube-prometheus-stack-values.yaml
```diff
@@ -121,6 +121,146 @@ alertmanager:
       zarf.dev/connect-name: alertmanager
     annotations:
       zarf.dev/connect-description: "Directly connect to the Alertmanager HTTP service"
+## UDS Core uptime recording rules
+additionalPrometheusRulesMap:
```
mjnagel: Going to check each of these rules and the dashboard changes but one comment I wanted to ask upfront - would it make more sense to use PrometheusRule CRs rather than values passed to the monitoring chart here? One concern I have is functional layer deployments which would be missing some of these components ("intentionally"). Since these rules all live in the monitoring layer, they'd still all get deployed. If instead we put the rules as CRs in the individual charts for falco, loki, ..., etc - they'd be "closer to the code" and only deploy if their components are present.

src/prometheus-stack/values/kube-prometheus-stack-values.yaml
mjnagel: So while I think this is "generally accurate" that the most important metric is Keycloak/Tenant gateway access (and ztunnel for networking) - I think each environment may have a slightly different interpretation of "core up". Maybe this is still fine as a baseline (and can be overridden in helm values), but wanted to bring it up for discussion.

src/prometheus-stack/chart/dashboards/uptime-core.json
```diff
@@ -0,0 +1,560 @@
...
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "dark-red",
+                "value": 0
+              },
+              {
+                "color": "dark-green",
+                "value": 99.9
+              }
```
mjnagel, line 240: Is it worth adding a step between red and green here (or multiple)? A target uptime % can vary widely from environment to environment and 99.9% + the default 3hour time range on this dashboard would likely mean ever slight blips drop into red. I do generally like the target for green here since UDS Core is and should be highly stable - just wondering if there's room for something in between 0 and 99.9 😅

src/prometheus-stack/chart/dashboards/uptime-core.json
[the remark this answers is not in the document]
mjnagel: I might slightly agree with copilot on this one, although I wouldn't say I feel strongly either.

src/prometheus-stack/chart/dashboards/uptime-probe.json
```diff
@@ -363,7 +366,7 @@
   },
   "timepicker": {},
   "timezone": "browser",
-  "title": "UDS / Monitoring / Uptime",
+  "title": "UDS / Monitoring / Probe Uptime",
```
mjnagel, line 369: We'll probably need a connected PR to adjust the doc for this on the `legacy-docs` branch ([here](https://github.com/defenseunicorns/uds-core/blob/legacy-docs/docs/reference/configuration/observability/blackbox-exporter.md#grafana-dashboard)) and probably also under the [old-reference folder](https://github.com/defenseunicorns/uds-core/blob/main/docs/dev/old-reference/configuration/observability/blackbox-exporter.md#grafana-dashboard).

src/prometheus-stack/chart/dashboards/uptime-core.json
mjnagel: Actually - after deploying I think I misinterpreted where this color was going. Going to resolve this one 🫣

src/prometheus-stack/values/kube-prometheus-stack-values.yaml
briantwatson: I _think_ and can test what this would look like.  I anticipate it would be 'No data' for those rules.  Would it make sense to extend this line of thought throughout our monitoring layer?   e.g. when monitoring is installed should dashboards only reflect the layers installed?  If we were to do separate PrometheusRule CRs tyring to think a pattern for the scenario where monitoring layer isn't present.  I think the PrometheusRule would break on helm install if the Prom stack isn't there?

I did try to group these by functional layer, and can more clearly do that on the dashboard too.  Let me know what direction you'd like me to investigate.

src/prometheus-stack/chart/dashboards/uptime-core.json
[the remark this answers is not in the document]
briantwatson: Works for me!  I was waiting for an opinion and will adjust 😄

src/prometheus-stack/values/kube-prometheus-stack-values.yaml
mjnagel: > If we were to do separate PrometheusRule CRs tyring to think a pattern for the scenario where monitoring layer isn't present. I think the PrometheusRule would break on helm install if the Prom stack isn't there?

We cheat a bit and the actual CRDs are part of the base layer so they will always be installable - although Helm also provides a handy trick here too to conditionally deploy templates based on "capabilities" in the cluster ([example](https://github.com/defenseunicorns/uds-core/blob/0c3e01b57e9698bcc6ca34d5ba6e1b8ccd5f8d1a/src/metrics-server/chart/templates/service-monitor.yaml#L4)).

mjnagel: I guess mostly just want to make sure we think carefully through the UX around functional layers and that it works as expected? Would want to make sure we don't have extra processing for rules that "shouldn't really exist" and/or strange displays in dashboards that show things are down when they aren't expected to exist.

src/prometheus-stack/chart/dashboards/uptime-probe.json
briantwatson: Thanks for catching -- draft PR here: https://github.com/defenseunicorns/uds-core/pull/2428

(I'll stage doc change for `main` on this PR)

src/prometheus-stack/values/kube-prometheus-stack-values.yaml
briantwatson: Totally makes sense and thank you for example link.  My comments basically were headed in the 'do we have an established pattern for handling this' direction. I'll do a bit more testing with separate rules and see what the dashboard UX is as well.

mjnagel: We also follow similar approaches already in a few spots:
- Dashboards for Keycloak: https://github.com/defenseunicorns/uds-core/blob/0c3e01b57e9698bcc6ca34d5ba6e1b8ccd5f8d1a/src/keycloak/chart/templates/keycloak-dashboards.yaml
- PrometheusRule for Keycloak logs: https://github.com/defenseunicorns/uds-core/blob/0c3e01b57e9698bcc6ca34d5ba6e1b8ccd5f8d1a/src/keycloak/chart/templates/keycloak-logging-alerting-rules.yaml
- Metrics-server servicemonitor: https://github.com/defenseunicorns/uds-core/blob/0c3e01b57e9698bcc6ca34d5ba6e1b8ccd5f8d1a/src/metrics-server/chart/templates/service-monitor.yaml

I think it's generally a good pattern if we can to keep these things deployed around when the workloads are to avoid any weirdness with missing layers and/or strange data due to timing of deploys? But emphasis on the `if we can` - we previously put an istio podmonitor in the [monitoring chart](https://github.com/defenseunicorns/uds-core/blob/0c3e01b57e9698bcc6ca34d5ba6e1b8ccd5f8d1a/src/prometheus-stack/chart/templates/istio-monitor.yaml) due to some CRD timing (although I think we could possibly move it now).

joelmccoy: Definitely would just advocate for putting these as `PrometheusRules` in ~~the `uds-prometheus-config` chart [here](https://github.com/defenseunicorns/uds-core/tree/main/src/prometheus-stack/chart/templates).~~ each of the components helm charts.

edit: missed a part of this thread.  I was focused on using raw templated `PrometheusRules`.  But it is definitely a good practice to include these next to the applications and what I would recommend

joelmccoy: updated/edited the above comment.  we should definitely put these as separate `PrometheusRules` next to each application.  This is the common best practice when shipping helm charts with prometheus CRs.

src/prometheus-stack/values/kube-prometheus-stack-values.yaml
briantwatson: I think we could simplify the composite to just the endpoint metric and ztunnel?  With the endpoint being more e2e, some of the others are redundant?

src/vector/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,21 @@
...
+    prometheus: kube-prometheus-stack-prometheus
+spec:
+  groups:
+    - name: uds-logging-vector-uptime
+      rules:
+        - record: uds:vector:up
+          labels:
+            uds_layer: logging
+            uds_component: vector
+          expr: kube_daemonset_status_number_ready{daemonset="vector", namespace="vector"} >= bool kube_daemonset_status_desired_number_scheduled{daemonset="vector", namespace="vector"}
```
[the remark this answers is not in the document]
briantwatson: Testing this and other daemonset changes below.

src/prometheus-stack/values/kube-prometheus-stack-values.yaml
briantwatson: Implemented and pushed up

src/vector/chart/templates/uptime-recording-rules.yaml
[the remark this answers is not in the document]
briantwatson: Took a slight different approach than proposed above to handle scenario daemonset can't schedule (selector constraints), but also has any unavailable.  If the daemsonset is not there shows as No data.  Here i patched vector with a nodeSelector so scheduable is zero (metric showing down), then after a period deleted the daemonset (no data)

image

src/prometheus-stack/values/kube-prometheus-stack-values.yaml
briantwatson: Also noting I adjusted dashboard so it only shows rules that are present, for example here I only installed monitoring layer on top of base:
image

src/vector/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,21 @@
...
+    prometheus: kube-prometheus-stack-prometheus
+spec:
+  groups:
+    - name: uds-logging-vector-uptime
+      rules:
+        - record: uds:vector:up
+          labels:
+            uds_layer: logging
+            uds_component: vector
+          expr: clamp_max(kube_daemonset_status_desired_number_scheduled{daemonset="vector", namespace="vector"}, 1) * on (daemonset, namespace) (kube_daemonset_status_number_unavailable{daemonset="vector", namespace="vector"} == bool 0)
```
joelmccoy: I think this is a good way to do this for daemonsets... but we are inevitably going to get some downtime for vector when we add a new node to the cluster based on this expression (will be waiting for the pod to be come ready on the new node).  I think this is ok?  Or will it mess with our overall metrics?

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,38 @@
...
+            uds_layer: monitoring
+            uds_component: blackbox_exporter
+          expr: clamp_max(kube_deployment_status_replicas_available{deployment="prometheus-blackbox-exporter", namespace="monitoring"}, 1)
+    - name: uds-core-uptime
+      rules:
+        - record: uds:core:up
+          labels:
+            uds_layer: core
+            uds_component: core
+          expr: uds:keycloak_endpoint:up * on() uds:ztunnel:up
```
joelmccoy: curious, are we leaving waypoint and tenant gateway off of this because the keycloak endpoint probe should technically capture the full endpoint path? or was there some other reasoning here? i think the design doc calls out these components as critical to uds core uptime?

src/istio/common/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - record: uds:admin_ingressgateway:up
+          labels:
+            uds_layer: base
+            uds_component: admin_ingressgateway
+          expr: clamp_max(kube_deployment_status_replicas_available{deployment="admin-ingressgateway", namespace="istio-admin-gateway"}, 1)
+        - record: uds:ztunnel:up
+          labels:
+            uds_layer: base
+            uds_component: ztunnel
+          expr: clamp_max(kube_daemonset_status_desired_number_scheduled{daemonset="ztunnel", namespace="istio-system"}, 1) * on (daemonset, namespace) (kube_daemonset_status_number_unavailable{daemonset="ztunnel", namespace="istio-system"} == bool 0)
```
joelmccoy: similar comment on daemonsets here (reference vector comment).  The ztunnel metric here is a bit critical here. I imagine if we add a new node to the cluster it will report down for this metric for a bit and then uds core total uptime will report down.  I wonder if there is a good way to lax our daemonset checks

src/istio/common/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - record: uds:istiod:up
+          labels:
+            uds_layer: base
+            uds_component: istiod
+          expr: clamp_max(kube_deployment_status_replicas_available{deployment="istiod", namespace="istio-system"}, 1)
+        - record: uds:istio_cni:up
+          labels:
+            uds_layer: base
+            uds_component: istio_cni
+          expr: clamp_max(kube_daemonset_status_desired_number_scheduled{daemonset="istio-cni-node", namespace="istio-system"}, 1) * on (daemonset, namespace) (kube_daemonset_status_number_unavailable{daemonset="istio-cni-node", namespace="istio-system"} == bool 0)
```
joelmccoy: similar comment to other daemonset comments i left

src/falco/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,26 @@
...
+    prometheus: kube-prometheus-stack-prometheus
+spec:
+  groups:
+    - name: uds-runtime-security-uptime
+      rules:
+        - record: uds:falco:up
+          labels:
+            uds_layer: runtime-security
+            uds_component: falco
+          expr: clamp_max(kube_daemonset_status_desired_number_scheduled{daemonset="falco", namespace="falco"}, 1) * on (daemonset, namespace) (kube_daemonset_status_number_unavailable{daemonset="falco", namespace="falco"} == bool 0)
```
joelmccoy: similar comment to other daemonsets

src/grafana/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,21 @@
...
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: uds-grafana-uptime
+  namespace: {{ .Release.Namespace }}
+  labels:
+    prometheus: kube-prometheus-stack-prometheus
+spec:
+  groups:
+    - name: uds-monitoring-grafana-uptime
```
joelmccoy: do we want to add the grafana endpoint probes as metrics too like we do for keyloak?

src/loki/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,21 @@
...
+    prometheus: kube-prometheus-stack-prometheus
+spec:
+  groups:
+    - name: uds-logging-loki-uptime
+      rules:
+        - record: uds:loki:up
+          labels:
+            uds_layer: logging
+            uds_component: loki
+          expr: clamp_max(kube_statefulset_status_replicas_available{statefulset="loki", namespace="loki"}, 1)
```
joelmccoy: this statefulset doesn't exist. we have `loki-backend` and `loki-write` which we should check for.  Also... we should probably update these checks to fail if it doesn't exist 😅 

I think we may need to add something like
```
absent(kube_statefulset_status_replicas{statefulset="loki", namespace="loki"}) * 0
  OR
clamp_max(kube_statefulset_status_replicas_available{statefulset="loki", namespace="loki"}, 1)
```
There may be a better way to do that, but AI just suggested that method.

We will need to update this everywhere because if a resource doesn't exist we want to count that as downtime (i.e. if I go into a cluster and delete a statefulset we will still report up based on this recording)

joelmccoy: To follow up we need to check:
- loki-backend statefulset
- loki-write statefulset
- loki-gateway deployment
- loki-read deployment

joelmccoy: It looks like loki is missing from the dashboard because of this mismatch. This rule returns no data and we need it to return a 0

src/prometheus-stack/chart/dashboards/uptime-core.json
joelmccoy, line 1: Right now the dashboard as i view it locally is missing:
- pepr (i can't pinpoint why though?)
- loki (wrong recording rules)

Image

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,38 @@
...
+            uds_component: alertmanager
+          expr: clamp_max(kube_statefulset_status_replicas_available{statefulset="alertmanager-kube-prometheus-stack-alertmanager", namespace="monitoring"}, 1)
+        - record: uds:blackbox_exporter:up
+          labels:
+            uds_layer: monitoring
+            uds_component: blackbox_exporter
+          expr: clamp_max(kube_deployment_status_replicas_available{deployment="prometheus-blackbox-exporter", namespace="monitoring"}, 1)
+    - name: uds-core-uptime
+      rules:
+        - record: uds:core:up
```
joelmccoy: I also wonder if there is a better name for this.  If i pretended to be a user who didn't know about this feature, i would assume core would reflect all of UDS Core.  AI gave me some alternative names: 

  That's an access/availability signal, not a "core health" signal. Some better options:

  - uds:access:up — direct, describes what it measures (user access works)
  - uds:ingress_auth:up — more specific: the ingress + authentication chain
  - uds:customer_access:up — matches the design doc's "Customer-Impacting" framing
  - uds:platform_access:up — similar but less tied to "customer" terminology

My pick would be uds:access:up — it's concise, accurate, and clearly distinct from the per-component uds:<component>:up metrics. It answers
  "can users access the platform?" without implying it represents the health of every Core subsystem.

src/velero/tasks.yaml
joelmccoy, line 1: commenting on this file because there is no other place -- We should add e2e tests for all of the recording metrics to prevent regressions here.  It should be fairly simple to check all of these metrics reporting up and we can probably just add that here: https://github.com/defenseunicorns/uds-core/blob/main/test/vitest/uptime.spec.ts

joelmccoy (commented):
Looking like the right direction here!

A couple overarching things to look at from my comments:
1. How do we want to handle daemonset recording? I called out the concerns of adding new nodes will report down until the pods come up. I don't have the perfect answer here and open to discussion
2. Make sure our recording rules return 0 if the metric doesn't return anything
3. Add e2e tests for all of these metrics to prevent regressions

src/vector/chart/templates/uptime-recording-rules.yaml
briantwatson: Yes, in that scenario I think we will have a period of the said daemonset being 'down'.  @mjnagel and I had discussed this and with daemonsets it's a bit trickier, but in the scenario a pod is missing on a node critical workloads _could_ be missed.  We essentially came up with this:

```
kube_daemonset_status_number_ready{daemonset="ds_name", namespace="ns"}
<
kube_daemonset_status_desired_number_scheduled{daemonset="ds_name", namespace="ns"}
```

After mulling feedback from Copilot review I had not considered the scenario of node taints etc where scheduled might compute to zero and we have a `0=0` scenario that leads to a false up.  I tried to capture that with what I implemented here.  I certainly think there is some inbetween here of 'degraded', but my understanding was we're going for a single binary metric.  Open to discussing changing this though.

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
briantwatson: Yeah, after @mjnagel comment above I made this change:  https://github.com/defenseunicorns/uds-core/pull/2426#discussion_r2886463235.

src/istio/common/chart/templates/uptime-recording-rules.yaml
briantwatson: Thanks, Going to resolve this and we can address path forward for all daemonsets in comments above.

src/falco/chart/templates/uptime-recording-rules.yaml
briantwatson: Thanks, Going to resolve this and we can address path forward for all daemonsets in comments above.

src/grafana/chart/templates/uptime-recording-rules.yaml
briantwatson: We didn't have that in our design document, but I don't see why not 😄 !  Adding it!

src/istio/common/chart/templates/uptime-recording-rules.yaml
briantwatson: Good callout, definitely more relevant even with this daemonset.  Don't have a good idea off the top of my head--I'll think on it

src/loki/chart/templates/uptime-recording-rules.yaml
briantwatson: > this statefulset doesn't exist. we have `loki-backend` and `loki-write` which we should check for. 

🤦 great catch, thanks.  Adding back in

> Also... we should probably update these checks to fail if it doesn't exist 😅

Co-pilot review called this out similarly and I disagreed here: https://github.com/defenseunicorns/uds-core/pull/2426#discussion_r2880620635

My personal preference is for us to treat `recording_rules` as we have here as just capturing/synthesizing the monitoring data. If there is no data, then we capture that as "no data".  _Then_ we use alerts to build of these for the absence scenario.  I believe measuring down when the metric is absent could also lead to false positives when certain metrics are unavailable.

Let me know what you think, happy to discuss further.

src/prometheus-stack/chart/dashboards/uptime-core.json
briantwatson: Ohhh, good catch on pepr too.  I'm thinking because the check for the capability check and it not being installed yet?

https://github.com/defenseunicorns/uds-core/blob/74ae9ce0ffbcdf6488f74257f099e4b7ff3509d3/src/pepr/uds-operator-config/templates/uptime-recording-rules.yaml#L4

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
briantwatson: Sounds good, I like it. Changing to `uds:access:up`!

src/vector/chart/templates/uptime-recording-rules.yaml
joelmccoy: I think my worry is that single binary metric with this setup will affect our SLA reporting.  Especially for the ztunnel daemonset we will report the Core End User SLA metric as down when a new node comes up when the end user path isn't really affected in that scenario.  I will need to think if there is a more sane way to do this but definitely open to other ideas/opinions here

src/loki/chart/templates/uptime-recording-rules.yaml
joelmccoy: I don't think I like that approach mainly because I can go into the cluster and delete a loki statefulset and our recording rules for loki up will not reflect downtime (and the dashboard will miss it too like it did in this case).

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
joelmccoy: hmmmmm. I think you could argue that the `keycloak_endpoint:up` metric captures everything e2e then right? Is there a reason why we left Ztunnel on?  And then at what point is the metric `uds:core:up` providing any useful aggregation?

Would love to maybe talk through this synchronously some more or get some thoughts from @mjnagel on this given ^

src/prometheus-stack/chart/dashboards/uptime-core.json
joelmccoy: hmmm but if i am tracing things correctly pepr should apply after the CRDS are loaded.... but maybe there is a delay and maybe we need to add some zarf wait actions to make sure they are truly ready??

briantwatson: Perhaps yes, I moved that template to the prometheus chart and see the PrometheusRule and looks like everything is there: 
image

src/loki/chart/templates/uptime-recording-rules.yaml
briantwatson: That's a good point about entire removal.  I think my concern is the scenario metrics are absent we report a false positive of down for a service. I've seen scenarios where kube-state-metrics is unhealthy.  If we incorporate a check for the kube-state-metrics job with the absent would that work?

something like (haven't tested):

```
(absent(kube_statefulset_status_replicas{statefulset="loki", namespace="loki"}) and on()
    (max(up{job="kube-state-metrics"}) == 1))
 * 0
  OR
clamp_max(kube_statefulset_status_replicas_available{statefulset="loki", namespace="loki"}, 1)
```

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
briantwatson: I think that's definitely valid and I'm not opposed to the single metric--I was going off our initial design notes.  A discussion would be good and I think I'd be deferring to you all on this one.

src/prometheus-stack/chart/dashboards/uptime-core.json
briantwatson: Dug a little further--I think the issue is currently they're placed in the operator-crds which is installed before prometheus crds here?

https://github.com/defenseunicorns/uds-core/blob/ff33062374d707925582e72ecf6ec20b1bb7ddde/packages/standard/zarf.yaml#L27-L36

and here in base:

https://github.com/defenseunicorns/uds-core/blob/ff33062374d707925582e72ecf6ec20b1bb7ddde/packages/base/zarf.yaml#L26-L35

I think we can just reorder these?  The operator config looks relevant just to the pepr-core component after the prometheus CRD component.

src/loki/chart/templates/uptime-recording-rules.yaml
joelmccoy: Failing closed makes more sense to me - if kube-state-metrics is down, Loki's availability is genuinely unknown and we shouldn't report it as up. A separate alert for kube-state-metrics being down gives on-call the context to distinguish a real outage from a metrics pipeline issue. That said, I'm not opposed to adding the `up{job="kube-state-metrics"}` guard if the team prefers to avoid the noise - happy to go either way.

src/authservice/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,21 @@
...
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: uds-authservice-uptime
+  namespace: {{ .Release.Namespace }}
+  labels:
+    prometheus: kube-prometheus-stack-prometheus
+spec:
+  groups:
+    - name: uds-authservice-uptime
```
mjnagel: nits on these names: A number follow `uds-<layer>-<component>-uptime`, but then:
- Keycloak/Falco/Promstack/Velero: just use layer
- Authservice: just component

Might be nice to keep this consistent.

src/authservice/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,21 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+{{- if .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" }}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: uds-authservice-uptime
+  namespace: {{ .Release.Namespace }}
+  labels:
+    prometheus: kube-prometheus-stack-prometheus
```
mjnagel: naive question - are these labels used/useful somewhere? I don't believe they're required for Prometheus/Operator to pick up the rules, but wasn't sure if they are useful somewhere I'm not seeing.

Definitely don't feel strongly so feel free to leave, just was thinking about this somewhat in relation to https://github.com/defenseunicorns/uds-core/issues/1569 - I think I tend to prefer identifying labels that would group with the layer/component rather than generic "prometheus" (which can be assumed by the CRD itself).

src/loki/chart/templates/uptime-recording-rules.yaml
mjnagel: > A separate alert for kube-state-metrics being down gives on-call the context to distinguish a real outage from a metrics pipeline issue.

I agree with this although I think the risk of showing everything as down (without the guard on `kube-state-metrics`) would be that you see a flood of everything showing down, when really it's just that our metrics pipeline isn't working.

I tend to like the idea of being able to differentiate downtime of Loki vs downtime of Kube-state-metrics. I think we really want to capture 3 states in most cases?
- up
- down
- unknown/no data (because kube-state-metrics / blackbox / ? down)

If I'm following this thread and thinking right that would mean adding `up{job="kube-state-metrics"})` to all of these? Down would definitely be an alert, but unknown probably wouldn't be?

src/prometheus-stack/chart/dashboards/uptime-core.json
mjnagel: Yeah shouldn't be any harm in reordering I don't think.

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
mjnagel: I think the rename here at least mostly addresses my concern in https://github.com/defenseunicorns/uds-core/pull/2426#discussion_r2884142264. although we may want to similarly adjust the Dashboard section for this which says `UDS Core Uptime` (similarly to access and could even go with available/unavailable instead of up/down?)

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
mjnagel: My comment was intended to be more in line with Joel's comment [here](https://github.com/defenseunicorns/uds-core/pull/2426/changes#r2892182319) (just poorly phrased) - basically critique on the name `core:up` given that could have different interpretation by end users.

Definitely agreed that we _could_ boil this down to just Keycloak endpoint - and if we want to do that I don't feel strongly one way or the other? That's ultimately the most important thing, and the other pod checks were mostly supporting pieces to help catch anything that contributes to the endpoint being down. Trying to think through if we get any benefits from the composite metric here if it will/should always surface in endpoint downtime...

After reading the other thread about daemonsets I think I might lean towards at least excluding ztunnel here unless we have a good way to handle the potential flap there with nodes.

src/vector/chart/templates/uptime-recording-rules.yaml
mjnagel: Great callout on node cycling...this starts to get into a more interpretive realm of "does the downtime matter". Ztunnel not being up when a brand new node is schedulable is technically "downtime" but only affects the pods on that node, which is likely very few if its a brand new node. There could be some complex ways we augment this metrics with `kube_node_` metrics (status, created time, etc) to effectively allow more "grace" when a node is coming up? I don't know if that's really the right answer here though and could start to get a bit messy.

src/loki/chart/templates/uptime-recording-rules.yaml
joelmccoy: Fair enough. Based on that, I think what Brian is suggesting makes sense!

src/velero/tasks.yaml
briantwatson: Added unit tests

src/prometheus-stack/chart/dashboards/uptime-core.json
briantwatson: Reordered

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
briantwatson: I think the benefits of the composite metric is if we decide to change what "access" means, we can change the definition of the rule and the other consumers (e.g. dashboards) don't need updating.

I've updated to "access" and have changed it to the single keycloak endpoint metric

src/istio/common/chart/templates/uptime-recording-rules.yaml
briantwatson: I've created several helper rules to consolidate the handling of different workload types.  For daemonsets, I've added a check for if the number of desired scheduled has changed for past 2 minutes to account for a node changing.  Welcome thoughts on that duration.

```
            kube_daemonset_status_desired_number_scheduled
                !=
                kube_daemonset_status_desired_number_scheduled offset 2m
```

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
briantwatson: Adjusted panel title/label as well

src/vector/chart/templates/uptime-recording-rules.yaml
briantwatson: Latest I've got up includes a 2m grace period for daemonsets if the number of desired scheduled changes.  I agree about the interpretive aspect of this potentially getting messy.  I wonder if it's worth defining some sort of in-between degraded state for daemonsets.  

https://github.com/defenseunicorns/uds-core/blob/95a7c339c81632169941fa0183abb7ed9add15ee/src/prometheus-stack/chart/templates/uptime-recording-rules.yaml#L23-L25

I also added an additional rule for daemonset availability ratio that might be helpful.  It's currently not used by any of the workload specific rules.

https://github.com/defenseunicorns/uds-core/blob/95a7c339c81632169941fa0183abb7ed9add15ee/src/prometheus-stack/chart/templates/uptime-recording-rules.yaml#L37-L41

src/authservice/chart/templates/uptime-recording-rules.yaml
briantwatson: removed (not used anywhere)

briantwatson:
@mjnagel @joelmccoy I think I've addressed all feedback and am ready for another pass. I added some details to description to help summarize the latest.  I opted for a different approach than using `absent()` that still captures the missing workload scenario as I was seeing duplicate rows (series) on the dashboard.

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,79 @@
...
+            and on()
+            (kube_state_metrics:up == 1)
+
+    - name: uds-identity-authorization-uptime
+      rules:
+        - record: uds:access:up
+          labels:
+            uds_layer: core
+            uds_component: core_access
+          expr: uds:keycloak_endpoint:up
```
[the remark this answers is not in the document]
joelmccoy: this is fine. KSM does not generate this metric

src/prometheus-stack/chart/dashboards/uptime-core.json
```diff
@@ -0,0 +1,340 @@
...
+  },
+  "time": {
+    "from": "now-3h",
+    "to": "now"
+  },
+  "timepicker": {},
+  "timezone": "browser",
+  "title": "UDS / Monitoring / Core Uptime",
+  "uid": "uds-uptime-status",
+  "version": 1
```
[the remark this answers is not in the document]
joelmccoy: I think it is fine to leave this off for now.  Will capture in new docs.

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,79 @@
...
+            (
+              (max by(daemonset, namespace)(kube_daemonset_status_number_unavailable) == 0) * 0 + 1
+              or on(daemonset, namespace)
+              clamp_max(
+                kube_daemonset_status_desired_number_scheduled
+                !=
+                kube_daemonset_status_desired_number_scheduled offset 2m
+              , 1)
+              or on(daemonset, namespace)
+              clamp_max(max by(daemonset, namespace)(kube_daemonset_status_desired_number_scheduled), 1) * 0
```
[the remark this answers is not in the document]
joelmccoy: agree this comment. should probably catch this edgecase

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,79 @@
...
+          expr: clamp_max(max by(deployment, namespace)(kube_deployment_status_replicas_available), 1)
+
+        - record: statefulset:up
+          expr: clamp_max(max by(statefulset, namespace)(kube_statefulset_status_replicas_available), 1)
+
+        - record: daemonset:availability_ratio
+          expr: |
+            max by(daemonset, namespace)(kube_daemonset_status_number_ready)
+            /
+            max by(daemonset, namespace)(kube_daemonset_status_desired_number_scheduled)
```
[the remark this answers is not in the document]
joelmccoy: agreed with this comment and should probably capture this edge case.  I prefer to `return 0 when desired is 0`

test/vitest/uptime.spec.ts
```diff
@@ -46,3 +46,67 @@ describe("Uptime Probes", { timeout: 210000 }, () => {
...
+  afterAll(async () => {
+    await closeForward(prometheusProxy.server);
+  });
+
+  const expectRecordingRuleUp = (metric: string) =>
+    pollUntilSuccess(
+      () => queryPrometheusMetric(prometheusProxy.url, metric),
+      result => result === 1,
+      `${metric} == 1`,
+      200000,
```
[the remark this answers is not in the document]
joelmccoy: eh. I think it is fine to leave this off. our metrics are pretty standardly named here

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+    - name: uds-monitoring-uptime
+      rules:
+        - record: uds:prometheus:up
```
joelmccoy, line 49: It looks like this comment didn't get submitted with my latest review :/ 

what about the deployments for:
- kube-prometheus-stack-kube-state-metrics
- kube-prometheus-stack-operator

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,79 @@
...
+          labels:
+            uds_layer: monitoring
+            uds_component: prometheus
+          expr: |
+            (max(statefulset:up{statefulset="prometheus-kube-prometheus-stack-prometheus", namespace="monitoring"}) or on() vector(0))
+            and on()
+            (kube_state_metrics:up == 1)
```
joelmccoy, line 56: nit: is there a reason why you left this logic out of the helper functions? it seems we repeat this logic for all our statefulset/daemonset checks and wondered if this could be DRYer? or if there was limitations of putting this in the helper?

joelmccoy (commented):
Love the refactor into helper functions.  Generally looking good minus a couple small things

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
briantwatson: Thanks! also adding node exporter daemonset after looking further

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
briantwatson: From what I've found it needs to be outside so we at least have two metrics in the rule (with the and condition) to properly return "no data" in the scenario KSM is down.  Otherwise the `vector(0)` catch for absence to return 0 would always return zero (down) when KSM is down.

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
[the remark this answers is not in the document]
briantwatson: Looked a few options and I think clamping the min of the denominator to 1 be simplest so we're always dividing by 1.

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,108 @@
...
+                max by(daemonset, namespace)(kube_daemonset_status_number_unavailable) == 0
+                and on(daemonset, namespace)
+                max by(daemonset, namespace)(kube_daemonset_status_desired_number_scheduled) > 0
+              ) * 0 + 1
+              or on(daemonset, namespace)
+              clamp_max(
+                kube_daemonset_status_desired_number_scheduled
+                !=
+                kube_daemonset_status_desired_number_scheduled offset 2m
+              , 1)
```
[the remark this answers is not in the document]
briantwatson: Will adjust this

docs/dev/old-reference/configuration/observability/blackbox-exporter.md
```diff
@@ -190,4 +190,4 @@ UDS Core includes an uptime monitoring dashboard that displays:
 - Percentage uptime over the selected time period
 - TLS certificate expiration dates
 
-Access it via Grafana under **UDS / Monitoring / Uptime**.
+Access it via Grafana under **UDS / Monitoring / Probe Uptime**.
```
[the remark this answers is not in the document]
briantwatson: This is a good point, however I'm unclear how much we want to get into detail into the Core uptime metrics in documentation?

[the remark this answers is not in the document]
briantwatson: renamed blackbox-exporter.md to uptime.md and made minimal adjustments to broaden scope of doc and detail these metrics

src/authservice/chart/templates/uptime-recording-rules.yaml
briantwatson: Updated these for consistency

docs/how-to-guides/monitoring-observability/overview.md
```diff
@@ -18,5 +18,5 @@ Relevant existing docs to draw from:
 - `src/content/docs-old/reference/configuration/observability/monitoring-metrics.md`
 - `src/content/docs-old/reference/configuration/observability/metrics-alerting.md`
 - `src/content/docs-old/reference/configuration/observability/dashboards.md`
-- `src/content/docs-old/reference/configuration/observability/blackbox-exporter.md`
+- `src/content/docs-old/reference/configuration/observability/uptime.md`
 - `src/content/docs-old/reference/configuration/observability/alert-management.md`
```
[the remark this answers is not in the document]
briantwatson: I think we're good here

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,108 @@
...
+            and on()
+            (kube_state_metrics:up == 1)
+
+    - name: uds-identity-authorization-uptime
+      rules:
+        - record: uds:access:up
+          labels:
+            uds_layer: core
+            uds_component: core-access
+          expr: uds:keycloak_endpoint:up
```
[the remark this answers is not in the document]
briantwatson: KSM isn't relevant to the exporter probe

joelmccoy (approved):

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,108 @@
...
+          expr: clamp_max(max by(deployment, namespace)(kube_deployment_status_replicas_available), 1)
+
+        - record: statefulset:up
+          expr: clamp_max(max by(statefulset, namespace)(kube_statefulset_status_replicas_available), 1)
+
+        - record: daemonset:availability_ratio
+          expr: |
+            max by(daemonset, namespace)(kube_daemonset_status_number_ready)
+            /
+            clamp_min(max by(daemonset, namespace)(kube_daemonset_status_desired_number_scheduled), 1)
```
mjnagel, line 45: nit: not seeing this actually used anywhere but might be mistaken?

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
```diff
@@ -0,0 +1,108 @@
...
+              ) * 0 + 1
+              or on(daemonset, namespace)
+              clamp_max(
+                max by(daemonset, namespace)(kube_daemonset_status_desired_number_scheduled)
+                !=
+                max by(daemonset, namespace)(kube_daemonset_status_desired_number_scheduled offset 2m)
+              , 1)
+              or on(daemonset, namespace)
+              clamp_max(max by(daemonset, namespace)(kube_daemonset_status_desired_number_scheduled), 1) * 0
+            )
```
mjnagel, line 33: Simplifying this... it's effectively `IF healthy OR "recent replica count change" THEN "up" ELSE "down"`. The `OR "recent replica count change"` means if we're scaling up/down then it returns healthy/up regardless of actual replica count. I know that's somewhat intentional here to not cause alert/dashboard flapping during scaling of nodes, but do want to call out this could mean a ~2min window where we lose visibility into the "actual status" during nodes rolling. I think that's pretty okay and acceptable though - especially since most daemonsets (maybe minus Ztunnel) don't strictly need 100% uptime.

mjnagel (approved):
LGTM overall - no real changes requested here just two small comments.

src/prometheus-stack/chart/templates/uptime-recording-rules.yaml
briantwatson: Yeah, I thought this might be helpful for human operators if we're looking at periods of 'down' on the daemonset:up.  Just a quick way to provide more context.  Also understand YAGNI here and happy to drop it.

PR #2427, opened 2026-03-03
chore(docs): add concepts/platform section to new docs site

joelmccoy:
## Description

Add content to the concepts/platform section of the new docs site. Replaces stub .md pages with full .mdx content covering supported distributions, environments, platform vs application layer, and flavors. Adds a CI lint task to error if the documented Kubernetes version is not sync with test infrastructure.

## Related Issue

Fixes Core-99
Fixes Core-79

## Type of change

## Steps to Validate
- `uds run dev-docs` to validate docs
- run `uds run lint:k8s-version-check` and verify it passes and change the version in supported-distributions.mdx to force a failure)

## Checklist before merging


tasks/lint.yaml
```diff
@@ -76,3 +76,42 @@ tasks:
...
+        shell:
+          darwin: bash
+          linux: bash
+        cmd: |
+          DOC_FILE="docs/concepts/platform/supported-distributions.mdx"
+
+          # Extract k8s versions from terraform config
+          EKS_VERSION=$(grep -A3 'variable "kubernetes_version"' .github/test-infra/aws/eks/variables.tf | sed -n 's/.*default *= *"\([0-9]*\.[0-9]*\)".*/\1/p')
+          AKS_VERSION=$(grep -A3 'variable "kubernetes_version"' .github/test-infra/azure/aks/variables.tf | sed -n 's/.*default *= *"\([0-9]*\.[0-9]*\)".*/\1/p')
+          RKE2_VERSION=$(sed -n 's/.*rke2_version *= *"\([0-9]*\.[0-9]*\)".*/\1/p' .github/test-infra/aws/rke2/terraform.tfvars)
```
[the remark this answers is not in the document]
joelmccoy: adjusted the script

tasks/lint.yaml
```diff
@@ -76,3 +76,42 @@ tasks:
...
+
+          # Extract k8s versions from the doc table (bold versions like **1.34**)
+          DOC_VERSIONS=$(sed -n 's/.*\*\*[^0-9]*\([0-9][0-9]*\.[0-9][0-9]*\)\*\*.*/\1/p' "$DOC_FILE" | sort -u)
+          DOC_VERSION_COUNT=$(echo "$DOC_VERSIONS" | wc -l | tr -d ' ')
+
```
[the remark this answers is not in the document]
joelmccoy: adjusted the script

docs/concepts/platform/environments.mdx
```diff
@@ -0,0 +1,37 @@
...
+> [!TIP]
+> For local development, Defense Unicorns publishes two pre-built bundles: **`k3d-core-slim-dev`** (Base + Identity & Authorization — lightweight, fast startup) and **`k3d-core-demo`** (Full Core — full-fidelity local environment). Both use the `upstream` flavor.
+
+## What Varies Between Environments
+
+Cluster-level configuration is the primary dimension that changes across environments:
+
+- **Cluster identity** — name and tags
+- **Networking** — API server CIDR, node CIDRs
+- **TLS** — custom CA certificates
```
mjnagel: These two both feel a bit weird to me. Networking might change, but Operator should generally handle that for you. TLS/certs are definitely a real change but maybe could be combined with domains below?

docs/concepts/platform/environments.mdx
```diff
@@ -0,0 +1,37 @@
...
+- **Domains** — tenant and admin domains
+- **External integrations** — database endpoints for Keycloak/Grafana HA, external object storage for Loki/Velero
+
+## What Stays the Same
+
+Across every environment tier you deploy the **same packages at the same version**, the **same policy baseline** (Pepr policies, Istio authorization), and the **same observability stack** (Prometheus, Loki, Grafana). This consistency makes staging-to-production promotion predictable — if it works in staging with production-like config, it will work in production.
```
mjnagel: > This consistency makes staging-to-production promotion predictable — if it works in staging with production-like config, it will work in production.

Maybe more even more important to call out "if it works in dev..." - I think that's the gap that more commonly is left by other platforms.

docs/concepts/platform/flavors.mdx
```diff
@@ -0,0 +1,33 @@
...
+## Available Flavors
+
+| Flavor | Image Source | Hardening | Availability | Typical Use |
+|--------|-------------|-----------|-------------|-------------|
+| **`upstream`** | Default chart sources (Docker Hub, GHCR, Quay) | Community-maintained | Public | Local development, CI, demos |
+| **`registry1`** | [Iron Bank](https://p1.dso.mil/services/iron-bank) (DoD hardened images) | STIG-hardened, CVE-scanned | Public | Production deployments requiring DoD compliance |
+| **`unicorn`** | Defense Unicorns curated registry | FIPS-validated, near-zero CVE posture | Private | Production deployments with Defense Unicorns support agreement |
+
+> [!NOTE]
+> The `unicorn` flavor is only available in a private container registry. It requires a Defense Unicorns support agreement. [Contact Defense Unicorns](https://www.defenseunicorns.com/contactus) for access.
```
mjnagel: small nit: `/contactus` redirects to `/contact` so we could just link directly there.

docs/concepts/platform/flavors.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+> [!CAUTION]
+> The `upstream` flavor is not recommended for production. Upstream images are community-maintained and may not meet the hardening or CVE-scanning requirements of regulated environments.
+
+> [!TIP]
+> **Compare CVE counts:** You can view current CVE counts for the `upstream` and `registry1` flavors on the [UDS Registry Core Package](https://registry.defenseunicorns.com/repo/public/core/versions). The `unicorn` flavor undergoes additional patching and curation by Defense Unicorns, resulting in significantly fewer CVEs — [contact Defense Unicorns](https://www.defenseunicorns.com/contactus) to learn more.
```
mjnagel: Same comment on `/contactus`.

tasks/lint.yaml
```diff
@@ -76,3 +76,57 @@ tasks:
     actions:
       - description: Add the SPDX license identifier to source files
         task: remote:fix-license
+
+  - name: k8s-version-check
+    description: "Verify documented Kubernetes version matches CI test config"
+    actions:
```
mjnagel, line 82: This script seems to work so I think I'm fine with it - but wonder if we can alternatively just use Renovate to update across docs and code (rather than manual updates + check for consistency)? There's some definite complexity there with AMI lookups and such though....so totally fine with this for now.

mjnagel (commented):
LGTM overall!

.github/workflows/docs-shim.yaml
```diff
@@ -67,6 +70,22 @@ jobs:
       - name: Check UDS Docs links
         run: uds run uds-docs-validate
 
+  k8s-version-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Environment setup
+        uses: ./.github/actions/setup
```
briantwatson, line 79: Is the setup action needed for k8s-version-check?  Also looking at the inputs had me wondering if we should consider breaking apart the registry auth steps into a separate action as a separate task? 🤔

tasks/lint.yaml
joelmccoy: Curious how you imagined a renovate flow to look like here... I was thinking about renovating this, but since we are always at `n-1` of kubernetes versions, I'm not sure there is a sane way in the renovate config to pin this version without manually updating the config?

mjnagel: I actually hadn't given that aspect much thought 🫣 - although the relatively new `minimumReleaseAge` ([docs](https://docs.renovatebot.com/key-concepts/minimum-release-age/)) could sort of help here since k8s releases are a standard cadence (although AKS/EKS/RKE2 might not follow exactly).

Could potentially overly simplify here and just lookup against upstream k8s with a `minimumReleaseAge` of 4 months +/- a bit, and then assume that AKS/EKS/RKE2 have already updated at that point? This would remove any complexity/weirdness with tracking 3 different upstream versions that may require AMI lookups (RKE2 AMI might still be a weird one since we manually build/update that separately).

joelmccoy: Actually I just thought of a hacky idea here that I have no idea if it would work...

1. set allowed versions in the config < 1.35 (latest minor) - which means we would resolve to 1.34 (our current n-1)
2. Put a renovate comment for `depName=kubernetes` in our renovate config (would need to switch to json5) and have this bump the minor versions
3. Renovate PR comes in to bump the renovate config separately < 1.35 -> < 1.36
4. Once that config is update, another renovate PR comes in to bump the docs & terraform variables for CI

A bit hacky/complicated and I have no idea if it would work. Open to other ideas here or we can just keep it simple and have a lint task here and we manually bump.

.github/workflows/docs-shim.yaml
joelmccoy: we need this step only for UDS CLI to run tasks... it may be worth breaking out the UDS CLI into it's own action. Just would require a refactor across our workflows.  Definitely might be beneficial security wise as these tasks do not need those creds and a supply chain compromise to any of the tools used in these tasks could grab them.  Just give me the thumbs up and I will refactor if we think that is worth while

tasks/lint.yaml
mjnagel: I guess mostly would like to call out that I'd love for these updates to _not_ be manual in the future - and if/when we can automate those it might make sense to change the way we do the docs too. Okay deferring that to later to eval/test separately though.

joelmccoy: Going to defer this to later. Created an issue capturing this: https://linear.app/defense-unicorns/issue/CORE-393/explore-renovate-automation-for-kubernetes-version-bumps

.github/workflows/docs-shim.yaml
joelmccoy: created an issue to do some refactoring here: [CORE-394](https://linear.app/defense-unicorns/issue/CORE-394/split-uds-cli-setup-into-a-separate-github-action)

docs/concepts/platform/flavors.mdx
joelmccoy: fixed

docs/concepts/platform/flavors.mdx
joelmccoy: fixed

docs/concepts/platform/environments.mdx
joelmccoy: adjusted the wording here. let me know if this reads better/what you were after

docs/concepts/platform/environments.mdx
joelmccoy: removed networking, combined TLS/Domains

mjnagel (approved):

PR #2428, opened 2026-03-04
docs: update blackbox-exporter doc to generalized uptime doc

briantwatson:
## Description

Updating dashboard name following comments in #2426

## Related Issue

Relates to INFRA-27 and INFRA-28

## Type of change

## Steps to Validate
- N/A

## Checklist before merging


mjnagel (approved):
LGTM - may also want to add a redirect on the docs side to preserve the old path as a redirect to the new one (old releases,etc might reference it).

That would be a change to here: https://github.com/defenseunicorns/uds-docs/blob/legacy-docs/astro.config.mjs#L15

PR #2433, opened 2026-03-05
chore(docs): add how-to guides for ha/resources on new docs site

joelmccoy:
## Description

  - Adds 6 new per-component how-to guides for configuring HA and resources: Keycloak, Logging (Loki/Vector), Monitoring (Grafana/Prometheus),
  Authservice, Runtime Security (Falco/Falcosidekick), and Service Mesh (Istio/gateways)
  - Adds an overview page with an HA capabilities summary table and links to each guide
  - Each guide covers bundle overrides for replica scaling, HPA configuration, resource tuning, and external dependency setup (PostgreSQL, S3,
  Redis) where applicable
  - All helm value paths verified against source charts; all upstream reference links validated

Note: The stubbed docs talked mainly about PDB and general Kubernetes concepts... but we don't have any existing docs/guidance for a lot of this.  So I opted for actionable HA/resource guides based on components.

## Related Issue

Fixes CORE-102

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Steps to Validate
- `uds run dev-docs` and review content

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md) followed

PR #2434, opened 2026-03-05
docs: add networking guides

mjnagel:
## Description

Adds how-to guides for networking (and service mesh).

## Type of change


docs/how-to-guides/networking/allow-permissive-mesh-traffic.mdx
```diff
@@ -0,0 +1,150 @@
...
+## Next steps
+
+- [Networking & Service Mesh Concepts](/concepts/core-features/networking/) for background on the authorization model
+- [Expose applications on gateways](/how-to-guides/networking/expose-apps-on-gateways/) to properly configure ingress through Package CRs
+
+## Related documentation
+
+- [Package CR Reference](/reference/operator-and-crds/package/)
+- [Istio Authorization Policy Documentation](https://istio.io/latest/docs/concepts/security/#authorization-policies)
+- [Istio PeerAuthentication Documentation](https://istio.io/latest/docs/reference/config/security/peer_authentication/)
```
joelmccoy: I think we should standardize on the format here for next steps and references. I have a PR in flight that also added a references section (I'm glad we were on the same page with adding something like this).  

Here is a ref for how I am doing it: https://github.com/defenseunicorns/uds-core/blob/17aadb6188b174d357b5ee127511d3bfbe3f825c/docs/how-to-guides/high-availability/configure-ha-keycloak.mdx?plain=1#L260-L282

Just so we standardize a couple questions:
1. Should we use the `CardGrid` for next steps? (I think this makes it a bit prettier)
2. Should we add the line: `These guides and concepts may be useful to explore next:` (I think this like is a bit helpful to show that these things aren't sequential)
3. Should related documentation/references come before or after next steps? (I think i like related documentation to come before next steps personally)
4. Related Documentation vs Next Steps (I think i like Related Documentation personally)

chance-coleman: Definitely agree the CardGrid looks nice! Also related to the removal of pagination, we can help better direct users with the next steps and related documentation. My leaning is for related documentation instead of references and before next steps as well.

mjnagel: Was just about to look at your PR and steal things 😏 - already did with the cards on the overview.
1. Yeah - cardgrid definitely prettier and I like that for next steps.
2. I like that :+1:
3. Yeah I think related docs/references before makes more sense.
4. I think these are two different things? 
    - Next Steps: We both have this right now, but I think your callout in 2 is pretty fair - things aren't necessarily sequential... so maybe "Related Documentation"
    - References vs Related Documentation: My PR currently calls the "reference" related documentation, but if we're using that for next steps...

mjnagel: Errr re-reading and processing how does this sound:
- Related Documentation: links to other external or internal docs. After any howto content, before next steps.
- Next Steps: Include the line you mentioned, use card grid, keep it named Next Steps. The line calling out "may be useful" clarifies its not sequential.

docs/how-to-guides/networking/overview.md
```diff
@@ -1,23 +0,0 @@
...
-
-### Source material from previous docs
-
-When authoring these guides, mine the following existing docs:
-
-- `src/content/docs-old/reference/configuration/Service Mesh/ingress.md`
-- `src/content/docs-old/reference/configuration/Service Mesh/egress.md`
-- `src/content/docs-old/reference/configuration/Service Mesh/custom-gateways.md`
-- `src/content/docs-old/reference/configuration/Service Mesh/non-http-ingress.md`
-- `src/content/docs-old/reference/configuration/uds-networking-configuration.md`
```
mjnagel, line 23: I initially left this off from my guides. Reviewing the doc again there may be some value in adding the "core portions" (how to add more network access for Core services) as a guide? The KubeAPI/KubeNodes pieces felt more reference-y or even operations (configuring a static kubeapi/kubenodes cidr is very much an "exception" since Pepr tracks these things.

docs/how-to-guides/networking/overview.mdx
```diff
@@ -0,0 +1,56 @@
+---
+title: Overview
+sidebar:
+  order: 1.0
```
mjnagel: Adjust to 2.x based on https://github.com/defenseunicorns/uds-core/pull/2433#discussion_r2892758755

docs/how-to-guides/networking/configure-tls-certificates.mdx
```diff
@@ -0,0 +1,242 @@
...
+## Verification
+
+Test the certificate chain:
+
+```bash
+curl -v https://my-app.yourdomain.com 2>&1 | grep -A 5 "Server certificate"
+```
+
+You should see your domain certificate and the correct certificate chain. You can also inspect the certificate in a browser by clicking the lock icon in the address bar.
+
```
ntwkninja, line 185: it also may be worth mentioning that the root CA may need to be loaded in your O/S trust store if using private or non-public CA's

Alternatively, we could specify a `publicly trusted` wildcard certificate in the pre-requisites section: http://localhost:4321/how-to-guides/networking/configure-tls-certificates/#prerequisites with a note there about adding a non-public CA to the o/s trust store

docs/how-to-guides/networking/configure-l7-load-balancer.mdx
```diff
@@ -0,0 +1,175 @@
...
+     version: "0.1.0"
+
+   packages:
+     - name: core
+       repository: registry.defenseunicorns.com/public/core
+       ref: x.x.x-upstream
+       overrides:
+         istio-tenant-gateway:
+           uds-istio-config:
+             values:
```
ntwkninja, line 53: do we also need a literal `"*"` for the ingress-gateway host entry since the SNI doesn't get passed from the TLS terminating / L7 loadbalancer?

docs/how-to-guides/networking/create-custom-gateways.mdx
```diff
@@ -0,0 +1,213 @@
...
+   ```
+
+</Steps>
+
+---
+
+## Troubleshooting
+
+### Traffic not routing through the custom gateway
+
```
ntwkninja, line 174: may be worth crossreferencing /configuration/trust-management/central-trust-bundle-management when tls_mode: is not SIMPLE

docs/how-to-guides/networking/configure-l7-load-balancer.mdx
```diff
@@ -0,0 +1,175 @@
...
+                 value: false
+               - path: tls.servers.tenant.enableHttpsRedirect
+                 value: false
+         # Uncomment if admin gateway is also behind the L7 load balancer:
+         # istio-admin-gateway:
+         #   uds-istio-config:
+         #     values:
+         #       - path: tls.servers.keycloak.enableHttpsRedirect
+         #         value: false
+         #       - path: tls.servers.tenant.enableHttpsRedirect
```
joelmccoy: **Bug: Wrong server name for admin gateway.** The admin gateway's servers are named `keycloak` and `admin` (per `src/istio/values/config-admin.yaml`), not `tenant`. This should be:
```yaml
- path: tls.servers.admin.enableHttpsRedirect
  value: false
```
The troubleshooting section (line 136) should also note the admin gateway uses `tls.servers.admin.enableHttpsRedirect`.

docs/how-to-guides/networking/define-network-access.mdx
```diff
@@ -0,0 +1,333 @@
...
+          remoteHost: api.example.com
+          remoteProtocol: TLS
+          selector:
+            app: my-app
+          serviceAccount: my-app
+  ```
+
+  The `serviceAccount` field is used by the operator to properly scope egress access and enforce least privilege. Include it for all ambient egress rules.
+
+  > [!WARNING]
```
joelmccoy: **Inaccurate conflict description.** The egress validator (`src/pepr/operator/controllers/istio/egress.ts:62-76`) only conflicts when the same host *and* port have different protocols. Two packages specifying the same host with *different* ports won't conflict.

Suggested rewrite: "If two packages specify the same host and port but with different protocols, the second package will fail to reconcile. Coordinate between package authors or consolidate egress rules when sharing host:port combinations."

docs/how-to-guides/networking/allow-permissive-mesh-traffic.mdx
```diff
@@ -0,0 +1,162 @@
+---
+title: Allow permissive traffic through the mesh
+sidebar:
+  order: 2.8
+---
+
+import { CardGrid, LinkCard } from '@astrojs/starlight/components';
```
joelmccoy: **Missing `Steps` component.** The how-to template requires a `## Steps` section using the `<Steps>` component. This file uses `## Option 1/2/3` headings instead and doesn't import or use `Steps`.

If option-based (pick-one) guides are an intentional pattern, that's fine — but the convention should be documented in `.ai/docs/how-to-guide-template.md` so future authors know when it applies.

docs/how-to-guides/networking/configure-l7-load-balancer.mdx
```diff
@@ -0,0 +1,175 @@
...
+   - Access an application through the load balancer URL and confirm it loads without redirect loops
+   - Verify Keycloak SSO works end-to-end by logging in through the tenant gateway
+   - If using mTLS, verify client certificate-based authentication works through Keycloak
+
+</Steps>
+
+---
+
+## Troubleshooting
+
```
joelmccoy, line 127: **Missing standalone `## Verification` section.** The how-to template expects a `## Verification` section after `</Steps>`. In this guide (and `configure-non-http-ingress.mdx`, `create-custom-gateways.mdx`, `enable-passthrough-gateway.mdx`, `expose-apps-on-gateways.mdx`), verification is embedded as a step inside `<Steps>` instead of being a standalone scannable section. Consider extracting or duplicating verification content into a dedicated section after `</Steps>`.

docs/how-to-guides/networking/define-network-access.mdx
```diff
@@ -0,0 +1,333 @@
...
+      allow:
+        - description: "Allow HTTPS to external API"
+          direction: Egress
+          port: 443
+          remoteHost: api.example.com
+          remoteProtocol: TLS
+          selector:
+            app: my-app
+          serviceAccount: my-app
+  ```
```
joelmccoy: **`serviceAccount` described as effectively required but is optional.** "Include it for all ambient egress rules" reads as a requirement. The validator shows it's optional (though recommended). It's also only valid on rules with `remoteHost` or `remoteGenerated: Anywhere` — setting it on other rule types causes validation failure. Consider clarifying: "The `serviceAccount` field is optional but strongly recommended for ambient egress rules with `remoteHost` or `remoteGenerated: Anywhere`."

joelmccoy (commented):
Review generated locally by Claude Code. Context given: review this PR for accuracy, consume the .ai directory for conventions and guidance, review for standardization against the how-to guide template, and check to make sure all helm values are correct overrides.

**Helm values:** All 28 override paths verified correct against source chart values.yaml files. All Zarf component names match `src/istio/zarf.yaml`.

**Cross-references:** All internal doc links verified — every referenced path has a corresponding file.

docs/how-to-guides/networking/define-network-access.mdx
```diff
@@ -0,0 +1,333 @@
...
+
+## Prerequisites
+
+- [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) installed
+- UDS Core deployed
+- Familiarity with the [Package CR](/reference/operator-and-crds/package/)
+
+## Before you begin
+
+UDS Core enforces strict network policies by default. All intra-cluster and external traffic must be explicitly declared in your Package CR's `allow` block. The UDS Operator translates these declarations into Kubernetes `NetworkPolicy` resources and, for external egress, into Istio `ServiceEntry` and routing resources.
```
joelmccoy: nit: is it worth calling out istio `AuthorizationPolicies` here too?

docs/how-to-guides/networking/define-network-access.mdx
joelmccoy, line 1: I think this document could use the `Steps` component?

docs/how-to-guides/networking/define-network-access.mdx
```diff
@@ -0,0 +1,333 @@
...
+        selector:
+          app.kubernetes.io/name: my-operator
+        remoteGenerated: KubeAPI
+```
+
+Other `remoteGenerated` values include `CloudMetadata` (for cloud provider metadata endpoints), `KubeNodes` (for node-level access), and `Anywhere` (unrestricted egress — use with caution).
+
+---
+
+## Allow egress to external services
```
joelmccoy: nit: I know we want to recommend least privilege for egress... but there may be some scenarios where the by host name egress isn't going to cut it (think wildcard domain limitations).  Is it worth adding an example where we do `remoteGenerated: Anywhere` with a `CAUTION` callout that this is not generally recommended unless host based egress rules don't fit your use case

joelmccoy (commented):
Couple other small comments. Otherwise content looks good.

docs/how-to-guides/networking/configure-l7-load-balancer.mdx
mjnagel: This example matches our current documented example (https://uds.defenseunicorns.com/reference/configuration/service-mesh/ingress/#using-an-l7-load-balancers-with-uds-core-gateways) which I know had been previously tested. I vaguely remember this SNI behavior, but we don't expose a way to add a true wildcard in the [hosts](https://github.com/defenseunicorns/uds-core/blob/main/src/istio/charts/uds-istio-config/templates/gateway.yaml#L21-L23) and I know we have users using L7 loadbalancers...so I'm hesitant to change anything?

docs/how-to-guides/networking/overview.md
mjnagel: Opted to add the core extension of networking.

docs/how-to-guides/networking/configure-tls-certificates.mdx
mjnagel: Added this in the prereq stuff to load the CA.

docs/how-to-guides/networking/create-custom-gateways.mdx
mjnagel: Decided to link to upstream Istio docs here for mutual gateway (non-simple). I think the central-trust-bundle stuff is/would be a bit different than the intent with mutual TLS?

docs/how-to-guides/networking/configure-l7-load-balancer.mdx
mjnagel: Fixed.

docs/how-to-guides/networking/define-network-access.mdx
mjnagel: Switched to use steps.

docs/how-to-guides/networking/define-network-access.mdx
mjnagel: Took the suggested wording change here.

docs/how-to-guides/networking/define-network-access.mdx
mjnagel: Good catch, fixed this wording.

docs/how-to-guides/networking/allow-permissive-mesh-traffic.mdx
mjnagel: Switched to steps + tabs here.

docs/how-to-guides/networking/configure-l7-load-balancer.mdx
mjnagel: Split off verification on all the guides.

docs/how-to-guides/networking/define-network-access.mdx
mjnagel: Yep, added!

docs/how-to-guides/networking/define-network-access.mdx
mjnagel: I reworked the way this doc was ordered/styled a fair bit in the last commit to make it more step based with options along the way. Let me know if that feels better?

docs/how-to-guides/networking/configure-l7-load-balancer.mdx
```diff
@@ -0,0 +1,169 @@
+---
+title: Configure an L7 load balancer
```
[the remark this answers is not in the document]
joelmccoy: This title is fine.  The guidance in the AI doc might be a bit restrictive.  We can probably remove that in a follow up

docs/how-to-guides/networking/allow-permissive-mesh-traffic.mdx
```diff
@@ -0,0 +1,235 @@
+---
+title: Allow permissive traffic through the mesh
```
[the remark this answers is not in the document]
joelmccoy: This title is fine. The guidance in the AI doc might be a bit restrictive. We can probably remove that in a follow up

docs/how-to-guides/networking/configure-tls-certificates.mdx
```diff
@@ -0,0 +1,230 @@
+---
+title: Configure TLS certificates for gateways
```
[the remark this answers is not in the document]
joelmccoy: This title is fine. The guidance in the AI doc might be a bit restrictive. We can probably remove that in a follow up

joelmccoy (dismissed):

joelmccoy (approved):

PR #2437, opened 2026-03-06
chore(docs): identity-access how to guides

chance-coleman:
## Description

how to guides for identity-access section

## Related Issue

Fixes CORE-103

## Type of change

## Checklist before merging


docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,184 @@
...
+**Solution:** Verify the plugin JAR was built and copied into the image. Check the build output of `uds run dev-build` for any Maven build errors. The plugin JAR should appear in `/opt/keycloak/providers/` inside the container:
+
+```bash
+uds zarf tools kubectl exec -n keycloak statefulset/keycloak -- ls /opt/keycloak/providers/
+```
+
+## Related documentation
+
+- [Testing custom image reference](/reference/uds-core/idam/testing-deployment-customizations/) — additional image management details including Zarf package transport
+- [Customize login page branding](/how-to-guides/identity-access/customize-branding/) — logo and T&C changes that do not require an image rebuild
```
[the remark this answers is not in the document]
joelmccoy: I think this is a good callout.  This is linking the old docs from the identity-config repo? Was that intentional?  I'm assuming we are keeping all identity docs living in this repo (or are there intentions to add docs to uds-identity-config)?

docs/how-to-guides/identity-access/protect-apps-with-authservice.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+       > [!NOTE]
+       > When using `network.expose` with Authservice-protected services, each expose entry must map to exactly one SSO client. Multiple services behind the same expose entry must share the same SSO configuration.
+
+    </Steps>
+  </TabItem>
+</Tabs>
+
+> [!NOTE]
+> **Ambient mode support:** If your Package CR sets `spec.network.serviceMesh.mode: ambient`, the UDS Operator automatically creates and manages an Istio [waypoint proxy](https://istio.io/latest/docs/ambient/usage/waypoint/) for Authservice to use. You do not need to configure the waypoint manually — the operator handles it. Ensure the `enableAuthserviceSelector` matches both the pod labels and the service's `spec.selector` for proper waypoint association.
```
joelmccoy: should we also add this caution here? https://uds.defenseunicorns.com/reference/configuration/single-sign-on/auth-service/#important-note-on-selector-matching

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,194 @@
...
+| `SOCIAL_AUTH_ENABLED` | `true` | Social/SSO login (Google, Azure AD, etc.) — requires an IdP to also be configured |
+| `OTP_ENABLED` | `true` | One-time password (TOTP) as a required MFA step for username/password login |
+| `WEBAUTHN_ENABLED` | `false` | WebAuthn/passkey as a required MFA step for username/password login |
+| `X509_MFA_ENABLED` | `false` | Require additional MFA (OTP or WebAuthn) after X.509 authentication |
+
+> [!CAUTION]
+> Disabling `USERNAME_PASSWORD_AUTH_ENABLED`, `X509_AUTH_ENABLED`, and `SOCIAL_AUTH_ENABLED` all at once will result in no authentication options on the login page. Users will not be able to log in or register. Also, disabling both `USERNAME_PASSWORD_AUTH_ENABLED` and `X509_AUTH_ENABLED` disables user self-registration.
+
+> [!NOTE]
+> `realmAuthFlows` values are applied only during initial realm import. To apply changes to an already-deployed Keycloak instance, a full redeploy of the Keycloak package is required.
```
joelmccoy: > a full redeploy of the Keycloak package is required

I think this wording might be confusing.  For the Realm Init settings this means you would need to destroy/recreate your keycloak deployment right? Or apply manual steps? Maybe we reword this a bit so we don't confuse Keycloak Package with UDS Package

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,194 @@
...
+
+   | Environment | Recommended configuration |
+   |---|---|
+   | CAC-only (no username/password) | Disable `USERNAME_PASSWORD_AUTH_ENABLED`, keep `X509_AUTH_ENABLED` |
+   | Username/password + OTP only | Keep defaults, disable `X509_AUTH_ENABLED` and `SOCIAL_AUTH_ENABLED` |
+   | Username/password + WebAuthn | Enable `WEBAUTHN_ENABLED`, disable `OTP_ENABLED` if desired |
+   | CAC + MFA | Enable `X509_MFA_ENABLED` (also requires `OTP_ENABLED` or `WEBAUTHN_ENABLED`) |
+
+   > [!NOTE]
+   > For X.509/CAC authentication to work, you must also configure the CA truststore. See [Configure the CA truststore](/how-to-guides/identity-access/configure-truststore/).
```
joelmccoy: I don't think this is accurate?  By default we provide CAC/DoD certs right? maybe we can reword this like:

```suggestion
   > By default, X.509/CAC authentication will work default DoD UNCLASSIFIED CA certs.  If you wish to update the truststore see [Configure the CA truststore](/how-to-guides/identity-access/configure-truststore/).
```

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,194 @@
...
+                   WEBAUTHN_ENABLED: false
+                   X509_MFA_ENABLED: false
+   ```
+
+   You only need to include the settings you want to change from default. However, for clarity and auditability, specifying all values explicitly is recommended.
+
+3. **Deploy your bundle**
+
+   ```bash
+   uds deploy uds-bundle.yaml
```
joelmccoy: bad syntax here
```suggestion
   uds create <path-to-bundle-dir>
   uds deploy uds-bundle-<name>-<arch>-<version>.tar.zst
```

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,194 @@
...
+   ```
+
+4. **(Optional) Enable X.509/CAC with MFA**
+
+   If you need to require MFA after CAC authentication, use this configuration in the `realmAuthFlows` block from step 2 in place of the values shown there, then run `uds deploy uds-bundle.yaml`:
```
joelmccoy:
```suggestion
   If you need to require MFA after CAC authentication, use this configuration in the `realmAuthFlows` block from step 2 in place of the values shown there, then recreate and deploy the bundle.
```

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,194 @@
...
+## Troubleshooting
+
+### Problem: Login page still shows disabled authentication options after deploy
+
+**Symptoms:** The login page displays username/password or CAC fields even though they were disabled.
+
+**Solution:** `realmAuthFlows` values are applied during initial realm import only. If Keycloak was already running before the override was applied, a full redeploy of the Keycloak package is required:
+
+```bash
+uds deploy uds-bundle.yaml
```
joelmccoy:
```suggestion
uds create <path-to-bundle-dir>
uds deploy uds-bundle-<name>-<arch>-<version>.tar.zst
```

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,194 @@
...
+
+   You only need to include the settings you want to change from default. However, for clarity and auditability, specifying all values explicitly is recommended.
+
+3. **Deploy your bundle**
+
+   ```bash
+   uds deploy uds-bundle.yaml
+   ```
+
+4. **(Optional) Enable X.509/CAC with MFA**
```
joelmccoy: should we even recommend this?  MFA with CaC auth feels weird. I think we had this enabled in the registry env and realized it was pointless/annoying (people in classified areas won't have their phone on them to do a web code).  And I think CaC is considered multi-factor with the pin anyways?

docs/how-to-guides/identity-access/connect-azure-ad-idp.mdx
```diff
@@ -0,0 +1,208 @@
...
+      | Claim    | Token type |
+      |----------|------------|
+      | `acct`   | SAML       |
+      | `email`  | SAML       |
+      | `ipaddr` | ID         |
+      | `upn`    | SAML       |
+
+      When prompted, enable the Microsoft Graph email and profile permissions.
+
+   3. Add a **Groups claim**: select **All groups**, accept the default values, and save.
```
joelmccoy, line 57: Is there some historical context around why we provide guidance for setting this up with the master realm?  And it looks like in this guide we allow all groups claims to be passed through? Is this intentional for the master realm. I think the old guide tried to scope this more to with `Under “Supported Account Types”, select “Accounts in this organizational directory ( only - Single tenant)”.` which is still a bit confusing? https://uds.defenseunicorns.com/reference/configuration/single-sign-on/identity-providers/azure-idp/#create-app-registration---master-realm

docs/how-to-guides/identity-access/connect-azure-ad-idp.mdx
```diff
@@ -0,0 +1,208 @@
...
+   > If UDS Core was deployed with `INSECURE_ADMIN_PASSWORD_GENERATION`, the username is `admin` and the password is in the `keycloak-admin-password` Kubernetes secret. Otherwise, register an admin user via `zarf connect keycloak`.
+
+   **Disable required actions** so Azure-federated users are not prompted to configure local credentials:
+   1. Go to **Authentication** → **Required actions**
+   2. Disable all required actions
+
+   **Create an admin group with realm admin role:**
+   1. Go to **Groups** → **Create Group**, name it `admin-group`
+   2. Open the group → **Role mapping** → **Assign role**
+   3. Switch to "Filter by realm roles" and assign the `admin` role
```
joelmccoy, line 85: Am I missing something here? How does an Azure identity get mapped to this admin-group?

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
```diff
@@ -0,0 +1,240 @@
...
+- [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) installed
+- Familiarity with [UDS bundle overrides](/how-to-guides/packaging-applications/overview/)
+
+## Before you begin
+
+UDS Core supports three layers of Keycloak customization:
+
+| Approach | Use for | Requires image rebuild? |
+|---|---|---|
+| **Helm chart values** (this guide) | Operational settings, session policies, resources | No |
```
joelmccoy: I don't think this is accurate?  We are setting values in the `realmInitEnv` in this doc and I'm pretty sure that would only happen on realm import?  Also we say that here: https://github.com/defenseunicorns/uds-identity-config/blob/037dddeab24e3bf28ddbe5e891e2634d7c6037ae/docs/reference/uds-core/idam/customization.md?plain=1#L302C1-L302C309

Maybe we summarize this table better and use the wording from that old doc.  And then maybe add a callout that changes in this how-to guide would require a complete realm rebuild (or apply manual settings manually)

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
```diff
@@ -0,0 +1,240 @@
...
+| **UDS Identity Config image** | Custom themes, plugins, realm structure changes | Yes |
```
joelmccoy: I think theme things can be updated on the fly without a rebuild? Also won't our plugins be updated automatically too? Either way, I think this whole section can be more accurately describe as to what requires a rebuild/as to what does not (see above comment)?

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
```diff
@@ -0,0 +1,240 @@
...
+               - path: realmInitEnv
+                 value:
+                   # Session idle timeout in seconds (default: 600 = 10 minutes)
+                   SSO_SESSION_IDLE_TIMEOUT: "1800"
+                   # Session max lifespan in seconds (default: 36000 = 10 hours)
+                   SSO_SESSION_MAX_LIFESPAN: "28800"
+   ```
+
+   > [!NOTE]
+   > `realmInitEnv` values are applied only during initial realm import. If Keycloak is already running, a full redeploy of the Keycloak package is required for these settings to take effect. To change timeouts on a live instance without redeploying, use the admin UI instead (see below).
```
joelmccoy:
```suggestion
   > `realmInitEnv` values are applied only during initial realm import. If Keycloak is already running, a full tear down and redeploy of Keycloak is required for these settings to take effect. To change timeouts on a live instance without redeploying, use the admin UI instead (see below).
```

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+   | Option | Default | Description |
+   |---|---|---|
+   | `maxRetries` | `0` (disabled) | Maximum retry attempts (set > 0 to enable) |
+   | `initialBackoffMillis` | `1000` | Initial backoff delay in milliseconds |
+   | `backoffMultiplier` | `2.0` | Exponential backoff multiplier |
+   | `applyJitter` | `true` | Adds randomness to prevent retry storms |
+   | `jitterFactor` | `0.5` | Jitter factor (0–1) for backoff variation |
+
+   Enable retries when Keycloak communicates with external services over unreliable networks. Keep them disabled in environments where fast failure detection is preferred.
```
joelmccoy: I'm not sure this fits under configure keycloak login policies as a guide? Maybe this should be it's own page? but open to hearing the argument for it living in this page.

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
```diff
@@ -0,0 +1,240 @@
...
+
+5. **Deploy your bundle**
+
+   ```bash
+   uds deploy uds-bundle.yaml
```
joelmccoy:
```suggestion
   uds create <path-to-bundle-dir>
   uds deploy uds-bundle-<name>-<arch>-<version>.tar.zst
```

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,184 @@
...
+## What you'll accomplish
+
+You'll build a custom uds-identity-config image containing your theme, plugin, or truststore changes, publish it to a container registry, and deploy it to UDS Core using the `configImage` Helm override. This guide covers the full workflow for any customization that requires an image rebuild.
+
+## Prerequisites
+
+- Docker installed and running
+- [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) installed
+- A container registry accessible from your cluster
+- Local clone of the [uds-identity-config](https://github.com/defenseunicorns/uds-identity-config) repository with your changes applied
```
joelmccoy: nit: i would remove this off the pre-reqs and include it in the steps

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,184 @@
...
+- CSS or FreeMarker templates in `src/theme/`
+- Custom Keycloak plugins in `src/plugin/`
+- The CA truststore (CA zip source in the Dockerfile)
+- Any file directly in the `src/` build context
+
+## Steps
+
+<Steps>
+
+1. **Make your changes to the uds-identity-config source**
```
joelmccoy: I would add a step to clone down the uds-identity-config repo and step into that directory.  Reading this guide as is, it is unclear that you need to be working in that repo to run the tasks.

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,200 @@
...
+   ```
+
+   This generates a `tls_cacert.yaml` file locally containing the base64-encoded CA certificate value.
+
+5. **Publish the image and configure the bundle override**
+
+   Tag and push the image to a registry your cluster can access:
+
+   ```bash
+   # For testing (ttl.sh is a short-lived public registry)
```
joelmccoy: nit: can we move this comment to a callout?  maybe add a caution that ttl.sh is a public open registry and should only be used in testing.  For production deploys, push this image to a secure container registry that can be accessed during your production builds. (or something like that)

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,200 @@
...
+   >   uds-istio-config:
+   >     values:
+   >       - path: tls.cacert
+   >         value: "<base64-encoded-ca-cert-from-tls_cacert.yaml>"
+   > ```
+
+6. **Deploy your bundle**
+
+   ```bash
+   uds deploy uds-bundle.yaml
```
joelmccoy:
```suggestion
   uds create <path-to-bundle-dir>
   uds deploy uds-bundle-<name>-<arch>-<version>.tar.zst
```

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,200 @@
...
+## What you'll accomplish
+
+You'll replace the default DoD CA certificate bundle in the uds-identity-config image with a custom CA bundle so that Keycloak can validate client certificates for X.509/CAC authentication in your environment. This requires building a custom uds-identity-config image.
+
+## Prerequisites
+
+- UDS Core deployed
+- Docker installed
+- [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) installed
+- Local clone of the [uds-identity-config](https://github.com/defenseunicorns/uds-identity-config) repository
```
joelmccoy: nit: would it make sense to move this out of the prereqs and pull into the steps?

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+The default uds-identity-config image includes DoD UNCLASS CA certificates, sourced at build time from a URL configured in the Dockerfile. To use your organization's own CA chain — or a locally generated test PKI — you must build a custom image with your certificates bundled in.
+
+The truststore is a Java KeyStore (JKS) file generated by the `ca-to-jks.sh` script during the image build. The Istio gateway also needs to know your CA so it can request client certificates from browsers.
+
+## Steps
+
+<Steps>
+
+1. **Prepare your CA certificate zip file**
```
joelmccoy: nit: recommend adding a step to clone down uds-identity-config and step into that directory (as it is unclear that you need to be in that directory to run the first task as written)

docs/how-to-guides/identity-access/configure-account-lockout.mdx
```diff
@@ -0,0 +1,154 @@
...
+   | Failure Reset Time (s) | `43200` | When to reset failure counters |
+   | Permanent Lockout | `ON` | Enable escalation to permanent lockout |
+   | Max Temporary Lockouts | Match your `MAX_TEMPORARY_LOCKOUTS` value | |
+
+   After configuring, save and test with a non-production account.
+
+3. **Deploy your bundle**
+
+   ```bash
+   uds deploy uds-bundle.yaml
```
joelmccoy:
```suggestion
   uds create <path-to-bundle-dir>
   uds deploy uds-bundle-<name>-<arch>-<version>.tar.zst
```

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,184 @@
...
+           keycloak:
+             values:
+               - path: configImage
+                 value: ttl.sh/<your-unique-id>:1h
+   ```
+
+5. **Deploy your bundle**
+
+   ```bash
+   uds deploy uds-bundle.yaml
```
joelmccoy:
```suggestion
   uds create <path-to-bundle-dir>
   uds deploy uds-bundle-<name>-<arch>-<version>.tar.zst
```

docs/how-to-guides/identity-access/customize-branding.mdx
```diff
@@ -0,0 +1,278 @@
...
+   - Pay Grade
+   - Unit, Organization or Company Name
+
+   > [!NOTE]
+   > Unlike `realmInitEnv`, `themeCustomizations.settings` values are applied at runtime and do not require a full redeploy of the Keycloak package to take effect.
+
+7. **Deploy your bundle**
+
+   ```bash
+   uds deploy uds-bundle.yaml
```
joelmccoy:
```suggestion
   uds create <path-to-bundle-dir>
   uds deploy uds-bundle-<name>-<arch>-<version>.tar.zst
```

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,184 @@
+---
+title: Build a custom Keycloak configuration image
+sidebar:
+  order: 12
```
joelmccoy: we are using sidebar ordering with decimals in our other how-to guides.  would suggest swapping to that. I think these docs would be `3.X`

joelmccoy: nit: i would also bump the build and deploy custom image up higher in this section as it is referenced by a couple other how-to guides and feels a bit more important

joelmccoy (commented):
Generally looking good! Love the coherence of these guides much better than our old ones. A couple questions/comments

docs/how-to-guides/identity-access/protect-apps-with-authservice.mdx
```diff
@@ -0,0 +1,210 @@
...
+             enableAuthserviceSelector:
+               app: httpbin
+       ```
+
+       Authservice will protect all pods labeled `app: httpbin` in the `httpbin` namespace.
+
+    2. **Apply the Package CR**
+
+       ```bash
+       kubectl apply -f package.yaml
```
joelmccoy: nit: we should default to do uds zarf tools kubectl everywhere
```suggestion
       uds zarf tools kubectl apply -f package.yaml
```

docs/how-to-guides/identity-access/protect-apps-with-authservice.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+import { Steps, Tabs, TabItem, LinkCard, CardGrid } from '@astrojs/starlight/components';
+
+## What you'll accomplish
+
+You'll add SSO protection to an application that has no native OIDC support. Authservice intercepts requests before they reach the application and handles the authentication flow on the application's behalf — requiring users to log in via Keycloak before they can access the app.
+
+## Prerequisites
+
+- UDS Core deployed (Authservice is included by default)
```
joelmccoy, line 15: nit: should add UDS cli as a prereq here.

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
chance-coleman: Agreed, CAC (certificate + PIN) already satisfies multi-factor requirements. The step is kept because X509_MFA_ENABLED is a real config option, but a NOTE has been added making clear it adds a second software factor on top of CAC, which is rarely needed and sometimes even problematic.

docs/how-to-guides/identity-access/connect-azure-ad-idp.mdx
chance-coleman: The master realm is configured so Keycloak administrators can log in to the admin console using enterprise Azure credentials instead of a local Keycloak account, this is/was the standard Day-2 setup to remove the bootstrap admin user. "All groups" is intentional: the SAML assertion must include the Entra group Object IDs for the mapper to work. the mapper then only acts on the specific OID you configure, so other group memberships are ignored. Single tenant is already set in the guide (step 1).

docs/how-to-guides/identity-access/connect-azure-ad-idp.mdx
chance-coleman: The Advanced Attribute to Group mapper reads the groups claim from the SAML assertion, which contains Object IDs of all Entra groups the user belongs to. If the claim contains the specific Entra group's Object ID (set in the mapper's Attribute value field), Keycloak adds the user to the mapped Keycloak group. A NOTE has been added to the guide clarifying this and noting that the Attribute value must be the group's Object ID (GUID), not the display name.

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
chance-coleman: Agreed, FIPS has been moved to its own dedicated guide (enable-fips-mode.mdx) with the full how-to format including pre-migration steps, verification, and troubleshooting.

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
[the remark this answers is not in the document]
chance-coleman: Yes, that was a stale link to the old uds-identity-config reference docs, it's been removed. All identity docs are consolidating into uds-core

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,190 @@
...
+   packages:
+     - name: core
+       repository: oci://registry.defenseunicorns.com/public/core
+       ref: x.x.x-upstream
+       overrides:
+         keycloak:
+           keycloak:
+             values:
+               - path: configImage
+                 value: ttl.sh/<your-unique-id>:1h
```
slaskawi: This probably should contain the production url (not ttl.sh)

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,190 @@
...
+**For truststore changes**, verify the gateway is requesting client certificates:
+
+```bash
+openssl s_client -connect sso.<domain>:443
+# Look for your CA in "Acceptable client certificate CA names"
+```
+
+## Troubleshooting
+
+### Problem: Init container fails to pull the image
```
slaskawi: Just a thought - maybe presenting this in a table would be more "compact" and friendly to the user?

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
slaskawi: I would love to keep it as a reference but maybe we should move it to the bottom of the page?

docs/how-to-guides/identity-access/configure-social-idp.mdx
```diff
@@ -0,0 +1,185 @@
+---
+title: Configure a social identity provider
```
slaskawi: I wonder if we should group configuring Azure idp with this page? After all, Entra is a social provider (although I prefer to call it an IdP, Identity Provider or External Identity Provider). Social Provider doesn't sound good (but it's a nit).

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+   ```bash
+   uds run regenerate-test-pki
+   ```
+
+   This creates `authorized_certs.zip` in the `src/` directory. If you need to adjust the subject or domain for your environment, edit `src/csr.conf` before running the task.
+
+   **For production**, assemble your organization's CA certificate chain into a zip file named `authorized_certs.zip` and place it in the `src/` directory of the uds-identity-config repository.
+
+3. **Update the Dockerfile CA source**
```
slaskawi: I wonder if this belong here or maybe it should be moved to the page describing how to build a custom Identity Config image?

docs/how-to-guides/identity-access/configure-user-account-settings.mdx
```diff
@@ -0,0 +1,197 @@
+---
+title: Configure Keycloak password policy and account settings
```
slaskawi: Looking at the content, I have a feeling that the title doesn't fully match it (e.g. we go through the Security Hardening settings which relate to Client Policies - they are not tied to the user settings per se). Perhaps it should be more generic or the Client Policy bits should be moved somewhere else.

docs/how-to-guides/identity-access/enable-fips-mode.mdx
```diff
@@ -0,0 +1,130 @@
+---
+title: Enable FIPS 140-2 strict mode
```
slaskawi: I wonder if we should keep this page or just remove it?

We migrated to FIPS a while back and I don't expect a lot of users looking for options how to disable it. I would remove most of this paragraph and just present a small guide how to disable FIPS (if someone has to).

docs/how-to-guides/identity-access/enforce-group-based-access.mdx
```diff
@@ -0,0 +1,155 @@
+---
+title: Enforce group-based access controls
```
slaskawi, line 2: Perhaps we could try to come up with a better title? Maybe "Authorization" or "User Access Control" or "User Authorization"?

docs/how-to-guides/identity-access/protect-apps-with-authservice.mdx
```diff
@@ -0,0 +1,214 @@
+---
+title: Protect apps with Authservice
```
slaskawi: How about changing the title to "Transparent OAuth 2.0 protection" or "Protecting application with SSO". I'd love to remove Authservice from the name as it's just an implementation detail.

slaskawi (commented):
The reworked pages look fantastic! I added a few comments and questions but we're definitely heading the right direction

docs/how-to-guides/identity-access/configure-social-idp.mdx
chance-coleman: Yeah i think this makes sense, i will keep them separate but rename to something more specific to Google IdP and if we add additional IdP support we would add a new doc for that.

docs/how-to-guides/identity-access/configure-truststore.mdx
chance-coleman: I prefer to keep these separated, one is the generic workflow while the other is a task specific workflow and i'd prefer to not make users have to jump around docs when follow a specific task workflow doc.

docs/how-to-guides/identity-access/configure-user-account-settings.mdx
chance-coleman: renamed to: Configure user accounts and security policies

docs/how-to-guides/identity-access/enable-fips-mode.mdx
chance-coleman: the guide has real value:
- The password migration step is critical for anyone upgrading from pre-FIPS deployments
- The verification steps help users confirm FIPS is active
- The troubleshooting section covers the lockout scenario

But agreed we should reframe it to "Manage FIPS 140-2 Mode" and add specific callout that uds-core ships default FIPS enabled.

docs/how-to-guides/identity-access/enforce-group-based-access.mdx
chance-coleman: I feel a bit differently here, i think those suggestions are too generic. "Enforce group-based access controls" is specific and action-oriented, it tells you what the guide does, which is solely focused on use of `group.anyOf` in package CRs.

docs/how-to-guides/identity-access/protect-apps-with-authservice.mdx
chance-coleman: went with "Protect non-OIDC apps with SSO"

docs/how-to-guides/identity-access/overview.mdx
```diff
@@ -0,0 +1,91 @@
+---
+title: Identity & access
```
joelmccoy: nit: we set this title to Overview in all the other pages since the top level is already called Identity & Access
```suggestion
title: Overview
```

docs/how-to-guides/identity-access/connect-azure-ad-idp.mdx
joelmccoy: thanks for adding that note. that definitely clears things up for me

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
joelmccoy: I think the http retries for this comment are still in the login policies guide.  Do we think these fit here?

joelmccoy (dismissed):
Generally LGTM! I had one comment not resolved and 1 nit for consideration.

docs/how-to-guides/identity-access/configure-user-account-settings.mdx
```diff
@@ -0,0 +1,197 @@
+---
+title: Configure user accounts and security policies
+sidebar:
+  order: 3.13
```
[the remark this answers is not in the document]
joelmccoy: LOL. I think this is unfortunately true.... it seems like sort doesn't treat this as a number 😠 .  maybe we could do 3.XX. so 3.00 - 3.99?  I believe that would work

docs/how-to-guides/identity-access/manage-keycloak-with-opentofu.mdx
```diff
@@ -0,0 +1,220 @@
+---
+title: Manage Keycloak with OpenTofu
+sidebar:
+  order: 3.14
```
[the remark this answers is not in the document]
joelmccoy: repeat address in: https://github.com/defenseunicorns/uds-core/pull/2437#discussion_r2919181972

docs/how-to-guides/identity-access/configure-google-idp.mdx
```diff
@@ -0,0 +1,185 @@
+---
+title: Configure Google SAML as an identity provider
+sidebar:
+  order: 3.15
```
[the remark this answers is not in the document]
joelmccoy: repeat address in: https://github.com/defenseunicorns/uds-core/pull/2437#discussion_r2919181972

docs/how-to-guides/identity-access/enable-fips-mode.mdx
```diff
@@ -0,0 +1,93 @@
+---
+title: Manage FIPS 140-2 mode
+sidebar:
+  order: 3.16
```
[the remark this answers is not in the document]
joelmccoy: repeat address in: https://github.com/defenseunicorns/uds-core/pull/2437#discussion_r2919181972

docs/how-to-guides/identity-access/configure-account-lockout.mdx
```diff
@@ -0,0 +1,155 @@
+---
+title: Configure Keycloak account lockout
+sidebar:
+  order: 3.11
```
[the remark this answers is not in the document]
joelmccoy: repeat address in: https://github.com/defenseunicorns/uds-core/pull/2437#discussion_r2919181972

docs/how-to-guides/identity-access/customize-branding.mdx
```diff
@@ -0,0 +1,278 @@
+---
+title: Customize Keycloak login page branding
+sidebar:
+  order: 3.12
```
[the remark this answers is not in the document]
joelmccoy: repeat address in: https://github.com/defenseunicorns/uds-core/pull/2437#discussion_r2919181972

docs/how-to-guides/identity-access/connect-azure-ad-idp.mdx
```diff
@@ -0,0 +1,218 @@
...
+
+   1. Go to **Manage** → **Token configuration**
+   2. Add the following optional claims (select **SAML** as the token type for each):
+
+      | Claim    | Token type |
+      |----------|------------|
+      | `acct`   | SAML       |
+      | `email`  | SAML       |
+      | `ipaddr` | ID         |
+      | `upn`    | SAML       |
```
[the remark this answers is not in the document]
joelmccoy: yea i think we can remove `(select **SAML** as the token type for each)` here?

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,158 @@
...
+3. **Build the custom image**
+
+   ```bash
+   uds run dev-build
+   ```
+
+   This builds the image using the `src/` directory and tags it locally as `uds-core-config:keycloak`.
+
+   > [!TIP]
+   > For air-gapped environments, you can also build a Zarf package that bundles the custom image for transport:
```
mjnagel:
```suggestion
   > For airgapped environments, you can also build a Zarf package that bundles the custom image for transport:
```

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,158 @@
...
+   > ```bash
+   > uds run build-zarf-pkg
+   > ```
+
+4. **Publish the image to a registry**
+
+   Tag and push to a registry your cluster can access.
```
mjnagel: imo we should have the airgap guidance be the primary steps here. Anyone deploying Core is using Zarf and will need to push the image inside of their cluster (to the Zarf registry). I think being more explicit about that in the steps would be good:
- build image + zarf package with `build-zarf-pkg` (passing in image name/version)
- publish zarf package optionally
- add zarf package to bundle along with the override

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+| Flow setting | Admin UI path |
+|---|---|
+| Disable username/password | **Authentication** → **Flows** → **UDS Authentication** → disable the **Deny Access** step below **Username Password Form** |
+| Disable credential reset | **Authentication** → **Flows** → **UDS Reset Credentials** → disable the **Reset Password** step |
+| Disable user registration | **Authentication** → **Flows** → **UDS Registration** → disable the **UDS Registration form** step |
+| Enable/disable OTP | **Authentication** → **Required Actions** tab → toggle **Configure OTP** |
+| Enable WebAuthn | 1. **Authentication** → **Required Actions** → toggle on **Webauthn Register Passwordless** under the **Enabled** column<br/>2. **Authentication** → **Flows** → **UDS Authentication** → set the **MFA** sub-flow to **Required**<br/>3. Inside the **MFA** sub-flow, set **WebAuthn Passwordless Authenticator** to **Required** |
+
+### Problem: X.509/CAC login fails with OCSP error in air-gapped environment
```
mjnagel:
```suggestion
### Problem: X.509/CAC login fails with OCSP error in airgapped environment
```

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+**Solution:** Two checks:
+1. Confirm the `tls.cacert` override was applied to `istio-tenant-gateway` and that the bundle was redeployed
+2. Confirm `X509_AUTH_ENABLED: true` is set in `realmAuthFlows` — if X.509 auth is disabled, the gateway will not request client certs even if the truststore is configured. See [Configure authentication flows](/how-to-guides/identity-access/configure-authentication-flows/).
+
+### Problem: Certificate authentication succeeds but OCSP errors appear in logs
+
+**Symptoms:** X.509 login works but Keycloak logs show OCSP revocation check failures.
+
+**Solution:** In air-gapped or restricted environments, the OCSP responder may be unreachable. Configure fail-open behavior or disable OCSP:
```
mjnagel:
```suggestion
**Solution:** In airgapped or restricted environments, the OCSP responder may be unreachable. Configure fail-open behavior or disable OCSP:
```

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,158 @@
...
+
+</Steps>
+
+## Verification
+
+Confirm the custom image was used:
+
+```bash
+# Check the init container image on the Keycloak pod
+uds zarf tools kubectl describe pod -n keycloak -l app.kubernetes.io/name=keycloak | grep -A 5 "Init Containers"
```
mjnagel: nit: we could use a more specific `kubectl` command to get the exact field, but don't feel strongly.

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,158 @@
...
+```
+
+## Troubleshooting
+
+| Problem | Symptoms | Solution |
+|---|---|---|
+| Init container fails to pull image | `ImagePullBackOff` or `ErrImagePull` on the Keycloak pod init container | Confirm the registry is reachable and the `configImage` value has no typos. For private registries, verify image pull secrets exist in the `keycloak` namespace. Debug with `uds zarf tools kubectl describe pod -n keycloak -l app.kubernetes.io/name=keycloak` |
+| Theme/truststore/plugin changes not reflected | Login page shows old branding, certificate auth fails, or plugin behavior unchanged despite new image | Themes, truststore, and plugins apply when the init container runs at pod start. Confirm the pod restarted: `uds zarf tools kubectl rollout status statefulset/keycloak -n keycloak`. If not, trigger a rollout: `uds zarf tools kubectl rollout restart statefulset/keycloak -n keycloak` |
+| Plugin JAR missing from providers | Custom plugin behavior not visible after deploy | Check `uds run dev-build` output for Maven errors. Verify the JAR exists: `uds zarf tools kubectl exec -n keycloak statefulset/keycloak -- ls /opt/keycloak/providers/` |
```
mjnagel: nit: stylistically this doesn't follow the other guides in terms of separate headers for each problem and then subheadings for symptom/solution.

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,158 @@
...
+  <LinkCard
+    title="Customize login page branding"
+    description="Replace logos and Terms & Conditions content via bundle overrides — no image rebuild needed."
+    href="/how-to-guides/identity-access/customize-branding/"
+  />
+  <LinkCard
+    title="Configure the CA truststore"
+    description="Build a custom image with your organization's CA certificates for X.509/CAC authentication."
+    href="/how-to-guides/identity-access/configure-truststore/"
+  />
```
mjnagel: I think ideally these would be different links - right now they are duplicated between next steps and related docs. Related docs could be upstream docs (plugin/theme references?) or concepts - imo they generally shouldn't be other guides.

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
chance-coleman: moved this into its own doc, didnt really fit well anywhere tbh.

docs/how-to-guides/identity-access/configure-account-lockout.mdx
```diff
@@ -0,0 +1,155 @@
...
+   ```
+
+</Steps>
+
+## Verification
+
+Confirm brute-force lockout is working:
+
+1. In a test browser session, attempt to log in with a valid username and incorrect password 3 times
+2. Confirm the account is locked — the login page should display a lockout message
```
mjnagel: Don't we hide lockout messages? So they may need to confirm on the admin side (step 4)? Maybe step 3/4 should be inverted?

docs/how-to-guides/identity-access/configure-account-lockout.mdx
```diff
@@ -0,0 +1,155 @@
...
+## Next steps
+
+These guides and concepts may be useful to explore next:
+
+<CardGrid>
+  <LinkCard
+    title="Configure Keycloak authentication methods"
+    description="Enable or disable X.509/CAC, OTP, WebAuthn, and social login via bundle overrides."
+    href="/how-to-guides/identity-access/configure-authentication-flows/"
+  />
```
mjnagel: Similar comment on duplicating the link here.

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,206 @@
...
+| `X509_MFA_ENABLED` | `false` | Require additional MFA (OTP or WebAuthn) after X.509 authentication |
+
+> [!CAUTION]
+> Disabling `USERNAME_PASSWORD_AUTH_ENABLED`, `X509_AUTH_ENABLED`, and `SOCIAL_AUTH_ENABLED` all at once will result in no authentication options on the login page. Users will not be able to log in or register. Also, disabling both `USERNAME_PASSWORD_AUTH_ENABLED` and `X509_AUTH_ENABLED` disables user self-registration.
+
+> [!NOTE]
+> `realmAuthFlows` values are only applied during the initial realm import. To change them on a running Keycloak instance, Keycloak must be fully torn down and redeployed so the realm is re-imported — or you can apply the changes manually in the admin UI (see the troubleshooting section below).
+
+> [!NOTE]
+> When upgrading UDS Core, realm configuration (`realmAuthFlows` and `realmInitEnv`) does **not** propagate automatically to a running Keycloak instance — Keycloak must be fully torn down and redeployed to pick up updated realm values. Theme files, truststore certificates, and custom plugin JARs **do** apply automatically on upgrade without a realm redeploy.
```
mjnagel: This seems repetitive?

docs/how-to-guides/identity-access/configure-keycloak-http-retries.mdx
```diff
@@ -0,0 +1,99 @@
...
+Confirm the bundle override applied successfully:
+
+1. Review your `uds deploy` output for the Keycloak release upgrade
+2. Confirm Keycloak is healthy and login flows that depend on external services (such as external IdPs) behave as expected during transient network failures
+
+## Related documentation
+
+- [Configure Keycloak login policies](/how-to-guides/identity-access/configure-keycloak-login-policies/) — session and logout behavior
+- [Connect Azure AD as an identity provider](/how-to-guides/identity-access/connect-azure-ad-idp/) — upstream IdP configuration
+- [Configure Google SAML as an identity provider](/how-to-guides/identity-access/configure-social-idp/) — upstream IdP configuration
```
joelmccoy:
```suggestion
- [Configure Keycloak outgoing HTTP requests](https://www.keycloak.org/server/outgoinghttp) — upstream Keycloak docs for outgoing HTTP requests
```

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,206 @@
...
+       repository: oci://registry.defenseunicorns.com/public/core
+       ref: x.x.x-upstream
+       overrides:
+         keycloak:
+           keycloak:
+             values:
+               - path: realmAuthFlows
+                 value:
+                   USERNAME_PASSWORD_AUTH_ENABLED: true
+                   X509_AUTH_ENABLED: false
```
mjnagel, line 71: If disabling the `X509_AUTH_ENABLED` toggle we probably should mention the recommended change to the Istio gateway(s) so that mutual TLS is not set (and users are not prompted for their CACs)?

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
```diff
@@ -0,0 +1,206 @@
...
+| Disable credential reset | **Authentication** → **Flows** → **UDS Reset Credentials** → disable the **Reset Password** step |
+| Disable user registration | **Authentication** → **Flows** → **UDS Registration** → disable the **UDS Registration form** step |
+| Enable/disable OTP | **Authentication** → **Required Actions** tab → toggle **Configure OTP** |
+| Enable WebAuthn | 1. **Authentication** → **Required Actions** → toggle on **Webauthn Register Passwordless** under the **Enabled** column<br/>2. **Authentication** → **Flows** → **UDS Authentication** → set the **MFA** sub-flow to **Required**<br/>3. Inside the **MFA** sub-flow, set **WebAuthn Passwordless Authenticator** to **Required** |
+
+### Problem: X.509/CAC login fails with OCSP error in air-gapped environment
+
+**Symptoms:** Certificate authentication fails with an OCSP revocation check error. Logs show the OCSP responder is unreachable.
+
+**Solution:** Configure OCSP fail-open behavior or disable OCSP checking via `realmInitEnv`.
```
mjnagel, line 131: Can we link to the reference guide for [airgap CRL config](https://github.com/defenseunicorns/uds-core/blob/main/docs/reference/configuration/single-sign-on/keycloak-crl-airgap.md) here? And/or should that also be a guide now?

docs/how-to-guides/identity-access/configure-device-flow.mdx
```diff
@@ -0,0 +1,147 @@
...
+import { Steps, LinkCard, CardGrid } from '@astrojs/starlight/components';
+
+## What you'll accomplish
+
+You'll configure a UDS Package to use the [OAuth 2.0 Device Authorization Grant](https://oauth.net/2/device-flow/) so that CLI tools, automation scripts, or headless devices can obtain tokens without a browser redirect. Once configured, the application can initiate a device code flow and present users with a short code to enter on a separate device.
+
+## Prerequisites
+
+- UDS Core deployed
+- A UDS Package CR for the application that needs device flow
```
mjnagel, line 17: Missing UDS CLI here. It might be good to rebase and review with the AI guidance to help catch and bulk fix these? (I had a separate comment on the `oci://` prefix that just got fixed in a commit as well)

docs/how-to-guides/identity-access/configure-device-flow.mdx
```diff
@@ -0,0 +1,147 @@
...
+   ```
+
+   > [!NOTE]
+   > No Kubernetes secret is created for public clients — there is no client secret to store. Your application initiates device flow by calling the Keycloak device authorization endpoint directly.
+
+2. **Apply the Package CR**
+
+   ```bash
+   uds zarf tools kubectl apply -f package.yaml
+   ```
```
mjnagel: Probably better to generally have build/deploy zarf package here. I did something similar in [this guide](https://github.com/defenseunicorns/uds-core/blob/main/docs/how-to-guides/networking/expose-apps-on-gateways.mdx) although I don't know that I love it so would welcome other approaches. Mostly just want to steer away from a direct kubectl apply since that's not really what we'd recommend here.

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,158 @@
...
+   ```
+
+5. **Set `configImage` in your bundle override**
+
+   In your `uds-bundle.yaml`, override the default identity config image:
+
+   ```yaml title="uds-bundle.yaml"
+   packages:
+     - name: core
+       repository: oci://registry.defenseunicorns.com/public/core
```
mjnagel: Also definitely would encourage updating the guidance if the rules there look too strict or don't work well as we add new guides.

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
```diff
@@ -0,0 +1,213 @@
...
+
+## Before you begin
+
+UDS Core supports three layers of Keycloak customization:
+
+| Approach | Use for | Requires image rebuild? |
+|---|---|---|
+| **Helm chart values** (this guide) | Session policies, account settings, auth flow toggles | No |
+| **UDS Identity Config image** | Custom themes, plugins, CA truststore | Yes — themes and plugins apply when the Keycloak pod restarts; no realm re-import needed |
+| **OpenTofu / IaC** | Managing groups, clients, IdPs post-deploy | No |
```
mjnagel: I wonder if this really fits better in a concepts doc (and then maybe add that as a prereq/before you begin reference)?

docs/how-to-guides/identity-access/configure-service-accounts.mdx
```diff
@@ -0,0 +1,216 @@
...
+
+## What you'll accomplish
+
+You'll configure a Keycloak client using the [OAuth 2.0 Client Credentials Grant](https://oauth.net/2/grant-types/client-credentials/) so that automated processes — CI/CD pipelines, backend services, and scripts — can obtain tokens and access Authservice-protected applications without a user session.
+
+## Prerequisites
+
+- UDS Core deployed
+- A UDS Package CR for the workload that needs machine-to-machine access
+- The `clientId` of the Authservice-protected application you want to access (used as the token audience)
```
mjnagel: nit: technically a service account client could be used for any SSO protected application, not just Authservice (although that is a common use-case).

docs/how-to-guides/identity-access/configure-service-accounts.mdx
```diff
@@ -0,0 +1,216 @@
...
+# Replace <domain>, <client-id>, and <client-secret> with your values
+curl -s -X POST \
+  "https://sso.<domain>/realms/uds/protocol/openid-connect/token" \
+  -d "grant_type=client_credentials" \
+  -d "client_id=<client-id>" \
+  -d "client_secret=<client-secret>" \
+  | jq .
+```
+
+A successful response includes an `access_token`. Decode it at [jwt.io](https://jwt.io) or with `jq` to verify the `aud` claim includes the expected audience.
```
mjnagel: Is it worth including an example of how to decode here (since we say "or with jq")?

docs/how-to-guides/identity-access/configure-google-idp.mdx
```diff
@@ -0,0 +1,185 @@
+---
+title: Configure Google SAML as an identity provider
```
mjnagel, line 2: nit: the actual filename of this doc probably could include `google` instead of generic `social`?

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+   **For production**, assemble your organization's CA certificate chain into a zip file named `authorized_certs.zip` and place it in the `src/` directory of the uds-identity-config repository.
+
+3. **Update the Dockerfile CA source**
+
+   In `src/Dockerfile`, update the `CA_ZIP_URL` build argument to reference your zip file:
+
+   ```dockerfile
+   ARG CA_ZIP_URL=authorized_certs.zip
+   ```
```
mjnagel: This is intentionally an `ARG` in the dockerfile to allow passing `--build-arg CA_ZIP_URL=some-other-file` to `docker build`. While that's not exposed with a task currently, I think it's still probably better to use that native arg mechanism rather than recommending file edit here?

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,210 @@
...
+
+   > [!CAUTION]
+   > `ttl.sh` is a public, ephemeral registry — images are accessible to anyone and expire after the specified duration. Only use it for local testing. For any shared or production environment, push to a private registry that your cluster can access securely.
+
+   **For local testing only:**
+
+   ```bash
+   docker tag uds-core-config:keycloak ttl.sh/<your-unique-id>:1h
+   docker push ttl.sh/<your-unique-id>:1h
+   ```
```
mjnagel, line 96: Similar to comment on the custom image doc - this should default to a zarf package workflow for the image.

docs/how-to-guides/identity-access/customize-branding.mdx
```diff
@@ -0,0 +1,278 @@
...
+     <li>Authorized use only</li>
+     <li>Activity may be monitored</li>
+   </ul>
+   ```
+
+   Convert to a single line (newlines replaced with `\n`):
+
+   ```bash
+   # Requires GNU sed
+   cat terms.html | sed ':a;N;$!ba;s/\n/\\n/g' > single-line.html
```
mjnagel: Might be nice to include a GNU sed and BSD sed version if we can?

docs/how-to-guides/identity-access/customize-branding.mdx
```diff
@@ -0,0 +1,278 @@
...
+   ```
+
+   Create a ConfigMap from the single-line file:
+
+   ```bash
+   uds zarf tools kubectl create configmap keycloak-tc-overrides \
+     --from-file=text=./single-line.html \
+     -n keycloak --dry-run=client -o yaml > terms-cm.yaml
+   ```
```
mjnagel, line 148: Should we have an explicit step here for making a zarf package (or adding to an existing one) and deploying before mounting in Keycloak?

docs/how-to-guides/identity-access/enable-fips-mode.mdx
```diff
@@ -0,0 +1,93 @@
...
+
+## Migrating from non-FIPS deployments
+
+If you are upgrading from an older deployment where FIPS was not enabled and user accounts already exist, you need to migrate password hashing before FIPS can activate cleanly.
+
+1. Before upgrading, connect to the Keycloak admin console (for example, via `uds zarf connect keycloak`)
+2. Switch to the **master** realm → **Authentication** → **Policies** → **Password Policy** → add `pbkdf2-sha512` as the hashing algorithm
+3. Reset the admin user's password to a 14+ character value with **Temporary** set to **Off**
+4. Repeat for any other local users whose passwords were hashed with `argon2`
+5. Upgrade to the latest UDS Core version
```
mjnagel: I think we discussed deleting this guide since FIPS mode is required - and currently the guide doesn't follow the standard format with any clear steps. This portion on migrating might have some value as the bulk of the guide/steps if we wanted or alternatively maybe just delete entirely?

docs/how-to-guides/identity-access/protect-apps-with-authservice.mdx
```diff
@@ -0,0 +1,217 @@
...
+
+Authservice works by matching a label selector on your application's pods. When a request comes in, Authservice intercepts it, validates the session, and redirects unauthenticated users to Keycloak. The first `redirectUris` entry you configure is used to populate the `match.prefix` hostname and the `callback_uri` in the Authservice chain.
+
+> [!CAUTION]
+> **Redirect URIs for Authservice clients cannot be root paths.** Using `https://myapp.example.com/` (a root path) is not allowed. Use a specific path like `https://myapp.example.com/login`.
+
+> [!NOTE]
+> **`enableAuthserviceSelector` must match both your pod labels and your Kubernetes Service's `spec.selector`.** If the selector matches pods but not the service, Authservice won't intercept traffic correctly. This is a common source of 503 errors and broken auth flows — double-check both before deploying.
```
mjnagel: These feel like they belong with the steps maybe rather than up front in the before you begin section?

joelmccoy (dismissed):

docs/how-to-guides/identity-access/protect-apps-with-authservice.mdx
```diff
@@ -0,0 +1,217 @@
...
+
+    </Steps>
+  </TabItem>
+</Tabs>
+
+> [!NOTE]
+> **Ambient mode support:** If your Package CR sets `spec.network.serviceMesh.mode: ambient`, the UDS Operator automatically creates and manages an Istio [waypoint proxy](https://istio.io/latest/docs/ambient/usage/waypoint/) for Authservice to use. You do not need to configure the waypoint manually — the operator handles it.
+
+> [!CAUTION]
+> **Selector matching in ambient mode:** The `enableAuthserviceSelector` must match both the pod labels **and** the Kubernetes Service's `spec.selector`. If the selector matches pods but not the service, the pod is mutated to use the waypoint but the service is not properly associated with it — traffic is blocked (503 errors) rather than routed through the SSO flow. Any `network.expose` entries should also use the same selector to ensure proper traffic flow from the gateway through the waypoint.
```
mjnagel, line 124: This could probably go up in the steps ~and/or as part of troubleshooting~ (already there ignore me)?

docs/how-to-guides/identity-access/configure-keycloak-login-policies.mdx
chance-coleman: moved this to the concepts identity-access.mdx file and added a callout in the before you begin section to reference some alternative docs to ready for more context.

docs/concepts/core-features/identity-access.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: Identity & authentication
+title: Identity & Authentication
```
mjnagel, line 2: Since this got changed I just realized...right now we have:
- concept/feature doc named Identity & Authentication
- how-tos names Identity-Access
- actual func layer names Identitiy-Authorization

Maybe should rename/standardize at some point (that can be a follow-on though if you want since it's not new in this PR).

docs/how-to-guides/identity-access/configure-authentication-flows.mdx
mjnagel, line 160: Resolved my other thread - I think I put my comment in the wrong place - this was more where I had in mind linking that other reference doc for CRL config.

docs/how-to-guides/identity-access/configure-service-accounts.mdx
```diff
@@ -8,13 +8,13 @@ import { Steps, LinkCard, CardGrid } from '@astrojs/starlight/components';
 
 ## What you'll accomplish
 
-You'll configure a Keycloak client using the [OAuth 2.0 Client Credentials Grant](https://oauth.net/2/grant-types/client-credentials/) so that automated processes — CI/CD pipelines, backend services, and scripts — can obtain tokens and access Authservice-protected applications without a user session.
+You'll configure a Keycloak client using the [OAuth 2.0 Client Credentials Grant](https://oauth.net/2/grant-types/client-credentials/) so that automated processes — CI/CD pipelines, backend services, and scripts — can obtain tokens and access SSO-protected applications without a user session.
 
 ## Prerequisites
 
 - UDS Core deployed
```
mjnagel, line 15: Probably need the CLI line here too?

docs/how-to-guides/identity-access/configure-device-flow.mdx
```diff
@@ -54,13 +55,35 @@ This flow creates a **public client** — a client with no secret. Two important
...
+     - name: my-app
+       required: true
+       manifests:
+         - name: uds-package
+           namespace: <app-namespace>
+           files:
+             - package.yaml
+   ```
+
+   Build and deploy the updated Zarf package through your UDS bundle as usual:
```
mjnagel: I think this diverges a bit from what we want in terms of showing both examples but leaving out the zarf.yaml (maybe good to make sure we're aligned based on the new guidance in [this PR](https://github.com/defenseunicorns/uds-core/pull/2445/changes)).

docs/how-to-guides/identity-access/customize-branding.mdx
```diff
@@ -136,8 +136,8 @@ This guide covers the bundle override approach. For CSS or structural theme chan
    Convert to a single line (newlines replaced with `\n`):
 
    ```bash
-   # Requires GNU sed
-   cat terms.html | sed ':a;N;$!ba;s/\n/\\n/g' > single-line.html
+   # Linux / macOS (requires perl, available by default on both)
+   perl -pe 's/\n/\\n/' terms.html > single-line.html
```
mjnagel: Perl feels weird to add here - I actually revisited and wonder if the comment was just wrong? I'm not sure that there were any GNU specific flags in there?

```
sed ':a;N;$!ba;s/\n/\\n/g' terms.html > single-line.html
```

Or the original cat | sed seem fine?

docs/how-to-guides/identity-access/enable-fips-mode.mdx
```diff
@@ -1,34 +1,84 @@
...
+
+2. **Add `pbkdf2-sha512` as the password hashing policy**
+
+   In the **master** realm:
+
+   1. Go to **Authentication** → **Policies** → **Password Policy**
+   2. Add a new policy: select **Hashing Algorithm** and set the value to `pbkdf2-sha512`
+   3. Save
+
+   Repeat this step for the **uds** realm if it contains local user accounts.
```
mjnagel: I may be mistaken but I thought we have always defaulted the UDS realm to a FIPS compatible hashing algo so this isn't applicable there? I went back to the [old upgrade steps](https://uds.defenseunicorns.com/reference/uds-core/idam/upgrading-versions/#v0140) and didn't see any `uds` realm mentions there.

I suppose this may be relevant if coming from non-uds keycloak, but in that case you wouldn't have the `uds` realm probably 😛

docs/how-to-guides/identity-access/build-deploy-custom-image.mdx
```diff
@@ -0,0 +1,190 @@
...
+   > [!CAUTION]
+   > `ttl.sh` is a public, ephemeral registry — images are accessible to anyone and expire after the specified duration. Only use it for local testing. For any shared or production environment, push to a private registry your cluster can access securely.
+
+   **Push the image to your registry:**
+
+   ```bash
+   docker push registry.example.com/uds/identity-config:1.0.0
+   ```
+
+   **For air-gapped environments**, publish the Zarf package to an OCI registry instead:
```
mjnagel:
```suggestion
   **For airgapped environments**, publish the Zarf package to an OCI registry instead:
```

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+2. **Prepare your CA certificate zip file**
+
+   The build script expects a zip file (`authorized_certs.zip`) containing your CA certificates.
+
+   **For test environments**, the repository includes a pre-configured `src/csr.conf` with a DoD certificate profile. Generate the test PKI directly:
+
+   ```bash
+   uds run regenerate-test-pki
+   ```
```
mjnagel: Not sure that this really fits in here now that I'm reading again? This is more about generating a test cert for experimentation - but doesn't really apply for actually building an image with custom trust?

docs/how-to-guides/identity-access/customize-branding.mdx
mjnagel, line 148: I think this is a case for the Zarf package (recommended) or kubectl apply structure for options?

docs/how-to-guides/identity-access/configure-truststore.mdx
```diff
@@ -0,0 +1,206 @@
...
+
+   This creates `authorized_certs.zip` in the `src/` directory. If you need to adjust the subject or domain for your environment, edit `src/csr.conf` before running the task.
+
+   **For production**, assemble your organization's CA certificate chain into a zip file named `authorized_certs.zip` and place it in the `src/` directory of the uds-identity-config repository.
+
+   > [!NOTE]
+   > The Dockerfile uses `CA_ZIP_URL=authorized_certs.zip` as the default build argument. If your file has a different name, pass it with `--build-arg CA_ZIP_URL=<your-filename>` when running `docker build` directly (rather than editing the Dockerfile). To exclude specific certificates from the final truststore, pass `CA_REGEX_EXCLUSION_FILTER=<pattern>` the same way.
```
mjnagel: The default in the dockerfile actually uses a [remote URL](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/Dockerfile#L45) so without passing `--build-arg` you'll always end up with the default. In other words - you always need to override this arg to get a different cert zip included/used.

Given the existing tasks I think it might make more sense to just use commands for `docker build` and `uds zarf package create` rather than using the existing tasks (and maybe some tech debt here for us to make a more flexible task)?

I do think overriding `CA_ZIP_URL` and/or `CA_REGEX_EXCLUSION_FILTER` are kind of the main things to do here though so they should be a more prominent part of the steps rather than just a note block.

docs/how-to-guides/identity-access/configure-device-flow.mdx
```diff
@@ -0,0 +1,168 @@
...
+   ```yaml title="zarf.yaml"
+   components:
+     - name: my-app
+       required: true
+       manifests:
+         - name: uds-package
+           namespace: <app-namespace>
+           files:
+             - package.yaml
+   ```
```
mjnagel:
```suggestion
```

mjnagel (dismissed):

mjnagel (approved):

PR #2440, opened 2026-03-06
docs: add documentation on resizing prometheus volumes

jasonwashburn:
## Description

Adds operations doc on resizing prometheus PVCs

## Type of change

## Steps to Validate
Can be mostly validated by running against k3d after running `uds run test-single-layer --set LAYER=monitoring` and then stepping through each step. Note: k3d does not actually care about resizing the volume, and it won't actually do a resize...but it will show all of the expected results with the exception of the storage capacity actually changing See: https://github.com/rancher/local-path-provisioner?tab=readme-ov-file#cons

Was also validated by running against an AWS EKS reference deployment

##  Checklist before merging

### DIFF
--- a/docs/operations/resize-prometheus-pvc.md
+++ b/docs/operations/resize-prometheus-pvc.md
@@ -20,7 +20,7 @@ This procedure follows upstream guidance from Prometheus Operator:
 
 ## Prerequisites
 
-- `kubectl` access with permissions to patch Prometheus CRs and PVCs and delete StatefulSets.
+- `kubectl` access with permissions to patch Prometheus CRs and PVCs, delete StatefulSets and Prometheus pods, and list/describe PVCs and StorageClasses.
 - Correct kube context selected.
 - Target Prometheus instance identified.
 - StorageClass for target PVCs supports expansion (`allowVolumeExpansion=true`).
--- a/docs/operations/resize-prometheus-pvc.md
+++ b/docs/operations/resize-prometheus-pvc.md
@@ -32,22 +32,6 @@ This runbook assumes UDS Core defaults:
 
 If your deployment uses non-default names, update the commands accordingly.
 
-## Target Size
-
-Set the target size before running commands:
-
-```bash
-export TARGET_SIZE=60Gi
-```
-
-Confirm the defaults exist in your cluster:
-
-This confirms the expected Prometheus CR exists before continuing.
-
-```bash
-kubectl get prometheus -n monitoring kube-prometheus-stack-prometheus
-```
-
 ## Update Bundle Configuration
 
 Set the target size in your UDS bundle so desired Prometheus storage is captured in code before running manual resize steps.
@@ -87,7 +71,15 @@ variables:
 
 ## Prechecks
 
-1. Confirm matching PVCs:
+1. Confirm the target Prometheus CR exists:
+
+This confirms the expected Prometheus CR exists before continuing.
+
+```bash
+kubectl get prometheus -n monitoring kube-prometheus-stack-prometheus
+```
+
+2. Confirm matching PVCs:
 
 This lists the PVCs that will be resized and verifies label selection is correct.
 
@@ -95,7 +87,7 @@ This lists the PVCs that will be resized and verifies label selection is correct
 kubectl get pvc -n monitoring -l "operator.prometheus.io/name=kube-prometheus-stack-prometheus"
 ```
 
-2. Confirm StorageClass and expansion support:
+3. Confirm StorageClass and expansion support:
 
 Use this output to verify each target PVC has a StorageClass and that class supports expansion.
 
@@ -110,24 +102,30 @@ kubectl get pvc -n monitoring -l "operator.prometheus.io/name=kube-prometheus-st
 kubectl get storageclass -o custom-columns=NAME:.metadata.name,ALLOWVOLUMEEXPANSION:.allowVolumeExpansion
 ```
 
-3. Confirm this is a size increase (never shrink):
+4. Confirm this is a size increase (never shrink):
 
-Compare current PVC request sizes to `TARGET_SIZE`; continue only for a size increase.
+Compare current PVC request sizes to your desired volume size; continue only if size will increase.
 
 ```bash
 kubectl get pvc -n monitoring -l "operator.prometheus.io/name=kube-prometheus-stack-prometheus" -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.spec.resources.requests.storage}{"\n"}{end}'
 ```
 
 > [!CAUTION]
-> If any target PVC is already larger than `TARGET_SIZE`, **stop and reassess**. PVC shrinking is not supported.
+> If any target PVC is already larger than your desired volume size, **stop and reassess**. PVC shrinking is not supported.
 
-4. Confirm the size configured in your `uds-bundle.yaml` and/or `uds-config.yaml` matches `TARGET_SIZE`.
+5. Confirm the size configured in your `uds-bundle.yaml` and/or `uds-config.yaml` matches your desired volume size.
 
 ## Procedure
 
-1. Update bundle configuration to the target size (see examples above).
+1. Set the `TARGET_SIZE` variable to your desired volume size. The `TARGET_SIZE` variable is used throughout this procedure:
 
-2. Pause Prometheus reconciliation:
+```bash
+export TARGET_SIZE=60Gi
+```
+
+2. Update bundle configuration to the target size (see examples above).
+
+3. Pause Prometheus reconciliation:
 
 Pausing prevents operator reconciliation churn while you patch PVCs and rotate the StatefulSet.
 
@@ -135,11 +133,11 @@ Pausing prevents operator reconciliation churn while you patch PVCs and rotate t
 kubectl patch prometheus kube-prometheus-stack-prometheus -n monitoring --type merge --patch '{"spec":{"paused":true}}'
 ```
 
-3. Create and deploy the updated bundle using your established UDS Core bundle creation and deployment workflows.
+4. Create and deploy the updated bundle using your established UDS Core bundle creation and deployment workflows.
 
 This applies the desired Prometheus storage size from code before patching existing PVCs.
 
-4. Patch each existing PVC to the new request size:
+5. Patch each existing PVC to the new request size:
 
 This updates the requested storage on all existing Prometheus PVCs to match bundle desired state.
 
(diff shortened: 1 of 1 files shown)


docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,227 @@
+---
+title: Resize Prometheus PVCs
+sidebar:
+  order: 3
```
jasonwashburn, line 4: model generated a metadata block that matches the existing docs in the same dir...but I'm honestly not sure what is using this...or what "order" we want this to be?

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,227 @@
...
+---
+
+## Purpose
+
+Use this runbook to manually resize Prometheus PVCs managed by Prometheus Operator when storage requirements increase.
```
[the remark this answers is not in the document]
jasonwashburn: agreed, but per previous conversations, that will get picked up by the currently ongoing docs reorg/reformatting effort

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,227 @@
...
+
+- If any step fails after pause, ensure unpause is restored:
+
+```bash
+kubectl patch prometheus kube-prometheus-stack-prometheus -n monitoring --type merge --patch '{"spec":{"paused":false}}'
+```
+
+- If StorageClass is not expandable, do not continue this runbook.
+- If one PVC patch fails, resolve that PVC issue first, then continue from Procedure step 4.
+- If PVCs remain in `ExternalExpanding` without capacity change for an extended period, stop and reassess.
```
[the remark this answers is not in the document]
jasonwashburn: added

docs/operations/resize-prometheus-pvc.md
chance-coleman: The metadata is is called [Frontmatter](https://starlight.astro.build/reference/frontmatter/#_top) used by the Starlight docs generator. `title` = Sidebar Title on the docsite and `sidebar.order` sets the precedence that this doc shows up in the directory its found. We don't have any real docs here yet, typically `order: 1` is preserved for the overview doc, so probably fine to leave this as 3 until we fill in the rest of the docs.

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Target Size
+
+Set the target size before running commands:
+
+```bash
+export TARGET_SIZE=60Gi
+```
+
+Confirm the defaults exist in your cluster:
```
chance-coleman: Would it make sense to keep only the `export TARGET_SIZE=60Gi` in "Target Size." Move the CR existence check into "Prechecks" as the first item. Target Size section seems to have mixed concerns.

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+
+This confirms the expected Prometheus CR exists before continuing.
+
+```bash
+kubectl get prometheus -n monitoring kube-prometheus-stack-prometheus
+```
+
+## Update Bundle Configuration
```
chance-coleman: Not sure i'm following why the bundle configuration step is before prechecks and then referenced from the procedure. Can we move this down into the procedure section (step one where it's referenced)?

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+
+```bash
+kubectl patch prometheus kube-prometheus-stack-prometheus -n monitoring --type merge --patch '{"spec":{"paused":true}}'
+```
+
+3. Create and deploy the updated bundle using your established UDS Core bundle creation and deployment workflows.
+
+This applies the desired Prometheus storage size from code before patching existing PVCs.
+
+4. Patch each existing PVC to the new request size:
```
chance-coleman: I think this step could broken up into two steps, step 4 patch PVCs step 5 monitor resize progress with a FSResize conditional callout.

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+
+3. Confirm this is a size increase (never shrink):
+
+Compare current PVC request sizes to `TARGET_SIZE`; continue only for a size increase.
+
+```bash
+kubectl get pvc -n monitoring -l "operator.prometheus.io/name=kube-prometheus-stack-prometheus" -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.spec.resources.requests.storage}{"\n"}{end}'
+```
+
+If any target PVC is already larger than `TARGET_SIZE`, stop and reassess.
```
chance-coleman: These kinds of warnings should utilize the [github alert syntax](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) to make them stand out more. 
```suggestion
> [!CAUTION]
> If any target PVC is already larger than `TARGET_SIZE`, **stop and reassess**. PVC shrinking is not supported.
```

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+  --patch "{\"spec\":{\"resources\":{\"requests\":{\"storage\":\"$TARGET_SIZE\"}}}}"
+```
+
+During and after patching, monitor PVC events for resize progress or errors:
+
+```bash
+kubectl describe pvc -n monitoring -l "operator.prometheus.io/name=kube-prometheus-stack-prometheus"
+```
+
+If controller-side expansion completes but the filesystem resize does not, PVCs may show `FileSystemResizePending` and `CAP` may remain below `REQ`. Check for that condition before proceeding:
```
chance-coleman: Similar comment, utilize github alert here.

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+
+This lists the PVCs that will be resized and verifies label selection is correct.
+
+```bash
+kubectl get pvc -n monitoring -l "operator.prometheus.io/name=kube-prometheus-stack-prometheus"
+```
+
+2. Confirm StorageClass and expansion support:
+
+Use this output to verify each target PVC has a StorageClass and that class supports expansion. If expansion is not supported, stop and reassess.
```
chance-coleman: Maybe break the warning out into its own line and add github alert.
```suggestion
Use this output to verify each target PVC has a StorageClass and that class supports expansion.

> [!CAUTION]
> If expansion is not supported, stop and reassess.
```

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+
+2. Pause Prometheus reconciliation:
+
+Pausing prevents operator reconciliation churn while you patch PVCs and rotate the StatefulSet.
+
+```bash
+kubectl patch prometheus kube-prometheus-stack-prometheus -n monitoring --type merge --patch '{"spec":{"paused":true}}'
+```
+
+3. Create and deploy the updated bundle using your established UDS Core bundle creation and deployment workflows.
```
chance-coleman: This feels kinda vague to me, could we add the `uds create` and `uds deploy` commands so that the user doesn't have to find that?

chance-coleman (commented):
First pass review, few comments to address. Generally looking good though.

joelmccoy (dismissed):
Content LGTM. I just established a PR that has guidance/templates for these types of docs here: https://github.com/defenseunicorns/uds-core/pull/2449. I'm fine merging this in and taking on the refactor in my PR to make this match the format/structure.

docs/operations/resize-prometheus-pvc.md
jasonwashburn: Its located above to be more reference help than procedural since everyone's bundle/preference for variables vs overrides is a little different and environment dependent...and bundle creation/approvals and deployment processes vary a little as well

docs/operations/resize-prometheus-pvc.md
jasonwashburn: agreed...this structure is a bit of hangover from when this section had a bunch more variables getting set and was named differently. Will move it into the procedure where it's needed, and move CR existence into prechecks

docs/operations/resize-prometheus-pvc.md
jasonwashburn: I left it this way intentionally because environments have different ways of creating and deploying bundles? For example, in registry....that happens when you merge your bundle changes into main via PR...there's not actually any commands being run to do that by the user...my worry was that by adding it to step by step procedure you might drive users to do something that goes against the norm or conventions enforced in their environment...while at the same time being authenticated to the cluster and able to actually do it?

docs/operations/resize-prometheus-pvc.md
jasonwashburn: works for me

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+          variables:
+            - name: PROMETHEUS_STORAGE_SIZE
+              description: Prometheus PVC requested storage size
+              path: prometheus.prometheusSpec.storageSpec.volumeClaimTemplate.spec.resources.requests.storage
+```
+
+```yaml title="uds-config.yaml"
+variables:
+  core:
+    prometheus_storage_size: "60Gi"
```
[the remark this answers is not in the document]
jasonwashburn: fixed

docs/operations/resize-prometheus-pvc.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Scope
+
+- Resize only (no retention tuning).
+- Intended for UDS Core deployments using `kube-prometheus-stack`.
+- Safe for mixed environments as long as prechecks pass.
+
+## Prerequisites
+
+- `kubectl` access with permissions to patch Prometheus CRs and PVCs and delete StatefulSets.
```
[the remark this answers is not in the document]
jasonwashburn: fixed

docs/operations/resize-prometheus-pvc.md
jasonwashburn: We could move it down into the step for updating your bundle in the procedure...I suppose the worst that could  come from that is that someone gets through prechecks, then has to stop when they realize they need to get a bundle PR or what not approved before they can actually do the procedure...but given nothing's really been started by then, that's probably fine

docs/operations/resize-prometheus-pvc.md
jasonwashburn: updated

docs/operations/resize-prometheus-pvc.md
jasonwashburn: went ahead and moved it

docs/operations/resize-prometheus-pvc.md
jasonwashburn: updated

docs/operations/resize-prometheus-pvc.md
jasonwashburn: fixed

joelmccoy (approved):
Content LGTM - will refactor to new runbook template in my PR

docs/operations/resize-prometheus-pvc.md
chance-coleman: Ahh okay yeah that makes sense to me. Thanks for the context.

chance-coleman (approved):
🚀 LGTM

PR #2452, opened 2026-03-10
chore: add polling/increase timeout to prometheus target e2e test

joelmccoy:
## Description

Fixes a flake on an e2e CI test: https://github.com/defenseunicorns/uds-core/actions/runs/22917963254/job/66511607227?pr=2414

Increase the timeout of this test to match the uptime tests (probes can take some time before getting loaded and showing metrics).

## Related Issue

Fixes CORE-405

## Type of change

## Steps to Validate
- e2e tests pass in CI

## Checklist before merging

### DIFF
--- a/test/vitest/prometheus.spec.ts
+++ b/test/vitest/prometheus.spec.ts
@@ -7,7 +7,7 @@ import { afterAll, beforeAll, describe, expect, test } from "vitest";
 import { closeForward, getForward } from "./helpers/forward";
 import { pollUntilSuccess } from "./helpers/polling";
 
-describe("Prometheus and Alertmanager", { retry: 1, timeout: 210000 }, () => {
+describe("Prometheus and Alertmanager", { timeout: 210000 }, () => {
   let prometheusProxy: { server: net.Server; url: string };
   let alertmanagerProxy: { server: net.Server; url: string };
 
@@ -49,7 +49,9 @@ describe("Prometheus and Alertmanager", { retry: 1, timeout: 210000 }, () => {
     await pollUntilSuccess(
       async () => {
         const response = await fetch(`${prometheusProxy.url}/api/v1/targets`);
-        expect(response.status).toBe(200);
+        if (!response.ok) {
+          throw new Error(`Prometheus targets API returned ${response.status}`);
+        }
 
         const body = (await response.json()) as {
           data: {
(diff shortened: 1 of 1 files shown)


test/vitest/prometheus.spec.ts
```diff
@@ -45,29 +46,46 @@ describe("Prometheus and Alertmanager", { retry: 1 }, () => {
   });
 
   test("all prometheus targets should be up", async () => {
-    const response = await fetch(`${prometheusProxy.url}/api/v1/targets`);
-    expect(response.status).toBe(200);
+    await pollUntilSuccess(
+      async () => {
```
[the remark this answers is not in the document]
joelmccoy: fixed.

test/vitest/prometheus.spec.ts
```diff
@@ -45,29 +46,46 @@ describe("Prometheus and Alertmanager", { retry: 1 }, () => {
...
+        if (failedTargets.length > 0) {
+          for (const target of failedTargets) {
+            console.warn(`Target health is down: ${target.scrapePool}`);
+          }
+          return false;
+        }
 
-    expect(failedTargets).toEqual([]);
+        return true;
+      },
```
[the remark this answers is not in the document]
joelmccoy: fixed.

test/vitest/prometheus.spec.ts
```diff
@@ -5,8 +5,9 @@
 import * as net from "net";
 import { afterAll, beforeAll, describe, expect, test } from "vitest";
 import { closeForward, getForward } from "./helpers/forward";
+import { pollUntilSuccess } from "./helpers/polling";
 
-describe("Prometheus and Alertmanager", { retry: 1 }, () => {
+describe("Prometheus and Alertmanager", { timeout: 210000 }, () => {
```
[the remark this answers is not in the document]
joelmccoy: removed the retry since we are polling now

test/vitest/prometheus.spec.ts
```diff
@@ -44,30 +45,39 @@ describe("Prometheus and Alertmanager", { retry: 1 }, () => {
...
+          throw new Error("No active targets returned from Prometheus");
+        }
 
-    expect(failedTargets).toEqual([]);
+        return body.data.activeTargets;
+      },
+      targets => {
+        const unhealthy = targets.filter(target => target.health !== "up");
+        for (const target of unhealthy) {
+          console.warn(`Target ${target.health}: ${target.scrapePool}`);
```
chance-coleman:
```suggestion
          console.warn(`Target ${target.scrapePool} is ${target.health}`);
```
super nit: i think this would read better

test/vitest/prometheus.spec.ts
```diff
@@ -44,30 +45,39 @@ describe("Prometheus and Alertmanager", { retry: 1 }, () => {
...
-      for (const target of unknownTargets) {
-        console.warn(`Target health is currently unknown: ${target.scrapePool}`);
-      }
-    }
+        if (!body.data?.activeTargets?.length) {
+          throw new Error("No active targets returned from Prometheus");
+        }
 
-    expect(failedTargets).toEqual([]);
+        return body.data.activeTargets;
```
chance-coleman, line 69: looking at other uses of the `pollUntilSuccess` in tests, i think all of them have an expect assertion on the returned result. Does it make sense to follow the same pattern?

chance-coleman (commented):
looks good to me, just two small comments

test/vitest/prometheus.spec.ts
```diff
@@ -44,30 +45,39 @@ describe("Prometheus and Alertmanager", { retry: 1 }, () => {
...
+      },
+      targets => {
+        const unhealthy = targets.filter(target => target.health !== "up");
```
mjnagel, line 72: I think this probably makes sense - just want to confirm it is intentional - previously we allowed "unknown" targets just with a log warning. With these changes we require everything to be up. I think with the retries and longer timeout this should be fine?

test/vitest/prometheus.spec.ts
joelmccoy: yeahh good catch here. updated.

test/vitest/prometheus.spec.ts
joelmccoy: fixed

test/vitest/prometheus.spec.ts
joelmccoy: correct. i think this is a more accurate check and we now have some grace period if unknown targets needs some time to come up

mjnagel (approved):

PR #2456, opened 2026-03-11
chore(docs): add backup restore how to guides

chance-coleman:
## Description
Add backup and restore how to guides

## Related Issue

Fixes core-107

## Type of change

## Checklist before merging


docs/how-to-guides/backup-restore/configure-storage-backends.mdx
```diff
@@ -0,0 +1,287 @@
...
+
+   > [!NOTE]
+   > The `bucket` field corresponds to the Azure Blob container name.
+
+</Steps>
+
+  </TabItem>
+</Tabs>
+
+### (Optional) Configure storage network egress
```
mjnagel: I think this (and the backup schedule + create/deploy bundle) probably belong as steps in the main flow of the guide?

docs/how-to-guides/backup-restore/configure-storage-backends.mdx
```diff
@@ -0,0 +1,287 @@
...
+```bash
+# Velero pod is running
+uds zarf tools kubectl get pods -n velero
+
+# Backup storage location shows "Available"
+uds zarf tools kubectl get backupstoragelocation -n velero
+
+# Backup schedule exists with correct cron expression
+uds zarf tools kubectl get schedule -n velero
+```
```
mjnagel, line 228: Might be worth adding a validation triggering an immediate backup based on the config and checking that it works? I don't think Velero tests storage locations until the cron triggers otherwise. Ref for how to do this: https://velero.io/docs/v1.10/backup-reference/#:~:text=velero%20backup%20create%20%2D%2Dfrom%2Dschedule%20example%2Dschedule (will just want to add Velero CLI to pre-reqs or validation steps if we do that)

docs/how-to-guides/backup-restore/enable-volume-snapshots.mdx
```diff
@@ -0,0 +1,357 @@
...
+
+> [!NOTE]
+> If your applications use PersistentVolumes and you need to restore the actual on-disk data (not just the PVC resource definitions), you must enable volume snapshots. Without them, a restore will recreate the PVC but the underlying data will be lost.
+
+## Steps
+
+<Tabs>
+  <TabItem label="AWS EBS">
+
+<Steps>
```
mjnagel: I don't have a strong opinion here but wanted to ask - given the lack of overlap between ebs and vsphere/csi - do these fit better as two separate guides rather than combined? I think if there were more overlap I can see combining in a single guide, but they currently feel quite divergent.

docs/how-to-guides/backup-restore/enable-volume-snapshots.mdx
```diff
@@ -0,0 +1,357 @@
...
+
+  </TabItem>
+</Tabs>
+
+**Create and deploy your bundle:**
+
+```bash
+uds create <path-to-bundle-dir>
+uds deploy uds-bundle-<name>-<arch>-<version>.tar.zst
+```
```
mjnagel: Should be part of the steps.

docs/how-to-guides/backup-restore/configure-storage-backends.mdx
mjnagel: After getting to the follow-on docs maybe we just link the perform backup/restore operations guide for validation 😅

docs/how-to-guides/backup-restore/perform-backup-restore.mdx
```diff
@@ -0,0 +1,215 @@
...
+title: Perform backup and restore operations
+sidebar:
+  order: 7.3
+---
+
+import { Steps, LinkCard, CardGrid } from '@astrojs/starlight/components';
+
+## What you'll accomplish
+
+You'll verify your scheduled backups are running, trigger a manual backup, restore specific namespaces, and confirm the restored state matches expectations.
```
mjnagel: I wonder if it makes sense to split backup/restore into two separate guides here? Maybe thinking too far ahead, but I would assume _eventually_ we would have operations docs on disaster recovery, which likely would link back to restore docs we have ideally (with backups already automated on a schedule). Don't feel strongly about this split right now though if you don't see the need.

mjnagel (commented):
Few organization comments here - content overall looks good.

docs/how-to-guides/backup-restore/enable-volume-snapshots.mdx
chance-coleman: My intention was that both answer the same question: "how do I get Velero to capture PV data?" I can see the main steps are quite different but the before and after would more or less be the same. So i went with the tabs pattern to handle the provider divergence. I could see this going either way and maybe we shouldnt use the tab pattern and instead just do separate docs with similar sections (specifically referring to before and after).

docs/how-to-guides/backup-restore/perform-backup-restore.mdx
chance-coleman: I guess in my head i was thinking you wouldn't have restore with out backup and if they were separate that meant having to switch between the docs for what i'd expect to be a single workflow. The future DR doc could link to a specific restore section in this doc still.

mjnagel: > I guess in my head i was thinking you wouldn't have restore with out backup and if they were separate that meant having to switch between the docs for what i'd expect to be a single workflow.

I think this is fair from a dev/test scenario (or flexing the muscle of backup/restore), but less true in a prod environment? You'd generally "always" be running backups (on a schedule and/or manually as needed) whereas the restore would be something that you run from existing backups in a disaster scenario or other rollback event.

Would probably welcome another opinion here (cc @joelmccoy?) since I can see both perspectives depending on how we want to frame this doc and others.

docs/how-to-guides/backup-restore/enable-volume-snapshots.mdx
mjnagel: I think this ended up in its own single step with the changes, but maybe can wait pending the other thread - https://github.com/defenseunicorns/uds-core/pull/2456#discussion_r2921025934

docs/how-to-guides/backup-restore/enable-volume-snapshots.mdx
mjnagel: I like the tab pattern when there's a lot of commonality between _most_ of the doc (maybe just a divergence in one or two steps). I think I could see this doc being a bit more focused and helpful if we split it though? Few examples:
- prerequisites could be a bit more focused (ex: AWS guide would require EBS CSI driver/storage class, RKE2/vsphere could have the cloud-provider-name in there)
- more specific verification steps (check EBS snapshots)
- troubleshooting better split out (currently the "snapshot limit reached" is only applicable to vSphere)
- more focused related docs rather than a long list that covers a mix of providers

I think especially if/when we add more documented storage providers here we might see more benefits from splitting them?

docs/how-to-guides/backup-restore/perform-backup-restore.mdx
joelmccoy: For searchability and single purpose docs I like the idea of splitting this into two separate how-to guides.  On the restore guide do you can add having a backup as a prerequisite and link the backup doc.

docs/how-to-guides/backup-restore/enable-volume-snapshots.mdx
mjnagel: Thinking again this morning of a different way to state this... I think the way I see it: the tab pattern feels right if it's _inside_ of the steps (a single step or two have options/tabs), but not when the entire steps section (or other sections) require tabs/options.

Maybe a good way to add this to AI guidance if we think it will come up more.

chance-coleman: I've split this into two docs as well as the backup and restore.

mjnagel (approved):

PR #2463, opened 2026-03-12
chore(docs): enable pagination override for getting started docs

chance-coleman:
## Description
After disabling pagination for all docs, we need to override that enable them for specific docs. this change does that for the getting started docs.

## Related Issue

Fixes [core-390](https://linear.app/defense-unicorns/issue/CORE-390/evaluate-disabling-pagination-on-most-docs-sections)

## Type of change

## Steps to Validate

## Checklist before merging

### DIFF
--- a/docs/getting-started/local-demo/basic-requirements.mdx
+++ b/docs/getting-started/local-demo/basic-requirements.mdx
@@ -1,5 +1,7 @@
 ---
 title: Set Up Your Environment
+prev: true
+next: true
 sidebar:
   order: 2.2
 ---
--- a/docs/getting-started/local-demo/basic-requirements.mdx
+++ b/docs/getting-started/local-demo/basic-requirements.mdx
@@ -1,6 +1,5 @@
 ---
 title: Set Up Your Environment
-prev: true
 next: true
 sidebar:
   order: 2.2
--- a/docs/getting-started/local-demo/integrate-your-package.mdx
+++ b/docs/getting-started/local-demo/integrate-your-package.mdx
@@ -1,7 +1,6 @@
 ---
 title: Add Your Own Package (Optional)
 prev: true
-next: true
 sidebar:
   order: 2.4
 ---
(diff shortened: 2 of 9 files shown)


docs/getting-started/local-demo/integrate-your-package.mdx
```diff
@@ -1,5 +1,7 @@
 ---
 title: Add Your Own Package (Optional)
+prev: true
+next: true
```
mjnagel: nit: I think we generally have "two paths" here for the getting-started so not sure if next (onto the prod deploy) makes sense - unless we custom title it (right now it says "Overview").

docs/getting-started/local-demo/overview.mdx
```diff
@@ -1,5 +1,7 @@
 ---
 title: Overview
+prev: true
```
mjnagel: Connected comment to ^ - I'm not sure if previous makes sense here.

docs/getting-started/local-demo/basic-requirements.mdx
```diff
@@ -1,5 +1,7 @@
 ---
 title: Set Up Your Environment
+prev: true
```
mjnagel: I wonder if prev really makes sense here (back to the overview). Don't feel strongly on this one, but would probably want consistency between local demo and prod since we present them as branching paths (not local demo first, then prod).

mjnagel (commented):
Small comments on the prev/next placement for a few docs.

docs/getting-started/local-demo/integrate-your-package.mdx
chance-coleman: thats a good distinction that i didn't think about. i think it makes sense to keep the two paths separated. Went with an `Overview` -> `prereqs` <-> `next doc` <-> `next doc` approach for both.

mjnagel (approved):

PR #2473, opened 2026-03-15
chore(docs): add docs.config.json file

chance-coleman:
### Description
Implement uds-docs configuration file and uds-cli landing page as well as a couple links


docs/index.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+UDS Core is the foundational runtime layer for the Unicorn Delivery Service platform. It delivers a curated,
+pre-integrated set of Kubernetes capabilities that enforce DoD and IC security requirements out of the box —
+so mission teams can focus on deploying applications rather than hardening infrastructure.
+
+## Key Capabilities
+
+<CardGrid>
+  <Card title="Security Enforcement">
+    Istio service mesh, Kyverno policy engine, NeuVector runtime security, and Keycloak-based identity
```
mjnagel: Kyverno/NeuVector are not accurate here (should be Pepr / UDS Policies and Falco).

docs/index.mdx
```diff
@@ -0,0 +1,81 @@
...
+    and access management -all pre-configured and integrated.
+  </Card>
+  <Card title="Observability">
+    Grafana, Prometheus, Loki, and Tempo provide metrics, logs, and traces across the cluster with
```
mjnagel: Tempo is not part of core. Missing Vector here (and Blackbox if we want to include everything).

docs/index.mdx
mjnagel: This also appears in other parts of the doc.

docs/index.mdx
```diff
@@ -0,0 +1,81 @@
+---
```
mjnagel, line 1: Any reason we wouldn't want to just move the existing `overview/uds-core.mdx` doc? It might be a bit more verbose (and we could add the logo in) but should be accurate and have most of the content we need already.

chance-coleman: this is now using the original uds-core.md file as the index.mdx.

mjnagel (approved):

PR #2477, opened 2026-03-16
chore(docs): cleanup promql and logql code blocks

chance-coleman:
## Description
UDS Docs `npm run build` has a few warnings about unsupported code block languages for:
- `logql`
- `promql`

This PR switches those to use the `text` language, which is what astro defaults to in the docsite.

## Related Issue

Fixes [core-426](https://linear.app/defense-unicorns/issue/CORE-426/cleaun-up-all-the-warnings-in-uds-docs-build)

## Type of change

## Checklist before merging


mjnagel (commented):
Approving for now - I wonder if we may want to leave these though and add/implement the grammars for these through expressive code config - https://expressive-code.com/key-features/syntax-highlighting/#configuration

mjnagel (approved):

PR #2479, opened 2026-03-16
feat: add keycloak realm display name customization

rodneyxr:
## Description

Currently, we allow users to override some theme settings such as the images displayed, registration fields, terms and conditions, etc. But there is no way to override the Realm's display name which shows up in the browser's page title tab when at the login page. To accomplish this today we would need to manually change the realm display name or use IaC or custom scripts to modify after deployment.

We should be able to customize the display name to allow full customization of the login page. To do this, I suggest 2 options.
1. Add a new option (`.Values.realmInitEnv.DISPLAY_NAME` which gets mapped to `REALM_DISPLAY_NAME` in the `uds-identity-config` image for realm init) to customize the realm name at install time. This would not affect existing installs.
2. Add a new option (`Values.themeCustomizations.settings.realmDisplayName`) to "dynamically" change the theme that allows overriding to a different display name (instead the "true realm name"). This would allow changing across upgrades.

This change adds the ability to set the `OVERRIDE_REALM_DISPLAY_NAME`  in core-identity-config via the new `.Values.themeCustomizations.settings.realmDisplayName` value.

It also adds a commented out value (`DISPLAY_NAME`) under the `realmInitEnv` in the `values.yaml`.

## Related Issue

Relates to https://github.com/defenseunicorns/uds-identity-config/pull/820

## Type of change

## Steps to Validate

## Checklist before merging


src/keycloak/chart/templates/secret-kc-realm.yaml
```diff
@@ -36,4 +36,5 @@ data:
   MFA_FLOW_ENABLED: {{ ternary "REQUIRED" "DISABLED" (or .Values.realmAuthFlows.OTP_ENABLED .Values.realmAuthFlows.WEBAUTHN_ENABLED) | b64enc }}
   ENABLE_REGISTRATION_FIELDS: {{ .Values.themeCustomizations.settings.enableRegistrationFields | toString | b64enc }}
   ENABLE_REGISTRATION_ACCESS_REQUEST_NOTES: {{ .Values.themeCustomizations.settings.enableAccessRequestNotes | toString | b64enc }}
+  OVERRIDE_REALM_DISPLAY_NAME: {{ .Values.themeCustomizations.settings.realmDisplayName | toString | b64enc }}
```
slaskawi: We probably don't need `OVERRIDE_` word here.

src/keycloak/chart/values.yaml
```diff
@@ -80,6 +81,8 @@ themeCustomizations:
     enableRegistrationFields: true
     # Enable Access Request Notes field on the registration page
     enableAccessRequestNotes: false
+    # Override realm display name (shown in the web page title; defaults to keycloak's realm display name if not set)
+    # realmDisplayName: "Unicorn Delivery Service"
```
slaskawi, line 87: The `Unicorn Delivery Service` seems like a valid name (this is what we use today). Could we uncomment it and leave it turned on?

slaskawi (commented):
Great work @rodneyxr ! 

I added two small comments and we should be ready to go!

src/keycloak/chart/templates/secret-kc-realm.yaml
rodneyxr: I wanted to leave the `OVERRIDE_` off but then it steps on the toes of the `DISPLAY_NAME` variable which gets prepended with `REALM_` in the realmInitEnv and would end up overriding the page title and realm init values.

- `OVERRIDE_REALM_DISPLAY_NAME` was meant to only allow overriding to a different display text (instead the "true realm name"). This would allow changing across upgrades.
- `REALM_DISPLAY_NAME` is meant to customize the realm display name at initial install time and not affect existing installs.

I could change it to be `REALM_DISPLAY_NAME` but then you lose the ability to specify them independently of each other.

src/keycloak/chart/values.yaml
rodneyxr: I would recommend against this because if we do this, it may be unintentionally override the display name if they change the display name via realm init variables. 

For example, if a user were to deploy with `My Cool Realm` in the realm init config, the actual realm value in Keycloak would be `My Cool Realm`. However, if they forget to also update the theme value, it would still display `Unicorn Delivery Service` in the login page title.

src/keycloak/chart/templates/secret-kc-realm.yaml
slaskawi: Aha! Thanks for explaining! I believe I understand the distinction now!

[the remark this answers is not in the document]
slaskawi: This probably is acceptable as it follows the convention we do with other variables.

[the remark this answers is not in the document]
slaskawi: A small Helm Chart test might be a good idea here.

src/keycloak/chart/values.yaml
slaskawi: Thanks for explaining! It makes sense!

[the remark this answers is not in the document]
slaskawi: This might be a good suggestion to add.

slaskawi (commented):
Overall LGTM! Could you please add a small Helm Chart test and look at the Github Copilot comments? Once addressed - I'll approve and merge this one.

src/keycloak/chart/values.yaml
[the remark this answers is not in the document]
rodneyxr: I've added some extra comments as per your suggestion.

src/keycloak/chart/templates/secret-kc-realm.yaml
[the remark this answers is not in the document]
rodneyxr: I have wrapped the variable in a `with` block to only render it when it is set. I've also added `src/keycloak/chart/tests/kc_realm_display_name_test.yaml` to test this.

slaskawi (dismissed):

slaskawi (commented):
LGTM

mjnagel (approved):

mjnagel (changes requested):
Could I ask for some docs updates to accompany this:
- https://github.com/defenseunicorns/uds-core/blob/override-keycloak-realm-display-name/docs/reference/configuration/identity-and-authorization.md: add the realm init variable to the `Realm initialization variables
` table/section, add the theme var to the `themeCustomizations.settings` table/section
- https://github.com/defenseunicorns/uds-core/blob/override-keycloak-realm-display-name/docs/how-to-guides/identity-and-authorization/customize-branding.mdx: add an optional step to customize the realm display name using this new value

mjnagel:
We'll also need to pull in the identity-config version bump when that is all released to make this work e2e, but that is not strictly required by this PR.

Note: There's a current CI problem with the upgrade jobs that is being resolved separately - those failures can be ignored for now.

docs/reference/configuration/identity-and-authorization.md
```diff
@@ -49,6 +49,7 @@ Bundle override path: `overrides.keycloak.keycloak.values[].path: realmInitEnv`
 | `OPENTOFU_CLIENT_ENABLED` | `false` | Enable the `uds-opentofu-client` Keycloak client for programmatic realm management |
 | `SECURITY_HARDENING_ADDITIONAL_PROTOCOL_MAPPERS` | `""` | Comma-separated additional Protocol Mappers to allow in the UDS client policy |
 | `SECURITY_HARDENING_ADDITIONAL_CLIENT_SCOPES` | `""` | Comma-separated additional Client Scopes to allow in the UDS client policy |
+| `DISPLAY_NAME` | unset | The display name for the realm. |
```
mjnagel: I think this should list the _actual_ default (not just the helm chart default), which would be `"Unicorn Delivery Service"`

rodneyxr: That makes sense! Change has been made.

mjnagel (dismissed):

mjnagel (approved):

PR #2483, opened 2026-03-17
chore!: remove keycloak FIPS switch

briantwatson:
BREAKING CHANGE: Removed `fipsAllowWeakPasswords`  and previously deprecated `fips` value from Keycloak helm chart. FIPS mode is now enabled by default. Users should remove any existing overrides utilizing the removed values.  See the [documentation](https://github.com/defenseunicorns/uds-core/blob/main/docs/how-to-guides/identity-and-authorization/enable-fips-mode.mdx)  for more details on handling password upgrades if you were not running in FIPS mode previously.

## Description

- Remove keycloak FIPS switch
- Update DEPRECATIONS.md

## Related Issue

Fixes # CORE-403

## Type of change

## Steps to Validate
- Ran e2e tests and helm unit tests

## Checklist before merging


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -99,21 +99,10 @@ spec:
             {{- end }}
             # This will only import the realm if it does not exist
             - "--import-realm"
-            # FIPS Mode
-            {{- if .Values.fips }}
-            # Full configuration might be found at https://www.keycloak.org/server/fips
+            # @lulaStart e7a920f8-e6e2-4a97-80a0-edda1fab0f8b
             - "--features=preview,fips"
             - "--fips-mode=strict"
```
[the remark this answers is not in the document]
briantwatson: Added helm-unittest

DEPRECATIONS.md
```diff
@@ -29,4 +28,4 @@ This section lists features that were removed in recent major releases for histo
 
 | Feature | Deprecated In | Removed In | Migration |
 |---------|---------------|------------|-----------|
-| _None_ | - | - | - |
+| Keycloak `fips` helm value | 0.43.0 | 1.0.0 | FIPS mode is now always enabled; remove any `fips` overrides from your values |
```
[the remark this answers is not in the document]
briantwatson: I believe we want this 1.0.0

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -99,21 +99,10 @@ spec:
...
-            {{- if .Values.fipsAllowWeakPasswords}}
```
mjnagel, line 107: @slaskawi / @chance-coleman do we want to remove the `fipsAllowWeakPasswords` value as well? I don't think we previously deprecated it explicitly and I think it might've been a separate case from fips toggle itself...but may be misremembering here.

src/keycloak/chart/templates/secret-kc-realm.yaml
```diff
@@ -36,4 +36,4 @@ data:
   MFA_FLOW_ENABLED: {{ ternary "REQUIRED" "DISABLED" (or .Values.realmAuthFlows.OTP_ENABLED .Values.realmAuthFlows.WEBAUTHN_ENABLED) | b64enc }}
   ENABLE_REGISTRATION_FIELDS: {{ .Values.themeCustomizations.settings.enableRegistrationFields | toString | b64enc }}
   ENABLE_REGISTRATION_ACCESS_REQUEST_NOTES: {{ .Values.themeCustomizations.settings.enableAccessRequestNotes | toString | b64enc }}
-  FIPS_ENABLED: {{ .Values.fips | toString | b64enc }}
+  FIPS_ENABLED: {{ "true" | b64enc }}
```
slaskawi, line 42: The reason this switch is here is that it control whether FIPS Bouncy Castle libraries should be copied to Keycloak (see https://github.com/defenseunicorns/uds-identity-config/blob/main/src/sync.sh#L22).

Copying these libraries is required by all the flavors as of today. So I would prefer to remove this flag altogether from here as UDS Identity Config.

src/keycloak/chart/templates/statefulset.yaml
slaskawi: This toggle controls if extra bits should be added to the short passwords making it artificially FIPS compliant. I do agree we haven't deprecated it, so I'm on the fence if we should remove it. 

My personal leaning would be to go ahead and clean it up. If someone really really complains, we can re-add it (with a better name) in UDS Core 1.x.y

slaskawi (commented):
Great work @briantwatson ! 

I suggested 1 change. It's also worth calling out that UDS Identity Config contains a few places where we also need to remove the FIPS mode.

src/keycloak/chart/templates/secret-kc-realm.yaml
briantwatson: Thanks for the review and reference!  Removed.

mjnagel: If I'm not mistaken we need an "upstream" PR to identity-config to strip this out of [the script](https://github.com/defenseunicorns/uds-identity-config/blob/main/src/sync.sh#L22) (so that it unconditionally copies the fips jar)? If we go that route (now) we'd need to make the changes there, release identity-config, then press with this PR.

Alternatively could do this in 2 phases:
- this PR with `FIPS_ENABLED` hardcoded to true
- followup work to remove it in identity-config

src/keycloak/chart/templates/statefulset.yaml
mjnagel: I'm good with this - put out a quick question to some known users to check. Let's make sure we explicitly note this in the removal table on our deprecation page though.

src/keycloak/chart/templates/secret-kc-realm.yaml
briantwatson: Came here to comment on that after circling back looking at Failing CI tests:

Digging into Keycloak logs from logs artifact download
```
2026-03-18T16:09:54.958944858Z stdout F Running the server in development mode. DO NOT use this configuration in production.
2026-03-18T16:09:56.47259856Z stderr F ERROR: Failed to start server in (development) mode
2026-03-18T16:09:56.472830833Z stderr F ERROR: Failed to start quarkus
2026-03-18T16:09:56.472837506Z stderr F ERROR: Failed to configure FIPS. Make sure you have added the Bouncy Castle FIPS dependencies to the 'providers' directory.
```

src/keycloak/chart/templates/statefulset.yaml
briantwatson: Added mention to this value specifically

src/keycloak/chart/templates/secret-kc-realm.yaml
slaskawi: Sorry I wasn't clear in the first place. 

By asking to remove this flag I mean going through the entire cycle:
- Setting it `true` for now
- Removing this flag in UDS Identity Config
- Cutting new UDS Identity Config release
- When the new release is pulled into the UDS Core, removing this flag entirely

It will be a series of PRs and we should definitely keep it set to true for this PR. We will remove it in a subsequent release.

slaskawi (commented):
LGTM once conflicts are resolved and CI is green

DEPRECATIONS.md
```diff
@@ -14,18 +14,19 @@ When adding a deprecation, include:
...
+| Feature                                                     | Deprecated In | Removed In | Migration                                                                                                                                                             |
+| ----------------------------------------------------------- | ------------- | ---------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Keycloak `x509LookupProvider`, `mtlsClientCert` helm values | 0.47.0        | 1.0.0      | Use `thirdPartyIntegration.tls.tlsCertificateHeader` and `thirdPartyIntegration.tls.tlsCertificateFormat`; remove any existing overrides utilizing the removed values |
+| Keycloak `fips` helm value                                  | 0.43.0        | 1.0.0      | FIPS mode is now always enabled; remove any `fips` overrides from your values including `fipsAllowWeakPasswords`.                                                     |
```
mjnagel: We could also link https://github.com/defenseunicorns/uds-core/blob/main/docs/how-to-guides/identity-and-authorization/enable-fips-mode.mdx here which provides a nice guide on password handling if someone is behind the times.

mjnagel (commented):
LGTM minus one docs comment and the deprecation doc conflict fix.

Also similar question/comment on the `refactor` commit title and adding a `BREAKING CHANGE` description (see https://github.com/defenseunicorns/uds-core/pull/2494#pullrequestreview-3977859431).

DEPRECATIONS.md
mjnagel: (would be great if we could link this on the actual docs site, but will need that to be served on the final domain/subdomain first)

DEPRECATIONS.md
```diff
@@ -16,18 +16,18 @@ When adding a deprecation, include:
...
-| `operator.KUBEAPI_CIDR`, `operator.KUBENODE_CIDRS` | 0.48.0 ([#1233](https://github.com/defenseunicorns/uds-core/pull/1233)) | **Reason:** Moved to ClusterConfig CRD.<br/>**Migration:** Use `cluster.networking.kubeApiCIDR` and `cluster.networking.kubeNodeCIDRs` instead | 1.0.0 |
-| `sso.secretName`, `sso.secretLabels`, `sso.secretAnnotations`, `sso.secretTemplate` | 0.60.0 ([#2264](https://github.com/defenseunicorns/uds-core/pull/2264)) | **Reason:** Simplified field structure.<br/>**Migration:** Use `sso.secretConfig.name`, `.labels`, `.annotations`, `.template` instead | Package `v1beta1` |
+| Feature                                                                             | PR                                                                      | Deprecated In                                                                                                                                            | Reason            | Migration | Removal Target |
```
mjnagel: I think the titles got messed up here (Details is the newer form, Reason/Migration is what we moved away from).

mjnagel (approved):

mjnagel:
Adding a BREAKING CHANGE description (see https://github.com/defenseunicorns/uds-core/pull/2494#pullrequestreview-3977859431) would be nice to format that a bit prettier for changelog. LGTM for actual PR content though!

PR #2491, opened 2026-03-18
chore(docs): add how to guides for additional core packages

chance-coleman:
## Description
- Adds **Register and customize SSO clients** how-to guide covering native OIDC/SAML client registration, secret customization, protocol mappers, client attributes, and JWT-based authentication
- Adds **Configure advanced HTTP routing** how-to guide covering match conditions, CORS policies, header manipulation, URI rewrites, redirects, direct responses, retries, and timeouts via the `advancedHTTP` Package CR field
- Wires both guides into their section overviews and existing related guides (`protect-apps-with-authservice`, `expose-apps-on-gateways`, `configure-pod-reload`)

## What's not included

Service mesh mode (`spec.network.serviceMesh.mode`) was considered for a third guide but omitted, the concepts page covers it adequately and the configuration is a single field.

## Related Issue

Fixes core-419

## Type of change

## Checklist before merging


docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
```diff
@@ -0,0 +1,398 @@
...
+           labels:
+             app.kubernetes.io/part-of: my-app
+           template:
+             config.json: |
+               {
+                 "client_id": "clientField(clientId)",
+                 "client_secret": "clientField(secret)",
+                 "issuer": "https://sso.uds.dev/realms/uds",
+                 "redirect_uris": clientField(redirectUris).json()
+               }
```
mjnagel, line 141: I think I'm good with this, but is there any value in including multiple examples (formats) similar to our old doc - https://uds.defenseunicorns.com/reference/configuration/single-sign-on/sso-templating/

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
```diff
@@ -0,0 +1,398 @@
...
+         clientId: my-app
+         clientAuthenticatorType: "client-jwt"
+         redirectUris:
+           - "https://my-app.uds.dev/auth/callback"
+   ```
+
+   With `client-jwt`, the application must present a signed JWT (using its private key) when requesting tokens. The operator still creates a Kubernetes secret, but the `secret` key is not used for authentication — the application uses its private key instead.
+
+   > [!NOTE]
+   > The Package CR only sets the authenticator type. Registering the client's public key or JWKS endpoint in Keycloak is a separate step — use the Keycloak admin UI or [manage Keycloak with OpenTofu](/how-to-guides/identity-and-authorization/manage-keycloak-with-opentofu/) to add the key after deployment.
```
mjnagel: Is this something we want to document (at least in this tutorial)? I feel like it could be useful but maybe goes beyond the basic tutorial here and could be a future separate tutorial (or reference material) as we understand/better support jwt client auth?

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
```diff
@@ -0,0 +1,398 @@
...
+
+6. **Deploy the Package CR**
```
mjnagel: Do we want to include a step after this for "configure your application to use the client"? I know it would be extremely generic but could just stub it out as "review your application documentation for how to configure SSO, use the client id/secret...etc".

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
```diff
@@ -0,0 +1,398 @@
...
+If your application expects credentials in a specific format (JSON config file, properties file, etc.), you can use `secretConfig.template` to control the secret layout.
+
+## Steps
+
+<Steps>
+
+1. **Register the SSO client in a Package CR**
+
+   <Tabs>
+     <TabItem label="OIDC client">
```
mjnagel, line 43: Is it worth any note here before the tabs about why you might choose SAML over OIDC? Or at minimum just note that you should pick whatever is supported by your app? For ref: uds-common [recommends SAML](https://github.com/defenseunicorns/uds-common/blob/4bcc9f30e65b80b51ce3cd88a59d3f219032ea64/docs/uds-packages/requirements/uds-package-requirements.md?plain=1#L32) - I think this is in our new packaging docs as well.

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
```diff
@@ -0,0 +1,398 @@
...
+             saml_single_logout_service_url_post: "https://my-saml-app.uds.dev/saml/slo"
+             saml_name_id_format: "username"
+             saml.assertion.signature: "true"
+             saml.client.signature: "false"
+     ```
+
+     The operator creates a SAML client in Keycloak and includes the IdP signing certificate as `samlIdpCertificate` in the generated Kubernetes secret. Your application uses this certificate to validate SAML assertions from Keycloak.
+
+     > [!NOTE]
+     > SAML clients do not use `redirectUris`. Set `standardFlowEnabled: false` to avoid the redirect URI validation. Configure assertion consumer and logout URLs via the `attributes` block instead.
```
mjnagel: Is this universally true? I see our [gitlab package](https://github.com/uds-packages/gitlab/blob/main/charts/config/templates/uds-package.yaml#L66-L69) sets redirect URIs + SAML (and standard flow) - I think that's actually been common across the SAML apps I've seen?

docs/how-to-guides/networking/configure-advanced-http-routing.mdx
```diff
@@ -0,0 +1,409 @@
...
+The `advancedHTTP` block is a property of each entry in the `spec.network.expose[]` array. It maps directly to [Istio VirtualService HTTPRoute](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRoute) fields. The UDS Operator translates your `advancedHTTP` configuration into the generated VirtualService for that expose entry.
+
+All `advancedHTTP` features are composable — you can combine match conditions, headers, retries, and timeouts in a single expose entry. However, `redirect` and `directResponse` are terminal actions: when either is set, traffic is not forwarded to the backend service.
+
+> [!CAUTION]
+> `advancedHTTP` cannot be used with the passthrough gateway. Passthrough gateways forward raw TLS without terminating it, so HTTP-level routing is not possible.
+
+## Steps
+
+<Steps>
```
mjnagel: This is maybe less a comment on this specific PR but more our formula/structure for how-tos: I think the steps sometimes feel forced because things like this are really a number of different settings/options more than sequential things to do. They _could be_ sequential in some cases (enabling matches, then header rules, etc) but not always. I don't know if that's an indication this content is better served as reference content or we need to be looser with our structure?

docs/how-to-guides/networking/configure-advanced-http-routing.mdx
```diff
@@ -0,0 +1,409 @@
+---
+title: Configure advanced HTTP routing
+sidebar:
+  order: 2.11
+---
```
mjnagel: Mixed feelings on this doc in general after reviewing - wondering if we could/should just add a note block in the existing `expose applications` doc and link to upstream istio docs for this (since everything just passes through to Istio directly)? Okay either way - just wanted to float that for consideration.

chance-coleman: I'm leaning towards folding this into the expose applications guide. Add an advancedHTTP section at the end of the existing expose-apps-on-gateways.mdx with one or two brief examples (e.g., match + headers) showing how to use it in a Package CR, then link out to the [Istio VirtualService HTTPRoute docs](https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPRoute) for the full field reference. Delete the standalone doc.

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
chance-coleman: Expanded the single JSON template into a single block showing four formats (raw key-value, JSON, properties, YAML)

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
chance-coleman: Removed the entire JWT step.

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
chance-coleman: brief generic stub: review your app's docs, reference the generated secret name, issuer URL, and samlIdpCertificate for SAML.

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
chance-coleman: Added one sentence before the tabs

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
chance-coleman: The note was rewritten to explain: SAML clients require redirectUris when standardFlowEnabled is true (the default); standardFlowEnabled: false is only for IdP-initiated SSO. Troubleshooting was updated to say "IdP-initiated SAML clients" instead of "SAML clients." Also added a SAML end-to-end test to the Verification section since it previously only covered OIDC.

docs/how-to-guides/networking/configure-advanced-http-routing.mdx
chance-coleman: OBE

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
```diff
@@ -37,6 +37,8 @@ If your application expects credentials in a specific format (JSON config file,
 
 1. **Register the SSO client in a Package CR**
 
+   Choose the protocol supported by your application. If your application supports both, [UDS package requirements](https://github.com/defenseunicorns/uds-common/blob/main/docs/uds-packages/requirements/uds-package-requirements.md) recommend considering SAML with SCIM as the more secure default.
```
mjnagel: nit: this link can be local to the docs site - https://github.com/defenseunicorns/uds-core/blob/main/docs/concepts/configuration-and-packaging/package-requirements.mdx

mjnagel (dismissed):

mjnagel (approved):

PR #2500, opened 2026-03-18
chore(docs): add likec4 diagram

chance-coleman:
## Description
Add back the main likec4 diagram. Also remove all other diagrams that are unused, i kept the indepth one since that could have some uses in referencing for support and otherwise.

Open to suggestions for other places this should live.

Also the diagram is quite small on some screens, so added a callout above the diagram as a way of tell users to hopefully interact with it and know that when clicked it fills the screen and allows for the fancy browsing features of likec4.

## Related Issue

Fixes [core-449](https://linear.app/defense-unicorns/issue/CORE-449/add-likec4-diagram)

## Type of change

## Checklist before merging


joelmccoy (commented):
If we are deleting the pngs... do we need to remove the views in likec4 so that these don't get autogenerated by our task again?

chance-coleman:
I don't think we autogen the diagrams anymore. We used to but then stopped because they kept changing for no reason.

joelmccoy:
> I don't think we autogen the diagrams anymore. We used to but then stopped because they kept changing for no reason.

I guess I am more talking about when we need to regenerate diagrams locally (if there are changes) with `uds run -f tasks/diagrams.yaml c4-update` it will add back in all these diagrams. Would suggest we update that task to exclude these? Or remove from the views in c4? Or even just leave them in.

joelmccoy (dismissed):
Changes LGTM. It might be worth addressing this comment though: https://github.com/defenseunicorns/uds-core/pull/2500#issuecomment-4086098983

mjnagel (approved):

PR #2503, opened 2026-03-19
fix: incr loki backend replicas to 3 in upgrade test

jasonwashburn:
## Description

Fixes issue where upgrade tests will fail when loki-backend stateful set is set to 1 replica. When deployed by itself, the loki-backend pod will hang when terminating (i.e. due to upgrade or restart) unless the loki-read is also cycled. If it does not terminate correctly, and is force killed, it leaves loki in a bad state and stops working correctly causing e2e tests to fail. 

In this particular instance, there were small, unrelated changes made to the loki-backend statefulset which triggered a restart which got hung as described above and caused the tests to fail.

Verified this behavior existed prior to recent helm chart changes but was likely masked previously by the fact that normally the loki pods are all cycling out at the same time instead of just the backend by itself.

We fix this by setting loki-backend replicas to 3 during the upgrade test. The upgrade test task was also refactored to use a default uds-upgrade-test-config.yaml rather than command line args to ensure consistent settings are used for both the initial deployment of the previous release in `deploy:latest-bundle-release` as well as during deployment of the current version.

Loki read, write, and backend replicas were set to 3 in the HA upgrade test to make it more representative of an HA configuration.

Additional commit title to include during merge for release-please:
chore: incr loki backend/read/write replicas to 3 in ha upgrade test
chore: switch loki helm chart to grafana-community chart
chore: bump loki helm chart to 6.57.0

## Related Issue

Fixes # CORE-451

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Steps to Validate
- Run  `uds run test-uds-core-upgrade`

## Checklist before merging

BEGIN_COMMIT_OVERRIDE
fix: incr loki backend replicas to 3 in upgrade test
chore: incr loki backend/read/write replicas to 3 in ha upgrade test
chore: switch loki helm chart to grafana-community chart
chore: bump loki helm chart to 6.57.0
END_COMMIT_OVERRIDE

### DIFF
--- a/bundles/k3d-standard/uds-ha-config.yaml
+++ b/bundles/k3d-standard/uds-ha-config.yaml
@@ -7,6 +7,10 @@ variables:
     AUTHSERVICE_REDIS_URI: redis://authservice:authservice@host.k3d.internal:6379
     AUTHSERVICE_REPLICA_COUNT: 2
 
+    # Enable Falco sandbox and incubating rules
+    FALCO_SANDBOX_RULES_ENABLED: true
+    FALCO_INCUBATING_RULES_ENABLED: true
+
     # Grafana variables
     grafana_ha: true
     grafana_pg_host: host.k3d.internal
--- a/bundles/k3d-standard/uds-ha-config.yaml
+++ b/bundles/k3d-standard/uds-ha-config.yaml
@@ -3,13 +3,9 @@
 
 variables:
   core:
-    # Keycloak variables
-    keycloak_ha: true
-    keycloak_pg_username: postgres
-    keycloak_pg_password: unicorn123!@#UN
-    keycloak_pg_database: keycloak
-    keycloak_pg_host: host.k3d.internal
-    keycloak_devmode: false
+    # Authservice variables
+    AUTHSERVICE_REDIS_URI: redis://authservice:authservice@host.k3d.internal:6379
+    AUTHSERVICE_REPLICA_COUNT: 2
 
     # Grafana variables
     grafana_ha: true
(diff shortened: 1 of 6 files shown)


mjnagel (dismissed):
Approved pending CI.

bundles/k3d-standard/uds-ha-config.yaml
```diff
@@ -3,13 +3,9 @@
...
-    # Keycloak variables
-    keycloak_ha: true
-    keycloak_pg_username: postgres
-    keycloak_pg_password: unicorn123!@#UN
-    keycloak_pg_database: keycloak
-    keycloak_pg_host: host.k3d.internal
-    keycloak_devmode: false
+    # Authservice variables
+    AUTHSERVICE_REDIS_URI: redis://authservice:authservice@host.k3d.internal:6379
+    AUTHSERVICE_REPLICA_COUNT: 2
```
mjnagel, line 9:
```suggestion

    # Enable Falco sandbox and incubating rules
    FALCO_SANDBOX_RULES_ENABLED: true
    FALCO_INCUBATING_RULES_ENABLED: true

```

jasonwashburn: good catch...should have been added here instead of private-pki-config...fixed

mjnagel (approved):

src/loki/common/zarf.yaml
```diff
@@ -15,8 +15,8 @@ components:
         version: 0.2.0
         localPath: ../chart
       - name: loki
-        url: https://grafana.github.io/helm-charts/
-        version: 6.55.0
+        url: https://grafana-community.github.io/helm-charts/
```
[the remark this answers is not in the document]
jasonwashburn: we'll catch that one next time Copilot...

PR #2526, opened 2026-03-24
chore(deps): update UDS CLI to 0.30.0, Zarf init to 0.74.0

mjnagel:
## Description

This PR primarily updates to UDS CLI 0.30.0. Along with the necessary changes to do this (some nontrivial operator changes), this PR also adds a CLI "matrix" of testing to test older CLI installs and upgrade scenarios. Full changes included:
- bump CLI to 0.30.0, zarf init to 0.74.0
- disable SSA for istio base/istiod chart due to [upstream issue](https://github.com/istio/istio/issues/59487)
- switch authservice restart annotation from a patch to SSA
- update istio namespace handling to cleanup "overmanaged fields" and only perform sparse applies
- update keycloak client secret to a full SSA apply
- update pod reloads to cleanup "overmanaged fields" and only perform sparse applies
- add CLI "matrix" testing across versions and upgrade scenarios (some task changes to support this)

To provide a bit more explanation around the cleanup "overmanaged fields" and only perform sparse applies...in those two spots previously in our operator code we were not handling Apply properly, applying the full resource (ex: deployment in reload code) with a Force. With KFC performing SSA this results in our "pepr" manager "owning" all fields on the resource. While not an issue historically, this causes a problem with the new versions of helm/zarf/cli. Helm upgrades to these same resources see a field conflict because we "overclaimed" (rather than only applying the single annotation we needed). To handle this for end users, this PR adds code to "remove over management" by scaling back our ownership to only what we expect to own for the given resource. This requires precise ordering - if subsequent `Apply`s only contain a label/annotation, this can result in other fields we "own" getting removed. Current code addresses this by adding retry handling and ultimately throwing errors on reconcile if the cleanup does not work.

Alternatively to this whole approach...in theory on the first upgrade where someone switches to SSA they could use Zarf's `--force-conflicts` option and have Zarf reclaim ownership of all expected fields. But...this is not end user friendly, and means that we would have to continue with our "overclaiming" since we can't safely sparsely apply.

## Related Issue

Fixes CORE-457

## Type of change

## Steps to Validate

CI should largely validate the 3 key paths:
- old -> old CLI upgrade
- clean install with old
- old -> new CLI upgrade
- new -> new CLI upgrade (baseline testing we already had)

## Checklist before merging


.github/workflows/test-cli-matrix.yaml
```diff
@@ -0,0 +1,138 @@
...
+        include:
+          # Both versions intentionally frozen on last compatible CLI version
+          - scenario: old-old
+            flavor: upstream
+            install_cli: v0.29.0
+            upgrade_cli: v0.29.0
+          # install_cli intentionally frozen (last compatible version); upgrade_cli tracks new releases via Renovate
+          - scenario: old-new
+            flavor: upstream
+            install_cli: v0.29.0
```
mjnagel, line 45: We could definitely set this further back. Right now this is just the oldest version before the SSA change to provide coverage of pre-SSA -> SSA world.

Mostly wanted to address the immediate issue I was working on and add some CI checks around this particular upgrade. There is a [separate ticket](https://linear.app/defense-unicorns/issue/CORE-455/test-core-against-older-cli-versions) where I think we could better design what this workflow should look like in the future and identify specific older version(s) to target here.

.github/workflows/test-cli-matrix.yaml
```diff
@@ -0,0 +1,138 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Environment setup
+        uses: ./.github/actions/setup
+        with:
+          registry1Username: ${{ secrets.IRON_BANK_ROBOT_USERNAME }}
+          registry1Password: ${{ secrets.IRON_BANK_ROBOT_PASSWORD }}
+          ghToken: ${{ secrets.GITHUB_TOKEN }}
+          rapidfortUsername: ${{ secrets.RAPIDFORT_USERNAME }}
+          rapidfortPassword: ${{ secrets.RAPIDFORT_PASSWORD }}
```
joelmccoy, line 60: nit: since we are running upstream only... would it maybe make sense to just move the k3d setup into setup-tools and just call that? ([ref](https://github.com/defenseunicorns/uds-core/blob/25bd0e78568e6c738c14d7881d89644e950deb38/.github/actions/setup/action.yaml#L30-L38)). This would minimize giving this workflow ironbank/unicorn creds (maybe a good idea given all the recent github supply chain things)

.github/workflows/test-cli-matrix.yaml
```diff
@@ -0,0 +1,138 @@
...
```
joelmccoy, line 115: nit: since we are running upstream only... would it maybe make sense to just move the k3d setup into setup-tools and just call that? (ref). This would minimize giving this workflow ironbank/unicorn creds (maybe a good idea given all the recent github supply chain things)

src/pepr/operator/controllers/reload/reload-utils.spec.ts
```diff
@@ -545,4 +553,497 @@ describe("restartController", () => {
...
+      metadata: {
+        name: "test-rs",
+        ownerReferences: [{ kind: "Deployment", name: "test-dep", controller: true }],
+      },
+    });
+    // Second Get: the Deployment (narrow entry, no cleanup needed)
+    mockK8sClient.Get.mockResolvedValueOnce({
+      ...makeTestDeployment(),
+      metadata: { name: "test-dep", namespace: "default", uid: "uid", managedFields: [] },
+    });
```
joelmccoy: nit: i think this would be a better test if we actually triggered a cleanup on the deployment and checked those calls

src/pepr/operator/controllers/reload/reload-utils.spec.ts
```diff
@@ -545,4 +553,497 @@ describe("restartController", () => {
...
+    mockK8sClient.Get.mockResolvedValueOnce({
+      ...makeTestReplicaSet(),
+      metadata: {
+        name: "test-rs",
+        namespace: "default",
+        uid: "uid",
+        ownerReferences: [],
+        managedFields: [],
+      },
+    });
```
joelmccoy, line 1022: nit: i think this would be a better test if we had an over claimed entry here and check that it was removed

src/pepr/operator/controllers/istio/namespace.ts
```diff
@@ -211,24 +256,39 @@ export async function applyNamespaceUpdates(
   annotations: Record<string, string>,
   originalLabels: Record<string, string> | undefined,
   originalAnnotations: Record<string, string> | undefined,
+  managedFields: V1ManagedFieldsEntry[] | undefined,
```
joelmccoy, line 261: nit: update tsdoc here with this field

joelmccoy (commented):
Generally LGTM at first pass. couple nits for consideration.  I do want to play around with this locally a bit some more before final approval

.github/workflows/test-cli-matrix.yaml
mjnagel: Hm, yeah I can do that since this is a net-new workflow. I wasn't sure if there would be other opinions on flavor coverage here - largely I think this kind of testing is "independent" of flavor, but wasn't sure if there'd be other opinions.

This would be a bit of refactor to the action/other workflows but can do this if we want.

joelmccoy: non blocking and don't feel too strongly about this refactor. feel free to dismiss

mjnagel: Went ahead and opened [CORE-468](https://linear.app/defense-unicorns/issue/CORE-468/refactor-setup-action-to-reduce-credential-usage) based on this - I see value here just would prefer to defer that to a separate PR to not overscope this one.

joelmccoy (dismissed):

.github/workflows/test-cli-matrix.yaml
```diff
@@ -0,0 +1,138 @@
...
+    - cron: "0 2 * * *"
+  pull_request:
+    # milestoned is added here as a workaround for release-please not triggering PR workflows (PRs should be added to a milestone to trigger the workflow).
+    # labeled is added to support renovate-ready labelling on PRs
+    types: [milestoned, labeled, opened, reopened, synchronize]
+    paths:
+      - ".github/workflows/test-cli-matrix.yaml"
+      - "bundles/k3d-standard/**"
+
+# Abort prior jobs in the same workflow / PR; use ref for PRs, sha for scheduled runs
```
chance-coleman: maybe im missing something here, is this comment correct? **I think** that the implementation only uses `github.ref`?

src/pepr/operator/controllers/ssa-field-cleanup.ts
```diff
@@ -0,0 +1,50 @@
...
+export async function removePeprManagedFieldsEntry(
+  resourceKind: Parameters<typeof K8s>[0],
+  name: string,
+  namespace: string | undefined,
+  managedFields: V1ManagedFieldsEntry[],
+  log: Logger,
+): Promise<void> {
+  const idx = managedFields.findIndex(
+    e => e.manager === PEPR_FIELD_MANAGER && e.operation === "Apply",
+  );
```
chance-coleman, line 36: for my own understanding, are multiple pepr apply entries possible here?

.github/workflows/test-cli-matrix.yaml
mjnagel: Yeah I can adjust the comment - always forget exactly what `ref` evaluates to depending on the context - I think it's branch name, not sha, so will clean this up and copy from other workflows.

src/pepr/operator/controllers/ssa-field-cleanup.ts
mjnagel: Having trouble finding this in docs...I know from initial testing that a given manager can have multiple entries with unique operations. I believe that multiple would also be possible if [other fields in the spec](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#System) differed (subresource/apiversion) which I don't think are a concern with the specific fields we're modifying (although I could probably make sure this handles multiple if we think there's cases where it might hit).

src/pepr/operator/controllers/reload/reload-utils.ts
```diff
@@ -202,26 +254,27 @@ export async function restartController(
...
   }
 
-  // Get the controller resource for the event
-  async function getController() {
-    return K8s(controllerKind).InNamespace(namespace).Get(name);
-  }
-  const controller = await retryWithDelay(getController, log);
+  // Fetch the controller and clean up any stale over-claimed Pepr SSA entry.
+  // Old code applied the full controller object, stealing SSA ownership of every spec field.
+  const controller = await cleanupControllerEntry(namespace, controllerKind, name, log);
```
mjnagel, line 266: Want to eval if this is still necessary - added here originally before adding the initial call in `pod-reload` (on first checksumming) which might make this call unnecessary.

src/pepr/operator/controllers/reload/pod-reload.ts
```diff
@@ -178,8 +178,23 @@ export async function handleResourceUpdate(
...
   checksumCache.set(cacheKey, currentChecksum);
 
-  // If this is the first time we're seeing this resource, or if the data hasn't changed, exit early
-  if (!previousChecksum || previousChecksum === currentChecksum) {
+  // First time we've seen this resource — proactively clean up any over-claimed controller
+  // fields so Helm upgrades don't conflict before the first reload fires.
+  if (!previousChecksum) {
+    try {
+      const pods = await discoverResourceConsumers(namespace, name);
+      await cleanupOverClaimedControllerFields(namespace, pods, log);
```
mjnagel: Going to look at the timing/ordering here - on startup of the Pepr watcher this does end up pulling all existing "reloaded" controllers (in parallel) which could be a concern at larger cluster scale. I want to evaluate if there's a better way to not cause a lot of stress here on the API (although I think it's probably OK at our normal scale).

src/pepr/operator/controllers/ssa-field-cleanup.ts
chance-coleman: so if there is multiple entries possible, does `findIndex` only get the first one? and should it or should it get all?

mjnagel: `findIndex`/`find` does just grab the first one so in theory there is an edge case here if there were multiple. I think it requires a unique combination of properties still though, and isn't really possible in our scenario(s) here? Willing to do the refactor here if there's a concrete case we can identify where it's relevant but ultimately don't see the added code/complexity here as worth it given I'm not sure it's actually a possible scenario specific to our handling here?

src/pepr/operator/controllers/reload/reload-utils.ts
mjnagel: Revisited - I think there's some safety here since the initial cleanup _could_ fail. This also doesn't add any additional API calls so no real harm.

src/pepr/operator/controllers/ssa-field-cleanup.ts
chance-coleman: yeah agreed, i'm also unsure if this is possible. fine with closing this out.

src/pepr/operator/controllers/reload/pod-reload.ts
mjnagel: Resolved by adding some queue handling of this + tracking "state of cleanup" via annotations. Open to thoughts on approach/implementation - at scale I didn't like the idea of these extra calls running everytime we restart the watcher, so at minimum the annotation seems like a nice to have. The queue also seemed simple to implement so I did that as well, although could revert if that seems overly complex.

chance-coleman (approved):

PR #2529, opened 2026-03-24
docs: loki storage configuration reference

briantwatson:
## Description

- Adds configuration reference for loki storage
- Small adjustment to style guide following our implemented pattern

## Related Issue

Fixes CORE-459

## Type of change

## Steps to Validate
- N/A

## Checklist before merging

### DIFF
--- a/docs/reference/configuration/loki-storage.md
+++ b/docs/reference/configuration/loki-storage.md
@@ -70,10 +70,10 @@ The `loki.storage` fields control the object storage type, endpoint, credentials
 | `loki.storage.s3.secretAccessKey`  | string  | `"uds-secret"`                                        | Secret access key for S3 authentication                                                                  |
 | `loki.storage.s3.s3ForcePathStyle` | boolean | `true`                                                | Use path-style URLs instead of virtual-hosted-style; required for MinIO and most S3-compatible providers |
 | `loki.storage.s3.insecure`         | boolean | `false`                                               | Allow HTTP (non-TLS) connections to the storage endpoint                                                 |
-| `loki.storage.s3.region`           | string  | —                                                     | AWS region for the S3 bucket; required for AWS S3, not needed for MinIO                                  |
+| `loki.storage.s3.region`           | string  | unset                                                 | AWS region for the S3 bucket; required for AWS S3, not needed for MinIO                                  |
 
 > [!NOTE]
-> The defaults target the internal MinIO dev stack deployed by `uds-dev-stack`. Production deployments must override the endpoint, credentials, and bucket names to point to external object storage. UDS Core does not set a default for `bucketNames.ruler`, but the Loki chart templates reference this field for ruler storage configuration. If you use Loki's ruler with external object storage, you may need to set `loki.storage.bucketNames.ruler` to a valid bucket name.
+> The defaults target the internal MinIO dev stack deployed by `uds-dev-stack`. Production deployments must override the endpoint, credentials, and bucket names to point to external object storage. The upstream Loki chart also supports a `bucketNames.ruler` field, but UDS Core does not use it. Ruler configuration is loaded from in-cluster ConfigMaps, so overriding this field is not necessary.
 
 The following example shows a minimal production override for S3-compatible storage:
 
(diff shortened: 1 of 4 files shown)


docs/reference/configuration/loki-storage.md
```diff
@@ -0,0 +1,137 @@
...
+| `loki.storage.s3.accessKeyId`      | string  | `"uds"`                                               | Access key ID for S3 authentication                                                                      |
+| `loki.storage.s3.secretAccessKey`  | string  | `"uds-secret"`                                        | Secret access key for S3 authentication                                                                  |
+| `loki.storage.s3.s3ForcePathStyle` | boolean | `true`                                                | Use path-style URLs instead of virtual-hosted-style; required for MinIO and most S3-compatible providers |
+| `loki.storage.s3.insecure`         | boolean | `false`                                               | Allow HTTP (non-TLS) connections to the storage endpoint                                                 |
+| `loki.storage.s3.region`           | string  | —                                                     | AWS region for the S3 bucket; required for AWS S3, not needed for MinIO                                  |
```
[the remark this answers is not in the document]
mjnagel: I think we can update the guidance to use `unset` for unset defaults.

docs/reference/configuration/loki-storage.md
```diff
@@ -0,0 +1,137 @@
...
+
+> [!NOTE]
+> The defaults target the internal MinIO dev stack deployed by `uds-dev-stack`. Production deployments must override the endpoint, credentials, and bucket names to point to external object storage. UDS Core does not set a default for `bucketNames.ruler`, but the Loki chart templates reference this field for ruler storage configuration. If you use Loki's ruler with external object storage, you may need to set `loki.storage.bucketNames.ruler` to a valid bucket name.
```
mjnagel: I think for the loki ruler callout here we can reword this a bit to mention that overriding it is not necessary since ruler configuration is stored in cluster (configmaps)?

docs/reference/configuration/loki-storage.md
```diff
@@ -0,0 +1,137 @@
...
+          path: loki.storage.s3.secretAccessKey
+          sensitive: true
+```
+
+## Additional configuration
+
+Operators may need to adjust query and replication settings depending on log volume and cluster topology.
+
+| Field                                                      | Type    | Default          | Description                                                                                          |
+| ---------------------------------------------------------- | ------- | ---------------- | ---------------------------------------------------------------------------------------------------- |
```
mjnagel: This table just ever so slightly introduces scrolling due to the width I think because of `loki.query_scheduler.max_outstanding_requests_per_tenant`. I do also wonder if it makes sense to even have this table (vs just linking to the chart upstream) since some of these feel a bit cherry picked (but maybe there's specific reasons behind each one)?

mjnagel (commented):
Two small comments - LGTM overall.

docs/reference/configuration/loki-storage.md
briantwatson: Was thinking this table might be a discussion point. I too saw that scroll bar and wasn't thrilled. 

> ...maybe there's specific reasons behind each one

I had selected a few I _thought_ might be common knobs to turn that we either override in our [values.yaml](https://github.com/defenseunicorns/uds-core/blob/main/src/loki/values/values.yaml) or would just be common to tailor to a given environment (replicas). I'll drop here and we can see if it makes sense to add something like this later!

docs/reference/configuration/loki-storage.md
```diff
@@ -0,0 +1,124 @@
...
+          sensitive: true
+        # S3 secret access key
+        - name: LOKI_S3_SECRET_ACCESS_KEY
+          path: loki.storage.s3.secretAccessKey
+          sensitive: true
+```
+
+## Additional configuration
+
+UDS Core deploys Loki in `SimpleScalable` mode with a `replication_factor` of `1`. It does not override upstream chart defaults for replica counts or most query settings. For the full set of available fields, see the [upstream Loki Helm chart values](https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml). For guidance on tuning replicas and resources for production workloads, see [Configure high-availability logging](/how-to-guides/high-availability/logging/). For compactor and retention settings, see [Configure log retention](/how-to-guides/logging/configure-log-retention/).
```
mjnagel:
```suggestion
UDS Core deploys Loki in `SimpleScalable` mode with a `replication_factor` of `1`. It does not override upstream chart defaults for replica counts or most query settings. For the full set of available fields, see the [upstream Loki Helm chart values](https://github.com/grafana-community/helm-charts/blob/main/charts/loki/values.yaml). For guidance on tuning replicas and resources for production workloads, see [Configure high-availability logging](/how-to-guides/high-availability/logging/). For compactor and retention settings, see [Configure log retention](/how-to-guides/logging/configure-log-retention/).
```

since we just changed the chart source 😅

mjnagel (approved):

PR #2530, opened 2026-03-24
feat: add default endpoint probe and TLS expiry alerts

jasonwashburn:
## Description

Adds default UDS Core probe alert rules for endpoint downtime and TLS certificate expiry, with Helm-configurable thresholds, durations, and severities. 

### Included changes
- Adds default probe alerts via the `uds-prometheus-config` chart:
  - `UDSProbeEndpointDown`
  - `UDSProbeTLSExpiryWarning`
  - `UDSProbeTLSExpiryCritical`
- Includes Helm unittest and Vitest coverage for default probe alerts
- Updates monitoring docs across how-to, reference, concepts, and overview pages to document the new defaults and configuration paths

## Related Issue

Fixes # CORE-72

## Type of change

## Steps to Validate
1. Run `uds run test-single-layer --set LAYER=monitoring` to deploy and run e2e tests
2. Use Grafana's web UI to verify presence and intended configuration of alerts

## Checklist before merging

### DIFF
--- a/docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
+++ b/docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
@@ -91,7 +91,7 @@ Runbooks for upstream defaults are available at [runbooks.prometheus-operator.de
 
    If default alerts are too noisy or not relevant to your environment, you can tune both upstream kube-prometheus-stack and UDS Core defaults through bundle overrides.
 
-   UDS Core default probe alerts can be tuned or disabled like so:
+   UDS Core default probe alerts can be tuned or disabled as follows:
 
    ```yaml title="uds-bundle.yaml"
    packages:
@@ -120,7 +120,7 @@ Runbooks for upstream defaults are available at [runbooks.prometheus-operator.de
                  value: critical
    ```
 
-   Upstream kube-prometheus-stack default rules can be disabled like so:
+   Upstream kube-prometheus-stack default rules can be disabled as follows:
 
    ```yaml title="uds-bundle.yaml"
    packages:
--- a/docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
+++ b/docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
@@ -94,53 +94,45 @@ Runbooks for upstream defaults are available at [runbooks.prometheus-operator.de
    UDS Core default probe alerts can be tuned or disabled as follows:
 
    ```yaml title="uds-bundle.yaml"
-   packages:
-     - name: core
-       repository: registry.defenseunicorns.com/public/core
-       ref: x.x.x-upstream
-       overrides:
-         kube-prometheus-stack:
-           uds-prometheus-config:
-             values:
-               # Disable all UDS Core probe default alerts
-               - path: udsCoreDefaultAlerts.enabled
-                 value: false
-               # Disable Endpoint Down alert
-               - path: udsCoreDefaultAlerts.probeEndpointDown.enabled
-                 value: false
-               # Tune threshold and severity for TLS expiry warning alerts
-               - path: udsCoreDefaultAlerts.probeTLSExpiryWarning.days
-                 value: 21
-               - path: udsCoreDefaultAlerts.probeTLSExpiryWarning.severity
-                 value: warning
-               # Tune threshold and severity for TLS expiry critical alerts
-               - path: udsCoreDefaultAlerts.probeTLSExpiryCritical.days
-                 value: 7
-               - path: udsCoreDefaultAlerts.probeTLSExpiryCritical.severity
-                 value: critical
+   overrides:
+     kube-prometheus-stack:
+       uds-prometheus-config:
+         values:
+           # Disable all UDS Core probe default alerts
+           - path: udsCoreDefaultAlerts.enabled
+             value: false
+           # Disable Endpoint Down alert
+           - path: udsCoreDefaultAlerts.probeEndpointDown.enabled
+             value: false
+           # Tune threshold and severity for TLS expiry warning alerts
+           - path: udsCoreDefaultAlerts.probeTLSExpiryWarning.days
+             value: 21
+           - path: udsCoreDefaultAlerts.probeTLSExpiryWarning.severity
+             value: warning
+           # Tune threshold and severity for TLS expiry critical alerts
+           - path: udsCoreDefaultAlerts.probeTLSExpiryCritical.days
+             value: 7
+           - path: udsCoreDefaultAlerts.probeTLSExpiryCritical.severity
+             value: critical
    ```
 
    Upstream kube-prometheus-stack default rules can be disabled as follows:
 
    ```yaml title="uds-bundle.yaml"
-   packages:
-     - name: core
-       repository: registry.defenseunicorns.com/public/core
-       ref: x.x.x-upstream
-       overrides:
-         kube-prometheus-stack:
-           kube-prometheus-stack:
-             values:
-               # Disable specific individual rules by name
-               - path: defaultRules.disabled
-                 value:
-                   KubeControllerManagerDown: true
-                   KubeSchedulerDown: true
-               # Disable entire rule groups with boolean toggles
-               - path: defaultRules.rules.kubeControllerManager
-                 value: false
-               - path: defaultRules.rules.kubeSchedulerAlerting
-                 value: false
+   overrides:
+     kube-prometheus-stack:
+       kube-prometheus-stack:
+         values:
+           # Disable specific individual rules by name
+           - path: defaultRules.disabled
+             value:
+               KubeControllerManagerDown: true
+               KubeSchedulerDown: true
+           # Disable entire rule groups with boolean toggles
+           - path: defaultRules.rules.kubeControllerManager
+             value: false
+           - path: defaultRules.rules.kubeSchedulerAlerting
+             value: false
    ```
 
    Use `defaultRules.disabled` for fine-tuned control over upstream individual rules. Use `defaultRules.rules.*` to disable upstream rule groups when broader changes are needed.
--- a/docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
+++ b/docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
@@ -19,7 +19,9 @@ Define alerting conditions based on Prometheus metrics using PrometheusRule CRDs
 
 ## Before you begin
 
-UDS Core ships default alerting rules from the upstream kube-prometheus-stack chart covering cluster health, node conditions, and platform components. Runbooks for these default rules are available at [runbooks.prometheus-operator.dev](https://runbooks.prometheus-operator.dev/). This guide covers creating custom rules for your applications and optionally tuning the defaults.
+UDS Core ships default alerting rules from two sources. The upstream `kube-prometheus-stack` chart provides cluster and node health alerts, and UDS Core provides default probe alerts for endpoint downtime and TLS certificate expiry.
+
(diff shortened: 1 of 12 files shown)


docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
```diff
@@ -85,9 +87,40 @@ UDS Core ships default alerting rules from the upstream kube-prometheus-stack ch
...
+
+   If default alerts are too noisy or not relevant to your environment, you can tune both upstream kube-prometheus-stack and UDS Core defaults through bundle overrides.
+
+   UDS Core default probe alerts can be tuned or disabled like so:
+
+   ```yaml title="uds-bundle.yaml"
+   packages:
+     - name: core
+       repository: registry.defenseunicorns.com/public/core
+       ref: x.x.x-upstream
```
briantwatson: nit: Would it still be valid and cleaner to remove these two lines?

docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
```diff
@@ -85,9 +87,40 @@ UDS Core ships default alerting rules from the upstream kube-prometheus-stack ch
...
+                 value: warning
+               # Tune threshold and severity for TLS expiry critical alerts
+               - path: udsCoreDefaultAlerts.probeTLSExpiryCritical.days
+                 value: 7
+               - path: udsCoreDefaultAlerts.probeTLSExpiryCritical.severity
+                 value: critical
+   ```
 
-   If default kube-prometheus-stack alerts are too noisy or not relevant to your environment, you can disable individual rules or entire rule groups through bundle overrides.
+   Upstream kube-prometheus-stack default rules can be disabled like so:
```
briantwatson:
```suggestion
   Upstream kube-prometheus-stack default rules can be disabled as follows:
```
nit: same here

docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
```diff
@@ -85,9 +87,40 @@ UDS Core ships default alerting rules from the upstream kube-prometheus-stack ch
 
    The Prometheus Operator picks up PrometheusRule CRs automatically.
 
-3. **Optional: Disable or tune default alert rules**
+3. **(Optional) Disable or tune default alert rules**
+
+   If default alerts are too noisy or not relevant to your environment, you can tune both upstream kube-prometheus-stack and UDS Core defaults through bundle overrides.
+
+   UDS Core default probe alerts can be tuned or disabled like so:
```
briantwatson:
```suggestion
   UDS Core default probe alerts can be tuned or disabled as follows:
```
nit: a little more direct phrasing

docs/how-to-guides/monitoring-and-observability/set-up-uptime-monitoring.mdx
```diff
@@ -24,6 +24,9 @@ Monitor HTTPS endpoint availability using Blackbox Exporter probes. Probes are c
 > [!NOTE]
 > Uptime checks for Authservice-protected applications are fully supported. The UDS Operator automatically creates a dedicated Keycloak service account client and configures OAuth2 authentication for the probe.
 
+> [!NOTE]
+> UDS Core ships default probe alerts through `PrometheusRule` resources in the `uds-prometheus-config` chart. By default, UDS Core creates `UDSProbeEndpointDown`, `UDSProbeTLSExpiryWarning`, and `UDSProbeTLSExpiryCritical` alerts. To tune or disable these defaults, reference [Create metric alerting rules](/how-to-guides/monitoring-and-observability/create-metric-alerting-rules/).
+
```
briantwatson, line 28:
```suggestion
> [!NOTE]
> Uptime checks for Authservice-protected applications are fully supported. The UDS Operator automatically creates a dedicated Keycloak service account client and configures OAuth2 authentication for the probe.
>
> UDS Core also ships default probe alerts (`UDSProbeEndpointDown`, `UDSProbeTLSExpiryWarning`, and `UDSProbeTLSExpiryCritical`) through `PrometheusRule` resources in the `uds-prometheus-config` chart. To tune or disable these defaults, see [Create metric alerting rules](/how-to-guides/monitoring-and-observability/create-metric-alerting-rules/).

```

The back to back notes looks a little busy:

Image

Wondering if combing them would look a little cleaner:
Image

src/prometheus-stack/chart/values.yaml
```diff
@@ -12,3 +12,21 @@ additionalNetworkAllow: []
...
+  probeTLSExpiryWarning:
+    enabled: true
+    for: 10m
+    days: 30
+    severity: warning
+  probeTLSExpiryCritical:
+    enabled: true
+    for: 10m
+    days: 14
+    severity: critical
```
briantwatson, line 32: This and the associated template parameters provides nice configuration options for alerts.  If we take this approach, we should probably consider standardizing some of the knobs we provide on all alerts and follow with updates to other alerts we define in the project, e.g. https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/chart/templates/keycloak-logging-alerting-rules.yaml

src/prometheus-stack/chart/values.yaml
mjnagel: nit: copyright update

```suggestion
# Copyright 2024-2026 Defense Unicorns
```

src/prometheus-stack/tasks.yaml
```diff
@@ -65,12 +65,20 @@ tasks:
...
-          npm ci && npx vitest run "prometheus"
+          npx vitest run "prometheus"
+        dir: test/vitest
+      - description: "Run UDS Core default probe alert tests"
+        cmd: |
+          npx vitest run "default-probe-alerts"
         dir: test/vitest
       - description: "Run Blackbox Exporter integration tests"
         cmd: |
           echo "Running Blackbox Exporter integration tests..."
```
mjnagel: very nit (and not introduced here): this echo isn't consistent with other tests which just use the description

```suggestion
```

src/prometheus-stack/chart/values.yaml
mjnagel, line 32: Can we add a values schema for this chart with types on each of these?

docs/reference/configuration/monitoring-and-observability.md
```diff
@@ -76,6 +76,78 @@ All endpoint probes (both built-in and application) produce standard Blackbox Ex
...
+overrides:
+  kube-prometheus-stack:
+    uds-prometheus-config:
+      values:
+        # Disable all UDS Core default probe alerts
+        - path: udsCoreDefaultAlerts.enabled
+          value: false
+        # Disable only the endpoint-down alert
+        - path: udsCoreDefaultAlerts.probeEndpointDown.enabled
+          value: false
```
mjnagel, line 138: nit: This kind of felt weird to me reading "common tuning patterns" and then it disables all the alerts. I think especially a bit weird since the two toggles here overlap. The table above this explains each value well so maybe a non-issue, just felt like a weird example in terms of "common patterns".

test/vitest/default-probe-alerts.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      .flatMap(group => group.rules ?? [])
+      .filter(rule => rule.type === "alerting" && Boolean(rule.name))
+      .map(rule => rule.name as string);
+  };
+
+  const expectAlertRulePresent = (alertName: string) =>
+    pollUntilSuccess(
+      fetchAlertRuleNames,
+      names => names.includes(alertName),
+      `${alertName} alert rule to be loaded`,
```
mjnagel: Should we also check the alert health here (would also need to add to the types/response above)? I think that might provide some sane regression checking in case something changes with metrics, etc.

mjnagel (commented):
Few smaller comments - nothing major overall.

docs/reference/configuration/monitoring-and-observability.md
jasonwashburn: yeah, that's fair...maybe it makes more sense to just trim the example down to something simpler like tuning the number of days on the TLS warning/critical and knocking down the severity of the warning or something...alternatively, we can remove this line and just make it "Some examples of tuning or disabling..."

test/vitest/default-probe-alerts.spec.ts
jasonwashburn: Yeah I can give that a try...that'll be testing a lot more than just the alerts got loaded...i.e. the endpoints are up, tls cert is good, blackbox is working, probe's are set up correctly... (well...assuming there isn't a break in the actual expression itself in the alerts that just make them pass...) ...so maybe doesn't test the actual probe functionality perfectly...but would provide some decent e2e testing for blackbox/probes/alerts etc

mjnagel: Yeah it could overload this test a bit, and we would have some [overlapping probe checks](https://github.com/defenseunicorns/uds-core/blob/feat/core-72-add-alertmanager-uptime-and-tls-alerts/test/vitest/uptime.spec.ts#L23-L48) in our tests already. Will defer to your opinion here - I think mostly would want to validate that "when probes are healthy" our alerts actually work (the underlying query isn't broken). The TLS cert expiry could be nice for us in CI as an early warning about our dev cert...or maybe really painful when it expires 😆

briantwatson (commented):
Nice addition, some comments below for consideration.

docs/reference/configuration/monitoring-and-observability.md
mjnagel: Don't feel strongly - could change the wording a bit and just make it more clear this a list of possible options/examples.

test/vitest/default-probe-alerts.spec.ts
jasonwashburn: oh I see...thats a maybe a little different than I was thinking...probably easy enough to test something like ok...manually I see that probe_success  == 1 for grafana and keycloak endpoints...let me verify that the alert is also happy/not-firing...though that doesn't really validate the alert expression...just that the expression is written in a way to not fire...we'd have to check both cases to make that assertion right...so somehow make an endpoint unhealthy long enough to fire (5 minutes)...or for a faster check...maybe unhealthy long enough to validate the alert goes to "pending" or whatever that specific status is when it's in that time period of failure before it fires 🤔

docs/how-to-guides/monitoring-and-observability/create-metric-alerting-rules.mdx
jasonwashburn: removed the boilerplate from both examples

docs/reference/configuration/monitoring-and-observability.md
jasonwashburn: fixed

src/prometheus-stack/chart/values.yaml
jasonwashburn: added...I limited the definition of `additionalNetworkAllow` to an array of objects rather than fully defining it...which seems to be the convention used elsewhere..and less fragile to later changes to the Package crd

jasonwashburn: makes sense to me

test/vitest/default-probe-alerts.spec.ts
jasonwashburn: alright...after looking at it a bit, went with the middle ground and added some assertions that verify health == ok and state == inactive...that should capture errors in the expression and if any alerts are pending/firing that aren't expected to be. I'm sure it'll be annoying when if/when it happens...but I don't hate the idea of an affirmative, cant miss it, alert if we allow it to get within 30 days of our dev cert expiring

mjnagel: LGTM.

mjnagel (approved):

briantwatson (approved):

PR #2547, opened 2026-03-27
chore(docs): add how to guides for ldap/ad/kerberos configurations

chance-coleman:
## Description
Add AD/LDAP federation and Kerberos/SPNEGO how-to guides

Adds three new how-to guides covering enterprise Active Directory integration patterns for UDS Core:

Configure AD/LDAP user federation with OpenTofu. declarative IaC approach using the keycloak/keycloak provider to define the federation provider

Configure AD/LDAP user federation in Keycloak UI. Admin Console walkthrough for environments where OpenTofu is not available. 

Configure Kerberos/SPNEGO SSO. end-to-end guide for transparent SSO on domain-joined clients

The two LDAP guides are cross-linked and describe the same configuration; the Kerberos guide is designed to be used alongside either LDAP guide for deployments that need both group sync and transparent SSO.

## Related Issue

Fixes [core-131](https://linear.app/defense-unicorns/issue/CORE-131/add-examples-and-detailed-steps-on-adding-adkerberos-user-federation) and [core-130](https://linear.app/defense-unicorns/issue/CORE-130/add-examples-and-detailed-steps-on-adding-ldap-user-federation)

## Type of change

## Checklist before merging


docs/how-to-guides/identity-and-authorization/configure-kerberos-spnego-sso.mdx
```diff
@@ -0,0 +1,335 @@
...
+                     ports:
+                       - 88
+                   - description: "Kerberos password change (omit if not needed)"
+                     direction: Egress
+                     selector:
+                       app.kubernetes.io/name: keycloak
+                     remoteCidr: "<KDC_IP>/32"
+                     ports:
+                       - 464
+               - path: extraVolumes
```
slaskawi, line 95: This one has been found by the Brutal Code Review:

```
  #1. extraVolumes override destroys the default CA trust bundle mount

  File: configure-kerberos-spnego-sso.mdx, lines 128-139

  - path: extraVolumes
    value:
      - name: keytab
        secret:
          secretName: keycloak-kerberos-keytab
          defaultMode: 0400

  The Keycloak chart's values.yaml defines a default extraVolumes entry that mounts the uds-trust-bundle ConfigMap at /tmp/ca-certs. Helm array overrides replace the entire array, they don't append. Following this guide
  as written silently removes the CA trust bundle, breaking all ldaps:// connections, OCSP checking, and any outbound TLS relying on the UDS trust bundle. The existing configure-x509-crl-airgap.mdx guide handles this
  correctly by including both the default ca-certs volume and the custom volume.

  Fix: Include the default ca-certs volume alongside the keytab in both YAML blocks (steps 3 and 4):

  - path: extraVolumes
    value:
      - name: ca-certs
        configMap:
          name: uds-trust-bundle
          optional: true
      - name: keytab
        secret:
          secretName: keycloak-kerberos-keytab
          defaultMode: 0400
  - path: extraVolumeMounts
    value:
      - name: ca-certs
        mountPath: /tmp/ca-certs
        readOnly: true
      - name: keytab
        mountPath: /opt/keycloak/conf/kerberos.keytab
        subPath: krb5.keytab
        readOnly: true

  Confidence: 97
```

This is actually a good point and I think we may want to either modify this example to vendor the `ca-certs` in, or even better, refactor the `ca-certs` to be always added to the array (maybe even hardcoded in the StatefulSet definition). I'm OK with addressing this as a follow up.

docs/how-to-guides/identity-and-authorization/configure-ad-ldap-user-federation-with-opentofu.mdx
```diff
@@ -0,0 +1,346 @@
+---
+title: Configure AD/LDAP user federation in Keycloak with OpenTofu
```
slaskawi, line 2: Just checking - do we need any additional authpols to unblock the traffic from Keycloak to the LDAP server? We can leverage `.Values.additionalNetworkAllow` for that.

docs/how-to-guides/identity-and-authorization/configure-kerberos-spnego-sso.mdx
```diff
@@ -0,0 +1,335 @@
...
+                       app.kubernetes.io/name: keycloak
+                     remoteCidr: "<KDC_IP>/32"
+                     ports:
+                       - 464
+   ```
+
+   > [!NOTE]
+   > Replace `<KDC_IP>/32` with your KDC's IP address. UDS network policies enforce TCP only; Kerberos defaults to UDP on port 88, so your KDC must accept TCP connections on port 88 (all modern AD Domain Controllers do). Port 464 (kpasswd) is only needed if Keycloak handles password changes; omit that entry if not required.
+
+4. **Mount the keytab into the Keycloak pod**
```
slaskawi: The examples from the steps 3 and 4 have a lot in common (specifically the YAML snippets). Could we combine them together?

docs/how-to-guides/identity-and-authorization/configure-ad-ldap-user-federation-with-opentofu.mdx
```diff
@@ -0,0 +1,346 @@
...
+     sync_registrations   = false
+     full_sync_period     = 86400  # full sync every 24 hours; omit this argument to disable
+     changed_sync_period  = 3600   # incremental sync every hour; omit this argument to disable; lower this value to reduce the window for disabled AD accounts to remain accessible in Keycloak
+
+     # TLS
+     use_truststore_spi = "ONLY_FOR_LDAPS"
+
+     # Timeouts
+     connection_timeout = "5s"
+     read_timeout       = "10s"
```
slaskawi: Even though LDAP servers are usually quite fast, 10s here might be a bit low for large user directories. I would change to 30s at least.

slaskawi (commented):
Generally looks great!

I believe I found two things that may require some changes - potentially missing LDAP networks polilices and replacing `ca-certs` volumes (found by the Brutal Code Review).

docs/how-to-guides/identity-and-authorization/configure-kerberos-spnego-sso.mdx
chance-coleman: combined

docs/how-to-guides/identity-and-authorization/configure-ad-ldap-user-federation-with-opentofu.mdx
chance-coleman: yes! that is a really good call

docs/how-to-guides/identity-and-authorization/configure-kerberos-spnego-sso.mdx
chance-coleman: i added this here.

slaskawi (commented):
LGTM!

docs/how-to-guides/identity-and-authorization/configure-ad-ldap-user-federation-with-opentofu.mdx
joelmccoy, line 1: i wonder if it is worth combining the opentofu and UI one into a single doc called `configure-ad-ldap-user-federation`.  And then you could wrap the top level steps in tab blocks for each method?

joelmccoy, line 1: Im also wondering what prompted us to do an opentofu version for this configuration specifically. A couple questions:
1. Did we have this guide before? or is this net new?
2. If we didn't have the guide before, what prompted us to create the opentofu version?
3. Do we have a sane way to test this?

I'm just a bit worried we are tossing up some tofu configuration that looks like it should work, but we don't test/use this anywhere and it might not be 100% perfect?  This does sound like a good guide if we start building a tofu solution/module for keycloak configuration and have testable/usuable code to use.

joelmccoy (commented):
I think things generally look good here, but have some questions on the approach/necessity of the tofu stuff.

docs/how-to-guides/identity-and-authorization/configure-ad-ldap-user-federation-with-opentofu.mdx
chance-coleman: Yeah totally get this concern. Definitely validated everything was working that i documented but there's no guarantee that continues to work nor do we have a CI test or anything like that. I think primarily this was a suggestion from @mjnagel in a DM that we should probably try to prioritize tofu as much as possible. Would be good to get @mjnagel thought's here i think.

mjnagel: Following up on some of the synchronous chats we had - I think we maybe jumped the gun scheduling this issue for work (my fault on that). We'll want to make sure we take the right approach holistically on these issues for "advanced keycloak config" - _likely_ building out a new repo with a number of modules for reference.

Nothing wrong with this doc necessarily, but we may want to revise how/where this lands based on some better planning around this overall effort.

chance-coleman:
We've decided to close this PR in favor of moving these docs to a future location for "advanced tofu" uses.

PR #2552, opened 2026-03-30
feat: add support for image volumes in policy

mjnagel:
## Description

Since Zarf [fully supports](https://docs.zarf.dev/best-practices/data-injections-migration/) image volumes, these should be an allowed volume type in our policies. The only reason they were previously excluded was because they didn't exist when our allow list was created.

## Related Issue

Fixes CORE-389

## Type of change

## Steps to Validate

Unit tests validate basic functionality. Note that for full e2e testing you would need to run k8s 1.35+ or the image volume feature flag on 1.34.

## Checklist before merging


docs/how-to-guides/identity-and-authorization/configure-x509-crl-airgap.mdx
```diff
@@ -34,7 +34,7 @@ This guide uses a **Kubernetes ImageVolume** to mount an OCI image containing th
 | 1.35+ | Enabled by default; no feature gate configuration needed |
 
 > [!NOTE]
-> ImageVolumes are currently blocked by UDS policies. You will need to add a policy exemption as part of this guide.
+> If you are running UDS Core < 1.1.0, `image` volumes are blocked by the `RestrictVolumeTypes` policy. Add a `RestrictVolumeTypes` `Exemption` targeting Keycloak pods to allow them. See [Configure infrastructure exemptions](/how-to-guides/policy-and-compliance/configure-infrastructure-exemptions/) for the exemption format.
```
mjnagel, line 37: Open to other thoughts here...for now our docs target `main` (pending some further thoughts on versioning docs maybe?) so while I did remove the exemption from the steps I also thought it would make sense to keep this note as a pointer for pre-1.1.0 versions.

Also placeholder comment in case this change doesn't land in 1.1.0 and we need it to be 1.2.0.

chance-coleman (approved):

PR #2566, opened 2026-04-02
feat: add user account disabled with inactivity

chance-coleman:
## Description
Allow for users account to be automatically disabled after being inactive for x number of days. This does not affect admin accounts. There is one new value:
-  `realmConfig.accountInactivityDays`
    - sets the number of days before account disabled 
    - when set to 0 this workflow is disabled

## Related Issue

Fixes [core-432](https://linear.app/defense-unicorns/issue/CORE-432/address-apsc-dv-000320-account-inactivity-disable-requirement)

## Type of change

## Checklist before merging


src/keycloak/chart/values.yaml
```diff
@@ -62,6 +62,9 @@ realmInitEnv:
 realmConfig:
   # Maximum number of in-flight (ongoing) login attempts per user to prevent login storms.
   maxInFlightLoginsPerUser: 300
+  # Number of days of inactivity after which a non-admin user account is automatically disabled (0 = disabled).
+  # Satisfies APSC-DV-000320 (ASD STIG). Requires Keycloak 26.5+ (Workflows preview feature).
+  accountInactivityDays: 0
```
slaskawi: Just wanted to check if this shouldn't go into the `realmInitEnv`?

These settings are related to the Realm JSON and are initialized only once at the import time. It might also be conceptually colocated with sessions, account lockouts etc.

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+import { Steps } from '@astrojs/starlight/components';
+
+## What you'll accomplish
+
+You'll configure UDS Core's Keycloak to automatically disable non-admin user accounts that have not authenticated for a configurable number of days. This satisfies **APSC-DV-000320** (ASD STIG), which requires applications to automatically disable accounts after 35 days of inactivity. Accounts belonging to realm administrators are excluded and are never automatically disabled.
+
+## Prerequisites
+
+- UDS Core deployed (Keycloak 26.5+)
```
slaskawi: Since Keycloak versions are tied with the UDS Core, maybe it's better to use UDS Core version here?

src/keycloak/chart/values.yaml
slaskawi: If you decided to move this, we'll need to update the manual (which call out that this setting gets activated at the import-time only).

docs/reference/configuration/identity-and-authorization.md
```diff
@@ -91,6 +91,7 @@ Bundle override path: `overrides.keycloak.keycloak.values[].path: realmConfig`
 | Field                      | Default | Description                                          |
 | -------------------------- | ------- | ---------------------------------------------------- |
 | `maxInFlightLoginsPerUser` | `300`   | Maximum concurrent in-flight login attempts per user |
+| `accountInactivityDays`    | `0`     | Days of inactivity before a non-admin account is automatically disabled. `0` disables the feature. Satisfies APSC-DV-000320 (ASD STIG). **Applied at initial realm import only.** Changes to a running instance require manual update via the Keycloak Admin Console. |
```
slaskawi: This is actually not correct. 

The https://www.stigviewer.com/stigs/application_security_and_development/2025-02-12/finding/V-222411 mentiones that it should be set to 35 days. So I think we have two options: set it to 35 by default or set it to 0 but clearly mention here that satisfying this requirement requires changing this setting.

src/keycloak/chart/values.schema.json
```diff
@@ -371,9 +371,15 @@
     },
     "realmConfig": {
       "type": "object",
+      "additionalProperties": false,
```
slaskawi: I think this might have slipped in.

src/keycloak/chart/templates/secret-kc-realm.yaml
```diff
@@ -40,3 +40,7 @@ data:
   OVERRIDE_REALM_DISPLAY_NAME: {{ . | toString | b64enc }}
   {{- end }}
   FIPS_ENABLED: {{ "true" | b64enc }}
+  {{- if gt (int .Values.realmConfig.accountInactivityDays) 0 }}
+  ACCOUNT_INACTIVITY_DAYS: {{ .Values.realmConfig.accountInactivityDays | toString | b64enc }}
```
slaskawi: Shall we protect against the negative values (user misconfiguration)?

slaskawi (commented):
Awesome stuff! I added a few comments but the most significant one is about moving this to the `realmInitEnv`. 

It's also worth double-checking the `additionalProperties` setting that probably slipped in.

src/keycloak/chart/values.yaml
chance-coleman: Good point on colocation. I kept it in realmConfig because the chart has logic that keys off this value (the conditional guard and ENABLED flag), and it gives us integer type + schema validation that realmInitEnv doesn't support. But open to moving it if you think the conceptual grouping is worth the tradeoff.

src/keycloak/chart/templates/secret-kc-realm.yaml
chance-coleman: I think this is already covered by the gt 0 check on line 43. negative values evaluate the same as 0 and won't emit the key. The schema also has minimum: 0 which catches it at validation time before the template renders.

docs/reference/configuration/identity-and-authorization.md
chance-coleman: I went with 0 but clearly mentioned that to satisfy the stig 35 days is required.

docs/operations/release-notes/1-2.mdx
joelmccoy, line 1: is the intention to just have this as a placeholder here? If it is, might want to wait until docs are only working off releases and not main right before merging this in so it doesn't show up on the docs site?

joelmccoy (dismissed):
LGTM

docs/reference/configuration/identity-and-authorization.md
```diff
@@ -91,6 +91,7 @@ Bundle override path: `overrides.keycloak.keycloak.values[].path: realmConfig`
 | Field                      | Default | Description                                          |
 | -------------------------- | ------- | ---------------------------------------------------- |
 | `maxInFlightLoginsPerUser` | `300`   | Maximum concurrent in-flight login attempts per user |
+| `accountInactivityDays`    | `0`     | Days of inactivity before a non-admin account is automatically disabled. `0` disables the feature. Set to `35` to satisfy APSC-DV-000320 (ASD STIG). **Applied at initial realm import only.** Changes to a running instance require manual update via the Keycloak Admin Console. |
```
mjnagel: We generally don't add compliance information in our docs/code comments (deferring to Lula mappings instead in the future). Might be good to remove the STIG mention here/in the values file?

docs/operations/release-notes/1-2.mdx
chance-coleman: yup thats a good call

docs/reference/configuration/identity-and-authorization.md
chance-coleman: removed the reference to the stig in a couple places.

docs/operations/release-notes/1-2.mdx
chance-coleman: I decided to just remove that doc until we're ready to release. Just creates too many complexities before we're ready for it.

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
```diff
@@ -0,0 +1,120 @@
...
+description: Configure Keycloak to automatically disable non-admin user accounts after a set number of days of inactivity, satisfying APSC-DV-000320 (ASD STIG).
+sidebar:
+  order: 3.020
+---
+
+import { Steps } from '@astrojs/starlight/components';
+
+## What you'll accomplish
+
+You'll configure UDS Core's Keycloak to automatically disable non-admin user accounts that have not authenticated for a configurable number of days. This satisfies **APSC-DV-000320** (ASD STIG), which requires applications to automatically disable accounts after 35 days of inactivity. Accounts belonging to realm administrators are excluded and are never automatically disabled.
```
joelmccoy: nit: think there is another reference to the STIG here that can be removed?

src/keycloak/chart/values.yaml
mjnagel: Hm, yeah I actually had the same comment as @slaskawi here - I think could be slightly confusing since the only other setting here (`maxInFlightLoginsPerUser`) is actually dynamic (can be changed post-install). But also see why this maybe doesn't fit right with the current `realmInitEnv` flow...

Maybe just a comment here noting that this cannot be set/changed on upgrades via the helm value? It's similar to `realmAuthFlows` in that way - feeds into the "realm envs" but not directly.

src/keycloak/chart/templates/secret-kc-realm.yaml
```diff
@@ -40,3 +40,7 @@ data:
   OVERRIDE_REALM_DISPLAY_NAME: {{ . | toString | b64enc }}
   {{- end }}
   FIPS_ENABLED: {{ "true" | b64enc }}
+  {{- if gt (int .Values.realmConfig.accountInactivityDays) 0 }}
```
mjnagel: Is this line necessary given the guard on the identity-config side? Don't know that it matters either way just want to make sure I understand if there's a hidden behavior with this workflow or if we can rely on the `ACCOUNT_INACTIVITY_DAYS_ENABLED` fully. Basically - is there a specific behavior you were trying to achieve with this conditional?

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+**What counts as activity:** The workflow tracks `user-authenticated` events. Any successful login through UDS Core SSO, including logins via federated identity providers (Azure AD, Google SAML), resets the inactivity timer.
+
+**Newly provisioned accounts:** Accounts that have never logged in do not generate a `user-authenticated` event and therefore never start a workflow instance. Accounts provisioned before this feature was enabled are also not retroactively evaluated. Run a one-time audit of last-login timestamps after enabling via **Keycloak Admin Console** → **Users**.
+
+**Admin accounts excluded:** Users with the `realm-management/realm-admin` role are never disabled by this workflow.
+
+> [!NOTE]
+> Keycloak Workflows is a [preview feature](https://www.keycloak.org/docs/latest/server_admin/index.html#preview-features). UDS Core enables it by default via `--features=preview,fips`. Preview features may change in future Keycloak releases.
```
mjnagel: nit: I don't know if this callout is really necessary in the docs but don't feel strongly.

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+## Steps
+
+<Steps>
+
+1. **Set `accountInactivityDays` in your bundle override**
+
+   Add the override to your `uds-bundle.yaml`. Set the value to the number of days of inactivity after which a non-admin account is disabled. The STIG-required value is `35`.
```
mjnagel:
```suggestion
   Add the override to your `uds-bundle.yaml`. Set the value to the number of days of inactivity after which you want non-admin accounts to be disabled.
```

reworded slightly + removed STIG mention

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
```diff
@@ -0,0 +1,120 @@
...
+           keycloak:
+             values:
+               - path: realmConfig.accountInactivityDays
+                 value: 35
+   ```
+
+   Setting the value to `0` (the default) leaves the workflow disabled.
+
+   > [!IMPORTANT]
+   > This setting is applied during initial realm import only. Changing `accountInactivityDays` in a bundle re-deploy updates the Kubernetes Secret but does **not** update the live workflow. Keycloak only reads the value during the first realm import. To update a running instance, use the Keycloak Admin Console to toggle **Enabled** on the `disable-inactive-users` workflow and set the **after** duration on the `disable-user` step to match your desired value.
```
mjnagel: Is it worth adding some more "steps" here in terms of where to navigate to (even screenshots maybe)? I could see this as being "optional" steps for changing after deploy.

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
```diff
@@ -0,0 +1,120 @@
...
+
+   > [!WARNING]
+   > Setting the **after** duration to `0` will disable non-admin accounts immediately on their next login. Always confirm the duration is correct before enabling the workflow.
```
mjnagel: This might be confusing to a reader as a few lines up they would've read this:
> Setting the value to `0` (the default) leaves the workflow disabled.

I think this is only applicable if setting in the Keycloak UI right?

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
```diff
@@ -0,0 +1,120 @@
...
+## Verification
+
+After deployment, confirm the workflow is active:
+
+1. Log in to the Keycloak Admin Console at `keycloak.<admin_domain>`
+2. Switch to the **uds** realm
+3. In the left sidebar under **Configure**, click **Workflows**
+4. Confirm the `disable-inactive-users` workflow is listed and **Enabled** is toggled on
+5. Log in as a non-admin test user through UDS Core SSO
+6. Return to the Admin Console, open the workflow, and click **View active instances**. A new instance should appear for the test user with a `disable-user` step scheduled at the number of days you configured (e.g., 35 days if you used the STIG-required value)
```
mjnagel:
```suggestion
6. Return to the Admin Console, open the workflow, and click **View active instances**. A new instance should appear for the test user with a `disable-user` step scheduled at the number of days you configured
```

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
```diff
@@ -0,0 +1,120 @@
+---
+title: Configure automatic account inactivity disable
+description: Configure Keycloak to automatically disable non-admin user accounts after a set number of days of inactivity, satisfying APSC-DV-000320 (ASD STIG).
```
mjnagel:
```suggestion
description: Configure Keycloak to automatically disable non-admin user accounts after a set number of days of inactivity.
```

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
mjnagel:
```suggestion
You'll configure UDS Core's Keycloak to automatically disable non-admin user accounts that have not authenticated for a configurable number of days. Accounts belonging to realm administrators are excluded and are never automatically disabled.
```

src/keycloak/chart/values.yaml
slaskawi: I agree with @mjnagel here and wanted to suggest accepting the tradeoff and actually moving it. 

> I think could be slightly confusing since the only other setting here (maxInFlightLoginsPerUser) is actually dynamic (can be changed post-install).

Yes, that's exactly the reason why I separated it out previously. The `maxInFlightLoginsPerUser` parameter needs to be passed through the SPI configuration in the StatefulSet. That's the primary reason (and what distinguishes it from the `realmInitEnv`) why it's separate.

src/keycloak/chart/templates/secret-kc-realm.yaml
slaskawi: Ok, I see we have similar comments elsewhere so let me resolve this one.

src/keycloak/chart/templates/secret-kc-realm.yaml
chance-coleman: The behavior i was trying to protect against `what if someone manually enables the workflow in the Keycloak Admin Console after deploy?` The realm template uses `${ACCOUNT_INACTIVITY_DAYS:35}d` (default 35 if unset). Without the guard, a Helm value of 0 would override that default to "0d", meaning any manually-enabled workflow would disable accounts immediately on login. Added a comment explaining this behavior.

src/keycloak/chart/values.yaml
chance-coleman: Moved this out to the `realmInitEnv`. I think both of you @mjnagel @slaskawi had a initial reaction that i just convinced myself not to listen too 😄

docs/how-to-guides/identity-and-authorization/configure-account-inactivity-disable.mdx
chance-coleman: i moved this into a separate optional step that's more specifically tied to manual admin realm actions.

src/keycloak/chart/values.yaml
slaskawi: Fantastic! Resolving.

slaskawi (approved):

mjnagel (approved):
LGTM

PR #2583, opened 2026-04-10
chore(docs): enable release branch versioned docs

chance-coleman:
## Description
Now that release branches are allowed in the uds-docs we want to utilize this and enable our versioned docs. This is set to `3` so that upon our next release `1.2.0` that new version will be pulled in.

## Type of change

## Steps to Validate
- With `uds-docs` and `uds-core` locally run the following command:
```
cd <uds-docs path>
DOCS_OVERRIDES="<uds-core path>" npm run build && npm run preview
```

## Checklist before merging

### DIFF
--- a/docs/docs.config.json
+++ b/docs/docs.config.json
@@ -3,6 +3,8 @@
   "label": "Core",
   "contentDir": "core",
   "description": "The platform layer deployed onto a Kubernetes cluster. Bundles Istio (service mesh and ingress), Keycloak (identity and SSO), Authservice (SSO proxy for apps without native OIDC), Prometheus + Grafana + Alertmanager (monitoring), Vector + Loki (logging), Falco (runtime security), and Pepr (policy engine and operator framework). Exposes UDS Package, UDS Exemption, and UDS ClusterConfig CRDs for application and platform integration.",
+  "versionSource": "branch",
+  "archiveCount": 3,
   "sidebarOrder": [
     "getting-started",
     "concepts",


docs/docs.config.json
```diff
@@ -3,6 +3,8 @@
   "label": "Core",
   "contentDir": "core",
   "description": "The platform layer deployed onto a Kubernetes cluster. Bundles Istio (service mesh and ingress), Keycloak (identity and SSO), Authservice (SSO proxy for apps without native OIDC), Prometheus + Grafana + Alertmanager (monitoring), Vector + Loki (logging), Falco (runtime security), and Pepr (policy engine and operator framework). Exposes UDS Package, UDS Exemption, and UDS ClusterConfig CRDs for application and platform integration.",
+  "versionSource": "branch",
+  "archiveCount": 3,
```
chance-coleman, line 7: We could actually set this to our desired number of versions. The docs integration doesnt allow for using versions and docs directories that don't have the `docs.config.json`. So that means that we can set this number to anything we want and it will only show our oldest doc version as 1.0.0 because that was the first iteration of having the `docs.config.json`.

joelmccoy (approved):
LGTM - This will become effective immediately on the next docs site build though correct? I traced through the logic and it appears so.  (i.e. we don't have to wait until 1.2 is release).  Also good call on just setting this to 3 now. I appears that the integration script will skip over old versions that don't have the config.

docs/docs.config.json
chance-coleman: @joelmccoy replying to your comment here. You're right that this will take immediate affect. Want to call out my earlier comment though that we could actually set this to 5 or 6 (i think we decided on 5 previously) because it wont show anything earlier than 1.0.0. Then we wont have to update this for future releases.

joelmccoy: Agree to set this to whatever that number is supposed to be now. I had in my head 3 (due to supported versions of core).  But maybe @mjnagel can confirm/deny.

I think also if you set to 3, you get Latest tag + 3 archived versions if I am not mistaken (so 4 versions total)?

mjnagel: I don't feel strongly on the right number here, I think the 3 supported versions is a good minimum (so 2) but I like the idea of one extra (unsupported) version for those who might be behind and upgrading. The truth is that _most_ docs don't change significantly across releases but I think there's some value in having 1+ unsupported versions for someone to check and see that they are missing something due to the version they're currently on, and should probably upgrade.

There's probably another angle here where we might want a "different" banner on docs for "unsupported" versions vs just "not latest".

chance-coleman: So i think for now we can just merge this in and update it later if we want to add or remove versions. it'll be 3 + latest as is. Quick fix if we're not liking it as is.

PR #2597, opened 2026-04-16
docs: add keycloak admin handling doc

mjnagel:
## Description

Adds a new doc around managing initial admin access. Also:
- updates order of all identity docs to be a bit more intuitive (basic -> advanced)
- changes `zarf connect keycloak` -> `uds zarf connect keycloak`
- updates credential recovery runbook with new symptoms/causes

### DIFF
--- a/docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
+++ b/docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
@@ -9,7 +9,7 @@ import { Steps, Tabs, TabItem } from '@astrojs/starlight/components';
 
 ## What you'll accomplish
 
-You'll create, use, and remove a Keycloak admin account in the master realm as a just-in-time credential. UDS Core ships without a default admin, so you must bootstrap one before you can access the admin console at `https://keycloak.<admin_domain>/`. This guide covers the two supported bootstrap methods and the recommended lifecycle for keeping admin credentials out of the cluster.
+You'll create, use, and remove a Keycloak admin account in the master realm. UDS Core ships without a default admin, so you must bootstrap one before you can access the admin console at `https://keycloak.<admin_domain>/`. This guide covers the two supported bootstrap methods and the recommended lifecycle for keeping admin credentials out of the cluster.
 
 ## Prerequisites
 
@@ -22,7 +22,7 @@ You'll create, use, and remove a Keycloak admin account in the master realm as a
 
 UDS Core does not create a default admin user in any realm. You must create the first admin in the **master realm** (Keycloak's built-in admin realm) before you can manage the `uds` realm or any of its clients. Two bootstrap methods are supported:
 
-- `uds zarf connect keycloak` opens a port-forward to the Keycloak HTTP service so you can register an admin through the browser-based first-run form. This is the preferred method for all interactive installs.
+- `uds zarf connect keycloak` opens a port-forward to the Keycloak HTTP service so you can register an admin through the browser-based Welcome Page. This is the preferred method for all interactive installs.
 - `insecureAdminPasswordGeneration` is a Helm value that provisions an admin user from a Kubernetes Secret at chart install time. Use this method only when you cannot run the interactive flow (for example, fully automated installs with no browser access).
 
 Treat every admin account as a break-glass credential. The goal is to have no standing admin user when administrative work is not in progress:
@@ -60,7 +60,7 @@ Treat every admin account as a break-glass credential. The goal is to have no st
    2026-04-16 17:11:00 INF Tunnel established, waiting for user to interrupt (ctrl-c to end) urls=http://127.0.0.1:56641
    ```
 
-   Open the URL shown next to `urls=` in a browser to reach Keycloak's first-run admin form. Fill in every field (**Username**, **Password**, **Email**, **First name**, **Last name**) and submit to create the master realm admin user. Leave the terminal running until you have submitted the form; pressing **Ctrl-C** closes the tunnel.
+   Open the URL shown next to `urls=` in a browser to reach Keycloak's Welcome Page. Fill in every field (**Username**, **Password**, **Email**, **First name**, **Last name**) and submit to create the master realm admin user. Leave the terminal running until you have submitted the form; pressing **Ctrl-C** closes the tunnel.
 
    > [!TIP]
    > Use a username tied to the individual performing the work (for example, `firstname.lastname`). Avoid generic names like `admin` or `root` so the account cannot be reused as a shared credential and so audit logs clearly attribute actions.
@@ -131,9 +131,9 @@ Log in at `https://keycloak.<admin_domain>/` with your admin credentials and con
 
 ## Troubleshooting
 
-### Problem: `uds zarf connect keycloak` does not show the first-run form
+### Problem: `uds zarf connect keycloak` does not show the Welcome Page
 
-**Symptom:** The tunnel URL loads the Keycloak sign-in page rather than the "Create administrative user" first-run form.
+**Symptom:** The tunnel URL loads the Keycloak sign-in page rather than the "Create administrative user" Welcome Page.
 
 **Solution:** A master realm admin already exists or did previously. Log in with those credentials, or follow the [Keycloak Credential Recovery](/operations/troubleshooting-and-runbooks/keycloak-credential-recovery/) runbook to bootstrap a temporary admin.
 
--- a/docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
+++ b/docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
@@ -15,7 +15,7 @@ You'll create, use, and remove a Keycloak admin account in the master realm as a
 
 - UDS Core deployed
 - [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) installed
-- Access to a Kubernetes cluster with `kubectl` permissions in the `keycloak` namespace
+- Access to a Kubernetes cluster with permissions in the `keycloak` namespace
 - For the headless option, familiarity with bundle overrides
 
 ## Before you begin
--- a/docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
+++ b/docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
@@ -113,9 +113,9 @@ Treat every admin account as a break-glass credential. The goal is to have no st
    > [!IMPORTANT]
    > Deleting the Secret does not remove or alter the admin user in Keycloak. It only clears the plaintext copy of the initial password from the cluster. The `KEYCLOAK_ADMIN` and `KEYCLOAK_ADMIN_PASSWORD` environment variables are only read on first startup, so deleting the Secret has no runtime effect after the initial bootstrap.
 
-3. **Delete the admin user when the work is complete**
+3. **(Optional) Delete the admin user when the work is complete**
 
-   After you finish the administrative task, remove the account so that no standing admin credential exists. Bootstrap a fresh one the next time you need admin access.
+   Removing the account after you finish the administrative task is recommended for security: no standing admin credential exists, and you bootstrap a fresh one the next time you need admin access. This step is optional; retain the account if your operational model requires a persistent master realm admin.
 
    1. In the Keycloak admin console, click the menu icon in the top-left and select **Users**.
    2. Select the admin user you created, click the three-dot menu, and choose **Delete**.
--- a/docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
+++ b/docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
@@ -34,6 +34,9 @@ Treat every admin account as a break-glass credential. The goal is to have no st
 > [!CAUTION]
 > A standing admin account or a leftover `keycloak-admin-password` Secret is equivalent to a cluster-wide superuser credential that anyone with `get secret` access in the `keycloak` namespace can read. Remove both as soon as your administrative task is complete.
 
+> [!NOTE]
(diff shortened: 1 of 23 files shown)


docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+3. **Delete the admin user when the work is complete**
+
+   After you finish the administrative task, remove the account so that no standing admin credential exists. Bootstrap a fresh one the next time you need admin access.
+
+   1. In the Keycloak admin console, click the menu icon in the top-left and select **Users**.
+   2. Select the admin user you created, click the three-dot menu, and choose **Delete**.
+
+   > [!NOTE]
+   > Once a master realm admin has existed, the `uds zarf connect keycloak` first-run form will not reappear and `insecureAdminPasswordGeneration` only fires on the initial install. To bootstrap a replacement admin later, follow the [Keycloak Credential Recovery](/operations/troubleshooting-and-runbooks/keycloak-credential-recovery/) runbook, which walks through `kc.sh bootstrap-admin` to create a temporary admin.
```
mjnagel: @chance-coleman / @slaskawi curious on thoughts here - I think we'd probably advocate for short-lived accounts (delete after usage) but my experience running through this is that you can't get back to the original bootstrap screen after a user has existed (even if one no longer exists). Maybe something I was missing in my process there...

Not sure if that friction might lean more towards changing this to an "option / recommendation" rather than an explicit step here?

chance-coleman: I think this is probably our best option (what you already have). The first-run bootstrap UI is a limitation upstream, not something we can re-trigger afaik.

I agree that its a good security practice to remove the admin users, but that's still a choice by end users. We could add a note right before the actual content of step 3 that says we recommend this for security reasons but its not required.

Step 3 in the keycloak credential recovery doc expects the admin user to be present (since its recovery of the password), might be worth a reword to account for this usecase.

mjnagel: Updated based on this - good callouts. Hopefully matches what you were thinking?

chance-coleman: I think this looks good! Would like to get @slaskawi 👀 on it as well, otherwise i'd approve this as is.

docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
```diff
@@ -0,0 +1,149 @@
...
+- [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) installed
+- Access to a Kubernetes cluster with `kubectl` permissions in the `keycloak` namespace
+- For the headless option, familiarity with bundle overrides
+
+## Before you begin
+
+UDS Core does not create a default admin user in any realm. You must create the first admin in the **master realm** (Keycloak's built-in admin realm) before you can manage the `uds` realm or any of its clients. Two bootstrap methods are supported:
+
+- `uds zarf connect keycloak` opens a port-forward to the Keycloak HTTP service so you can register an admin through the browser-based first-run form. This is the preferred method for all interactive installs.
+- `insecureAdminPasswordGeneration` is a Helm value that provisions an admin user from a Kubernetes Secret at chart install time. Use this method only when you cannot run the interactive flow (for example, fully automated installs with no browser access).
```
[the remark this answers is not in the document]
mjnagel: Yeah...I intentionally didn't change these to keep this PR focused on just the main docs (and only run the docs CI 🫣). But can edit if we want to now.

docs/operations/troubleshooting-and-runbooks/keycloak-credential-recovery.mdx
```diff
@@ -78,9 +81,12 @@ This runbook uses the Keycloak [Admin bootstrap and recovery](https://www.keyclo
...
    Navigate to `https://keycloak.<admin_domain>/` and log in with the `temp-admin` user and the password you set in the previous step.
 
-3. **Reset the admin password**
+3. **Restore a permanent admin user**
 
-   Once logged in, navigate to the **Users** tab, select the **admin** user, go to the **Credentials** tab, and click **Reset Password**. Set a new password for the admin account.
+   Use whichever case applies:
+
+   - **Reset the password on an existing admin user:** Navigate to the **Users** tab, select the admin user, go to the **Credentials** tab, and click **Reset Password**. Set a new password for the account.
+   - **Create a new admin user** (for example, the previous break-glass admin was deleted): Navigate to the **Users** tab, click **Add user**, and create an account. Open the **Role mapping** tab, click **Assign role**, and assign the `admin` role from the master realm.
```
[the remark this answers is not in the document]
mjnagel: Hm - will have to validate this one. Not 100% sure if this is correct or not 🤷

docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+## What you'll accomplish
+
+You'll create, use, and remove a Keycloak admin account in the master realm as a just-in-time credential. UDS Core ships without a default admin, so you must bootstrap one before you can access the admin console at `https://keycloak.<admin_domain>/`. This guide covers the two supported bootstrap methods and the recommended lifecycle for keeping admin credentials out of the cluster.
+
+## Prerequisites
+
+- UDS Core deployed
+- [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) installed
+- Access to a Kubernetes cluster with `kubectl` permissions in the `keycloak` namespace
```
joelmccoy: nit: kubectl permissions feels like weird wording here
```suggestion
- Access to a Kubernetes cluster with permissions in the `keycloak` namespace
```

docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
```diff
@@ -0,0 +1,149 @@
...
+
+3. **(Optional) Delete the admin user when the work is complete**
+
+   Removing the account after you finish the administrative task is recommended for security: no standing admin credential exists, and you bootstrap a fresh one the next time you need admin access. This step is optional; retain the account if your operational model requires a persistent master realm admin.
+
+   1. In the Keycloak admin console, click the menu icon in the top-left and select **Users**.
+   2. Select the admin user you created, click the three-dot menu, and choose **Delete**.
+
+   > [!NOTE]
+   > Once a master realm admin has existed, the `uds zarf connect keycloak` first-run form will not reappear and `insecureAdminPasswordGeneration` only fires on the initial install. To bootstrap a replacement admin later, follow the [Keycloak Credential Recovery](/operations/troubleshooting-and-runbooks/keycloak-credential-recovery/) runbook, which walks through `kc.sh bootstrap-admin` to create a temporary admin.
```
joelmccoy: nit: I almost feel like we should call this higher up before the steps and outside of the tabs (this note calls out both methods, but when in tabs is confusing too). Basically call out to only use this runbook for the first time use of admin user

joelmccoy (dismissed):
Generally LGTM with some nits for consideration

docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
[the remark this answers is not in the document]
joelmccoy: fine to defer

docs/operations/troubleshooting-and-runbooks/keycloak-credential-recovery.mdx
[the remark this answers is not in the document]
mjnagel: Based on parity with the "bootstrap admin" will adjust wording, but to note here - `Realm role`, `admin` is what the bootstrap admin gets.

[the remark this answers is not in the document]
mjnagel: Fixed.

docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
mjnagel: Moved to above the steps - agreed on the placement.

joelmccoy (dismissed):

docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
```diff
@@ -0,0 +1,149 @@
...
+description: Create, rotate, and remove the Keycloak master realm admin account using `uds zarf connect keycloak` or the headless `insecureAdminPasswordGeneration` bundle override.
+sidebar:
+  order: 3.001
+---
+
+import { Steps, Tabs, TabItem } from '@astrojs/starlight/components';
+
+## What you'll accomplish
+
+You'll create, use, and remove a Keycloak admin account in the master realm as a just-in-time credential. UDS Core ships without a default admin, so you must bootstrap one before you can access the admin console at `https://keycloak.<admin_domain>/`. This guide covers the two supported bootstrap methods and the recommended lifecycle for keeping admin credentials out of the cluster.
```
slaskawi: > remove a Keycloak admin account in the master realm as a just-in-time credential

This doesn't sound good - epecifically with the User Auth space. Just in time credentials might be interpreted as passwordless approach like FIDO Tokens. Can we somehow reword it?

docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
```diff
@@ -0,0 +1,149 @@
...
+- UDS Core deployed
+- [UDS CLI](https://github.com/defenseunicorns/uds-cli/releases) installed
+- Access to a Kubernetes cluster with permissions in the `keycloak` namespace
+- For the headless option, familiarity with bundle overrides
+
+## Before you begin
+
+UDS Core does not create a default admin user in any realm. You must create the first admin in the **master realm** (Keycloak's built-in admin realm) before you can manage the `uds` realm or any of its clients. Two bootstrap methods are supported:
+
+- `uds zarf connect keycloak` opens a port-forward to the Keycloak HTTP service so you can register an admin through the browser-based first-run form. This is the preferred method for all interactive installs.
```
slaskawi: > first-run form

The official name for thie page is the "Welcome Page". Maybe we could reword it to use that term?

docs/operations/troubleshooting-and-runbooks/keycloak-credential-recovery.mdx
```diff
@@ -78,9 +81,12 @@ This runbook uses the Keycloak [Admin bootstrap and recovery](https://www.keyclo
 
    Navigate to `https://keycloak.<admin_domain>/` and log in with the `temp-admin` user and the password you set in the previous step.
 
-3. **Reset the admin password**
+3. **Restore a permanent admin user**
```
slaskawi: I would prefer if we called it restoring admin credentials or something similar. Technically, the credentials are lost in this scenario, not the account itself. Maybe "Recoving lost admin credentials"?

docs/how-to-guides/identity-and-authorization/manage-admin-access.mdx
mjnagel: I'll just remove this wording.

docs/operations/troubleshooting-and-runbooks/keycloak-credential-recovery.mdx
mjnagel: So the reason I made this change is because of the other doc that I am adding. _IF_ we advise users to delete the admin account, then from what I could tell, they would need to go through this same "recovery" process to create a new user. In this case they wouldn't have "lost" credentials, they just wouldn't have an admin user currently.

Is there a better way to advise here for scenarios where we recommend short-lived admin credentials? It seems like getting a new admin user requires going through this same process.

chance-coleman (approved):

docs/operations/troubleshooting-and-runbooks/keycloak-credential-recovery.mdx
slaskawi: But then the other doc says:

> **(Optional) Delete the admin user when the work is complete**

So it's optional and we don't know if the System Administrator deleted it. I won't go further into the rabbit hole and leave it up to you how to address this. 

> Is there a better way to advise here for scenarios where we recommend short-lived admin credentials? It seems like getting a new admin user requires going through this same process.

We might want to cut a new ticket to investigate this but there are interesting options on the table (there were not there a few versions ago)!

We could use Keycloak Workflows to deactivate the admin user after a day or two after the installation. I'm not a big fan of deleting the admin account as it makes following the audit trail harder. I'm a big fan of removing the credentials though (and leaving the account there). 

I also think we might want to use `KC_BOOTSTRAP_ADMIN_USERNAME` and `KC_BOOTSTRAP_ADMIN_PASSWORD` ([docs](https://www.keycloak.org/server/bootstrap-admin-recovery#_create_an_admin_user)) which would produce that yellow bar saying that you're using a Temporary Admin User at the top (which is the intention here).

slaskawi (approved):
Added a comment and suggested slighly different path forward for short-lived credentials. It would be great to address this but it's something certainly out of the scope for this work

docs/operations/troubleshooting-and-runbooks/keycloak-credential-recovery.mdx
mjnagel: > So it's optional and we don't know if the System Administrator deleted it. I won't go further into the rabbit hole and leave it up to you how to address this.

Agreed on this - I was hoping the "use whichever case applies" would kind of cover this but I did also change the wording to `Restore admin access` rather than the other two options.

mjnagel: The bootstrap stuff is interesting - probably deserves a more full evaluation of what we want to support and recommend in UDS.

Issue #1292, opened 2025-02-17
Consider allowing Public Clients


slaskawi:
### Is your feature request related to a problem? Please describe.

Currently, creating Public Clients (Clients without any credentials) with the Standard Flow is prohibited by the UDS Platform and results in an error. It has been implemented this way intentionally as part of the Security Hardening, see [https://github.com/defenseunicorns/uds-core/pull/630](<https://github.com/defenseunicorns/uds-core/pull/630>). The only use case we've considered so far is the Device Flow.

Recently we have discovered a new use case, where a React app configuration requires using Standard Flow with PKCE. The goal of this ticket is to reconsider our previous choice and consider allowing this type of flows.

Hard requirements:

* Since Public Clients cannot store the Client Credentials securely, using PKCE and protecting against stealing Code is necessary. The Platform should enforce this.

### Additional context

* AuthService v1.0.3+ support PKCE, see [https://github.com/istio-ecosystem/authservice/pull/265](<https://github.com/istio-ecosystem/authservice/pull/265>). UDS uses  v1.0.4
* Here's a tutorial how to manually enforce PKCE in Keycloak: [https://skycloak.io/blog/keycloak-how-to-create-a-pkce-authorization-flow-client/](<https://skycloak.io/blog/keycloak-how-to-create-a-pkce-authorization-flow-client/>)

Outcome of this ticket should be an ADR for the approach in handling public clients, with implementation to follow in a separate ticket

mjnagel:
I think it's good for us to evaluate what is possible and what is easiest/best for the mission applications. To date I'd say we typically say "if you can't manage client credentials/login flow in your app, offload that responsibility to authservice". This works in most cases from what we've seen so far, and is generally a pleasant experience for the end users. If they want to they don't even have to implement any auth logic on their end - Istio will validate the JWT and only pass authenticated users through to the application.

It would be good to identify a concrete scenario where public clients provide more/are easier for users to make sure we're targeting a real use case here (and that it's one we would encourage, instead of Authservice). Definitely worth revisiting this - just providing context on what I think we should clearly define to make a decision here.

bburky:
The description I was recently provided was:

> We are using the Keycloak Javascript Adapter, which supports PKCE, and so we're wondering if the publicClient + standardFlowEnabled combination could be reconsidered for our use case.

I'll try to get additional detail, but is there any progress to allow publicClient + standardFlowEnabled + PKCE which is safe as far as I know?

PR #2598, opened 2026-04-20
feat(CORE-27): allow opt-in public clients

slaskawi:
## Description

This Pull Request adds a new ADR as well as the implementation for Public Keycloak Clients. 

## Related Issue

Relates to https://linear.app/defense-unicorns/issue/CORE-27/consider-allowing-public-clients

## Type of change

## Steps to Validate

Tested automatically:

* The negative case is tested in this repo
* The positive case is tested in UDS Identity Config

## Checklist before merging


adrs/0010-allow-opt-in-public-clients.md
```diff
@@ -0,0 +1,71 @@
...
+
+Date: 2026-04-20
+
+## Status
+
+Accepted
+
+## Context
+
+UDS Core has historically rejected Keycloak SSO clients with `publicClient: true` unless they are configured exclusively for the OAuth 2.0 Device Authorization Grant. The restriction was introduced in [uds-core#630](https://github.com/defenseunicorns/uds-core/pull/630) as a defensive measure: device flow was the only well-understood use case we had at the time, and there was no pressing demand from Mission Heroes for browser-based public clients. [CORE-27](https://linear.app/defense-unicorns/issue/CORE-27/consider-allowing-public-clients) and [uds-core#1292](https://github.com/defenseunicorns/uds-core/issues/1292) have tracked the arguments for and against revisiting that choice.
```
slaskawi: Rework to "Keycloak Public Clients"

slaskawi: Apart from highlights:

> [CORE-27](https://linear.app/defense-unicorns/issue/CORE-27/consider-allowing-public-clients) and [uds-core#1292](https://github.com/defenseunicorns/uds-core/issues/1292) have tracked the arguments for and against revisiting that choice.

We need a very short summary of the arguments here.

adrs/0010-allow-opt-in-public-clients.md
```diff
@@ -0,0 +1,71 @@
...
+
+The same question has now surfaced in Slack at least three times (June 2025, September 2025, April 2026), with increasingly concrete use cases from Mission Heroes. This ADR records the arguments, the decision to allow public clients as an explicit opt-in, and the controls that accompany the opt-in to keep the platform's default security posture intact.
```
slaskawi: Omit the information about Slack conversation.

adrs/0010-allow-opt-in-public-clients.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Arguments for keeping public clients restricted
+
+These arguments come from the June 2025 and April 2026 product-support discussions, principally from Blake Burkhart, Micah Nagel, and Sebastian Laskawiec.
+
+- **Public clients have the highest attack surface of the available auth patterns.** Blake laid out a hierarchy for UDS Core: admin-gateway apps (network-restricted) → Authservice-protected apps (no unauthenticated network access) → confidential-client apps (backend holds credentials) → public-client apps (browser holds no credentials). Every step down that ladder reduces the blast radius of both misconfiguration and CVE exploitation. Authservice in particular turns every app into a "no pre-auth CVE" target because unauthenticated traffic never reaches the workload.
```
slaskawi: This piece needs to be shortened and more direct.

adrs/0010-allow-opt-in-public-clients.md
```diff
@@ -0,0 +1,71 @@
...
+- **PKCE is a floor, not a ceiling.** Sebastian noted that at a minimum PKCE is required, but federated client authentication (Signed JWT + Kubernetes Service Accounts, per [Keycloak's federated client authentication](https://www.keycloak.org/2026/01/federated-client-authentication)) is strictly stronger and should be preferred when the architecture allows.
+- **The original decision was "not yet," not "never."** In the October 2025 thread, Micah explicitly said "there was nothing wrong with us exposing PKCE + public though." The issue was cost-of-support versus demand, and Authservice was sufficient for all the cases that had reached us.
+
+### Arguments in favor of Mission Heroes using public clients
+
+These are the concrete use cases and pressures surfaced on Slack.
+
+- **HNCD / Tim Seagren (June 2025):** A mission hero wanted `publicClient: true` + `standardFlowEnabled: true` with PKCE and redirects for user auth. The UDS Operator rejected the Package CR. This was the first signal that our blanket rejection was creating friction for app teams who had already built around OIDC + PKCE.
```
slaskawi: Same here.

adrs/0010-allow-opt-in-public-clients.md
```diff
@@ -0,0 +1,71 @@
...
+- **Orca / Andrew Roth (September 2025):** A mission hero with a split frontend/backend deployment (static HTML/JS frontend + FastAPI backend on a different origin) could not make the two talk to each other cleanly under Authservice. Two clients produced two audiences and CORS pain; one client over two virtual services conflicted with Authservice's use of the first `redirectUri` to match hostnames. After a very long troubleshooting thread the app had to be restructured to co-host frontend and backend under `/` and `/api`. The thread explicitly asked "what needs to be worked on so that UDS Core supports what we are trying to do without needing to write custom stuff."
```
slaskawi: No Mission names please.

adrs/0010-allow-opt-in-public-clients.md
```diff
@@ -0,0 +1,71 @@
...
+3. **Always-enforced incompatibilities remain.** Public clients combined with `secret`, `secretConfig`, `serviceAccountsEnabled`, `enableAuthserviceSelector`, or `protocol: saml` are rejected regardless of the flag. These combinations are fundamentally invalid for a credential-less client.
+4. **Two-layer enforcement.** The flag is enforced by the UDS Operator admission webhook today, so Package CRs are rejected at apply time without hitting Keycloak. We will also expose the same toggle through the `UDSClientPolicyPermissionsExecutor` plugin in [uds-identity-config](https://github.com/defenseunicorns/uds-identity-config) so the setting is visible and editable in the Keycloak Admin UI (under the UDS Client Profile), mirroring how `SECURITY_HARDENING_ADDITIONAL_*` is surfaced today. That gives consistent behavior whether a client is created via the UDS Operator, via direct Keycloak Admin API calls, or via Terraform / OpenTofu.
```
slaskawi: No - let's stick to one layer

docs/how-to-guides/identity-and-authorization/configure-user-account-settings.mdx
```diff
@@ -126,7 +127,30 @@ All settings in this guide are applied via `realmInitEnv` in a bundle override.
...
+                 value:
+                   SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED: "true"
+   ```
+
+   The procedure is identical to extending the protocol mapper and client scope allow lists above — the switch name is `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED`.
+
+   > [!WARNING]
+   > Public clients cannot securely store client credentials, so any browser or device that runs the client can present its `client_id` without proof of identity. That makes public clients a common stepping stone for attacks — authorization code interception, redirect URI injection, token theft from browser storage, and open redirect abuse. Even with PKCE, a public client expands the attack surface compared with a confidential client. Prefer a confidential client (backend-for-frontend, BFF, or server-side token exchange) for any application that can hold a secret. Enable this flag only when the application architecture makes a confidential client impossible — for example, a device-flow CLI or a fully static single-page app with no server component — and review [CORE-27](https://linear.app/defense-unicorns/issue/CORE-27/consider-allowing-public-clients) for the full risk analysis.
```
slaskawi: We also need a piece of documentation describing the UDS Identity Config validation rules from https://github.com/defenseunicorns/uds-identity-config/pull/841

src/keycloak/chart/values.yaml
```diff
@@ -56,6 +56,10 @@ realmInitEnv:
   # @lulaEnd ef276ec4-7921-4287-a18c-80dec5f8e742
   SECURITY_HARDENING_ADDITIONAL_PROTOCOL_MAPPERS: ""
   SECURITY_HARDENING_ADDITIONAL_CLIENT_SCOPES: ""
+  # When "true", the UDS Operator allows SSO clients with publicClient: true beyond the
+  # default device-flow-only case. Public clients cannot securely store client credentials,
+  # so this is disabled by default. See the Security hardening docs for risk considerations.
```
slaskawi: Add a short info about the validation rules from https://github.com/defenseunicorns/uds-identity-config/pull/841

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -310,22 +311,37 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
-        client.protocol === Protocol.Saml ||
-        client.attributes?.["oauth2.device.authorization.grant.enabled"] !== "true")
+        client.protocol === Protocol.Saml)
     ) {
       return req.Deny(
         `The client ID "${client.clientId}" sets options incompatible with publicClient`,
       );
     }
+
+    // Non-device-flow public clients are denied unless SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED
```
slaskawi, line 342: Add a comment that this implementation follows the backwards behavioral compatibility rules. As it was previously - all Device Flow requests are being sent to Keycloak, whereas the regular Public Clients are now being considered based on the flag.

docs/how-to-guides/identity-and-authorization/configure-device-flow.mdx
```diff
@@ -21,13 +21,13 @@ You'll configure a UDS Package to use the [OAuth 2.0 Device Authorization Grant]
...
+This flow creates a **public client** (a client with no secret). Device flow is the only public client flow allowed by default — it requires no browser redirect, so it avoids the authorization-code interception risks that apply to other public client flows. The following constraints still apply:
 
-- `standardFlowEnabled` must be explicitly set to `false`. The UDS operator will reject the `Package` CR if it is not. Public clients in UDS Core are restricted to device flow only.
-- `publicClient: true` is incompatible with `serviceAccountsEnabled: true`
+- `standardFlowEnabled` must be set to `false`. Enabling the standard flow on a public client is gated by `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED` in `realmInitEnv`. See [Configure user accounts and security policies](/how-to-guides/identity-and-authorization/configure-user-account-settings/) for how to opt in, and note the warning about public client risk before doing so.
+- `publicClient: true` is incompatible with `serviceAccountsEnabled: true`, `secret`, `secretConfig`, `enableAuthserviceSelector`, and `protocol: saml` — these always reject regardless of the hardening flag.
```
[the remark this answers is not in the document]
slaskawi: Good catch, fixed. Replaced the em dash with a period/new sentence.

docs/how-to-guides/identity-and-authorization/configure-device-flow.mdx
```diff
@@ -21,13 +21,13 @@ You'll configure a UDS Package to use the [OAuth 2.0 Device Authorization Grant]
...
 
 > [!NOTE]
-> If your application needs **both** device flow and a standard browser redirect flow, create two separate SSO clients in the same `Package` CR, one for each flow. They cannot be combined in a single client.
+> If your application needs **both** device flow and a standard browser redirect flow, create two separate SSO clients in the same `Package` CR, one for each flow. They cannot be combined in a single client. The redirect-flow client should typically be a **confidential** (non-public) client — enabling `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED` is only needed when the application cannot hold a secret.
```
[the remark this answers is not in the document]
slaskawi: Good catch, fixed per the suggestion.

docs/how-to-guides/identity-and-authorization/configure-user-account-settings.mdx
```diff
@@ -126,7 +127,30 @@ All settings in this guide are applied via `realmInitEnv` in a bundle override.
    > [!CAUTION]
    > Only add protocol mappers and client scopes that your applications explicitly require. Each addition expands the set of capabilities packages in the realm are permitted to use.
 
-4. **Create and deploy your bundle**
+4. **(Optional) Allow non-device-flow public clients**
+
+   By default, the UDS Operator rejects SSO clients with `publicClient: true` unless they are configured for OAuth 2.0 Device Authorization Grant (device flow). To permit other uses — for example, a single-page app using Standard Flow with PKCE — set `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED` to `"true"`:
+
```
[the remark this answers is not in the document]
slaskawi: Fixed. Replaced em dashes on the step intro and the follow-up sentence with parentheses/period splits.

docs/how-to-guides/identity-and-authorization/configure-user-account-settings.mdx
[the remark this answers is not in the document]
slaskawi: Fixed. Replaced em dashes in the warning with a colon and parentheses.

slaskawi: Good point. Added a section after the warning documenting the defense-in-depth rules enforced by the UDS Identity Config Keycloak plugin (protocol, implicit/ROPC, redirect URI hygiene, loopback http exception, PKCE requirement, device-flow exemption).

docs/reference/configuration/identity-and-authorization.md
```diff
@@ -145,6 +146,31 @@ To extend the allow-list, set `SECURITY_HARDENING_ADDITIONAL_PROTOCOL_MAPPERS` o
 > [!CAUTION]
 > Do not use the `bare-groups` claim to protect applications. Because it strips path information, two groups with the same name but in different parent groups are indistinguishable, which creates authorization vulnerabilities.
 
+#### Public clients
+
+By default, the UDS Operator rejects SSO clients with `publicClient: true` unless they are configured for OAuth 2.0 Device Authorization Grant (device flow). Other uses — including Standard Flow with PKCE for single-page apps — must be opted into by setting `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED` to `"true"` in `realmInitEnv`.
```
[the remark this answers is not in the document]
slaskawi: Fixed per the suggestion. Also swept one more em dash in the warning block on line 172 while I was in the file.

src/keycloak/chart/values.yaml
slaskawi: Added a comment block above the setting summarizing the plugin-enforced validation rules (OIDC-only, no implicit/ROPC, redirect URI hygiene, loopback http exception, PKCE requirement with device-flow exemption).

slaskawi: Reworked the implementation per your guidance. The UDS Operator no longer reads the keycloak-realm-env secret or gates on the flag — it always forwards Package CR requests to Keycloak. Validation is fully delegated to the UDS Identity Config Keycloak plugin / UDS Client Profile, which means bypasses via Admin API or Terraform get the same enforcement. Changes: removed refreshPublicClientsFlag / handleKeycloakRealmEnvUpdate / KEYCLOAK_REALM_ENV_* constants in config.ts, dropped the securityHardeningPublicClientsEnabled field from the Config type, removed the Secret Watch in index.ts, stripped the publicClient validation block in package-validator.ts (replaced with a comment explaining delegation to Keycloak), and deleted the related Pepr tests and the test/vitest integration spec. The values.yaml still surfaces SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED for the plugin to consume.

slaskawi: Reverted my prior rework — on reflection, deleting the Pepr-side validation now would create a window where a cluster could upgrade to this UDS Core version before its pinned uds-identity-config image contains the plugin-side executor (from uds-identity-config#841), causing non-compliant clients to be silently accepted. Added a comment in package-validator.ts explaining the rationale and pointing to [CORE-497](https://linear.app/defense-unicorns/issue/CORE-497/remove-pepr-side-public-client-validation-after-uds-identity-config), which tracks removing the Pepr logic once configImage is guaranteed to be v0.27.0+.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -39,8 +39,43 @@ export const UDSConfig: Config = {
...
+
+/**
+ * Refreshes UDSConfig.securityHardeningPublicClientsEnabled from the keycloak-realm-env Secret.
+ * Missing secret or key falls back to false. Any non-"true" value is treated as false.
+ */
+export async function refreshPublicClientsFlag(): Promise<void> {
+  try {
+    const secret = await K8s(kind.Secret)
+      .InNamespace(KEYCLOAK_REALM_ENV_NAMESPACE)
+      .Get(KEYCLOAK_REALM_ENV_SECRET_NAME);
```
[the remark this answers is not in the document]
slaskawi: Added the guard per the suggestion — if `.Get()` succeeds but returns an object without `.data`, we now log at debug and default the flag to false instead of letting decodeSecret throw.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -39,8 +39,43 @@ export const UDSConfig: Config = {
...
+      return;
+    }
+    configLog.error(e, "Failed to read keycloak-realm-env Secret");
+    throw e;
+  }
+}
+
+export function handleKeycloakRealmEnvUpdate(secret: kind.Secret) {
+  const decoded = decodeSecret(secret);
+  UDSConfig.securityHardeningPublicClientsEnabled = decoded[PUBLIC_CLIENTS_KEY] === "true";
```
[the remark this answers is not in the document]
slaskawi: Added the same defensive guard here — missing `.data` on the Watch payload now logs at debug and defaults the flag to false instead of throwing.

src/pepr/operator/controllers/config/types.ts
```diff
@@ -27,4 +27,7 @@ export type Config = {
   kubeApiCIDR: string;
   kubeNodeCIDRs: string[];
   isIdentityDeployed: boolean;
+  // Mirrors SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED from Keycloak realmInitEnv.
+  // When false (default), the Package validator rejects SSO clients with publicClient: true.
```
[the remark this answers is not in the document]
slaskawi: Fixed per the suggestion.

src/pepr/operator/index.ts
```diff
@@ -152,6 +155,13 @@ When(a.Secret)
   .WithName(KEYCLOAK_CLIENTS_SECRET_NAME)
   .Reconcile(s => updateKeycloakClientsSecret(s, false));
 
+// Watch the Keycloak realm env Secret for SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED changes
+When(a.Secret)
+  .IsCreatedOrUpdated()
+  .InNamespace(KEYCLOAK_REALM_ENV_NAMESPACE)
+  .WithName(KEYCLOAK_REALM_ENV_SECRET_NAME)
+  .Reconcile(handleKeycloakRealmEnvUpdate);
+
```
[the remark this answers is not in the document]
slaskawi: Added the IsDeleted() watch per the suggestion — on Secret deletion the flag resets to false so stale in-memory state can't keep public clients enabled after an uninstall/upgrade.

src/pepr/operator/controllers/config/config.ts
```diff
@@ -39,8 +39,59 @@ export const UDSConfig: Config = {
...
+
+/**
+ * Refreshes UDSConfig.securityHardeningPublicClientsEnabled from the keycloak-realm-env Secret.
+ * Missing secret or key falls back to false. Any non-"true" value is treated as false.
+ */
+export async function refreshPublicClientsFlag(): Promise<void> {
+  try {
+    const secret = await K8s(kind.Secret)
+      .InNamespace(KEYCLOAK_REALM_ENV_NAMESPACE)
+      .Get(KEYCLOAK_REALM_ENV_SECRET_NAME);
```
bburky, line 57: There is some risk this can get out of sync with the actual Keycloak policy configuration. For existing clusters, setting variables like `REALM_SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED` in uds-identity-config usually does nothing (they're only applied at initial import), so the admin has to manually change the setting in their Keycloak configuration.

How does this operator handle the policy being _enabled_ in the Keycloak settings but `REALM_SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED` _disabled_? (And admin might do this on accident. And normally it's not a problem to do this because the uds-identity-config values don't matter after initial install)

I assume it's not possible for the UDS Operator to use Keycloak Admin APIs to query for the actual policy configuration directly, instead of reading this secret? (I assume not, this would need extra permissions for the operator?)

This all probably fine, just want to make sure we've thought through it and handle or document the edge cases.

mjnagel: I actually was about to comment something similar about this config and associated validation seeming "extra" and being potentially problematic. I did note this comment on the validation:
```
    // The UDS Identity Config Keycloak plugin (UDSClientPolicyPermissionsExecutor)
    // enforces an equivalent (and stricter) policy on the Keycloak side, so in principle
    // this Operator-side validation is redundant. We keep it here because UDS Core and
    // uds-identity-config ship as separate images with independent version pins: a
    // cluster can be upgraded to a UDS Core carrying this flag before its pinned
    // configImage is bumped to a version containing the plugin-side executor, and
    // relying solely on the plugin would silently accept non-compliant clients during
```

I think my 2 cents is that could just have an "unconditional" validation of public client + pkce in our validator (basically mirror the check on identity-config if we keep that) rather than conditionally doing it on potentially an out of sync config (out of sync with actual Keycloak API even if in sync with the secret).

Related comment: https://github.com/defenseunicorns/uds-identity-config/pull/841#discussion_r3118974257

slaskawi: I think it's a good idea. I reworked the code to omit most of the validation bits (and giving the Mission Heroes the flexibility they need) but created [CORE-498](https://linear.app/defense-unicorns/issue/CORE-498/uds-operator-policy-for-validating-keycloak-client) to introduce a new Keycloak Client Policy in UDS Operator and perform strictier validation.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -304,27 +304,33 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
-    // If this is a public client ensure that it only sets itself up as an OAuth Device Flow client
-    if (
-      client.publicClient &&
-      (client.standardFlowEnabled !== false /* default true */ ||
-        client.serviceAccountsEnabled /* default false */ ||
-        client.secret !== undefined ||
-        client.secretConfig?.name !== undefined ||
-        client.secretConfig?.template !== undefined ||
-        client.enableAuthserviceSelector !== undefined ||
-        client.protocol === Protocol.Saml ||
```
mjnagel, line 322: I think we still want to enforce that users aren't setting some of these options?
- Service account: seems like something to block
- Secret: doesn't make sense for public client / might be blocked by keycloak but we can block here?
- secretConfig: currently we [don't make a secret for public clients](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/keycloak/client-sync.ts#L176) so these wouldn't have any effect. I actually wonder if there's a use-case for making a secret (or configmap) since the client representation includes other fields that could be useful by the app for "dynamic" configuration (endpoints, etc). Definitely defer to a future issue if there is value here.
- enableAuthserviceSelector: definitely shouldn't allow public clients with this
- protocol = saml: PKCE is OIDC only I think, so we'd want this still (probably would be blocked on Keycloak side too)

test/vitest/keycloak-public-clients.spec.ts
```diff
@@ -0,0 +1,73 @@
+/**
```
mjnagel, line 1: I don't know that we need an e2e test like this here? I think this is really just testing the validator again, which is tested in our unit tests.

I could see some value in a "true" e2e test of an app using a public+pkce client (and maybe device flow too) but think we can defer that to later.

adrs/0010-allow-opt-in-public-clients.md
```diff
@@ -0,0 +1,59 @@
...
+- **Keycloak's own admin console is public + standard flow.** The built-in `security-admin-console` is already used in production for CAC-authenticated realm management.
+- **Existing workarounds bypass operator lifecycle management.** Public clients were created manually via the Admin UI or Terraform, avoiding the UDS Operator entirely and reducing platform visibility.
+
+## Decision
+
+We will allow public SSO clients in UDS Core as an **explicit, default-off opt-in** controlled by a new `realmInitEnv` flag, `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED`, following the same pattern as the existing `SECURITY_HARDENING_ADDITIONAL_PROTOCOL_MAPPERS` and `SECURITY_HARDENING_ADDITIONAL_CLIENT_SCOPES` settings.
+
+The above translates technically to:
+
+1. **Default off.** `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED` defaults to `"false"`. UDS Core's out-of-the-box posture is unchanged: non-device-flow public clients are rejected by admission.
```
mjnagel: I know there's a parallel convo on the identity-config side but I do want to call out here in the ADR as well - this is actually a breaking change since previously we only enforced this restriction in the Operator / Package CR (not in Keycloak itself).

At minimum this should probably be called out in the ADR consequences.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -304,27 +304,33 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+    // standardFlowEnabled) are intentionally NOT enforced here. They are tracked in
+    // CORE-498 and will ship as a separate UDS Operator Policy with Exemption
+    // support to avoid breaking existing clusters:
+    // https://linear.app/defense-unicorns/issue/CORE-498/uds-operator-policy-for-validating-keycloak-client
+    if (client.publicClient) {
+      const isDeviceFlowOnly =
+        client.standardFlowEnabled === false &&
+        client.attributes?.["oauth2.device.authorization.grant.enabled"] === "true";
+      const pkceMethod = client.attributes?.["pkce.code.challenge.method"]?.trim();
+      if (!isDeviceFlowOnly && !pkceMethod) {
```
chance-coleman: This accepts any non-empty string. `pkce.code.challenge.method: "plain"` passes. RFC 7636 plain means the code challenge equals the code verifier on the wire, so anyone who intercepted the authorization code can just use it directly. That's exactly the attack PKCE is supposed to prevent. The ADR at line 39 says PKCE is the only effective defense against code interception, but "plain" doesn't actually defend against it.

chance-coleman: tests for using `plain` and lowercase `s256` might be worth checking as well.

adrs/0010-allow-opt-in-public-clients.md
chance-coleman: At first read this confused me, if im understanding it correctly: the flag only gates the Keycloak plugin at reconcile time. An admin reading this will think the webhook blocks the Package CR when the flag is off. It doesn't. The CR gets stored and then Keycloak rejects the client later during reconciliation.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -304,27 +304,33 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+    //
+    // All other Keycloak Client hygiene rules (implicit flow, ROPC, redirect-URI
+    // rules including wildcards and non-loopback http://, built-in client shadowing,
+    // SAML public clients, publicClient combined with secret / secretConfig /
+    // serviceAccountsEnabled / enableAuthserviceSelector, serviceAccountsEnabled +
+    // standardFlowEnabled) are intentionally NOT enforced here. They are tracked in
+    // CORE-498 and will ship as a separate UDS Operator Policy with Exemption
+    // support to avoid breaking existing clusters:
+    // https://linear.app/defense-unicorns/issue/CORE-498/uds-operator-policy-for-validating-keycloak-client
+    if (client.publicClient) {
```
chance-coleman, line 318: possibly related to Micah's comment: `publicClient: true, protocol: saml, attributes: { "pkce.code.challenge.method": "S256" }` passes the validator. PKCE is an OAuth concept and does nothing on a SAML client in Keycloak. On top of that, Keycloak Client Policies (including the `allow-public-clients` executor) only apply to OIDC clients, not SAML. So a SAML public client with a PKCE attribute set bypasses both the operator check and the Keycloak plugin.

src/pepr/operator/crd/validators/package-validator.ts
slaskawi: Covered — the operator-side validator now blocks all of these explicitly for both public-client shapes (device-flow-only and flag-on non-device-flow). Specifically, `src/pepr/operator/crd/validators/package-validator.ts`:

- `serviceAccountsEnabled` → denied ([L335](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L335), [L372](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L372))
- `secret` → denied ([L341](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L341), [L373](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L373))
- `secretConfig` (`.name` and `.template`) → denied ([L342-343](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L342-L343), [L374-375](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L374-L375)). Blocking it now to keep the input surface clean; the "maybe emit a ConfigMap/Secret with non-credential client metadata for dynamic config" idea I'll track as a separate follow-up rather than silently accept the fields today.
- `enableAuthserviceSelector` → denied ([L349](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L349), [L376](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L376))
- `protocol == saml` → denied ([L330](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L330), [L377](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L377))

Each case also has a unit test in `package-validator.spec.ts`.

test/vitest/keycloak-public-clients.spec.ts
slaskawi: The negative case is covered by the e2e test in `test/vitest/keycloak-public-clients.spec.ts`, whereas the positive case is tested in the UDS Identity Config repo.

adrs/0010-allow-opt-in-public-clients.md
slaskawi: Addressed — we reverted the Keycloak-side enforcement (the `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED` realmInitEnv flag and the `allow-public-clients` executor tightening) and moved the gate back to the UDS Operator admission validator via a new `ALLOW_PUBLIC_CLIENTS` key in `uds-operator-config`. Net effect:

- No Keycloak-layer behavior change, so this is backwards compatible — clusters that already have public clients in Keycloak (Admin API, OpenTofu, Admin UI) keep reconciling.
- Device Flow is unaffected: device-flow-only public clients are always admitted, regardless of the flag.
- Non-device-flow public clients are still denied by default; the flag is an opt-in expansion of what the validator accepts, not a new restriction.

The ADR has been rewritten along these lines — see the updated Decision points (1-5) and the "deferred" alternative at the bottom.

src/pepr/operator/crd/validators/package-validator.ts
slaskawi: Thanks for the careful read. The design intent here is deliberate: the operator admission check only asserts `pkce.code.challenge.method` is set to a non-blank value, and we defer the *specific method string* to Keycloak, which enforces RFC 7636 at the authorization endpoint (exact, case-sensitive `plain` or `S256`).

You're right that `plain` doesn't defend against authorization-code interception. I updated the ADR to drop the "S256-only" framing and document that the specific method is an app-owner/Keycloak responsibility, with an explicit recommendation to prefer `S256`. The reference docs now also call this out and recommend `S256`. We considered pinning `S256` in the operator but chose not to duplicate Keycloak's enforcement for one detail of the spec — same reason we don't validate redirect URI wildcards here.

Tests: there's already a `plain` admission case; I added a lowercase `s256` case in the latest push — it passes admission but will be rejected by Keycloak at runtime, which is the intended split.

slaskawi: Covered in the reply above — a `plain` test already existed and a lowercase `s256` test is added in the latest push.

src/pepr/operator/crd/validators/package-validator.ts
slaskawi: Yes — addressed alongside Micah's restriction list. The validator now denies `protocol: saml` public clients even when PKCE is set and `ALLOW_PUBLIC_CLIENTS` is on (`package-validator.ts` [L330](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L330) for the flag-on path and [L377](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-27/src/pepr/operator/crd/validators/package-validator.ts#L377) for the device-flow-only path). Unit test: `"denies SAML public clients even with PKCE when ALLOW_PUBLIC_CLIENTS is on"` in `package-validator.spec.ts`.

adrs/0010-allow-opt-in-public-clients.md
slaskawi: Resolved by the broader rework — the gate is now at the UDS Operator (Pepr) admission webhook, not the Keycloak plugin. When `ALLOW_PUBLIC_CLIENTS="false"` (the default), the Package CR is rejected at admission and never stored, so the confusion you called out no longer applies. The ADR has been rewritten accordingly; see Decision point 5 ("Enforcement stops at UDS Operator admission") for the explicit scope call-out.

src/pepr/uds-operator-config/values.schema.json
```diff
@@ -15,6 +15,11 @@
           "type": "string",
           "enum": ["AUTO", "CLIENT_SECRET", "SIGNED_JWT"],
           "description": "Authentication mode for the UDS Operator Keycloak client. AUTO tries signed JWT first, then falls back to client secret."
+        },
+        "ALLOW_PUBLIC_CLIENTS": {
+          "type": ["string", "boolean"],
+          "enum": ["true", "false", true, false],
+          "description": "When 'true', the UDS Operator admits SSO clients with publicClient: true that use flows beyond OAuth 2.0 Device Authorization Grant (PKCE required). Accepts 'true'/'false' strings or YAML booleans; Helm stringifies either into the Kubernetes Secret value. Defaults to 'false'."
```
slaskawi: It was a bit surprising to see this being resolved to a boolean from a UDS Identity Config override: https://github.com/defenseunicorns/uds-identity-config/pull/841/changes#diff-b44c35cd9a89ecff0105a5929c0f4f0e1a2ec6a622dcff99ce43b38b0caabba8R31-R35

So I decided to make it a bit more "tolerant" and react properly on both booleans and strings.

mjnagel: Could we alternatively set it to a boolean only in the values, and have the secret template use helm `toString`?

src/pepr/operator/crd/validators/package-validator.ts
mjnagel: It seems like enforcing S256 here would be wise? We don't have any enforcement on the Keycloak/identity-config side (now) but this seems like a best practice that we'd want to require?

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
```diff
@@ -256,7 +256,7 @@ If your application expects credentials in a specific format (JSON config file,
    </Tabs>
 
    > [!NOTE]
-   > Custom protocol mappers and client scopes are subject to Keycloak's security hardening policy. If Keycloak rejects your mapper or scope, add it to the allow list via `SECURITY_HARDENING_ADDITIONAL_PROTOCOL_MAPPERS` or `SECURITY_HARDENING_ADDITIONAL_CLIENT_SCOPES`. See [Configure user accounts and security policies](/how-to-guides/identity-and-authorization/configure-user-account-settings/).
+   > Custom protocol mappers and client scopes are subject to Keycloak's security hardening policy. If Keycloak rejects your mapper or scope, add it to the allow list via `SECURITY_HARDENING_ADDITIONAL_PROTOCOL_MAPPERS` or `SECURITY_HARDENING_ADDITIONAL_CLIENT_SCOPES`. The same mechanism gates non-device-flow public clients on the Keycloak side via `SECURITY_HARDENING_PUBLIC_CLIENTS_ENABLED`. See [Configure user accounts and security policies](/how-to-guides/identity-and-authorization/configure-user-account-settings/).
```
chance-coleman: Reference `ALLOW_PUBLIC_CLIENTS` in `uds-operator-config` and link to `/reference/configuration/identity-and-authorization/#public-clients`

src/pepr/operator/crd/validators/package-validator.spec.ts
```diff
@@ -796,6 +797,258 @@ describe("Test validation of Package CRs", () => {
...
+    await validator(mockReq);
+    expect(mockReq.Approve).toHaveBeenCalledTimes(1);
+    UDSConfig.allowPublicClients = false;
+  });
+
+  // The admission check only verifies that the attribute is set to a non-blank
+  // value. Keycloak is the authoritative gate for the specific method string
+  // per RFC 7636 (exact "plain" or "S256", case-sensitive), so a lowercase
+  // "s256" passes the operator but will be rejected at the Keycloak authorization endpoint.
+  it("allows non-device-flow public clients with a lowercase 's256' PKCE value (Keycloak enforces the exact string)", async () => {
```
chance-coleman: My understanding is that `s256` (lowercase) will result in no protections because Keycloak's PKCE enforcer expects `S256` (capitalized) based on the [RFC 7636](https://www.rfc-editor.org/rfc/rfc7636#section-4.2). I think this test is allowing a footgun when we should not allow for it. I was expecting this to be a `deny` test.

mjnagel (commented):
LGTM pending two open threads.

src/pepr/operator/crd/validators/package-validator.ts
slaskawi: So I thought about leaving `plain` is it may be used as a workaround for Clients that don't provide the correct PKCE implementation. But after sleeping on it - I think you're right - we should enforce the most secure method here and enable Mission Heroes to manually create Keycloak Clients (via Tofu or ClickOps) if they wish to use incompatible implementation.

src/pepr/uds-operator-config/values.schema.json
slaskawi: Looking how other values are configured (like `cluster.caBundle.includeDoDCerts`) I think we should leave this `["boolean", "string"]`. This way we can switch to boolean in the config (like it should be in the first place) and leave string for more compatibility.

docs/how-to-guides/identity-and-authorization/register-and-customize-sso-clients.mdx
slaskawi: Done!

src/pepr/operator/crd/validators/package-validator.spec.ts
slaskawi: Absolutely!

mjnagel (approved):

PR #2602, opened 2026-04-20
chore: add retry to multi-arch manifest check

joelmccoy:
## Description

We have experienced some network flakes (specifically for registry1 flavors) on multi-arch image check workflow. Adding 1 retry to each request on failure.

## Type of change

## Steps to Validate
- CI passes

## Checklist before merging

### DIFF
--- a/scripts/check-multiarch.sh
+++ b/scripts/check-multiarch.sh
@@ -1,11 +1,35 @@
 #!/bin/bash
-# Copyright 2025 Defense Unicorns
+# Copyright 2025-2026 Defense Unicorns
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 set -e  # Exit on error
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 SRC_DIR="${1:-$SCRIPT_DIR/../src}"
+
+# Retry crane manifest against transient registry failures.
+MAX_ATTEMPTS="${CRANE_MAX_ATTEMPTS:-2}"
+RETRY_DELAY="${CRANE_RETRY_DELAY:-3}"
+
+fetch_manifest() {
+    local image="$1"
+    local attempt=1
+    local manifest=""
+    local delay="$RETRY_DELAY"
+    while [ "$attempt" -le "$MAX_ATTEMPTS" ]; do
+        if manifest=$(crane manifest "$image" 2>/dev/null) && [ -n "$manifest" ]; then
+            printf '%s' "$manifest"
+            return 0
+        fi
+        if [ "$attempt" -lt "$MAX_ATTEMPTS" ]; then
+            echo "  retry $attempt/$((MAX_ATTEMPTS - 1)) after ${delay}s" >&2
+            sleep "$delay"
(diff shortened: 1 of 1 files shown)


scripts/check-multiarch.sh
```diff
@@ -1,12 +1,36 @@
...
+    local delay="$RETRY_DELAY"
+    while [ "$attempt" -le "$MAX_ATTEMPTS" ]; do
+        if manifest=$(crane manifest "$image" 2>/dev/null) && [ -n "$manifest" ]; then
+            printf '%s' "$manifest"
+            return 0
+        fi
+        if [ "$attempt" -lt "$MAX_ATTEMPTS" ]; then
+            echo "  retry $attempt/$((MAX_ATTEMPTS - 1)) after ${delay}s" >&2
+            sleep "$delay"
+            delay=$((delay * 2))
```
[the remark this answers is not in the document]
joelmccoy: eh overkill for this script

scripts/check-multiarch.sh
```diff
@@ -1,12 +1,36 @@
...
+    local attempt=1
+    local manifest=""
+    local delay="$RETRY_DELAY"
+    while [ "$attempt" -le "$MAX_ATTEMPTS" ]; do
+        if manifest=$(crane manifest "$image" 2>/dev/null) && [ -n "$manifest" ]; then
+            printf '%s' "$manifest"
+            return 0
+        fi
+        if [ "$attempt" -lt "$MAX_ATTEMPTS" ]; then
+            echo "  retry $attempt/$((MAX_ATTEMPTS - 1)) after ${delay}s" >&2
```
mjnagel, line 25: Probably hard to know/see without it failing, but wonder if this extra log would just be extra noise when trying to review the output?

mjnagel (approved):

scripts/check-multiarch.sh
slaskawi: Or maybe a report at the end? Something along the lines: "Maximum retries: x, the longest wait time: y" ?

This way, after a few runs, we could make an educated guess on the default settings we should use here.

joelmccoy: I think the current log format isn't crazy and is pretty intuitive to me. Here is an example of what the run might look like: 
```
Example run with 150 images, image #42 fails both attempts, #87 flakes then succeeds:

Scanning ./src for unicorn and registry1 images...
Found 150 images

[1/150] quay.io/rfcurated/foo:1.0
  ✓ OK
[2/150] quay.io/rfcurated/bar:2.1
  ✓ OK
...
[42/150] registry1.dso.mil/ironbank/flaky/thing:v3
  retry 1/1 after 3s
  ERROR: Failed to fetch after 2 attempts
[43/150] quay.io/rfcurated/next:1.0
  ✓ OK
...
[87/150] registry1.dso.mil/ironbank/slow/thing:v2
  retry 1/1 after 3s
  ✓ OK
...
[150/150] quay.io/rfcurated/last:9.9
  ✓ OK

==========================================
Images missing amd64/arm64: 1

Missing multi-arch support:
registry1.dso.mil/ironbank/flaky/thing:v3
```

I think anything extra is overkill for this. I just wanted to add a quick retry because we randomly get flakes sometimes (although not all the time).

PR #2605, opened 2026-04-21
fix(ci): run istio gen-crds in autogenerated-check

briantwatson:
## Description

Working on istio v1 update I came across what I believe is a gap in our autogenerated-check.

We updated to 1.29.1 here: https://github.com/defenseunicorns/uds-core/pull/2387

## Related Issue

Relates to CORE-44

## Type of change

## Steps to Validate
n/a

## Checklist before merging


src/pepr/operator/crd/generated/k8s/gateway-v1.ts
```diff
@@ -1457,7 +1540,7 @@ export interface StatusListener {
    * and invalid Route kinds are specified, the implementation MUST
    * reference the valid Route kinds that have been specified.
    */
-  supportedKinds: SupportedKind[];
+  supportedKinds?: SupportedKind[];
```
briantwatson, line 1543: This change to optional is the most notable change here. From what I found no operator code references this.

src/istio/tasks.yaml
```diff
@@ -56,6 +56,9 @@ tasks:
             fi
           done
       - description: "Generate Gateway CRD types via KFC"
+        shell:
```
briantwatson, line 59: Failed in CI on the double `[[` https://github.com/defenseunicorns/uds-core/blob/4e8ba231384f8ef2515d042674251bd85e5d9b5c/src/istio/tasks.yaml#L54

mjnagel (dismissed):
This probably makes sense. We didn't do this previously really only because we generally don't need to 🫣. I think this is also a "problem" for the Prometheus CRD types that we generate if we want to change that now/later as well.

briantwatson:
> This probably makes sense. We didn't do this previously really only because we generally don't need to 🫣. I think this is also a "problem" for the Prometheus CRD types that we generate if we want to change that now/later as well.

@mjnagel went ahead and added prometheus-stack.  No diff on the tasks run

joelmccoy (approved):

slaskawi (approved):

PR #2614, opened 2026-04-27
fix: re-add keycloak client registration authpol

mjnagel:
## Description

As a defense in depth measure this re-adds protections to prevent client registrations via the "public" (tenant) client registration endpoint for the uds realm.

### DIFF
--- a/src/keycloak/chart/templates/istio-authpol-admin.yaml
+++ b/src/keycloak/chart/templates/istio-authpol-admin.yaml
@@ -41,7 +41,7 @@ spec:
             ports:
               - "8080"
             paths:
-              # Block client registration on the public (tenant) API
+              # Block client registration API access. Users can use Admin API via admin gateway instead.
               - "/realms/{{ .Values.realm }}/clients-registrations/*"
     - to:
         - operation:
--- a/src/keycloak/chart/templates/istio-authpol-admin.yaml
+++ b/src/keycloak/chart/templates/istio-authpol-admin.yaml
@@ -41,6 +41,13 @@ spec:
             ports:
               - "8080"
             paths:
+              # Block client registration on the public (tenant) API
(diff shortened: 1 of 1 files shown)


src/keycloak/chart/templates/istio-authpol-admin.yaml
```diff
@@ -36,6 +36,13 @@ spec:
         - source:
             notNamespaces:
             - monitoring
+    - to:
+        - operation:
+            ports:
+              - "8080"
+            paths:
+              # Block client registration on the public (tenant) API
+              - "/realms/{{ .Values.realm }}/clients-registrations/*"
```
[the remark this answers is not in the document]
mjnagel: This is intentional. Client registrations can still be performed, but exclusively through the admin API / gateway.

src/keycloak/chart/templates/istio-authpol-admin.yaml
```diff
@@ -36,6 +36,13 @@ spec:
         - source:
             notNamespaces:
             - monitoring
+    - to:
+        - operation:
+            ports:
+              - "8080"
+            paths:
+              # Block client registration on the public (tenant) API
```
bburky: The current comment suggests this is only on the tenant gateway, but I think this code blocks it everywhere?
```suggestion
              # Block client registration API access. Users can use Admin API via admin gateway instead.
```

mjnagel: Yeah this is probably better wording - I was thinking `/realms/uds/clients-registration` = "public" (served via tenant gateway) vs `/admin/realms/uds/clients` = "private" (served via admin gateway).

joelmccoy (approved):
LGTM. tested locally to ensure this is blocked on tenant gateway path

PR #2629, opened 2026-04-28
docs: update release notes to include newest patches

mjnagel:
## Description

Adds info on the newest patches to the docs site.

### DIFF
--- a/docs/operations/release-notes/1-0.mdx
+++ b/docs/operations/release-notes/1-0.mdx
@@ -49,6 +49,9 @@ This release removes the following deprecated fields: the legacy `CA_CERT` Zarf 
 
 ## Upgrade considerations
 
+> [!IMPORTANT]
+> Upgrade directly to v1.0.1 for security fixes.
+
 ### Pre-upgrade steps
 
 The following steps only apply if your bundle overrides the specific deprecated values being removed. If you are not using any of these overrides, no action is required.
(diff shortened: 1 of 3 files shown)


docs/operations/release-notes/1-0.mdx
```diff
@@ -49,6 +49,9 @@ This release removes the following deprecated fields: the legacy `CA_CERT` Zarf
 
 ## Upgrade considerations
 
+> [!IMPORTANT]
+> Upgrade directly to v1.0.1 for security fixes.
```
jasonwashburn, line 53: nit - consider adding a link to the patch release something like 
`Upgrade directly to v1.0.1 for security fixes. (Release)` or the like...first thing that comes to mind when I see a note about security fixes...I want to know what security fixes. I don't know that I would even think to scroll all the way to the bottom to use the updated diff link...it'd be nice to just have a release link right there to click on (applies to all 3)

jasonwashburn (dismissed):
One comment for consideration, but otherwise lgtm

docs/operations/release-notes/1-0.mdx
mjnagel: Not sure if I handled this in a sufficient way - open to other thoughts/suggestions here. We're leaning more into release notes on the docs site so the actual github release descriptions will generally just link to the docs site. With that approach...linking back to github here would feel a bit weird.

jasonwashburn (approved):

PR #2636, opened 2026-04-30
docs: add more clarity around tenant/admin lbs

mjnagel:
## Description

Adds some additional documentation around setting the admin gateway to an internal load balancer with upstream links/references.

The main guidance is in the build your bundle getting started guide, with links here from concepts and exposing app how-to.


docs/concepts/core-features/networking.mdx
```diff
@@ -46,10 +46,10 @@ UDS Core deploys two required gateways and one optional gateway:
...
 
-> [!TIP]
-> A common pattern is to expose the Tenant Gateway publicly (or broadly within a network) while keeping the Admin Gateway accessible only via private/internal networking, behind a VPN, bastion, or restricted subnet. This lets end users reach applications normally (including Keycloak for SSO, which is on the Tenant Gateway) while ensuring that admin interfaces like Grafana and the Keycloak admin console are never reachable from the public internet.
+> [!IMPORTANT]
+> The Admin Gateway should be reachable only from trusted networks. Expose the Tenant Gateway as your public (or broadly reachable) ingress and restrict the Admin Gateway to private/internal networking, a VPN, a bastion, or a restricted subnet. This pattern allows end users to still reach applications (including Keycloak for SSO), while ensuring that admin interfaces are never reachable from the public internet. This is typically configurable through loadbalancer controller specific annotations, see [restrict the admin gateway to internal networks](/getting-started/production/build-your-bundle/#restrict-the-admin-gateway-to-internal-networks) for examples and references on implementing this.
```
chance-coleman: super nit: should `loadbalancer` be one or two words? in the earlier paragraph its `load balancers`.

mjnagel: Good catch - I think convention is 2 words. Will update this.

mjnagel:
```suggestion
> The Admin Gateway should be reachable only from trusted networks. Expose the Tenant Gateway as your public (or broadly reachable) ingress and restrict the Admin Gateway to private/internal networking, a VPN, a bastion, or a restricted subnet. This pattern allows end users to still reach applications (including Keycloak for SSO), while ensuring that admin interfaces are never reachable from the public internet. This is typically configurable through load balancer controller specific annotations, see [restrict the admin gateway to internal networks](/getting-started/production/build-your-bundle/#restrict-the-admin-gateway-to-internal-networks) for examples and references on implementing this.
```

docs/getting-started/production/build-your-bundle.mdx
```diff
@@ -187,6 +187,34 @@ variables:
...
+  - name: core
+    overrides:
+      istio-admin-gateway:
+        gateway:
+          values:
+            # Provision the admin gateway's load balancer as an internal NLB
+            # so the admin domain is reachable only from inside the VPC.
+            - path: service.annotations
+              value:
+                service.beta.kubernetes.io/aws-load-balancer-type: "external"
```
chance-coleman: maybe we should add a comment explaining how external means internal lol still confused me sometimes. My understanding is that external means `use the AWS LBC controller` rather than `internet facing`.

mjnagel: I might just delete this line and the one below and replace with "add other annotations as needed"? I think they're largely irrelevant to the purpose of this doc after looking at it a bit more... `service.beta.kubernetes.io/aws-load-balancer-scheme` is the only one that matters for internal vs internet facing.

chance-coleman: i can definitely get on board with that, something claude pointed out was that the nlb target type might also require other setup that we didnt mention. Using a more "...add other annotations as needed" will be a good signal to readers this isn't a verbatim solution.

chance-coleman (dismissed):
LGTM

bburky:
This seems good, "Admin Gateway should be reachable only from trusted networks." But we could write this even stronger as something like "Admin Gateway should be reachable only from trusted networks and only by privileged users" or similar. We could more clearly state that standard end-users should not have any network access to the admin gateway.

Basically, if your end users also use a VPN to access the tenant network, this doesn't count as "trusted network" enough to also allow access to the admin gateway. It's _okay_ to do this, but is less strong security isolation.

Really, this is a spectrum of choices and level of accepted risk. As long as it's not on the internet, it's probably ok, but we should encourage the strongest configuration when possible.

bburky (approved):
LGTM

joelmccoy (approved):

PR #2641, opened 2026-04-30
feat: add TCP and UDP support to remoteProtocol on Allow entries

chance-coleman:
## Description
Extends the remoteProtocol field on the Allow type to accept TCP and UDP
alongside the existing TLS and HTTP values.

- TCP and UDP now set the transport protocol on generated NetworkPolicy
  ports; TLS and HTTP remain Istio ServiceEntry concerns only
- Port 15008 (ztunnel HBONE) is no longer injected into UDP-only policies
- Removes the post-generation protocol mutation in allow-egress-dns.ts
- Adds validator guards for invalid TCP/UDP combinations (e.g. UDP with
  remoteHost, TCP/UDP without a port, UDP with KubeAPI/KubeNodes/CloudMetadata)
- Updates CRD schema, generated types, and reference docs
- Adds unit tests and an e2e UDP NetworkPolicy enforcement test

## Related Issue

Fixes [Core-481](https://linear.app/defense-unicorns/issue/CORE-481/add-protocol-field-to-package-cr-allow-type)

## Type of change

## Steps to Validate
- `npx vitest run`
- `uds run -f tasks/test.yaml e2e-tests`

This deploys the test resources (including the new UDP echo server/client pods), runs the full vitest E2E suite, then removes test resources. The two new tests to watch are:
- UDP NetworkPolicy - DNS allowed, TCP DNS blocked
- UDP NetworkPolicy - custom allow and deny

## Checklist before merging


chance-coleman:
@greptileai

mjnagel:
@codex review

test/vitest/network.spec.ts
```diff
@@ -476,6 +480,99 @@ describe("Network Policy Validation", { retry: 2 }, () => {
...
+        "sh",
+        "-c",
+        "echo ping | nc -u -w 1 udp-echo-server.curl-ns-udp-server.svc.cluster.local 5000 2>/dev/null || true",
+      ]);
+    })(),
+  ]);
+
+  const deniedDebug = `UDP blocked: server stdout="${deniedServer.stdout}"`;
+  expect(deniedServer.stdout.trim(), deniedDebug).toBe("");
+});
```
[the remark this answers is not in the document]
chance-coleman: @greptileai Promise.all + 2500ms delay is intentional. the listener and sender start concurrently so the delay gives nc time to bind before the first packet lands. polling until the port is bound would need a shell loop inside execInPod which feels like overkill here. I think it makes sense to revisit if it actually flakes in CI.

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -97,19 +97,33 @@ const allow = {
         description: "Remote host to allow traffic out to",
         type: "string",
       },
+      // TODO: remoteProtocol mixes L7 application protocols (TLS/HTTP, used to generate
+      // Istio ServiceEntry) with L4 transport protocols (TCP/UDP, used to set NetworkPolicy port
+      // protocol). Before v1 this should be split into separate fields (e.g. remoteProtocol for
```
joelmccoy: I'm confused by this TODO comment, especially "Before v1".  Are you suggesting that when we bump our CRD version we make this split?

chance-coleman: yeah good catch, we decided to not adopt the multiple fields in the TDD. I need to remove this. I'll remove this

docs/how-to-guides/networking/define-network-access.mdx
```diff
@@ -130,7 +130,13 @@ Every `allow` entry must also specify at least one remote field: `remoteGenerate
...
 
    > [!NOTE]
-   > The egress protocol defaults to TLS if not specified. Only HTTP and TLS protocols are currently supported.
+   > When `remoteHost` is specified, `remoteProtocol` defaults to `TLS` if omitted. Use `TLS` or `HTTP` for external HTTPS/HTTP egress via an Istio ServiceEntry. Use `TCP` for raw TCP egress to an external host.
+
+   > [!CAUTION]
+   > Using `remoteProtocol: TCP` with `remoteHost` generates a TCP-mode Istio `ServiceEntry`. This disables SNI-based routing and Istio L7 `AuthorizationPolicy` enforcement for that host. Istio can only apply L4 controls (IP/port) in this mode. Prefer `TLS` unless the upstream service does not support TLS.
+
+   > [!NOTE]
+   > Use `remoteProtocol: UDP` (without `remoteHost`) for in-cluster UDP traffic such as DNS. `UDP` cannot be combined with `remoteHost`. To allow the same port on both TCP and UDP, create two separate Allow entries.
```
mjnagel: nit: "such as DNS" might be a misleading example since we allow DNS by default? UDP usage could also be broader than just in-cluster.

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -98,18 +98,28 @@ const allow = {
...
-        description: "Protocol used for external connection",
+        description:
+          "The protocol for this Allow entry. " +
+          "`TLS` and `HTTP` control Istio ServiceEntry generation for egress entries with `remoteHost`; " +
+          "`TCP` also generates a TCP Istio ServiceEntry when combined with `remoteHost`. " +
+          "`TCP` and `UDP` explicitly set the transport protocol on the generated NetworkPolicy port. " +
+          "When omitted with `remoteHost`, defaults to TLS for ServiceEntry generation; " +
+          "otherwise NetworkPolicy ports carry no explicit protocol (Kubernetes defaults to TCP). " +
+          "`UDP` cannot be combined with `remoteHost`. " +
+          "To allow the same port on both TCP and UDP, use two separate Allow entries.",
```
mjnagel: Maybe just my opinion but I think this is leaking too many "internals" into our API descriptions/schema. The description should be a simple description of how/when to use this field - an end user shouldn't need to know/care about what resulting resources are generated as a result.

I'd suggest keeping this simpler with a few key details:
- the protocol to use for this allow entry
- prefer TLS or HTTP when used with `remoteHost`, default of TLS if using `remoteHost`
- use TCP or UDP based on the connection type for all other remotes, default of TCP if using other remotes

I think that would largely cover it without getting into the resulting technicals of netpols, service entries, etc? It might end up being a similar length, but just trimming out some of the "Istio ServiceEntry", etc details.

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -98,18 +98,28 @@ const allow = {
...
         type: "string",
-        enum: ["TLS", "HTTP"],
+        enum: ["TLS", "HTTP", "TCP", "UDP"],
       },
       port: {
-        description: "The port to allow (protocol is always TCP)",
+        description:
+          "The port to allow (transport protocol defaults to TCP unless `remoteProtocol` is set to `TCP` or `UDP`)",
```
mjnagel: nit: this reads funny "defaults to TCP" + "unless set to TCP"

```suggestion
          "The port to allow (transport protocol defaults to TCP unless `remoteProtocol` is set)",
```

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -98,18 +98,28 @@ const allow = {
...
         minimum: 1,
         maximum: 65535,
         type: "number",
       },
       ports: {
-        description: "A list of ports to allow (protocol is always TCP)",
+        description:
+          "A list of ports to allow (transport protocol defaults to TCP unless `remoteProtocol` is set to `TCP` or `UDP`)",
```
mjnagel: nit: this reads funny "defaults to TCP" + "unless set to TCP"

```suggestion
          "A list of ports to allow (transport protocol defaults to TCP unless `remoteProtocol` is set)",
```

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -137,74 +141,82 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
-      (policy.remoteGenerated ||
-        policy.remoteNamespace !== undefined ||
-        policy.remoteSelector ||
-        policy.remoteCidr ||
-        !policy.remoteHost)
-    ) {
+    // Istio ServiceEntry does not support UDP; use remoteNamespace/remoteSelector/remoteCidr instead.
+    if (policy.remoteProtocol === RemoteProtocol.UDP && policy.remoteHost) {
+      return req.Deny(
+        "UDP remoteProtocol cannot be combined with remoteHost (Istio ServiceEntry does not support UDP)",
```
mjnagel: Similar to other comment - I'd keep this more generic/try to not include the internal implementation details.

```suggestion
        "UDP remoteProtocol cannot be combined with remoteHost",
```

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -137,74 +141,82 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
-    if ((policy.remoteHost || policy.remoteProtocol) && policy.direction === "Ingress") {
-      return req.Deny("remoteHost and/or remoteProtocol cannot be used with Ingress");
+    // Without ports, TCP/UDP remoteProtocol silently broadens the policy (all ports allowed, no protocol filter applied).
+    if (
+      (policy.remoteProtocol === RemoteProtocol.TCP ||
+        policy.remoteProtocol === RemoteProtocol.UDP) &&
+      policy.port === undefined &&
+      !policy.ports?.length
+    ) {
+      return req.Deny("TCP/UDP remoteProtocol requires at least one port or ports entry");
```
mjnagel: This validation is bordering on a breaking change. It's _technically_ not one because it explicitly checks for TCP `remoteProtocol`, but it feels strange to validate this in that case and not when `remoteProtocol` is empty (which ends up having the same behavior).

I'd probably suggest removing this. It's something we could evaluate adding as a security improvement down the road, but it would have wide effects downstream.

test/vitest/network.spec.ts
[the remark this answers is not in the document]
mjnagel: This does feel like a valid comment imo - given some of the IaC failures we've seen with differing levels of delays/timeouts I'd be hesitant to ship a hardcoded "sleep" of 2.5s. The overall retries may cover this, just wonder if there's a cleaner approach here.

src/test/app-curl-packages.yaml
```diff
@@ -149,3 +149,50 @@ spec:
...
+  network:
+    allow:
+      - direction: Ingress
+        remoteNamespace: curl-ns-udp-allow
+        remoteSelector:
+          app: udp-echo-client
+        selector:
+          app: udp-echo-server
+        port: 5000
+        remoteProtocol: UDP
```
mjnagel, line 175: Might recommend we also add explicit TCP to a test package (could do it on an existing one). That would help us catch any deltas between the explicit and implicit TCP paths.

docs/how-to-guides/networking/define-network-access.mdx
```diff
@@ -130,7 +130,13 @@ Every `allow` entry must also specify at least one remote field: `remoteGenerate
    By default, workloads in the mesh cannot reach the internet. Choose the approach that fits your use case:
 
    > [!NOTE]
-   > The egress protocol defaults to TLS if not specified. Only HTTP and TLS protocols are currently supported.
+   > When `remoteHost` is specified, `remoteProtocol` defaults to `TLS` if omitted. Use `TLS` or `HTTP` for external HTTPS/HTTP egress via an Istio ServiceEntry. Use `TCP` for raw TCP egress to an external host.
+
+   > [!CAUTION]
+   > Using `remoteProtocol: TCP` with `remoteHost` generates a TCP-mode Istio `ServiceEntry`. This disables SNI-based routing and Istio L7 `AuthorizationPolicy` enforcement for that host. Istio can only apply L4 controls (IP/port) in this mode. Prefer `TLS` unless the upstream service does not support TLS.
```
mjnagel: I think we should probably deny TCP + remoteHost for now (and maybe scope this as future work). We currently don't have any testing of this behavior and from a quick review I think this would not currently work in sidecar mode.

I think there's some real utility here for things like a remote redis instance, but probably better to separately build and test that support intentionally?

docs/how-to-guides/networking/define-network-access.mdx
mjnagel: I think we're missing a clear warning note _somewhere_ about UDP usage. The main thing to call out is no encryption through the mesh - so I think we'd probably want some guidance around ensuring the app is handling encryption?

This might also be a good thing to callout in the [audit doc](https://github.com/defenseunicorns/uds-core/blob/8f3ca192baf83ae4d3bc0d7db4b8e1097ec10b2e/docs/how-to-guides/policy-and-compliance/audit-security-posture.mdx#L2).

chance-coleman: done. DNS is a bad example since we allow it by default. removed it.

src/pepr/operator/crd/sources/package/v1alpha1.ts
chance-coleman: done. rewrote it using your suggested outline. no ServiceEntry or NetworkPolicy references.

src/pepr/operator/crd/validators/package-validator.ts
chance-coleman: done. simplified to match your suggestion.

src/pepr/operator/crd/validators/package-validator.ts
chance-coleman: agreed. removed it. we can revisit as a deliberate security improvement with proper downstream testing when the time comes.

src/test/app-curl-packages.yaml
chance-coleman: done. added remoteProtocol: TCP to the existing ingress entry on curl-pkg-allow-all.

docs/how-to-guides/networking/define-network-access.mdx
chance-coleman: agreed. denied TCP+remoteHost in the validator and removed the docs for it. can follow up with proper sidecar testing as a separate feature when we're ready to support it intentionally.

docs/how-to-guides/networking/define-network-access.mdx
chance-coleman: added a CAUTION in the networking doc and a row in the audit table.

test/vitest/network.spec.ts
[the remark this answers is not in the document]
chance-coleman: replaced both sleeps with a 5-iteration retry loop at 600ms each. UDP gives no signal if the server isn't ready yet, so retrying is more reliable than one delayed send.

src/pepr/operator/controllers/istio/egress-ports.ts
```diff
@@ -24,5 +24,12 @@ export function getPortsForHostAllow(allow: {
   if (ports) {
     return ports;
   }
+  // TCP and UDP without ports are rejected by the validator; this branch should not be reached.
```
joelmccoy: I don't think this statement is correct?  And maybe the validator should block this?  I deployed this package and it applied (validator didn't block), but it did fail reconciliation because of the logic here: 
```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: curl-pkg-udp-allow
  namespace: test-package
spec:
  network:
    allow:
      - direction: Egress
        remoteNamespace: curl-ns-udp-server
        remoteProtocol: UDP
```

I'm getting a little confused here...
1. Do we support UDP/TCP without ports in core? (i think these technically can be valid network policies)
2. If we don't support UDP/TCP without ports then i think we should block this at the validator and add some robust tests around this?

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
```diff
@@ -1058,3 +1058,25 @@ describe("test purgeAmbientEgressResources", () => {
...
+  });
+
+  it("should default to 443 for TLS without explicit ports", () => {
+    expect(getPortsForHostAllow({ remoteProtocol: RemoteProtocol.TLS })).toEqual([443]);
+  });
+
+  it("should default to 80 for HTTP without explicit ports", () => {
+    expect(getPortsForHostAllow({ remoteProtocol: RemoteProtocol.HTTP })).toEqual([80]);
+  });
+});
```
joelmccoy: nit: should this be in a new `egress-ports.spec.ts` file for consistency with what file we are testing? or is there a technical reason why these are included here?

src/pepr/operator/controllers/istio/egress-ports.ts
joelmccoy: Also this should never reach the egress code because there is no remoteHost right? we should probably have a guard for that up the stack in the egress code so that we don't process egress code if `remoteHost` is not defined.

src/pepr/operator/crd/sources/istio/virtualservice-v1.ts
```diff
@@ -4,7 +4,7 @@
  */
 
 /**
- * Auto-generated from Istio 1.29.1 CRD (virtualservices.networking.istio.io v1).
+ * Auto-generated from Istio 1.29.2 CRD (virtualservices.networking.istio.io v1).
```
joelmccoy, line 7: hmmm. was this not caught in the istio 1.29.2 bump? this is auto generated right?

src/pepr/operator/controllers/istio/egress-ports.ts
joelmccoy: ahh i see this comment says not to deny this: https://github.com/defenseunicorns/uds-core/pull/2641#discussion_r3189892478

If that is the case... we shouldn't fail it here. The validator doesn't block it but it reaches this code and fails reconcile.

src/pepr/operator/crd/validators/package-validator.spec.ts
```diff
@@ -456,13 +690,34 @@ describe("Test validation of Package CRs", () => {
...
+          remoteProtocol: RemoteProtocol.TCP,
+          port: 8080,
+        },
+      ],
+      [],
+      [],
+    );
+    await validator(mockReq);
+    expect(mockReq.Approve).toHaveBeenCalledTimes(1);
+  });
```
joelmccoy, line 746: can we add a test(s) for this comment: https://github.com/defenseunicorns/uds-core/pull/2641#discussion_r3189892478

Since we are allowing tcp/udp without a port, can we add explicit allow tests?

src/pepr/operator/crd/generated/package-v1alpha1.ts
```diff
@@ -225,7 +226,10 @@ export interface Allow {
    */
   remotePodLabels?: { [key: string]: string };
   /**
-   * Protocol used for external connection
+   * The protocol for this Allow entry. When using `remoteHost`, prefer `TLS` or `HTTP`; `TLS`
+   * is the default if omitted. For all other remotes, use `TCP` or `UDP` based on the
+   * connection type; `TCP` is the default if omitted. `UDP` cannot be combined with
+   * `remoteHost`. To allow the same port on both TCP and UDP, use two separate Allow entries.
```
joelmccoy:
```suggestion
   * `remoteHost`. To allow the same port on both TCP and UDP, use two separate Allow entries with different descriptions
```

Interesting edge case:
```yaml
apiVersion: uds.dev/v1alpha1
kind: Package
metadata:
  name: curl-pkg-udp-allow
  namespace: test-package
spec:
  network:
    allow:
    - direction: Egress
      remoteNamespace: kube-system
      port: 53
      remoteProtocol: UDP
    - direction: Egress
      remoteNamespace: kube-system
      port: 53
      remoteProtocol: TCP
```
gets: 

```
Error from server: error when creating "package-test.yaml": admission webhook "pepr-uds-core.pepr.dev" denied the request: The combination of characteristics of this network allow rule would create a duplicate NetworkPolicy. Verify you do not have duplicate allow rules, or add a unique "description" field for this rule. The duplicate rule would be named "allow-curl-pkg-udp-allow-egress-all-pods-kube-system-all-pods".
```

I think it is fine to require having an description here instead of redoing our naming for UDP? or do we want to put `udp` in our netpol name generation for those netpols??

joelmccoy (commented):
Mostly looking good! I found possibly two bugs we want to address? And a couple other small comments for consideration.

src/pepr/operator/crd/generated/package-v1alpha1.ts
mjnagel: I kind of like adding the protocol to the name (at least in UDP case)?

src/pepr/operator/controllers/istio/egress-ports.ts
mjnagel: > Also this should never reach the egress code because there is no remoteHost right?

I'm also trying to trace this down - I believe we do have to traverse some of the egress code for the purposes of `Anywhere` entries and "cleanup" (current Package might not have `remoteHost` but we need to process to remove previous hosts). I'm not sure this specific function needs to be hit in either of those cases though? Might be a bit of a reworking of what level we have conditionals at...

Regardless agreed on not failing here.

src/pepr/operator/crd/sources/istio/virtualservice-v1.ts
mjnagel: I think [this issue](https://github.com/defenseunicorns/uds-core/pull/2605) was fixed in parallel to the [1.29.2 update](https://github.com/defenseunicorns/uds-core/pull/2587). The Renovate branch was likely just behind `main` so it didn't run the diff check in CI.

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
```diff
@@ -1058,3 +1058,25 @@ describe("test purgeAmbientEgressResources", () => {
     );
   });
 });
```
mjnagel, line 1060: Can't comment on the specific file/line since it's not in this diff but I think we need to filter out UDP allow rules in our egress anywhere handling ([here](https://github.com/defenseunicorns/uds-core/blob/chance/core-481/src/pepr/operator/controllers/istio/egress.ts#L139-L147)). TCP should still be allowed there (since TCP would support HTTP/TLS), but UDP anywhere rules should not be added to the TLS/HTTP authpols for fine-grained egress.

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -262,15 +262,20 @@ export async function networkPolicies(pkg: UDSPackage, namespace: string, istioM
       policy.metadata.name = `allow-${pkgName}-${policy.metadata.name}`;
     }
```
mjnagel, line 264: Similar to [this comment](https://github.com/defenseunicorns/uds-core/pull/2641/changes#r3202498797), can't comment on the diff but we need to filter UDP allow out of our [authpol handling](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/network/authorizationPolicies.ts#L267-L269) - should be a simple conditional like how we handle Egress rules.

mjnagel (commented):
Two edge cases I identified while testing a variety of packages - still iterating a bit locally but in both of these cases we are provisioning resources we don't need to for UDP (since authpols are TCP handling only).

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
chance-coleman: Added check in createAmbientPackageEntry() to skip UDP Anywhere rules since UDP doesn't work with TLS/HTTP auth policies for fine-grained egress. Exported the function and added test.

src/pepr/operator/controllers/network/policies.ts
chance-coleman: Added check in generateAuthorizationPolicies() to skip UDP rules since UDP doesn't work with L7 authorization policies. Added missing RemoteProtocol import and test.

src/pepr/operator/controllers/istio/egress-ports.ts
chance-coleman: Added early return in getHostPortsProtocol() so it returns undefined immediately when no remoteHost is set (before calling getPortsForHostAllow()). Removed the unreachable throw in egress-ports.ts since TCP/UDP without ports is allowed by the validator.

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
chance-coleman: Moved getPortsForHostAllow and getAllowedPorts tests into new egress-ports.spec.ts.

src/pepr/operator/crd/validators/package-validator.spec.ts
chance-coleman: Added tests in package-validator.spec.ts confirming TCP/UDP without ports are allowed (matching the validator behavior from mjnagel's suggestion).

src/pepr/operator/crd/generated/package-v1alpha1.ts
chance-coleman: Updated the docstring to include "with different descriptions". Also updated generateName() to include the protocol in the NetworkPolicy name when remoteProtocol is TCP or UDP, preventing duplicate names when the same port is used with different protocols.

src/pepr/operator/controllers/istio/egress-ports.ts
```diff
@@ -24,12 +24,11 @@ export function getPortsForHostAllow(allow: {
...
-  // TCP and UDP without ports are rejected by the validator; this branch should not be reached.
   // TLS/HTTP (or omitted) default to 443/80 respectively for ServiceEntry generation.
+  // TCP/UDP without ports return empty array (allowed by validator since NetworkPolicy
+  // can be port-less, and this function is only called when remoteHost is set).
   if (allow.remoteProtocol === RemoteProtocol.TCP || allow.remoteProtocol === RemoteProtocol.UDP) {
-    throw new Error(
-      `remoteProtocol ${allow.remoteProtocol} requires at least one port or ports entry`,
-    );
+    return [];
   }
```
joelmccoy: But should we not even reach this code if TCP/UDP is set? This is specific to external egress right with remoteHost? Or do we need this here to support `remoteGenerated: AnyWhere`?

src/pepr/operator/crd/generated/package-v1alpha1.ts
```diff
@@ -229,7 +229,7 @@ export interface Allow {
    * The protocol for this Allow entry. When using `remoteHost`, prefer `TLS` or `HTTP`; `TLS`
    * is the default if omitted. For all other remotes, use `TCP` or `UDP` based on the
    * connection type; `TCP` is the default if omitted. `UDP` cannot be combined with
-   * `remoteHost`. To allow the same port on both TCP and UDP, use two separate Allow entries.
+   * `remoteHost`. To allow the same port on both TCP and UDP, use two separate Allow entries with different descriptions.
```
joelmccoy: I don't think you need `with a different description` here now that you are naming netpols based on remote protocol?

src/pepr/operator/crd/generated/package-v1alpha1.ts
joelmccoy: I think you can leave out the `with different descriptions` if you are naming netpols based on tcp/udp?  I think @mjnagel maybe suggested just adding the naming to the UDP? But I think it is safe/fine to include TCP on the TCP policies too?

src/pepr/operator/crd/generated/package-v1alpha1.ts
chance-coleman: Agreed. I removed the "with different descriptions" text and regenerated CRDs/docs. We are keeping protocol-based naming as implemented, including TCP and UDP, so collision guidance is no longer needed.

src/pepr/operator/controllers/istio/egress-ports.ts
chance-coleman: Good call. This was dead code. getPortsForHostAllow is only used on remoteHost paths, and validator blocks TCP/UDP with remoteHost, so that branch is unreachable. I removed it.

src/pepr/operator/crd/generated/package-v1alpha1.ts
mjnagel: Either is probably okay - I don't think names of netpols are part of our "API" so changing them is okay. Mostly was thinking that UDP is a bit more of the "unique" so including in the name there is where it's most important imo.

src/pepr/operator/controllers/istio/egress-ambient.spec.ts
mjnagel: LGTM now from code inspection.

src/pepr/operator/controllers/network/policies.ts
mjnagel: LGTM now from code inspection.

src/pepr/operator/controllers/istio/egress-ports.ts
mjnagel: Did something else change here? I probably need to re-deploy and test but this code was reachable before. I think the call stack that leads to this is reconciler -> `istioEgressResources` -> `createHostResourceMap` -> `getHostPortsProtocol` -> `getPortsForHostAllow`. If I followed correctly there's no checks for remoteHost through this codepath so it runs on _all_ `allow` rules.

chance-coleman: I changed `getHostPortsProtocol` to return early when remoteHost is not set, before calling `getPortsForHostAllow`. So now `createHostResourceMap` only evaluates host-scoped allows. Combined with validator blocking TCP/UDP + remoteHost, that TCP/UDP branch in `getPortsForHostAllow` became unreachable and was removed

mjnagel: ~I'll probably just have to run some scenarios locally - tracing the codepath described above there's no remoteHost check _before_ you get to this part in core ([`getHostPortsProtocol` for reference](https://github.com/defenseunicorns/uds-core/blob/main/src/pepr/operator/controllers/istio/egress.ts#L400-L410) - no conditional on remoteHost so `getPortsForHostAllow` always gets called from here).~

Edit: Bleh - my diff was outdated or github had me on the wrong branch - early return looks good now [here](https://github.com/defenseunicorns/uds-core/blob/chance/core-481/src/pepr/operator/controllers/istio/egress.ts#L404-L407).

src/pepr/operator/controllers/network/authorizationPolicies.ts
```diff
@@ -267,6 +268,8 @@ export async function generateAuthorizationPolicies(
   if (pkg.spec?.network?.allow) {
     for (const rule of pkg.spec.network.allow) {
       if (rule.direction === "Egress") continue;
+      // Skip UDP - UDP doesn't work with L7 authorization policies
```
mjnagel: super nit: these are L4 authpols - real problem is that Istio doesn't manage UDP so authpols in general don't work 😄 

```suggestion
      // Skip UDP - authorization policies do not handle UDP traffic
```

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -114,23 +114,28 @@ export function generate(namespace: string, policy: Allow, istioMode?: Mode): ki
...
-  // Define the ports to allow from the ports property
-  const ports: V1NetworkPolicyPort[] = (policy.ports ?? []).map(port => ({ port }));
+  // Only TCP/UDP map to NetworkPolicyPort.protocol; TLS/HTTP are Istio ServiceEntry concerns.
+  // policies.ts ztunnel injection uses port.protocol to detect UDP-only policies; it skips
+  // port 15008 when all ports carry protocol:"UDP". Do not remove this stamping without
+  // updating that condition.
+  const protocol =
+    policy.remoteProtocol === RemoteProtocol.UDP || policy.remoteProtocol === RemoteProtocol.TCP
+      ? policy.remoteProtocol
+      : undefined;
```
mjnagel: This is 100% technically equivalent but now that we're supporting UDP in netpols it might be good to explicitly include the protocol in all of our netpols. This would just mean changing this ternary to:

```suggestion
  const protocol =
    policy.remoteProtocol === RemoteProtocol.UDP || policy.remoteProtocol === RemoteProtocol.TCP
      ? policy.remoteProtocol
      : RemoteProtocol.TCP;
```

rather than the undefined fallback/false path. Okay to not do this if we want to skip. This would also simplify the ports/toPort handling below if you know protocol is always defined.

src/pepr/operator/controllers/network/generate.ts
```diff
@@ -156,9 +161,14 @@ export function generateName(policy: Allow) {
         policy.remoteNamespace,
         Object.values(policy.remoteSelector || ["all pods"]),
       ],
+      // Include protocol in name for TCP/UDP to avoid duplicate NetworkPolicies
+      policy.remoteProtocol === RemoteProtocol.TCP || policy.remoteProtocol === RemoteProtocol.UDP
+        ? policy.remoteProtocol
+        : undefined,
```
mjnagel: Related to [comment above](https://github.com/defenseunicorns/uds-core/pull/2641/changes#r3205447029), I think we might want to explicitly put TCP as the fallback/false value here. This one may be almost more important here since this is "user-facing" for name consistency between UDP/TCP/"implicit TCP" policies.

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -262,15 +262,20 @@ export async function networkPolicies(pkg: UDSPackage, namespace: string, istioM
...
 
-    // Loop through all ports in ingress/egress policies and add port 15008 for ztunnel
+    // Add port 15008 (ztunnel HBONE) to all ingress/egress port lists.
+    // Skipped for UDP-only port lists; ztunnel uses TCP HBONE and cannot carry UDP.
+    // generate.ts stamps protocol:"UDP" on ports when remoteProtocol is UDP; that stamping
+    // is what makes the some() predicate below return false for UDP-only policies.
+    // Both ingress and egress are evaluated independently (not else-if) so a policy that
+    // has both spec.ingress and spec.egress (rare but valid) gets 15008 injected into both.
```
mjnagel: This comment and change caught me off guard - I don't think there's any code path that can actually produce ingress and egress in a combined policy? This has historically been intentional so we clearly split policies by direction. It's not super important because it won't get hit or change how this evaluates, but might be worth reverting this comment change and/or logic.

mjnagel (dismissed):
LGTM overall - I had a few final nits reviewing the diffs but everything seemed to function as expected in testing.

src/pepr/operator/controllers/network/policies.ts
chance-coleman: Agreed. I reverted the wording to avoid implying combined ingress+egress is an expected generated path.

src/pepr/operator/controllers/network/generate.ts
chance-coleman: Implemented. Name generation now uses the same protocol normalization as policy generation, so implicit TCP and explicit TCP are consistent, and UDP remains explicit.

src/pepr/operator/controllers/network/generate.ts
chance-coleman: Implemented. NetworkPolicy ports now always carry explicit L4 protocol. UDP/TCP stay explicit, and TLS/HTTP/omitted now resolve to TCP in generated netpol ports.

src/pepr/operator/controllers/network/authorizationPolicies.ts
chance-coleman: Good catch. Updated comment

mjnagel (approved):

src/pepr/operator/controllers/istio/egress-ports.ts
joelmccoy: looks resolved on my end too

joelmccoy (approved):
LGTM - validated the edge cases we identified locally and things seem to working great

PR #2650, opened 2026-05-05
chore(deps): migrate unicorn flavor images from RapidFort to Chainguard

chance-coleman:
## Description
Migrates the unicorn flavor from RapidFort to Chainguard FIPS images across all components.

### Keycloak FIPS JCE configuration

Chainguard's Keycloak image requires an explicit JCE security provider override to enable FIPS mode, unlike the RapidFort image which handled this internally. A new ConfigMap is injected as a volume mount and `JAVA_OPTS_APPEND` is set to load the override file at startup. The old `JDK_JAVA_OPTIONS` injection for rfcurated images is removed.

### CI authentication

Replaced RapidFort username/password credentials with a Chainguard federated identity (`chainguardIdentity`) using the `chainguard-dev/setup-chainctl` action across all workflows.

### fapolicyd on RHEL 9

Chainguard's CNI image writes `libcrypto.so.3` into `/opt/cni/bin/.cgr/` at runtime. On RHEL 9 with fapolicyd in enforcing mode this blocks CNI startup. Added a prerequisite note with the required fapolicyd rule and the RKE2 user data script now configures this automatically.

### Pepr image policy

Updated the `validateIstioImage` policy to recognize `cgr.dev/defenseunicorns.com/istio-proxy-fips` as the valid unicorn proxy registry/repository.

### Renovate and label cleanup

Updated the renovate-readiness action and scripts to use `waiting on unicorn` instead of `waiting on rapidfort`. Updated the multi-arch check script to scan `cgr.dev/defenseunicorns.com` images and fixed comment stripping to prevent inline YAML comments from being passed as image names to `crane`.

## Related Issue

Fixes [Core-522](https://linear.app/defense-unicorns/issue/CORE-522/switch-core-unicorn-flavor-to-chainguard-images)

## Type of change

## Checklist before merging


.github/actions/renovate-readiness/action.yaml
```diff
@@ -120,7 +120,7 @@ runs:
             echo "Upstream package.json is behind image versions - possible Pepr release issue"
             gh pr edit ${{ github.event.pull_request.number }} --add-label "waiting on upstream"
             gh pr edit ${{ github.event.pull_request.number }} --remove-label "waiting on ironbank" || true
-            gh pr edit ${{ github.event.pull_request.number }} --remove-label "waiting on unicorn" || true
+            gh pr edit ${{ github.event.pull_request.number }} --remove-label "waiting on chainguard" || true
```
mjnagel: I am thinking we should maybe just set this label as `unicorn` rather than being rapidfort/chainguard specific? That way we keep it generic and don't have to change labels if we use different providers. This would require changes across a number of files which I won't comment on.

docs/dev/unicorn-registry-auth.md
```diff
@@ -0,0 +1,39 @@
+# Unicorn registry authentication
+
+The `unicorn` flavor images are hosted on `cgr.dev/defenseunicorns.com`, a private Chainguard organization. Working with the unicorn flavor locally requires authentication to pull images from this registry.
```
mjnagel: Don't have a strong opinion here, but in the past we had [this in Notion](https://www.notion.so/defense-unicorns/Unicorn-Flavor-Development-197e512f24fc80f488dfe5d2baf71e34?source=copy_link) which might be a bit more accessible for others and ultimately this is only for internal devs (opensource contributors wouldn't have this access).

docs/dev/unicorn-registry-auth.md
```diff
@@ -0,0 +1,39 @@
...
+chainctl auth login
+```
+
+## Configure the credential helper
+
+Register `chainctl` as the credential helper for `cgr.dev`:
+
+```bash
+chainctl auth configure-docker
+```
```
mjnagel: Do these commands need the org based on some of our newer findings/SSO changes?

src/grafana/zarf.yaml
```diff
@@ -62,7 +62,7 @@ components:
...
           - values/unicorn-values.yaml
     images:
-      - quay.io/rfcurated/grafana:12.4.2-jammy-scratch-fips-rfcurated
-      - quay.io/rfcurated/busybox:1.37.0-musl-rf.1-fips-rfcurated
-      - quay.io/rfcurated/curl:8.19.0-jammy-scratch-fips-rfcurated
-      - quay.io/rfcurated/kiwigrid/k8s-sidecar:2.5.0-jammy-scratch-fips-rfcurated-rfhardened
+      - cgr.dev/defenseunicorns.com/grafana-fips:12.4.2
+      - cgr.dev/defenseunicorns.com/busybox-fips:1.37.0
+      - cgr.dev/defenseunicorns.com/curl-fips:8.19.0
+      - cgr.dev/defenseunicorns.com/k8s-sidecar-fips:2.5.1
```
mjnagel: Might be okay but noticed this version diff and started looking more...registry1 has a `2.5.4` tag it looks like and upstream doesn't even have a `2.5.1` tag. Definitely something to track here, not sure if we need to resolve right now or not but it is a bit strange that the cgr image is tagged for something that isn't upstream (maybe got yanked?).

src/istio/zarf.yaml
```diff
@@ -82,10 +82,10 @@ components:
...
     # @lulaStart b4367e52-bef0-4463-a906-e5af6b4aa015
     images:
-      - quay.io/rfcurated/istio/pilot:1.29.2-jammy-fips-rfcurated-rfhardened
-      - quay.io/rfcurated/istio/proxyv2:1.29.2-jammy-fips-rfcurated-rfhardened
-      - quay.io/rfcurated/istio/install-cni:1.29.2-jammy-fips-rfcurated-rfhardened
-      - quay.io/rfcurated/istio/ztunnel:1.29.2-jammy-scratch-fips-rfcurated
+      - cgr.dev/defenseunicorns.com/istio-pilot-fips:1.29.1
+      - cgr.dev/defenseunicorns.com/istio-proxy-fips:1.29.1
+      - cgr.dev/defenseunicorns.com/istio-install-cni-fips:1.29.1
+      - cgr.dev/defenseunicorns.com/ztunnel-fips:1.29.1
```
mjnagel: To track: ideally this should be 1.29.2 before merge so we don't downgrade istio.

src/keycloak/chart/values.yaml
```diff
@@ -415,6 +415,13 @@ env: []
 migrations:
   # Deletes the generated truststore to enable recreating it by Keycloak
   deleteGeneratedTrustStore: false
+  # Mounts a JVM security properties override that adds SunJCE as provider.99.
+  # Required for cgr.dev keycloak-fips images, which omit SunJCE from java.security,
+  # causing Keycloak's hardcoded PKCS12 truststore creation to fail with
+  # "Algorithm HmacPBESHA256 not available". BC FIPS remains the primary provider.
+  # Remove this flag and configmap-fips-jce-security.yaml once Chainguard ships
+  # keycloak-fips with SunJCE included in java.security.
+  writeFipsJceSecurity: false
```
mjnagel: Maybe a bit nit but putting this as a "migration" feels a bit weird for something that needs to stay enabled forever. I wonder if we should just manage the conditional similar to what we did on the old one (`contains "rfcurated" .Values.image.repository`)?

src/loki/zarf.yaml
```diff
@@ -59,7 +59,7 @@ components:
         valuesFiles:
           - ./values/unicorn-values.yaml
     images:
-      - quay.io/rfcurated/grafana/loki:3.7.1-jammy-fips-rfcurated-rfhardened
-      - quay.io/rfcurated/nginx:1.29.7-slim-jammy-fips-rfcurated-rfhardened
-      - quay.io/rfcurated/memcached:1.6.41-jammy-fips-rfcurated-rfhardened
-      - quay.io/rfcurated/kiwigrid/k8s-sidecar:2.5.0-jammy-scratch-fips-rfcurated-rfhardened
+      - cgr.dev/defenseunicorns.com/loki-fips:3.6.10 # NOTE: exact version 3.7.1 not available at cgr.dev/defenseunicorns.com; using 3.6.10
```
mjnagel: I see 3.7.1 available now 🚀

src/prometheus-stack/zarf.yaml
```diff
@@ -72,13 +72,13 @@ components:
...
-      - "quay.io/rfcurated/prometheus-operator:0.90.1-jammy-scratch-fips-rfcurated"
-      - "quay.io/rfcurated/kube-state-metrics:2.18.0-jammy-scratch-fips-rfcurated-rfhardened"
-      - "quay.io/rfcurated/prometheus/alertmanager:0.32.1-jammy-fips-rfcurated-rfhardened"
-      - "quay.io/rfcurated/prometheus-operator/prometheus-config-reloader:0.90.1-jammy-scratch-fips-rfcurated"
-      - "quay.io/rfcurated/prometheus:3.11.3-jammy-fips-rfcurated-rfhardened"
-      - "quay.io/rfcurated/ingress-nginx/kube-webhook-certgen:1.6.9-jammy-fips-rfcurated-rfhardened"
+      - "cgr.dev/defenseunicorns.com/prometheus-node-exporter-fips:1.11.1"
+      - "cgr.dev/defenseunicorns.com/prometheus-operator-fips:0.90.1"
+      - "cgr.dev/defenseunicorns.com/kube-state-metrics-fips:2.18.0"
+      - "cgr.dev/defenseunicorns.com/prometheus-alertmanager-fips:0.31.1" # NOTE: exact version 0.32.1 not available at cgr.dev/defenseunicorns.com; using 0.31.1
```
mjnagel: This should also be something to bump pre-merge if we can to not cause a downgrade.

src/vector/zarf.yaml
```diff
@@ -52,4 +52,4 @@ components:
         valuesFiles:
           - values/unicorn-values.yaml
     images:
-      - quay.io/rfcurated/vector:0.55.0-jammy-fips-rfcurated-rfhardened
+      - cgr.dev/defenseunicorns.com/vector-fips:0.54.0
```
mjnagel: 0.55.0 seems to be available 🚀

mjnagel (commented):
Few initial comments, looking good overall.

tasks/deploy.yaml
```diff
@@ -126,11 +126,16 @@ tasks:
...
+          UDS_K3D_REF=$(./uds zarf tools yq '.packages[] | select(.name == "uds-k3d-dev") | .ref' bundles/k3d-standard/uds-bundle.yaml)
+
+          # Update the bundle: pin core to the latest release OCI ref, and k3d to the current branch's version
           ./uds zarf tools yq e -i "
             del(.packages[] | select(.name == \"core\").path) |
             .packages[] |= (select(.name == \"core\") |
-            .repository = \"${TARGET_REPO}/core\" | .ref = \"${LATEST_VERSION}\")" \
+            .repository = \"${TARGET_REPO}/core\" | .ref = \"${LATEST_VERSION}\") |
+            .packages[] |= (select(.name == \"uds-k3d-dev\") |
+            .ref = \"${UDS_K3D_REF}\")" \
```
chance-coleman: this will need to be reverted before merge, here for testing 1.35 k3d version in upgrade tests

src/istio/zarf.yaml
chance-coleman: yeah definitely agree. ztunnel is technically the only istio image that is on 1.29.1, 1.29.2 is available for everything else.

.github/actions/renovate-readiness/action.yaml
chance-coleman: i originally had `unicorn` and then saw that rapidfort was using its actual name and switch 🙃 agreed. will switch this back.

src/grafana/zarf.yaml
chance-coleman: if you're referring to `k8s-sidecar` it does seem like everything is going to point at 2.7.1 soon ([link](https://github.com/defenseunicorns/uds-core/pull/2584/changes#diff-654e0a21d8866be9c2da53a65e116cb22b23cdbc7fdcf7fc5c8e0d8e756fff3d))

docs/dev/unicorn-registry-auth.md
chance-coleman: I have updated this notion with chainguard specific information. Moved the RapidFort information into an archived toggle for the time being, we can come back and clean this up later, once the switch has officially happened.

src/keycloak/chart/values.yaml
chance-coleman: Removed the migrations.writeFipsJceSecurity boolean toggle. Replaced all template checks with contains "cgr.dev" .Values.image.repository. Now the FIPS JCE override activates when the Keycloak image is from Chainguard

chance-coleman:
@greptileai review this PR please

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -185,13 +185,19 @@ spec:
             - secretRef:
                 name: {{ include "keycloak.fullname" . }}-realm-env
           env:
+          {{- if and (contains "cgr.dev" .Values.image.repository) .Values.env }}
+            {{- range .Values.env }}
+              {{- if eq .name "JAVA_OPTS_APPEND" }}
+                {{- fail "cgr.dev image detected with JAVA_OPTS_APPEND in .Values.env; the chart injects JAVA_OPTS_APPEND to load the FIPS JCE security override" }}
+              {{- end }}
+            {{- end }}
```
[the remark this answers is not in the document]
chance-coleman: The `envFrom` secret (`keycloak-realm-env`) is chart-managed, its contents are controlled by the chart, not by user-supplied values. An external actor injecting `JAVA_OPTS_APPEND` into that secret would already have enough cluster access to cause far worse problems. The guard covers the only realistic conflict vector: a deployer explicitly setting `JAVA_OPTS_APPEND` in `.Values.env`. Closing without a fix.

src/prometheus-stack/values/unicorn/kube-prometheus-stack.yaml
```diff
@@ -39,19 +39,20 @@ prometheusOperator:
     patch:
       enabled: true
       image:
-        registry: quay.io
-        repository: rfcurated/ingress-nginx/kube-webhook-certgen
-        tag: 1.6.9-jammy-fips-rfcurated-rfhardened
+        registry: cgr.dev
+        repository: defenseunicorns.com/kube-webhook-certgen-fips
+        # Chainguard provides kube-webhook-certgen from their ingress-nginx package, which follows a different versioning scheme than upstream's ingress-nginx/kube-webhook-certgen
+        tag: 1.15.2
```
joelmccoy, line 45: nit: i think this is the latest
```suggestion
        tag: 1.15.5
```

joelmccoy: But totally ok not making this change for this PR. renovate will bump. dont need to retrigger CI just for this

joelmccoy (approved):
LGTM!

PR #2651, opened 2026-05-06
docs: add patch 1.3.1 to release notes (#2649) (backport-1.3)

mjnagel:
cherry picked from commit 2eb9763ba13fc1f084a686449275a974830d93e9

PR #2655, opened 2026-05-06
docs: clarity on 1.3 release notes

mjnagel:
## Description

Switched release callout to mention `1.3` instead of `1.3.0`. This was feedback [here](https://github.com/defenseunicorns/uds-core/pull/2651#discussion_r3195453674) on the backport PR to the release branch.

`1.3` mirrors other callouts in this same doc and other release notes, for example `UDS Core 1.3 introduces...`.

### DIFF
--- a/docs/operations/release-notes/1-3.mdx
+++ b/docs/operations/release-notes/1-3.mdx
@@ -36,13 +36,13 @@ UDS Core 1.3 introduces opt-in support for public Keycloak clients (PKCE-enforce
 
 ### Identity Config updates (0.26.1)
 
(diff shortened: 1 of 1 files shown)


briantwatson (commented):
Thinking it might make sense to drop the `.0` in the template here in the spirit of these changes?  https://github.com/defenseunicorns/uds-core/blob/main/.ai/docs/release-notes-template.md?plain=1#L96

mjnagel:
> Thinking it might make sense to drop the `.0` in the template here in the spirit of these changes? https://github.com/defenseunicorns/uds-core/blob/main/.ai/docs/release-notes-template.md?plain=1#L96

The changelog is explicitly for the first release in each minor, so that link should actually stay as `.0` (note the [callout on our changelog](https://github.com/defenseunicorns/uds-core/blob/main/CHANGELOG.md)).

briantwatson (approved):

PR #2666, opened 2026-05-12
fix(eks): stabilize UDP NetworkPolicy e2e coverage

chance-coleman:
### Problem
The UDP NetworkPolicy E2E test was flaky on EKS for two independent reasons:
1. Cross-node UDP traffic was not actually allowed by the EKS node security group configuration.  
The EKS test infra only allowed node-to-node UDP/53 by default, and the earlier SG change used the cluster security group as the source instead of a self-referencing node security group rule. That left cross-node UDP/5000 traffic unreliable.

2. The test itself depended on a one-shot UDP listener started via execInPod.  
The old test launched nc -u -l -p 5000 at assertion time and immediately raced client sends against listener startup and exec WebSocket timing. That made the test flaky even when networking was healthy.

### Fix
This PR fixes both issues with the smallest set of real changes:
- adds a self-referencing UDP ingress rule to the EKS node security group so cross-node pod-to-pod UDP is allowed
- changes the UDP server test workload to run a persistent nc loop inside the container
- updates the UDP E2E test to:
  - fetch pod names in parallel in beforeAll() to avoid EKS hook timeout
  - assert delivery by polling the server log instead of relying on listener stdout
  - send UDP directly to the server pod IP so the test stays focused on NetworkPolicy behavior instead of kube-proxy UDP ClusterIP routing behavior

### Why this approach
This keeps the test focused on what UDS Core owns:
-  UDP allow/deny NetworkPolicy behavior between pods
It avoids unrelated sources of flake:
- incorrect EKS cross-node UDP security group rules
- execInPod listener startup timing
- EKS UDP ClusterIP/DNAT service-path instability

### Validation
Validated successfully across CI flavors after these changes.

## Type of change

## Checklist before merging

### DIFF
--- a/.github/test-infra/aws/eks/cluster.tf
+++ b/.github/test-infra/aws/eks/cluster.tf
@@ -1,4 +1,4 @@
-# Copyright 2025 Defense Unicorns
+# Copyright 2025-2026 Defense Unicorns
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 locals {
@@ -53,6 +53,17 @@ module "eks" {
       description                   = "Cluster API Ingress on non-privileged ports"
       protocol                      = "tcp"
       from_port                     = 1025
+      to_port                       = 65535
+      type                          = "ingress"
+      source_cluster_security_group = true
+    }
+
+    # The terraform-aws-eks recommended rules only open UDP/53 (CoreDNS) between nodes,
+    # so cross-node pod-to-pod UDP on any other port is silently dropped at the SG.
+    clusterapi_ingress_udp = {
+      description                   = "Pod-to-pod UDP across nodes"
+      protocol                      = "udp"
+      from_port                     = 0
       to_port                       = 65535
       type                          = "ingress"
       source_cluster_security_group = true
--- a/.github/test-infra/aws/eks/cluster.tf
+++ b/.github/test-infra/aws/eks/cluster.tf
@@ -59,14 +59,15 @@ module "eks" {
     }
 
     # The terraform-aws-eks recommended rules only open UDP/53 (CoreDNS) between nodes,
-    # so cross-node pod-to-pod UDP on any other port is silently dropped at the SG.
+    # so cross-node pod-to-pod UDP on any other port is silently dropped at the node SG.
+    # This must be self-referencing node-to-node traffic, not cluster-SG sourced traffic.
     clusterapi_ingress_udp = {
-      description                   = "Pod-to-pod UDP across nodes"
-      protocol                      = "udp"
-      from_port                     = 0
-      to_port                       = 65535
-      type                          = "ingress"
-      source_cluster_security_group = true
+      description = "Pod-to-pod UDP across nodes"
+      protocol    = "udp"
+      from_port   = 0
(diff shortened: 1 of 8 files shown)


test/vitest/network.spec.ts
```diff
@@ -83,22 +83,82 @@ let curlPodNameEgress2 = "";
 let udpServerPodName = "";
 let udpClientPodName = "";
 
+async function waitForPodByLabel(namespace: string, labelSelector: string): Promise<string> {
+  const deadline = Date.now() + 120000;
```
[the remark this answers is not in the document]
chance-coleman: I shortened the helper deadline so it fails with a descriptive pod-readiness error before Vitest times out the test, and I increased the outer test timeout to leave enough budget for the UDP assertions and retries.

test/vitest/network.spec.ts
```diff
@@ -83,22 +83,82 @@ let curlPodNameEgress2 = "";
...
+    .WithLabel("app=udp-echo-client")
+    .Get();
+  for (const pod of clientPods.items) {
+    if (pod.metadata?.name) {
+      await K8s(kind.Pod).InNamespace("curl-ns-udp-allow").Delete(pod.metadata.name);
+    }
+  }
+
+  udpServerPodName = await waitForPodByLabel("curl-ns-udp-server", "app=udp-echo-server");
+  udpClientPodName = await waitForPodByLabel("curl-ns-udp-allow", "app=udp-echo-client");
```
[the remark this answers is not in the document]
chance-coleman: Since both UDP pods are deleted together, the readiness waits should run in parallel too. I updated restartUdpPods() to wait for both replacement pods with Promise.all().

chance-coleman:
@greptileai review this PR

briantwatson (dismissed):

joelmccoy (dismissed):

test/vitest/network.spec.ts
```diff
@@ -68,6 +68,55 @@ function isResponseError(curlOutput: { stdout: string; stderr: string }) {
...
+  let log = await readUdpLog(serverPodName);
+
+  while (Date.now() < deadline) {
+    if (log === expected) {
+      return log;
+    }
+
+    await new Promise(resolve => setTimeout(resolve, intervalMs));
+    log = await readUdpLog(serverPodName);
+  }
```
jasonwashburn: definitely nit-worthy, but I believe we could simplify the logic a bit here by just calling `readUdpLog` in the beginning of the loop
```suggestion

  while (Date.now() < deadline) {
    const log = await readUdpLog(serverPodName);
    if (log === expected) return log;
    await new Promise(resolve => setTimeout(resolve, intervalMs));
  }
```

.github/test-infra/aws/eks/cluster.tf
jasonwashburn:
```suggestion
# Copyright 2025-2026 Defense Unicorns
```

jasonwashburn (commented):
One nit and a copyright update, otherwise everything looks good to me!

test/vitest/network.spec.ts
chance-coleman: definitely cleaner, updating this

jasonwashburn (approved):
Looks good (pending CI)

PR #2673, opened 2026-05-15
chore(ci): add test to verify loki able to flush to s3

jasonwashburn:
## Description

Adds test to EKS workflow to verify that Loki can flush to S3.

## Related Issue

Relates to CORE-495

## Type of change

## Steps to Validate
- Happy Path - Successful workflow completion
- Sad Path - Break Loki write s3 permissions. Run workflow. Validate test fails as expected.
  - example failing run created by changing loki-write service account in commit [756131d](https://github.com/defenseunicorns/uds-core/pull/2673/commits/756131dc7d3583e405685f2aa1752e8328591e0a) - https://github.com/defenseunicorns/uds-core/actions/runs/26062032260/job/76624324060?pr=2673#step:16:102

## Checklist before merging

- [X] [Contributor Guide](https://github.com/defenseunicorns/uds-core/blob/main/CONTRIBUTING.md) followed

### DIFF
--- a/.github/workflows/test-iac.yaml
+++ b/.github/workflows/test-iac.yaml
@@ -1,4 +1,4 @@
-# Copyright 2024 Defense Unicorns
+# Copyright 2024-2026 Defense Unicorns
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
 name: Filter IaC
@@ -77,6 +77,7 @@ jobs:
               - ".github/workflows/test-rke2.yaml"
             eks:
               - "tasks/iac.yaml"
+              - "src/loki/tasks.yaml"
               - ".github/bundles/eks/**"
               - ".github/test-infra/aws/eks/**"
               - ".github/workflows/test-eks.yaml"
--- a/src/loki/tasks.yaml
+++ b/src/loki/tasks.yaml
@@ -216,21 +216,23 @@ tasks:
           ATTEMPT=1
           while [ "$(date +%s)" -lt "${END_TIME}" ]; do
             REMAINING_SECONDS=$(( END_TIME - $(date +%s) ))
-            echo "Attempt ${ATTEMPT}: checking for Loki chunks object written after ${LOKI_OBJECT_STORAGE_TEST_START} (${REMAINING_SECONDS}s remaining)..."
+            echo "Attempt ${ATTEMPT}: checking for Loki chunk objects under fake/ written after ${LOKI_OBJECT_STORAGE_TEST_START} (${REMAINING_SECONDS}s remaining)..."
 
-            RECENT_OBJECT_KEY="$(aws s3api list-objects-v2 \
+            # auth_enabled is false, so Loki writes tenant chunk objects under fake/.
+            RECENT_CHUNK_OBJECT_COUNT="$(aws s3api list-objects-v2 \
               --bucket "${LOKI_CHUNKS_OBJECT_STORAGE_BUCKET}" \
-              --query "Contents[?LastModified>=\`${LOKI_OBJECT_STORAGE_TEST_START}\`].Key | [0]" \
+              --prefix "fake/" \
+              --query "length(Contents[?LastModified>=\`${LOKI_OBJECT_STORAGE_TEST_START}\`])" \
               --output text 2>"${S3_LIST_ERROR}" || true)"
 
-            if [ -n "${RECENT_OBJECT_KEY}" ] && [ "${RECENT_OBJECT_KEY}" != "None" ]; then
-              echo "Validated Loki wrote a recent object to configured chunks object storage."
+            if [ -n "${RECENT_CHUNK_OBJECT_COUNT}" ] && [ "${RECENT_CHUNK_OBJECT_COUNT}" -gt 0 ] 2>/dev/null; then
+              echo "Validated Loki wrote recent chunk objects to configured object storage."
               exit 0
             fi
 
-            echo "No recent Loki chunks object found yet; waiting ${{ .inputs.poll_interval }}s before retrying."
+            echo "No recent Loki chunk objects found under fake/ yet; waiting ${{ .inputs.poll_interval }}s before retrying."
             ATTEMPT=$(( ATTEMPT + 1 ))
             sleep "${{ .inputs.poll_interval }}"
           done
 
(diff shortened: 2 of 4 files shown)


jasonwashburn:
@greptileai review

src/loki/tasks.yaml
```diff
@@ -31,3 +35,178 @@ tasks:
...
+
+          curl -fsS \
+            -H "Content-Type: application/json" \
+            -X POST \
+            --data-raw "${PAYLOAD}" \
+            "http://127.0.0.1:${LOKI_PORT}/loki/api/v1/push" \
+            || fail "Failed to push test log to Loki write endpoint."
+
+          curl -fsS -X POST "http://127.0.0.1:${LOKI_PORT}/flush" >/dev/null \
+            || fail "Failed to flush Loki write endpoint."
```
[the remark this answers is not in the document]
jasonwashburn: Confirmed that /flush endpoint expects POST, reference here: https://grafana.com/docs/loki/latest/reference/loki-http-api/#flush-in-memory-chunks-to-backing-store

.github/workflows/test-iac.yaml
```diff
@@ -77,6 +77,7 @@ jobs:
               - ".github/workflows/test-rke2.yaml"
             eks:
               - "tasks/iac.yaml"
+              - "src/loki/tasks.yaml"
```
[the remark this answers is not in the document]
jasonwashburn: I think I'm going to leave this as-is...the intent of adding `src/loki/tasks.yaml` was to catch changes to the test task itself and run CI on PRs that make changes to that file. Adding `./src/loki/**` would be running CI for EVERY change to `./src/loki` which is not our current CI pattern.

src/loki/tasks.yaml
```diff
@@ -31,3 +35,202 @@ tasks:
...
+            REMAINING_SECONDS=$(( END_TIME - $(date +%s) ))
+            echo "Attempt ${ATTEMPT}: checking for Loki chunks object written after ${LOKI_OBJECT_STORAGE_TEST_START} (${REMAINING_SECONDS}s remaining)..."
+
+            RECENT_OBJECT_KEY="$(aws s3api list-objects-v2 \
+              --bucket "${LOKI_CHUNKS_OBJECT_STORAGE_BUCKET}" \
+              --query "Contents[?LastModified>=\`${LOKI_OBJECT_STORAGE_TEST_START}\`].Key | [0]" \
+              --output text 2>"${S3_LIST_ERROR}" || true)"
+
+            if [ -n "${RECENT_OBJECT_KEY}" ] && [ "${RECENT_OBJECT_KEY}" != "None" ]; then
+              echo "Validated Loki wrote a recent object to configured chunks object storage."
```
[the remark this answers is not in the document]
jasonwashburn: This is acceptable, we're validating that loki can write any logs to the s3 bucket, not that our specific log message made it.

src/loki/tasks.yaml
```diff
@@ -31,3 +35,202 @@ tasks:
...
+        setVariables:
+          - name: LOKI_CHUNKS_OBJECT_STORAGE_BUCKET
+
+      - description: Set object storage validation start time
+        cmd: date -u +'%Y-%m-%dT%H:%M:%S+00:00'
+        mute: true
+        setVariables:
+          - name: LOKI_OBJECT_STORAGE_TEST_START
+
+      - description: Push a test log and flush Loki chunks
```
chance-coleman, line 93: i think this can false-pass. Check only asks "did any object land in bucket after start time", not "did object from this test land". In EKS config, Loki chunks and admin both point at same S3 bucket, and Loki can already be writing unrelated data. So this can succeed even if injected log never makes it to chunk storage.

src/loki/tasks.yaml
```diff
@@ -31,3 +35,202 @@ tasks:
...
+          END_TIME=$(( $(date +%s) + ${{ .inputs.poll_timeout }} ))
+          ATTEMPT=1
+          while [ "$(date +%s)" -lt "${END_TIME}" ]; do
+            REMAINING_SECONDS=$(( END_TIME - $(date +%s) ))
+            echo "Attempt ${ATTEMPT}: checking for Loki chunks object written after ${LOKI_OBJECT_STORAGE_TEST_START} (${REMAINING_SECONDS}s remaining)..."
+
+            RECENT_OBJECT_KEY="$(aws s3api list-objects-v2 \
+              --bucket "${LOKI_CHUNKS_OBJECT_STORAGE_BUCKET}" \
+              --query "Contents[?LastModified>=\`${LOKI_OBJECT_STORAGE_TEST_START}\`].Key | [0]" \
+              --output text 2>"${S3_LIST_ERROR}" || true)"
```
chance-coleman: something my local codex review mentioned, This only looks at one list-objects-v2 page. S3 paginates at 1000 keys, and result order is by key, not LastModified. Once bucket has enough objects, newer writes can sit off first page and test will fail intermittently.

src/loki/tasks.yaml
jasonwashburn: Yeah, I looked at this a bit, but after looking into it a bit...found that by default nothing is being written to loki admin bucket ...for example, staging has been running a year and is still empty 

...that being said...there is maybe a case to be made for either splitting the buckets in CI...or trying to make the test a little more write aware (maybe look for presence of `fake/` or `index/` prefixes) ...in case we start adding things in the future...

chance-coleman: yeah, i think this is fine for now. There’s still some theoretical false-pass risk, but probably not enough to block on without seeing it happen. We can revisit if the check gets flaky or storage behavior changes.

jasonwashburn: well...actually...looking more into it..seems that maybe the admin bucket only get's used for GEL - https://github.com/grafana/loki/blob/88cdd1badd8b7251622790b59513c8d53a948e5d/production/helm/loki/values.yaml#L432

jasonwashburn: as discussed in slack... pivoting to searching the bucket for objects using the `fake/` prefix that are newer than test start time as evidence of a pass

src/loki/tasks.yaml
jasonwashburn: as discussed in slack... pivoting to searching the bucket for objects using the `fake/` prefix that are newer than test start time as evidence of a pass

chance-coleman (dismissed):

chance-coleman (approved):

PR #2682, opened 2026-05-19
fix: set KC_CACHE=local for single-node postgres to prevent Infinispan deadlock

chance-coleman:
## Description

When `autoscaling.enabled: false` (the default) and postgres is configured, `KC_CACHE` was not set. Keycloak defaults to `ispn`, which starts Infinispan with JGroups and JDBC_PING cluster discovery. On a single-node deployment, `StateTransferManagerImpl.start()` blocks on a `CompletableFuture` that never resolves which results in the node waiting for cluster members that will never join.

The bug became deterministic in Keycloak 26.6.0 due to the Infinispan 15 → 16 upgrade ([#45380](https://github.com/keycloak/keycloak/pull/45380)), which introduced more complex state transfer coordination that consistently deadlocks on singleton clusters.

## Fix

Set `KC_CACHE=local` when postgres is configured and autoscaling is disabled. Local cache skips JGroups/Infinispan clustering entirely, which is correct for single-replica deployments where there are no cluster peers to coordinate with.

HA mode (`autoscaling.enabled: true`) is unchanged, it continues to use `KC_CACHE=ispn`.

## Steps to Validate

Testing:
setup the uds-ha-config.yaml to look like this:
```

variables:
  core-identity-authorization:
    keycloak_ha: false
    keycloak_pg_username: postgres
    keycloak_pg_password: unicorn123!@#UN
    keycloak_pg_database: keycloak
    keycloak_pg_host: host.k3d.internal
    keycloak_devmode: false
    insecure_admin_password_generation: true
    AUTHSERVICE_REPLICA_COUNT: 2

  # Authservice config is managed by the operator in the base layer
  # core-base:
  #   AUTHSERVICE_REDIS_URI: redis://authservice:authservice@host.k3d.internal:6379
```
and then create a docker container for postgres:
```
docker run -d \
  --name postgres \
  -e POSTGRES_USER=keycloak \
  -e POSTGRES_PASSWORD=keycloakpass1234 \
  -e POSTGRES_DB=keycloak \
  -p 5432:5432 \
  postgres:16
```

then build the k3d-slim-dev:
```
uds run create:k3d-slim-dev-bundle
```

and then deploy:
```
UDS_CONFIG=bundles/k3d-slim-dev/uds-external-db-config.yaml \
  uds deploy bundles/k3d-slim-dev/uds-bundle.yaml --confirm
```

The fix:
use my PR branch and do the same thing again.

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -300,6 +300,10 @@ spec:
             - name: KC_CACHE_EMBEDDED_MTLS_ENABLED
               value: "false"
           {{- end }}
+          {{- if and (not .Values.autoscaling.enabled) (eq (include "keycloak.postgresql.config" .) "true") }}
+            - name: KC_CACHE
+              value: local
+          {{- end }}
           {{- if eq (include "keycloak.postgresql.config" .) "true" }}
             # Postgres database configuration
             - name: KC_DB
--- a/src/keycloak/chart/templates/uds-package.yaml
+++ b/src/keycloak/chart/templates/uds-package.yaml
@@ -105,8 +105,7 @@ spec:
           app: istiod
         port: 15012
 
-      {{- if .Values.autoscaling.enabled }}
-      # HA for keycloak
+      # JGroups ports for Keycloak clustering (required for multi-replica deployments)
       - direction: Ingress
         remoteGenerated: IntraNamespace
         ports:
@@ -117,7 +116,6 @@ spec:
         ports:
           - 7800
           - 57800
-      {{- end }}
 
       # Custom rules for additional networking access
       {{- with .Values.additionalNetworkAllow }}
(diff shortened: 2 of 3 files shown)


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -300,6 +300,10 @@ spec:
             - name: KC_CACHE_EMBEDDED_MTLS_ENABLED
               value: "false"
           {{- end }}
+          {{- if and (not .Values.autoscaling.enabled) (eq (include "keycloak.postgresql.config" .) "true") }}
```
slaskawi, line 303: Could we branch it based on `.Values.autoscaling.enabled` only? 

And then if yes, could we colocate it with an "if" statement for autoscaling? This way we'll improve the readability of the entire block

slaskawi (approved):

slaskawi (commented):
After discussing this with Keycloak Maintainers, we'll probably need another approach. 

The root issue is https://github.com/keycloak/keycloak/issues/48792. The Virtual Thread Checker has a small bug and doesn't react correctly for environments with less than 4 CPUs/vCPUs. In these configurations, there are higher chances of deadlock. 

In order to address this (and make the entire Keycloak less prone to the deadlocks with limited resources), it is recommended to add `-Dorg.infinispan.threads.virtual=false` to the continer args section.

PR #2685, opened 2026-05-20
feat: add Envoy Gateway Component

chance-coleman:
## Description
Adds Envoy Gateway as an optional UDS Core component, providing the controller and GatewayClass infrastructure needed for UDP ingress support.

- New `src/envoy-gateway/` component wrapping `oci://docker.io/envoyproxy/gateway-helm:v1.8.0` with upstream, registry1, and unicorn flavor variants
- Deploys Envoy Gateway controller in `envoy-gateway-system` with a `GatewayClass` named `envoy-gateway`
- UDS Package CR with ambient mesh mode and network policies covering controller, webhook, and certgen job
- ServiceMonitor targeting port 19001 for Prometheus scraping
- Wired into `packages/standard` as an optional component and `bundles/k3d-standard` as an opt-in via `optionalComponents`

Chart version pinned to v1.8.0, this ships Gateway API CRDs at v1.5.1, which is required to pass the `safe-upgrades.gateway.networking.k8s.io` ValidatingAdmissionPolicy installed by Istio. Earlier chart versions bundle v1.4.1 CRDs and will be blocked on install.

On-demand Gateway lifecycle (Pepr operator) and UDPRoute generation are scoped to CORE-502.

## Related Issue

Fixes Core-482

## Type of change

## Steps to Validate
1. Deploy the standard package with the `envoy-gateway` component enabled
2. Run `uds run src/envoy-gateway:validate`,  waits for the `envoy-gateway` Deployment to be available and `GatewayClass envoy-gateway` to be `Accepted`

## Checklist before merging


src/envoy-gateway/tasks.yaml
```diff
@@ -0,0 +1,26 @@
...
+            kind: GatewayClass
+            name: envoy-gateway
+            condition: Accepted
+
+  - name: e2e-test
+    actions:
+      - description: "Run Envoy Gateway E2E tests"
+        cmd: |
+          npm ci && npx vitest run "envoy-gateway"
+        dir: test/vitest
```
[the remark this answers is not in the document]
chance-coleman: envoy-gateway.spec.ts covers controller deployment availability and GatewayClass acceptance. E2e tests for UDPRoute traffic flows are scoped to the follow-on PR (CORE-502) since there's no Gateway or UDPRoutes to test against yet.

src/envoy-gateway/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,45 @@
+# Copyright 2024-2026 Defense Unicorns
```
briantwatson: For net new, I think we should we just do 2026 on license headers?

src/envoy-gateway/chart/templates/_helpers.tpl
briantwatson, line 1: I think we can delete this file if it's only the comment?

briantwatson (commented):
Nice addition!  A couple small comments

slaskawi (dismissed):

packages/envoy-gateway/zarf.yaml
joelmccoy, line 1: shouldn't we put envoy-gateway under `base`?  declaring it as it's own folder here makes it look like a functional layer which I think is not the case?

packages/envoy-gateway/tasks.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright 2024-2026 Defense Unicorns
```
joelmccoy:
```suggestion
# Copyright 2026 Defense Unicorns
```

Change this for all net new files. there is a lot in this PR

src/envoy-gateway/chart/templates/uds-package.yaml
```diff
@@ -0,0 +1,45 @@
...
+        description: "Intra-namespace communication for managed proxies and cert generation"
+      - direction: Ingress
+        selector:
+          control-plane: envoy-gateway
+        remoteGenerated: IntraNamespace
+        description: "Intra-namespace communication from managed proxies"
+      - direction: Ingress
+        selector:
+          control-plane: envoy-gateway
+        remoteGenerated: KubeAPI
```
joelmccoy: I wonder if this needs to be Anywhere for managed clusters like EKS (as the IPs can be outside the cluster).  This might work fine, but might be good to validate based on the comment we have here on the promtheus webhooks: https://github.com/defenseunicorns/uds-core/blob/62cd29eeff8b18ab9610745e9992c97313fc6f98/src/prometheus-stack/chart/templates/uds-package.yaml#L47

src/envoy-gateway/common/zarf.yaml
```diff
@@ -0,0 +1,32 @@
...
+  url: https://github.com/envoyproxy/gateway
+
+components:
+  - name: envoy-gateway
+    required: false
+    manifests:
+      - name: envoy-gateway-class
+        namespace: envoy-gateway-system
+        files:
+          - ../manifests/gatewayclass.yaml
```
joelmccoy: nit: is there a strong reason why we don't just include this in the `uds-envoy-gateway-config` chart? feels cleaner that way and will be managed via helm vs a raw manifest.

src/envoy-gateway/common/zarf.yaml
```diff
@@ -0,0 +1,32 @@
...
+        namespace: envoy-gateway-system
+        version: 0.1.0
+        localPath: ../chart
+      - name: envoy-gateway
+        namespace: envoy-gateway-system
+        url: oci://docker.io/envoyproxy/gateway-helm
+        version: v1.8.0
+        # Disable SSA to prevent field manager conflicts with Istio's Gateway API CRDs.
+        # Istio installs safe-upgrades.gateway.networking.k8s.io VAP with field manager "uds";
+        # SSA mode causes a conflict when this chart tries to apply the same VAP on install.
```
joelmccoy: This comment is a bit confusing to me. I thought this fails because of it trying to reinstall CRDs at a lower version: https://www.notion.so/UDP-Envoy-Gateway-Component-34be512f24fc8082ac78d3e5c41ce991?d=35ae512f24fc802ba17a001cc44bf6fc&source=copy_link#350e512f24fc808a9442d86200ad3e5b

Or is something else happening?

Ideally I like to remove the need to turn SSA off.  I think once this [issue](https://github.com/envoyproxy/gateway/issues/8560) is merged (there is a PR out there for it already), we should just be able to skip CRDs completely?

If so, maybe cut a ticket for this and drop this issue reference in this comment with the context so we can go clean this up later.

test/vitest/envoy-gateway.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+      plural: "gatewayclasses",
+      name: "envoy-gateway",
+    });
+    const gatewayClass = res as {
+      status?: { conditions?: Array<{ type: string; status: string }> };
+    };
+    const accepted = gatewayClass.status?.conditions?.find(c => c.type === "Accepted");
+    expect(accepted?.status).toBe("True");
+  });
+});
```
joelmccoy, line 198: these e2e tests are useless in my opinion... the validate task already checks them. and zarf already confirms deployments are ready with kstatus checks.  These should be removed and replace with a better e2e test.

A true e2e test here would involve applying a gateway manifest of type `envoy-gateway` and check to see if it resolves correctly.

And then eventually once we have the operator logic, we remove the applying of the gateway manifest and just apply and check the package CR.

joelmccoy (commented):
Generally looking good. A couple comments and suggested changes.

src/envoy-gateway/common/zarf.yaml
chance-coleman: Updated. To clarify: the chart ships `safe-upgrades.gateway.networking.k8s.io` as a VAP in its `crds/` directory. Istio already installed that VAP with field manager `uds`. When Helm applies it via SSA, it conflicts on `.spec.matchConstraints` due to competing field ownership, not a CRD version rejection. Disabling SSA switches Helm to client-side apply which does not enforce field ownership, so the apply becomes a no-op when the resource already exists. https://github.com/envoyproxy/gateway/issues/8560 will let us skip CRDs entirely and remove this.

src/envoy-gateway/common/zarf.yaml
chance-coleman: Moved to `chart/templates/gatewayclass.yaml`. The design doc originally specified a Zarf manifest but the chart template is cleaner and Helm manages the lifecycle correctly.

src/envoy-gateway/chart/templates/uds-package.yaml
chance-coleman: Updated to `Anywhere` with the same TODO comment pattern as prometheus-stack. In managed clusters like EKS the API server IP is external, so `KubeAPI` on an ingress rule won't cover it. The egress certgen rule stays as `KubeAPI` since that direction resolves correctly

packages/envoy-gateway/tasks.yaml
chance-coleman: fixed 👍🏼

packages/envoy-gateway/zarf.yaml
chance-coleman: Good call, the design doc calls this a base layer component. Moved `envoy-gateway` into `packages/base/zarf.yaml` as an optional component, same pattern as `istio-egress-gateway`. Removed the standalone `packages/envoy-gateway/` directory and updated `packages/standard/` to import from `../base`.

test/vitest/envoy-gateway.spec.ts
chance-coleman: Removed the basic tests since they were redundant with validate/kstatus and the Gateway test requires the proxy image which isn't bundled yet. Real e2e tests with traffic flows are scoped to CORE-502

joelmccoy (changes requested):
After playing around with this, I realize that we don't bundle the envoy proxy image in this PR, so this feature is not really usable/testable e2e yet. It was called out in the design doc to include this in this part and I think we would include the proxy image in zarf and also include the CRD to override the envoy proxy image: https://www.notion.so/defense-unicorns/UDP-Envoy-Gateway-Component-34be512f24fc8082ac78d3e5c41ce991?source=copy_link#35ae512f24fc8010a8eccb8d6398a766.

chance-coleman:
@greptileai review this

joelmccoy:
@greptileai review

test/vitest/envoy-gateway.spec.ts
joelmccoy, line 1: is there a reason why you removed this test? I think it can ultimately be replaced once we have the operator logic in... but it might be good to keep in for now so that envoy gateway creation is tested on all flavors and IAC as well?

chance-coleman: No strong reasons, mainly because they'll be improved/replaced with future testing. I've added them back in.

src/envoy-gateway/common/zarf.yaml
```diff
@@ -0,0 +1,52 @@
...
+    files:
+      # Fetched at package create time and applied at deploy time to avoid Helm's
+      # release secret size limit with the generated EnvoyProxy CRD.
+      - source: https://github.com/envoyproxy/gateway/releases/download/v1.8.0/envoy-gateway-crds.yaml
+        target: /tmp/uds-core-envoy-gateway-crds.yaml
+    actions:
+      onDeploy:
+        after:
+          - description: Apply Envoy Gateway CRDs
+            cmd: ./zarf tools kubectl apply --server-side=true --force-conflicts --field-manager=uds -f /tmp/uds-core-envoy-gateway-crds.yaml
```
joelmccoy: Do we really even need this if ssa is disabled below?? feels like we are reapplying for no reason.

src/envoy-gateway/manifests/envoyproxy-registry1.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: uds-default
+  namespace: envoy-gateway-system
+spec:
+  provider:
+    type: Kubernetes
+    kubernetes:
+      envoyDeployment:
+        container:
+          image: registry1.dso.mil/ironbank/opensource/envoy_proxy/envoy-distroless:v1.38.0
```
joelmccoy: am i missing something here? why do we need this manifest? can't we just set the value in the helm chart?     https://github.com/envoyproxy/gateway/blob/b96a2d3f7210850e113ec3274e0ee5180c498ea8/charts/gateway-helm/values.tmpl.yaml#L31

src/envoy-gateway/zarf.yaml
```diff
@@ -0,0 +1,119 @@
...
+        target: /tmp/uds-core-envoy-gateway-envoyproxy.yaml
+      - source: manifests/gatewayclass.yaml
+        target: /tmp/uds-core-envoy-gateway-gatewayclass.yaml
+    actions:
+      onDeploy:
+        after:
+          - description: Apply EnvoyProxy configuration
+            cmd: ./zarf tools kubectl apply --server-side=true --force-conflicts --field-manager=uds -f /tmp/uds-core-envoy-gateway-envoyproxy.yaml
+          - description: Apply GatewayClass
+            cmd: ./zarf tools kubectl apply --server-side=true --force-conflicts --field-manager=uds -f /tmp/uds-core-envoy-gateway-gatewayclass.yaml
```
joelmccoy: not a fan of this hack at all and not sure it is needed. why can't we have this in the config chart? It was passing CI at one point and not sure I fully understand why this is needed.

joelmccoy: These will never get cleaned up on a remove too.  We need to have these managed via a helm chart ideally in the config chart.

joelmccoy (changes requested):
Feels like we added a couple hacks here that I'm am not sure I fully understand.  The raw kubectl apply probably isn't going to fly and will mostly like cause us some issues. Best practice is to have all resources managed by a helm chart.

src/envoy-gateway/zarf.yaml
chance-coleman: sorry i should have moved this to draft while i was testing different solutions here. definitely agreed this manifest approach is not the right one, i actually didnt expect it to work but was curious how it worked.

src/envoy-gateway/common/zarf.yaml
chance-coleman: You're right, removed it. The chart's charts/crds/ subchart applies all the CRDs (including the generated EnvoyProxy one) as part of helm install. The release secret size concern in the comment was from the old manifests: approach and no longer applies. Also dropped the before-action wait that was only there to check for the CRD the pre-apply was creating.

chance-coleman:
@greptileai review this again

joelmccoy (commented):
Generally looking good here! However, I think we need to enable this optional component for our IAC bundles (for example [here](https://github.com/defenseunicorns/uds-core/blob/main/.github/bundles/eks/uds-bundle.yaml#L26)).  Will also want to see these pass the IAC tests before feeling confident to ship this. I think updating the bundles will trigger thoughts

joelmccoy (approved):

PR #2686, opened 2026-05-21
fix: avoid virtual threads in Keycloak

slaskawi:
## Description

This Pull Request uses an alternative approach to https://github.com/defenseunicorns/uds-core/pull/2682

Instead of conditionally disabling clustering, it disables the Virtual Threads. 

See these links for more details:

- https://github.com/keycloak/keycloak/issues/48792
- https://github.com/keycloak/keycloak/issues/49203

## Type of change

## Steps to Validate

Run a single replica Keycloak, with PostreSQL and without HA. 

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -177,9 +177,12 @@ spec:
             - "--spi-connections-http-client-default-jitter-factor={{ .Values.httpRetry.jitterFactor }}"
             {{- end }}
             {{- end }}
+            {{- if .Values.disableVirtualThreads }}
             # Infinispan 16 introduced virtual threads, enabled by default. In constrained environments, such as Kubernetes with fewer than 4 vCPUs, it is highly recommended
             # to disable this feature to reduce the risk of deadlocks. See https://github.com/keycloak/keycloak/issues/48792 for more details.
+            # The switch if officially supported, see https://infinispan.org/blog/2025/11/10/infinispan-16-0
             - "-Dorg.infinispan.threads.virtual=false"
+            {{- end }}
           {{- with .Values.lifecycleHooks }}
           lifecycle:
           {{- toYaml . | nindent 12 }}
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -179,7 +179,7 @@ spec:
             {{- end }}
             # Infinispan 16 introduced virtual threads, enabled by default. In constrained environments, such as Kubernetes with fewer than 4 vCPUs, it is highly recommended
             # to disable this feature to reduce the risk of deadlocks. See https://github.com/keycloak/keycloak/issues/48792 for more details.
-            - Dorg.infinispan.threads.virtual=false
+            - "-Dorg.infinispan.threads.virtual=false"
           {{- with .Values.lifecycleHooks }}
           lifecycle:
           {{- toYaml . | nindent 12 }}
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -177,6 +177,9 @@ spec:
             - "--spi-connections-http-client-default-jitter-factor={{ .Values.httpRetry.jitterFactor }}"
             {{- end }}
             {{- end }}
+            # Infinispan 16 introduced virtual threads, enabled by default. In constrained environments, such as Kubernetes with fewer than 4 vCPUs, it is highly recommended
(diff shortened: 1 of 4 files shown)


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -177,6 +177,9 @@ spec:
             - "--spi-connections-http-client-default-jitter-factor={{ .Values.httpRetry.jitterFactor }}"
             {{- end }}
             {{- end }}
+            # Infinispan 16 introduced virtual threads, enabled by default. In constrained environments, such as Kubernetes with fewer than 4 vCPUs, it is highly recommended
+            # to disable this feature to reduce the risk of deadlocks. See https://github.com/keycloak/keycloak/issues/48792 for more details.
+            - Dorg.infinispan.threads.virtual=false
```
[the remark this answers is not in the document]
slaskawi: Good catch, the flag is unconditional so one assertion is enough. Added kc_infinispan_virtual_threads_test.yaml mirroring the kc_http_retry_test.yaml style and verified with helm unittest src/keycloak/chart. Test passes.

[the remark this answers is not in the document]
slaskawi: Already fixed in b38524e3 — the leading dash is present (`- "-Dorg.infinispan.threads.virtual=false"`). Keeping it as a CLI arg rather than JAVA_OPTS_APPEND because that env path is currently cgr.dev-gated with a fail-guard against user overrides, and kc.sh forwards -D flags to the JVM regardless. Happy to revisit if we ever generalize the JAVA_OPTS_APPEND injection across flavors.

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -177,6 +177,9 @@ spec:
             - "--spi-connections-http-client-default-jitter-factor={{ .Values.httpRetry.jitterFactor }}"
             {{- end }}
             {{- end }}
+            # Infinispan 16 introduced virtual threads, enabled by default. In constrained environments, such as Kubernetes with fewer than 4 vCPUs, it is highly recommended
+            # to disable this feature to reduce the risk of deadlocks. See https://github.com/keycloak/keycloak/issues/48792 for more details.
+            - "-Dorg.infinispan.threads.virtual=false"
```
chance-coleman, line 184: This flag is unconditional, it applies to all deployments including HA clusters. Disabling globally seems like a regression for HA users who are not affected by this bug.

chance-coleman, line 184: Is this a documented API endpoint? i was having a hard time finding it, just want to make sure we don't get rug pulled if we have to keep this implementation for awhile.

src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -177,6 +177,9 @@ spec:
             - "--spi-connections-http-client-default-jitter-factor={{ .Values.httpRetry.jitterFactor }}"
             {{- end }}
             {{- end }}
+            # Infinispan 16 introduced virtual threads, enabled by default. In constrained environments, such as Kubernetes with fewer than 4 vCPUs, it is highly recommended
```
chance-coleman, line 181: Are you think of combining your solution with mine? Or would you rather implement only yours?

chance-coleman (commented):
I think this works, but im concerned its solving the symptom instead of the problem. Maybe that's okay cause we're just waiting for an upstream fix.

src/keycloak/chart/templates/statefulset.yaml
slaskawi: The Java [Virtual Threads](https://openjdk.org/jeps/444) feature has been introduced in JDK 21 and is 100% transparent to the end user. Because of the complexity of the Keycloak stack (JGroups, Infinispan and finally Keycloak layer), switching to that new mechanism has been challanging. 

Users with signifficant amount of CPU in Resources and Limits will see slight performance increase. 

But maybe you're right - I'll introduce a switch for it.

slaskawi: The only reference I could find is https://infinispan.org/blog/2025/11/10/infinispan-16-0. 

Let me put it into the comment.

slaskawi: Added a reference to the docs.

slaskawi:
@chance-coleman Since the switch seems to be supported (at least officially called in the Infinispan Release Notes), I think it makes sense to leave it there. Large deployments (more than 4 vCPUs) can probably benefit from it but I honestly never heard about performance problems so far.

chance-coleman (approved):
LGTM, confirmed working in my reproduction

PR #2688, opened 2026-05-21
feat(checkpoint-dev): make checkpoint work on macOS and without sudo

zachariahmiller:
## Description

The checkpoint package previously relied on host bind-mounts and sudo to copy k3s/kubelet volume data, which breaks on macOS because Docker volume paths live inside the VM and aren't accessible from the host.

This PR reworks both sides of the checkpoint:

- **Create**: checkpoint.sh now uses docker cp to stream volume data directly from the container over the Docker socket — no host path access needed, no sudo. The container is explicitly paused across the commit and both copies for a consistent snapshot.
- **Deploy**: zarf.yaml now creates named Docker volumes and populates them by piping the bundled tarballs through a busybox helper container. Named volumes live inside the Docker VM so k3d can mount them on any platform. The old sudo rm -rf host cleanup is replaced with docker volume rm.
- **Docs**: Removed the macOS limitation notice and sudo callouts. Added an OrbStack subnet config tip for intermittent networking issues.

Tested on macOS with Lima and OrbStack and Linux.

## Related Issue

Fixes CORE-552

## Type of change

## Steps to Validate
- Follow the steps in the README.md in packages/checkpoint-dev on MacOS and/or Linux (without sudo). Observe that it works. 

## Checklist before merging


jasonwashburn:
@greptileai review

packages/checkpoint-dev/zarf.yaml
```diff
@@ -30,54 +29,63 @@ components:
...
         onSuccess:
-          - cmd: |
-              if [ -z "$TMPDIR" ]; then
-                #  macOS sets TMPDIR to a user temp directory - this also provides more options to linux
-                TMPDIR="/tmp"
-              fi
-              DATA_DIR="${TMPDIR}/uds-checkpoint-data"
-              sudo rm -rf "$DATA_DIR" uds-checkpoint.tar
+          - cmd: rm -f uds-checkpoint.tar
       onDeploy:
```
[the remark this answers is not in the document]
zachariahmiller: Yeah i added this to the onDeploy and missed in the onCreate.

packages/checkpoint-dev/checkpoint.sh
```diff
@@ -1,77 +1,49 @@
...
+echo "Saving checkpoint image ..."
+docker save -o uds-k3d-checkpoint-latest.tar "$IMAGE_NAME"
 
-# Step 5: Commit and save the current container as a new image
-IMAGE_NAME="ghcr.io/defenseunicorns/uds-core/checkpoint:latest"
-echo "Committing container $CONTAINER_ID to image $IMAGE_NAME:latest..."
-docker commit -p "$CONTAINER_ID" "$IMAGE_NAME"
+echo "Saving busybox helper image ..."
+docker pull busybox
+docker save busybox -o busybox.tar
```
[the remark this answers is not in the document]
zachariahmiller: This wasn't tagged because i was unsure what actual source we would want to use for the image. When that is determined it should obviously been tagged accordingly.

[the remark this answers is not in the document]
Racer159: I would just use `cgr.dev/chainguard/busybox:latest` for this - that should be lightweight and would match how Zarf does things - https://github.com/zarf-dev/zarf/blob/main/Dockerfile#L1

packages/checkpoint-dev/zarf.yaml
```diff
@@ -30,54 +29,63 @@ components:
...
+              # (Lima, OrbStack, Docker Desktop) — no host bind-mount or sudo
+              # required.
+              docker volume create "$VOL_K3S" >/dev/null
+              docker volume create "$VOL_KUBELET" >/dev/null
 
+              tar -xOf uds-checkpoint.tar k3s_data.tar | \
+                docker run --rm -i -v "${VOL_K3S}:/data" busybox tar -x -C /data
+
+              tar -xOf uds-checkpoint.tar kubelet_data.tar | \
+                docker run --rm -i -v "${VOL_KUBELET}:/data" busybox tar -x -C /data
```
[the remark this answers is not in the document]
zachariahmiller: This is working, so i don't believe this is technically correct. Busybox tar reads from stdin by default. Happy to change this for clarity/portability, but the comments assertion is incorrect.

packages/checkpoint-dev/zarf.yaml
```diff
@@ -87,18 +95,27 @@ components:
...
               cluster:
                 kind: Pod
                 name: app.kubernetes.io/name=keycloak
                 namespace: keycloak
                 condition: Ready
+          - description: Pepr to be Healthy
+            wait:
+              cluster:
+                kind: Pod
+                name: app=pepr-uds-core
```
joelmccoy, line 121: nit: this technically only watches for the pepr admission pods (pepr policies). if we want for pepr to be fully healthy we should also watch for `app=pepr-uds-core-watcher` (operator logic) pod as well .

packages/checkpoint-dev/zarf.yaml
```diff
@@ -30,54 +29,63 @@ components:
...
+              # helper container that mounts the volume.
+              #
+              # Named volumes live inside the Docker VM, so this works on macOS
+              # (Lima, OrbStack, Docker Desktop) — no host bind-mount or sudo
+              # required.
+              docker volume create "$VOL_K3S" >/dev/null
+              docker volume create "$VOL_KUBELET" >/dev/null
 
+              tar -xOf uds-checkpoint.tar k3s_data.tar | \
+                docker run --rm -i -v "${VOL_K3S}:/data" busybox tar -x -C /data -f -
```
joelmccoy: i think the docker run command will default to upstream busybox? I think we want this to be the chainguard busybox we load in?
```suggestion
                docker run --rm -i -v "${VOL_K3S}:/data" cgr.dev/chainguard/busybox tar -x -C /data -f -
```

packages/checkpoint-dev/zarf.yaml
```diff
@@ -30,54 +29,63 @@ components:
...
+
+              tar -xOf uds-checkpoint.tar kubelet_data.tar | \
+                docker run --rm -i -v "${VOL_KUBELET}:/data" busybox tar -x -C /data -f -
```
joelmccoy:
```suggestion
                docker run --rm -i -v "${VOL_KUBELET}:/data" cgr.dev/chainguard/busybox tar -x -C /data -f -
```

packages/checkpoint-dev/zarf.yaml
```diff
@@ -30,54 +29,63 @@ components:
...
+
+              # Read the cluster token from the restored volume — no host
+              # filesystem access needed.
+              K8S_TOKEN=$(docker run --rm \
+                -v "${VOL_K3S}:/data" busybox \
```
joelmccoy:
```suggestion
                -v "${VOL_K3S}:/data" cgr.dev/chainguard/busybox \
```

joelmccoy (commented):
This is awesome and going to be very useful! Generally looking good to me. A couple small comments. Fixing the busybox docker runs to use the chainguard images is the only blocking thing from my perspective IMO.

packages/checkpoint-dev/zarf.yaml
zachariahmiller: oops... yeah good catch 🤦  Will test again with these changed just to make sure and update.

zachariahmiller: have to pass `--user root` to these for them to work correctly now, which doesn't really change anything, but definitely validates the correct image being used.

zachariahmiller: fixed

packages/checkpoint-dev/zarf.yaml
```diff
@@ -30,54 +29,63 @@ components:
...
       onDeploy:
         after:
           - cmd: |
-              if [ -z "$TMPDIR" ]; then
-                #  macOS sets TMPDIR to a user temp directory - this also provides more options to linux
-                TMPDIR="/tmp"
-              fi
-              DATA_DIR="${TMPDIR}/uds-checkpoint-data"
-              mkdir -p "$DATA_DIR"
+              VOL_K3S="k3d-${ZARF_VAR_CLUSTER_NAME}-k3s"
```
briantwatson, line 57: nit: noticed there are several piped commands below in this block here. I don't know if `pipefail` is set for zarf actions for inline scripts, I could find is a reference to `set -e` https://docs.zarf.dev/ref/actions/#cmd-action-configuration.  If possible set pipefail in the script or consider breaking up the pipes?

zachariahmiller: breaking up the pipes has other implications i'd rather avoid happy to add pipefail.

joelmccoy (dismissed):

zachariahmiller:
yup looking into it

joelmccoy (approved):

briantwatson (approved):
💯 

Image

PR #2699, opened 2026-05-26
feat: uds portal as layer in upstream flavor

briantwatson:
## Description

  - Add core-portal as a new layer (packages/portal/, src/portal/) via helm chart
  - Upstream flavor only via git-based helm chart from defenseunicorns/uds-portal
  - Wire tasks, README, renovate grouping, and release-please version bumping
  - DOMAIN/ADMIN_DOMAIN variables follow established src/-level pattern
 
> [!IMPORTANT]
> Currently NOT in the PR is addition to `publish-uds-core-layers` job in the publish workflow.  Do we want Portal as a layer published like these others?  https://github.com/defenseunicorns/uds-core/blob/feat/portal-in-upstream/.github/workflows/publish.yaml#L106

## Related Issue

Implements CORE-536

## Type of change

## Steps to Validate
- Created registry1 flavor standard package and confirmed portal was NOT present
- Created upstream and confirmed Portal present

## Checklist before merging


src/portal/common/zarf.yaml
```diff
@@ -0,0 +1,32 @@
...
+        valuesFiles:
+          - ../values/values.yaml
+    actions:
+      onDeploy:
+        after:
+          - description: Validate Portal Package
+            maxTotalSeconds: 300
+            wait:
+              cluster:
+                kind: packages.uds.dev
```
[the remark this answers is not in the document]
briantwatson: Had based this off the package/helm chart pattern from runtime that was shared during design:
- https://github.com/defenseunicorns/uds-core/blob/0011852dd6c8f1305e2fa0c837db45f3c1801c31/src/runtime/common/zarf.yaml#L26-L29

src/portal/tasks.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+tasks:
+  - name: validate
+    actions:
+      - description: Wait for uds-portal deployment to be ready
+        wait:
+          cluster:
+            kind: Deployment
+            name: uds-portal
+            namespace: uds-portal
```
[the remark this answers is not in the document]
briantwatson: Added

packages/portal/zarf.yaml
```diff
@@ -0,0 +1,25 @@
+# Copyright 2026 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+kind: ZarfPackageConfig
+metadata:
+  name: core-portal
+  description: "UDS Core (Portal)"
+  authors: "Defense Unicorns - Product"
```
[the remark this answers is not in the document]
briantwatson: Do we want to document this as a functional layer?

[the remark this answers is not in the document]
joelmccoy: yes we should document this as a functional layer

joelmccoy (commented):
I think we should also have a very basic e2e playwright test that checks that the UI loads properly when deployed on top of core.

src/portal/common/zarf.yaml
joelmccoy: you shouldn't need this check. zarf automatically waits for this validation based on kstatus checks.  zarf deploy won't finish until the package is ready.

src/portal/tasks.yaml
```diff
@@ -0,0 +1,13 @@
...
+            condition: Available
```
joelmccoy: this is unnecessary IMO since zarf checks this with kstatus on the deployment.  I would leave this blank or add an http status check to make sure authservice redirects. can use maru wait network tools like here: https://github.com/defenseunicorns/uds-core/blob/main/src/keycloak/tasks.yaml#L10-L15

src/portal/zarf.yaml
```diff
@@ -0,0 +1,30 @@
...
+    default: "uds.dev"
+
+  - name: ADMIN_DOMAIN
+    description: "Domain for admin services, defaults to `admin.DOMAIN`"
+
+components:
+  - name: uds-portal
+    required: true
+    only:
+      flavor: upstream
```
joelmccoy, line 22: we should add the unicorn in this PR too. unicorn is the exact same as upstream so I would just mirror an entry here.

renovate.json
```diff
@@ -131,6 +131,11 @@
       "groupName": "support-deps",
       "commitMessageTopic": "support-deps"
     },
+    {
+      "matchFileNames": ["src/portal/**", "packages/portal/**"],
```
joelmccoy: nit: shouldn't need the packages dir right?
```suggestion
      "matchFileNames": ["src/portal/**"],
```

packages/standard/zarf.yaml
```diff
@@ -128,3 +128,10 @@ components:
     required: true
     import:
       path: ../backup-restore
+
+  - name: uds-portal
+    required: true
+    only:
+      flavor: upstream
```
joelmccoy, line 135: add unicorn too

joelmccoy (commented):
Generally looking good. A couple things:
1. Add unicorn flavor (same as upstream)
2. Add simple e2e playwright test for UI sanity check
3. Add documentation and have this published as a functional layer

renovate.json
briantwatson: Removed

src/portal/zarf.yaml
briantwatson: Added unicorn

packages/standard/zarf.yaml
briantwatson: Added unicorn

src/portal/tasks.yaml
briantwatson: Added a `302` check like the keycloak one you shared

src/portal/common/zarf.yaml
briantwatson: Removed!

packages/portal/zarf.yaml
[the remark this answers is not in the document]
briantwatson: Added documentation

src/portal/common/zarf.yaml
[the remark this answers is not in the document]
briantwatson: Removed following discussion below

.github/workflows/test.yaml
```diff
@@ -64,7 +65,7 @@ jobs:
           installK3d: 'true'
 
       - name: Test a single layer package
-        if: ${{ inputs.package != 'all' && inputs.test_type == 'install' }}
+        if: ${{ inputs.package != 'all' && inputs.test_type == 'install' && !(inputs.package == 'portal' && inputs.flavor == 'registry1') }}
```
joelmccoy, line 68: isn't this already filtered out in the pull-request-conditionals filter, do we actually need this? https://github.com/defenseunicorns/uds-core/blob/b9696a92251e184fcbdc82376c239db47818e99f/.github/workflows/pull-request-conditionals.yaml#L177-L179

packages/portal/zarf.yaml
```diff
@@ -0,0 +1,32 @@
...
+  # x-release-please-start-version
+  version: "1.4.0"
+  # x-release-please-end
+  x-uds-dependencies: ["base", "identity-authorization"]
+  annotations:
+    dev.uds.title: UDS Core (Portal)
+    dev.uds.tagline: Provides a user-facing application portal for UDS environments
+    dev.uds.categories: UDS-Core,Portal
+    dev.uds.keywords: uds,portal
```
joelmccoy, line 17: We should have the Doug icon here like we do for the other functional layers: https://github.com/defenseunicorns/uds-core/blob/b9696a92251e184fcbdc82376c239db47818e99f/packages/backup-restore/zarf.yaml#L18

.github/workflows/test.yaml
```diff
@@ -49,6 +49,7 @@ jobs:
     name: Test
     env:
       UDS_LAYER: ${{ inputs.package }}
+      EXCLUDED_PACKAGES: ${{ inputs.flavor == 'registry1' && 'portal' || '' }}
```
joelmccoy: is this used somewhere? I can't seem to trace to see where this is used?

.github/workflows/publish.yaml
```diff
@@ -75,6 +75,8 @@ jobs:
       # Standard Package by default tests full core
       - name: Test amd64 Bundle
         if: ${{ !inputs.snapshot }}
+        env:
+          EXCLUDED_PACKAGES: ${{ matrix.flavor == 'registry1' && 'portal' || '' }}
```
joelmccoy: is this used somewhere? I can't seem to trace to see where this is used?

packages/portal/tasks.yaml
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+includes:
+  - portal: ../../src/portal/tasks.yaml
+
+tasks:
+  - name: validate
+    actions:
+      - description: "Validate Portal"
+        task: portal:validate
```
joelmccoy, line 11: I think we will want to run the e2e tests here in the functional layer validate. I would add an e2e task in `portal/tasks.yaml` and call it here.  See how we do logging/loki here: https://github.com/defenseunicorns/uds-core/blob/main/packages/logging/tasks.yaml#L13

joelmccoy (commented):
Couple other small things here but generally LGTM

joelmccoy:
@greptileai review

.github/workflows/test.yaml
briantwatson: Moved to using --set in the task for use by the [validate-packages job](https://github.com/defenseunicorns/uds-core/blob/202c8acfee8c678957c9ebce77fcee37ec395cb8/tasks/test.yaml#L71)

.github/workflows/publish.yaml
briantwatson: Moved to using --set in the task for use by the [validate-packages job](https://github.com/defenseunicorns/uds-core/blob/202c8acfee8c678957c9ebce77fcee37ec395cb8/tasks/test.yaml#L71)

.github/workflows/test.yaml
briantwatson: This workflow is triggered by both `workflow_dispatch` & `workflow_call`.  Guards against manual workflow dispatches primarily and any future workflows that run it.

packages/portal/zarf.yaml
briantwatson: Added

packages/portal/tasks.yaml
briantwatson: added

briantwatson:
@greptileai review

.github/workflows/test.yaml
joelmccoy: Generally I don't think this should ever really be called from a workflow call. and if you did you would just expect it to fail if the user put in portal and registry 1 from that workflow call.

Totally fine to keep this in, it's just another place to manage conditionals that might not be necessary.

joelmccoy (approved):
LGTM!

PR #2701, opened 2026-05-27
Security: YAML Parsing Without Schema Validation in Istio ConfigMap Sync

tomaioo:
## Summary

Security: YAML Parsing Without Schema Validation in Istio ConfigMap Sync

## Problem

**Severity**: `Medium` | **File**: `src/pepr/operator/controllers/istio/istio-configmap-sync.ts:L45`

The `restartGatewayPods` function in `src/pepr/operator/controllers/istio/istio-configmap-sync.ts` parses YAML from a ConfigMap using `yaml.load()` without specifying a schema or safe loading options. While the input comes from a ConfigMap within the cluster, this could potentially be exploited if an attacker gains access to modify the ConfigMap, allowing arbitrary object construction through YAML tags.

## Solution

Use `yaml.load()` with the `schema: yaml.SAFE_SCHEMA` option or use `yaml.safeLoad()` to prevent potential arbitrary object construction. Validate the parsed object structure against a strict schema before use.

## Changes

- `src/pepr/operator/controllers/istio/istio-configmap-sync.ts` (modified)

### DIFF
--- a/src/pepr/operator/controllers/istio/istio-configmap-sync.ts
+++ b/src/pepr/operator/controllers/istio/istio-configmap-sync.ts
@@ -33,7 +33,7 @@ export async function restartGatewayPods(istioConfig: kind.ConfigMap): Promise<v
   const mesh = istioConfig?.data?.["mesh"];
 
   if (mesh) {
-    const meshConfig = yaml.load(mesh) as IstioConfiguration;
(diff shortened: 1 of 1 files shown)


joelmccoy:
@tomaioo were you running into a particular issue here? Curious if there is a reason why you need this specific change?

joelmccoy:
closing this out. contributor is a bot that is spamming multiple repos right now with no human intervention.

PR #2708, opened 2026-05-28
chore: rename uds to unified defense stack

joelmccoy:
## Description

Renames Unicorn Delivery Service to Unified Defense Stack.

## Related Issue

CORE-550

## Type of change

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/values.yaml
+++ b/src/keycloak/chart/values.yaml
@@ -56,7 +56,7 @@ realmInitEnv:
   # @lulaEnd ef276ec4-7921-4287-a18c-80dec5f8e742
   SECURITY_HARDENING_ADDITIONAL_PROTOCOL_MAPPERS: ""
   SECURITY_HARDENING_ADDITIONAL_CLIENT_SCOPES: ""
-  # DISPLAY_NAME: "Unicorn Delivery Service"
+  # DISPLAY_NAME: "Unified Defense Stack"
   # Number of days of inactivity after which a non-admin user account is automatically disabled.
   # When unset, the workflow is disabled. Set to a positive integer to enable (e.g., 35).
   # ACCOUNT_INACTIVITY_DAYS: "35"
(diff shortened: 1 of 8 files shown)


src/keycloak/chart/values.yaml
```diff
@@ -87,7 +87,7 @@ themeCustomizations:
     # Override the realm display name at runtime for the UI (e.g., browser page title).
     # This is a theme-level override; if not set, the UI falls back to the Keycloak realm
     # display name configured in the realm itself (which may come from realmInitEnv.DISPLAY_NAME).
-    # realmDisplayName: "Unicorn Delivery Service"
+    realmDisplayName: "Unified Defense Stack"
```
joelmccoy: note: i uncommented this so that in live environments we override the UX theme. The realm config changes won't happen automatically.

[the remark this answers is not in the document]
joelmccoy: this is intentional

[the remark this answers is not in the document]
joelmccoy: this is intentional

jasonwashburn: Concerned this could be a problem in environments where the realm was initialized with a custom display name...if I understand how it works correctly...since `realmDisplayName` didn't have a default value before...it would have been using the custom name used for init...and now by adding a default value it would change those existing environments to use "Unified Defense Stack" I believe?

joelmccoy: yeah it's a good callout. I think that would be less common and we could just provide directions to update if they aren't setting the theme customization now.... if we leave it uncommented, then we have to provide upgrade instructions basically for everyone who doesn't override the theme currently.  Does this current approach still seem sane?  If so, I can prep some early release notes for calling this out

chance-coleman (dismissed):

src/keycloak/chart/values.yaml
joelmccoy: As discussed, there is a way we can make this backwards compatible with helm templating.  Let me know if you think this approach is good/correct.

joelmccoy:
@greptileai review

src/keycloak/chart/values.yaml
jasonwashburn: I like it!

jasonwashburn (approved):

PR #2710, opened 2026-05-29
fix: classification banner working for SPAs

joelmccoy:
## Description

The classification banner was noticed to block traffic to Grafana in uds-identity-config. See: https://github.com/defenseunicorns/uds-identity-config/pull/870.

Updates the classification banner injection to support SPAs and full-viewport applications by wrapping page content in a CSS grid with dedicated banner rows and a scrollable center. It handles both buffered and streamed HTML responses while skipping redirects and other non-200 responses.

I took inspiration from here: https://github.com/defenseunicorns-udm/uds-dot-mil/blob/main/packages/demo-banner/chart/templates/classification-banner.yaml

I tested this locally on:
- keycloak
- portal
- uds ui
- headlamp
- grafana
- minio dev console
- neuvector
- uds registry

NOTE: People will see the SAMPLE BANNER classification banner for demo bundle deploys.  If we feel this is weird I would be open to changing this. I also set this to SAMPLE BANNER instead of UNCLASSIFIED as I didn't want to assume classification for where people are deploying the demo bundle locally.

image

 
## Related Issue

Fixes CORE-590

## Type of change

## Steps to Validate
- `uds run` and check the banners on portal, sso, and grafana

## Checklist before merging


src/istio/common/chart/templates/classification-banner.yaml
```diff
@@ -153,26 +176,51 @@ spec:
...
+              if response_handle:headers():get("content-length") ~= nil then
+                -- Bounded response: buffer and rewrite in one shot.
+                local body = response_handle:body()
+                response_handle:body():setBytes(inject_banner(tostring(body:getBytes(0, body:length()))))
+              else
+                -- Streamed response: rewrite chunks as they pass through (bodyChunks does not
+                -- buffer), so we never wait for end-of-stream. Stop once the banner is in; if a
+                -- tag straddles a chunk boundary the chunk is left untouched rather than stalling.
+                local injected = false
+                for chunk in response_handle:bodyChunks() do
```
chance-coleman, line 281: This can silently skip the banner on streamed pages. Chunk boundaries are arbitrary, so matching <body> inside one chunk is not safe. Keep a small carry-over buffer until the tag is found.

src/istio/common/chart/templates/classification-banner.yaml
```diff
@@ -153,26 +176,51 @@ spec:
...
+              {{- else }}
+              html = html:gsub("<body([^>]*)>", "<body%1>" .. header, 1)
+              {{- end }}
+              return (html:gsub("<head>", "<head>" .. bodyPaddingScript, 1))
+            end
+
             -- Inject the banner for any hosts where it is enabled
             function envoy_on_response(response_handle)
               local content_type = response_handle:headers():get("Content-Type") or ""
               local host = response_handle:streamInfo():dynamicMetadata():get("envoy.lua")["host"]
```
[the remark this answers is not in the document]
joelmccoy: fixed.

src/istio/common/chart/templates/classification-banner.yaml
```diff
@@ -153,26 +176,51 @@ spec:
...
+                response_handle:body():setBytes(inject_banner(tostring(body:getBytes(0, body:length()))))
+              else
+                -- Streamed response: rewrite chunks as they pass through (bodyChunks does not
+                -- buffer), so we never wait for end-of-stream. Stop once the banner is in; if a
+                -- tag straddles a chunk boundary the chunk is left untouched rather than stalling.
+                local injected = false
+                for chunk in response_handle:bodyChunks() do
+                  if chunk:length() > 0 and not injected then
+                    local data = inject_banner(chunk:getBytes(0, chunk:length()))
+                    chunk:setBytes(data)
```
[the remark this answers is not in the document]
joelmccoy: fixed.

src/istio/common/chart/templates/classification-banner.yaml
```diff
@@ -153,26 +176,51 @@ spec:
...
 
-                response_handle:body():setBytes(body_text)
+              -- response_handle:body() buffers the whole body and blocks until end-of-stream.
+              -- That is fine for a bounded response (Content-Length present), but it hangs the
+              -- gateway on a streamed/chunked response with no Content-Length (e.g. an SPA that
+              -- renders its HTML directly to the response). Pick the strategy accordingly.
+              if response_handle:headers():get("content-length") ~= nil then
+                -- Bounded response: buffer and rewrite in one shot.
+                local body = response_handle:body()
+                response_handle:body():setBytes(inject_banner(tostring(body:getBytes(0, body:length()))))
```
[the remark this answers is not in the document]
joelmccoy: fixed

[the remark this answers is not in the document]
joelmccoy: fixed.

src/istio/common/chart/templates/classification-banner.yaml
joelmccoy: Added a carry-over buffer. When a chunk doesn't contain the complete tag, I hold back the trailing unterminated `<...` fragment and prepend it to the next chunk. This ensures that a `<body>` or `<head>` split across a boundary still gets matched. Since valid HTML always ends with a closed tag, nothing is held back at the end of the stream, eliminating any risk of truncation.

chance-coleman (dismissed):
LGTM

briantwatson:
If I'm tracking correctly, some apps that more aggressively rebuild pages (e.g. React, Vue) might appear behind the banner as the `place()` function only triggers on first load. Untested, but some sort of observer pattern for window Mutation might work:

>  1. Declare observer (after var H = '24px';):
  var observer;
 > 2. Disconnect at top of place(), right after the if (!top && !bottom) { return; } guard:
  if (observer) { observer.disconnect(); }
  > 3. Reconnect at end of place(), before its closing }:
  if (window.MutationObserver) {
    if (!observer) { observer = new MutationObserver(function () { place(); }); }
    observer.observe(b, { attributes: true, attributeFilter: ['style'] });
    observer.observe(document.documentElement, { childList: true });
  }
  
  Makes me think as the Classification banner use cases/needs become more complex, maybe adding a playwright test with a more complex or tailored app that offers more coverage in this sense would be helpful if we don't have one already.

slaskawi (dismissed):
LGTM

Tobyhb:
Could we change 'UNKNOWN' to either 'SAMPLE BANNER' or 'NOTIONAL'? Seeing 'UNKNOWN' implies that the environment's actual classification hasn't been determined, which can look like a compliance issue. 'SAMPLE BANNER' or 'NOTIONAL' makes it clear that it's a deliberate placeholder for the demo.

joelmccoy:
@greptileai review

joelmccoy:
@greptileai review

src/istio/common/chart/templates/classification-banner.yaml
```diff
@@ -147,32 +170,136 @@ spec:
...
+            end
+
+            -- Rewrite one complete body or one streamed fragment. State prevents duplicate
+            -- replacements when the response arrives in multiple chunks.
+            local function inject_banner(html, state)
+              local matches
+
+              if not state.head then
+                html, matches = html:gsub(
+                  "<head([^>]*)>",
```
chance-coleman: I believe that tag names in HTML are not case sensitive, so if an app uses `<HEAD>` the banner markup would be added but i think the CSS for grid and scroll container wouldnt be, causing some malformed layout.

src/istio/common/chart/templates/classification-banner.yaml
```diff
@@ -147,32 +170,136 @@ spec:
...
+                      pending = string.sub(data, lt)
+                    else
+                      chunk:setBytes(data)
+                      pending = ""
+                    end
+                  end
+                end
 
-                response_handle:body():setBytes(body_text)
+                local updated_headers = response_handle:headers()
```
chance-coleman: These header mutations happen after `bodyChunks()` has run. Envoy continues the response headers after the first streamed chunk, so these calls can fail with `header map can no longer be modified` on multi-chunk responses.

Move all response-header mutations before entering `bodyChunks()`.

src/istio/common/chart/templates/classification-banner.yaml
joelmccoy: do a case insensitive match for these tags now

src/istio/common/chart/templates/classification-banner.yaml
joelmccoy: moved

chance-coleman (approved):

PR #2714, opened 2026-06-02
feat: allow a list of principals to access keycloak admin API

bunchmj:
## Description

Adds a new adminApiAllowedPrincipals value to the Keycloak chart that lets operators allowlist specific Istio source principals (mTLS SPIFFE identities) to reach the Keycloak Admin API.

By default, the keycloak-block-admin-access-from-public-gateway AuthorizationPolicy denies Admin API access (/admin*, /realms/master*, and the clients endpoint) to all sources outside the admin gateway and pepr-system. This change renders a notPrincipals exclusion on those DENY rules for each configured principal, so trusted in-cluster workloads (e.g. an automation/CI service account) can be granted Admin API access without routing through the admin gateway.

Key points:
- New value adminApiAllowedPrincipals (array of strings), defaulting to [] — no behavioral change when unset.
- Exclusion is scoped only to the two Admin API DENY rules; the metrics (:9000) and mTLS catch-all rules are unaffected.
- Principal values are rendered with | quote so wildcard SPIFFE identities (e.g. */ns/argo/sa/foo) don't misparse as YAML aliases.
- values.schema.json updated to validate the new value.

## Related Issue

Fixes CORE-591

## Type of change

## Steps to Validate
1. Set the value, e.g.:
```
adminApiAllowedPrincipals:
  - "cluster.local/ns/argo/sa/workflow-service-account"
```
2. Deploy Keycloak and confirm the keycloak-block-admin-access-from-public-gateway AuthorizationPolicy renders notPrincipals on the /admin* and /admin/realms/<realm>/clients rules.
3. From a workload running as the configured principal, confirm the Admin API is reachable; confirm a workload with a different identity is still denied.
4. Run the chart unit tests: `uds run lint:helm-unittest` (or helm unittest src/keycloak/chart).

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/values.yaml
+++ b/src/keycloak/chart/values.yaml
@@ -23,6 +23,12 @@ additionalGatewayNamespaces: []
 # Example
 # - "istio-login-gateway"
 
+# Istio source principals (mTLS SPIFFE identities) allowed to reach the Keycloak Admin API,
+# in addition to the istio-admin-gateway and pepr-system namespaces. Entries are used verbatim.
+adminApiAllowedPrincipals: []
+# Example
+# - "cluster.local/ns/argo/sa/workflow-service-account"
+
 # Enables an EnvoyFilter to block path-parameter bypasses of AuthorizationPolicy rules
 # Should only be disabled if this is incorrectly blocking paths you expect to be allowed
 pathParameterProtection: true
--- a/src/keycloak/chart/values.yaml
+++ b/src/keycloak/chart/values.yaml
@@ -23,8 +23,7 @@ additionalGatewayNamespaces: []
 # Example
 # - "istio-login-gateway"
 
-# Istio source principals (mTLS SPIFFE identities) allowed to reach the Keycloak Admin API,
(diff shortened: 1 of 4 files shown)


joelmccoy:
@greptileai review

src/keycloak/chart/values.yaml
```diff
@@ -23,6 +23,12 @@ additionalGatewayNamespaces: []
 # Example
 # - "istio-login-gateway"
 
+# Istio source principals (mTLS SPIFFE identities) allowed to reach the Keycloak Admin API,
+# in addition to the istio-admin-gateway and pepr-system namespaces. Entries are used verbatim.
```
joelmccoy: nit: i would remove the second line of this comment as it could become stale
```suggestion
# Istio source principals (mTLS SPIFFE identities) allowed to reach the Keycloak Admin API
```

joelmccoy (commented):
Generally looking good. Could you just some very basic helm unit tests for this here: https://github.com/defenseunicorns/uds-core/tree/main/src/keycloak/chart/tests?

bunchmj:
@greptileai review

joelmccoy (approved):

PR #2752, opened 2026-06-18
feat(istio): add option to use post-quantum key exchange

lemmarathon:
## Description

Add an opt-in, default-off knob that forces post-quantum hybrid key exchange (X25519MLKEM768) across the Istio mesh by setting COMPLIANCE_POLICY=pqc on istiod and ztunnel. Supported only on the non-FIPS upstream flavor.

- src/istio/common/zarf.yaml: shared ISTIO_COMPLIANCE_POLICY chart variable on the istiod (pilot.env.COMPLIANCE_POLICY) and ztunnel (env.COMPLIANCE_POLICY) charts, no default so the unset case renders identically to main
- bundles/k3d-standard and k3d-slim-dev: expose the variable as a bundle override on both charts and document it in each bundle README
- docs: new reference page and networking how-to (accepted value pqc, default-off, ambient + upstream-only support matrix, force-everywhere caveats)
- test/vitest/istio-pqc.spec.ts: PQC force-mode scenario (gated by PQC_TESTS) asserting the negotiated group, classical-client rejection, and COMPLIANCE_POLICY=pqc on both workloads
- tasks and .github/workflows/test-pqc.yaml: deploy/test wiring for the separate upstream PQC scenario

Validated on k3d: the default-off suite is green (120 vitest + 15 playwright, 0 failed; no COMPLIANCE_POLICY env on either workload) and the PQC scenario is green (3/3 assertions; the tenant gateway negotiates X25519MLKEM768 and rejects classical-only clients).

## Type of change

## Steps to Validate

- Run the full e2e tests (no regressions)
- Run the PQC specific tests in `istio-pqc.spec.ts`

## Checklist before merging

### DIFF
--- a/.github/workflows/test-pqc.yaml
+++ b/.github/workflows/test-pqc.yaml
@@ -1,0 +1,46 @@
+# Copyright 2026 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+name: Test PQC
+
+# Runs the dedicated post-quantum cryptography (COMPLIANCE_POLICY=pqc) force-mode scenario on the
+# upstream flavor. Kept out of the standard test matrix because forcing PQC mesh-wide makes the
+# tenant gateway reject the non-PQC clients the standard suite relies on. Trigger manually, or call
+# it from an orchestrating workflow.
+on:
+  workflow_dispatch: {}
+  workflow_call: {}
+
+permissions:
+  contents: read
+  id-token: write # This is needed for OIDC federation.
+  packages: read # Allows reading the published GHCR packages
+
+jobs:
+  test-pqc:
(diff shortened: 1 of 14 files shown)


joelmccoy:
Thanks for this PR. This is something we probably want to consider adding into Core at some point!  However, we probably won't want to explore and test this until we can also get this feature tested for our unicorn images.  We are working with our secure image provider to get this working.

Additionally, the workflow additions are a probably a bit heavy in this PR for testing this particular feature and we might want to take a different approach. Our team will want to decide and create a design around how we will want enable/expose this feature to end users.  We are tracking this on our roadmap, but for now we are going to close this PR in preference of a different implementation!  

Link to internal tracking: https://linear.app/defense-unicorns/issue/CORE-635/investigate-addingtesting-istio-post-quantum-key-exchange-pqc-in-uds

PR #2760, opened 2026-06-22
feat: add UDP expose reconciliation with Envoy Gateway

chance-coleman:
## Description
Adds UDP ingress support to the Package CR's `spec.network.expose[]` via `protocol: HTTP | UDP`. UDP entries route through Envoy Gateway instead of Istio.

**Schema:**
- `protocol: HTTP | UDP` on expose entries
- `anyOf` restructured: `host` required only on HTTP branches
- Admission validation rejects UDP entries with HTTP-only fields (host, domain, advancedHTTP, match, uptime, podLabels)

**Operator:**
- New `udp-route-resources.ts` controller creates `UDPRoute` (Gateway API v1alpha2) and UDP ingress NetworkPolicy per expose entry
- Default mode: manages `Gateway/envoy-default-gateway` in `envoy-default-gateway` with port-based listeners, `allowedRoutes` scoped to declaring namespace
- User-managed mode: uses existing `gateway` field, generates UDPRoute + NetworkPolicy only
- Port conflict resolution: earliest `creationTimestamp` wins, ties broken by lexicographic `uid`
- Listener writes serialized through mutex queue
- Orphan purge for UDPRoutes and NetworkPolicies
- Status conditions: `EnvoyGatewayDefaultDisabled`, `UDPPortConflict` (Phase.Failed)
- Watches `Package/envoy-gateway` in `envoy-gateway-system` for availability flag
- Live fallback re-checks flag on restart
- Finalizer recomputes shared Gateway listener list on package deletion

**Istio/Network:**
- Istio resources controller skips UDP expose entries
- Network policy generator skips UDP entries for Istio gateway policies

**E2E:**
- UDP traffic test through default Envoy Gateway with resource assertions (Gateway listener, UDPRoute parentRef, NetworkPolicy)

## Related Issue
Relates to CORE-504, CORE-507

## Type of change

## Checklist before merging


chance-coleman:
@greptileai

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
```diff
@@ -0,0 +1,437 @@
...
+      await performDefaultGatewayReconciliation(packages);
+    }
+  })();
+
+  try {
+    await defaultGatewayReconcileInFlight;
+  } finally {
+    defaultGatewayReconcileInFlight = null;
+  }
+}
```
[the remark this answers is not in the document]
chance-coleman: Updated the serialized default Gateway reconciliation so only the first pass can use the caller-provided package snapshot. Dirty re-runs now fetch fresh package state, which handles finalizer-driven listener cleanup correctly.

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
```diff
@@ -0,0 +1,437 @@
...
+      !otherPkg.metadata?.deletionTimestamp &&
+      getUDPExposeEntries(otherPkg).some(
+        entry =>
+          resolveGatewayTarget(entry.expose).defaultMode && entry.expose.port === expose.port,
+      ),
+  );
+  const winner = conflictingPackages.sort(comparePackagePrecedence)[0];
+
+  return winner !== undefined && winner.metadata?.uid !== pkg.metadata?.uid;
+}
```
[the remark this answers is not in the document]
chance-coleman: Added validator coverage to reject duplicate default-mode UDP expose ports within the same package. User-managed gateway entries are still allowed to reuse the same port because they target a separate gateway/listener owner.

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -306,6 +342,27 @@ export async function packageFinalizer(pkg: UDSPackage) {
...
+      reason: "RemovalFailed",
+      type: "Warning",
+    });
+    await updateStatus(pkg, { phase: Phase.RemovalFailed });
+    return false;
+  }
+
   // Indicate success - all other resources (network policies, virtual services, etc) are cleaned up through owner references
   // See https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/#ownership-and-finalizers
   log.debug(`Package ${pkg.metadata?.namespace}/${pkg.metadata?.name} removed successfully`);
```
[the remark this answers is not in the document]
chance-coleman: Updated finalizer cleanup to force-reconcile remaining default-mode UDP packages instead of only recomputing shared Gateway listeners. That lets packages previously failed with UDPPortConflict recover when the winning package is deleted.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -63,6 +79,62 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+
+      if (!expose.gateway && defaultModeUDPPorts.has(expose.port!)) {
+        return req.Deny(
+          `Only one default-mode UDP expose entry can use port ${expose.port} in a package. ` +
+            `Use a different port or a user-managed gateway for additional UDP entries.`,
+        );
+      }
+      if (!expose.gateway) {
+        defaultModeUDPPorts.add(expose.port!);
+      }
```
joelmccoy: nit: Instead of only doing validation for default gateway... can't we just key of the gateway name (or default).  If someone provides expose entries on the same port for their custom gateway there will be a collision if they have separate descriptions... thinking something like this:
```ts
// Track (gateway, port) pairs to prevent conflicts on any gateway
const udpGatewayPorts = new Map<string, Set<number>>();

// ...inside the UDP block:
const gatewayKey = expose.gateway ?? "__default__";
const portsForGateway = udpGatewayPorts.get(gatewayKey) ?? new Set();
if (portsForGateway.has(expose.port!)) {
  return req.Deny(
    `Only one UDP expose entry can use port ${expose.port} on the same gateway. ` +
    `Use a different port or a different gateway.`,
  );
}
portsForGateway.add(expose.port!);
udpGatewayPorts.set(gatewayKey, portsForGateway);
```

Definitely non blocking, but a nice to have and implementation would be rather simple.

src/pepr/operator/crd/validators/package-validator.spec.ts
```diff
@@ -405,6 +419,172 @@ describe("Test validation of Package CRs", () => {
...
+        [],
+        [],
+      );
+      await validator(mockReq);
+      expect(mockReq.Deny).toHaveBeenCalledWith(
+        "podLabels cannot be set when protocol is UDP; use selector",
+      );
+    });
+
+    it("denies duplicate UDP expose entries that would generate the same UDPRoute name", async () => {
```
joelmccoy, line 517: this test is basically the same as the below test... but it actually doesn't test the udproute based on description. I think this test should be something like:
```ts
it("denies duplicate UDP expose entries that would generate the same UDPRoute name", async () => {
    const mockReq = makeMockReq(
      {},
      [
        { ...validUDPExpose, description: "udp-game", port: 7777 },
        { ...validUDPExpose, description: "udp-game", port: 8888 },
      ],
      [], [], [],
    );
    await validator(mockReq);
    expect(mockReq.Deny).toHaveBeenCalledWith(expect.stringContaining("duplicate UDPRoute"));
  });
```

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -306,6 +341,27 @@ export async function packageFinalizer(pkg: UDSPackage) {
     return false;
   }
 
+  // Recompute the shared default UDP Gateway after package removal. Owner references clean up
+  // generated UDPRoutes and NetworkPolicies, but the shared Gateway has no package owner.
+  try {
+    if (hasDefaultModeUDPExpose(pkg)) {
+      await retryWithDelay(async function cleanupEnvoyGatewayResources() {
+        await reconcileEnvoyGatewayDefaultPackages();
```
joelmccoy: Not a fan of the extra cycles of reconciling all packages here.  This is a side effect on the trigger of one package and should be avoided at all costs.  I'm pretty sure we can get rid of this by removing the check if egress gateway is deployed (see other comment) and blocking non unique udp port/gateway combinations at admission (see other comment in package validator logic). 

And then I believe this should just become:
```ts
  try {
    if (hasDefaultModeUDPExpose(pkg)) {
      await retryWithDelay(async function recomputeDefaultGateway() {
        await reconcileDefaultGatewayListeners();
      }, log);
    }
  }
```

src/pepr/operator/controllers/config/config.ts
```diff
@@ -39,6 +39,7 @@ export const UDSConfig: Config = {
   kubeApiCIDR: "",
   kubeNodeCIDRs: [],
   isIdentityDeployed: false,
+  isEnvoyGatewayDefaultEnabled: false,
```
joelmccoy: I don't think we need this config property and gate. We don't do this for the egress gateway.  The envoygateway will always be deployed before a package CR has a UDP expose entry.  The only case where it isn't is when they don't toggle the optional component and I'd rather just fail a package reconciliation in that case instead of skipping and coming back to reconciliation.

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -82,6 +88,25 @@ export async function packageReconciler(pkg: UDSPackage) {
...
+export async function forcePackageReconcile(pkg: UDSPackage) {
+  migrate(pkg);
+  await updateStatus(pkg, { phase: Phase.Pending, conditions: getReadinessConditions(false) });
+  try {
+    await reconcilePackageFlow(pkg);
+  } catch (err) {
+    await handleFailure(err, pkg);
+  }
+}
+
```
joelmccoy: similar to a comment later in this file. we should never trigger package reconciliations from the deletion of one package (it's a bad side effect and breaks the package/namespace ownership model). we should remove this and chose a different architecture. I proposed something in a later comment for something that I think should work

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -140,20 +168,27 @@ async function reconcilePackageFlow(pkg: UDSPackage): Promise<void> {
   // Create the CA Bundle Config Map if needed
   await caBundleConfigMap(pkg, namespace!);
 
-  await updateStatus(pkg, {
+  const envoyGatewayConditions = getEnvoyGatewayStatusConditions(envoyGatewayResult);
+  const readyStatus: StatusObject = {
     phase: Phase.Ready,
-    conditions: getReadinessConditions(true),
+    conditions: envoyGatewayConditions ?? getReadinessConditions(true),
```
joelmccoy: I strongly recommend against having special conditions for UDP failures. We can achieve this by simply allowing package reconciliation to fail.

If you remove the condition to "skip reconciliation when Envoy Gateway isn't deployed" and let the UDPRoute and Gateway creation fail naturally (the `Apply` command will throw an error because no GatewayClass exists), then:

* The `EnvoyGatewayDefaultDisabled` condition will be eliminated. The package will retry using standard error handling until the Envoy Gateway is available, similar to how egress functions.
* The `UDPPortConflict` we should handle at admission via the package (see my comment in the validator logic)

If we do that we don't have to have these weird special envoy gateway conditions.

src/pepr/operator/crd/validators/package-validator.ts
```diff
@@ -63,6 +79,62 @@ export async function validator(req: PeprValidateRequest<UDSPackage>) {
...
+      );
+      if (udpRouteNames.has(udpRouteName)) {
+        return req.Deny(
+          `The combination of characteristics of this expose entry would create a duplicate UDPRoute. ` +
+            `Verify you do not have duplicate values, or add a unique "description" field for this rule. ` +
+            `The duplicate rule would be named "${udpRouteName}".`,
+        );
+      }
+      udpRouteNames.add(udpRouteName);
+
```
joelmccoy, line 115: We also should block port conflicts across all packages. UDP Port + gateway combinations should be unique to the cluster. We keep information in the package store and perform denies in conflicts just like we do for sso client ids: https://github.com/defenseunicorns/uds-core/blob/ea32261b24ac23cb950de57176758ee623d04b5a/src/pepr/operator/crd/validators/package-validator.ts#L286-L297

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -99,7 +99,9 @@ export async function networkPolicies(pkg: UDSPackage, namespace: string, istioM
   // Generate NetworkPolicies for any VirtualServices that are generated
   const exposeList = pkg.spec?.network?.expose ?? [];
   // Iterate over each exposed service, excluding directResponse services
-  for (const expose of exposeList.filter(exp => !exp.advancedHTTP?.directResponse)) {
+  for (const expose of exposeList.filter(
+    exp => exp.protocol !== ExposeProtocol.UDP && !exp.advancedHTTP?.directResponse,
```
joelmccoy, line 111: wouldn't it just make sense to have the netpol generation from here: https://github.com/defenseunicorns/uds-core/blob/4cfb00eb6b8ef3d6a46a564998b5fe844db667e9/src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts#L118-L125 into this file instead of skipping udp?

I think you would also update the purge logic to be treated like a normal netpol purge and not the `"uds/managed-by":"envoy-gateway"` label

We already generate expose network policies in this logic and feels weird to exclude it here and handle it in another path unless I am missing some sort of limitation here.

src/pepr/operator/index.ts
```diff
@@ -115,6 +119,29 @@ When(UDSPackage)
...
+
+When(UDSPackage)
+  .IsDeleted()
+  .InNamespace("envoy-gateway-system")
+  .WithName("envoy-gateway")
+  .Watch(() => {
+    log.info("Envoy Gateway removed, operator will NOT handle default UDP expose entries.");
+    UDSConfig.isEnvoyGatewayDefaultEnabled = false;
+  });
+
```
joelmccoy: based on earlier comment. don't think we need to track if this is deployed. we don't do this for sidecar egress gateway for example.

src/test/tasks.yaml
```diff
@@ -24,7 +24,7 @@ tasks:
           validate_egress: ${{ .inputs.validate_egress }}
       - description: "Run Network Tests"
         cmd: |
-          EGRESS_TESTS=${{ .inputs.validate_egress }} npx vitest run "network.spec.ts"
+          EGRESS_TESTS=${{ .inputs.validate_egress }} UDP_EXPOSE_TESTS=${{ .inputs.validate_egress }} npx vitest run "network.spec.ts"
```
joelmccoy: I think we need to key these tests not on the same thing as egress. Maybe this should be something like:

```suggestion
          EGRESS_TESTS=${{ .inputs.validate_egress }} UDP_EXPOSE_TESTS=${{ .inputs.validate_udp_expose }} npx vitest run "network.spec.ts"
```

And make sure this is passed through on the proper ci workflows and tasks

src/test/tasks.yaml
```diff
@@ -71,6 +71,10 @@ tasks:
         if: ${{ eq .inputs.validate_egress "true" }}
         cmd: "uds zarf package deploy build/zarf-package-uds-core-test-apps-*.zst --confirm --components egress-app"
 
+      - description: Apply Envoy Gateway UDP test overlay
+        if: ${{ eq .inputs.validate_egress "true" }}
```
joelmccoy: similar to other comment. just make sure we pass this variable true in the right places
```suggestion
        if: ${{ eq .inputs.validate_udp_expose"true" }}
```

joelmccoy (commented):
First pass at this. I didn't get into the gateway reconciliation logic yet and haven't played with this locally yet, but wanted to provide this feedback as it will most likely change the architecture here a bit.

src/pepr/operator/reconcilers/package-reconciler.ts
chance-coleman: Removed package re-reconciliation triggered from another Package’s deletion

src/pepr/operator/controllers/config/config.ts
chance-coleman: Removed the cached/default-enabled config path

src/pepr/operator/reconcilers/package-reconciler.ts
chance-coleman: Removed the forced re-reconcile path. Finalizer now only cleans up UDP gateway NetworkPolicies and recomputes the shared default Gateway listeners when the deleted Package used default UDP expose.

src/pepr/operator/crd/validators/package-validator.spec.ts
chance-coleman: Updated the test to use the same `description` with different ports so it actually covers duplicate UDPRoute name generation.

src/pepr/operator/crd/validators/package-validator.ts
chance-coleman: UDP expose validation now keys uniqueness by `(gateway, port)` using `getUDPGatewayPortKey()`. Same port on different gateways is allowed. Same gateway and port is denied within the package.

src/pepr/operator/reconcilers/package-reconciler.ts
chance-coleman: Removed `EnvoyGatewayDefaultDisabled` and `UDPPortConflict`. Port conflicts are denied at admission.

src/pepr/operator/crd/validators/package-validator.ts
chance-coleman: Added a PackageStore index for UDP `(gateway, port)` and the validator now denies conflicts across namespaces

src/pepr/operator/controllers/network/policies.ts
chance-coleman: UDP expose NetworkPolicies now come from `networkPolicies()`

src/pepr/operator/controllers/network/policies.ts
```diff
@@ -46,6 +51,39 @@ export function findMatchingClient(pkg: UDSPackage, podLabels: Record<string, st
...
+}
+
+function getUDPGatewaySelector(gateway?: string) {
+  return {
+    "gateway.envoyproxy.io/owning-gateway-name": getUDPGatewayName(gateway),
+    "gateway.envoyproxy.io/owning-gateway-namespace": getUDPGatewayNamespace(gateway),
+  };
+}
+
+async function purgeUDPGatewayNetworkPolicies(
```
joelmccoy: do we really need a special function and labels to clean up for this? Can they be cleaned up with the generic generation id cleanup?

joelmccoy: Oh I think I see what is happening now... We create netpols in the envoy-gateway namespace with a package CR... This is because the envoy-gateway has a UDS Package and there is a default deny generated there (different from istio gateways).

But I would probably vote for just adding the netpols to the envoy-gateway-package CR so we don't have to manage them on a per package basic.  We basically need to allow all traffic internally anyways like we do for the istio gateways. I think we can just add them like so and we can get rid of all this logic and special labels?

```yaml
# Allow external UDP traffic into managed Envoy Gateway proxies
  - direction: Ingress
    selector:
      app.kubernetes.io/managed-by: envoy-gateway
    remoteGenerated: Anywhere
    remoteProtocol: UDP
    description: "UDP ingress for managed Gateway proxies"

  # Allow managed Envoy Gateway proxies to forward UDP to backend services
  - direction: Egress
    selector:
      app.kubernetes.io/managed-by: envoy-gateway
    remoteGenerated: Anywhere
    remoteProtocol: UDP
    description: "UDP egress for managed Gateway proxies to backends"
```

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -306,6 +321,35 @@ export async function packageFinalizer(pkg: UDSPackage) {
...
+  // generated UDPRoutes, but the shared Gateway has no package owner.
+  try {
+    const hasUDPExpose = pkg.spec?.network?.expose?.some(
+      expose => expose.protocol === ExposeProtocol.UDP,
+    );
+    const hasDefaultUDPExpose = hasDefaultModeUDPExpose(pkg);
+
+    if (hasUDPExpose) {
+      await retryWithDelay(async function cleanupEnvoyGatewayUDPResources() {
+        await cleanupUDPGatewayNetworkPolicies(pkg);
```
joelmccoy: relevant to what i mentioned earlier. if we get rid of the network policy generation in the envoy-gateway namespace we can let owner references just clean up the network policies.

src/pepr/operator/index.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2024 Defense Unicorns
+ * Copyright 2024-2026 Defense Unicorns
```
joelmccoy: nit: nothing changed in the file, i think we can keep this as 2024?

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
```diff
@@ -0,0 +1,280 @@
...
+  const ownerRefs = getOwnerRef(pkg);
+  const udpEntries = getUDPExposeEntries(pkg);
+
+  if (udpEntries.length === 0) {
+    const hasGeneratedUDPResources = await hasGeneratedUDPRoutes(namespace, pkgName);
+
+    if (!hasGeneratedUDPResources) {
+      return;
+    }
+  }
```
joelmccoy: I'm confused by this block. I don't think we need it at all. When `udpEntries` is empty, the for loop does nothing, and `purgeOrphans` already takes care of cleaning up stale `UDPRoutes` from previous generations.

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  await purgeOrphans(generation, namespace, pkgName, K8sUDPRoute, log);
+
+  await reconcileDefaultGatewayListeners();
+}
+
+export async function reconcileDefaultGatewayListeners(packages?: UDSPackage[]): Promise<void> {
+  if (defaultGatewayReconcileInFlight) {
+    defaultGatewayReconcileDirty = true;
```
joelmccoy: Can we replace this with the `Mutex` in `controllers/utils.ts`? I feel like a mutex is sufficient here (unless I am missing something other than a speed optimation). As written I believe there's a race condition where a caller can sneak in after the while loop exits but before `inFlight` is set to null. That caller sees the promise, sets dirty, awaits it, and returns immediately since it has already resolved, leaving it under the impression that someone handled its update. In reality, no one did. A simple mutex prevents this issue and simplifies the code a bit.  I don't think we get much speed optimization here worth the complexity.

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
```diff
@@ -0,0 +1,280 @@
...
+            },
+          },
+        })),
+      },
+    },
+    { force: true },
+  );
+}
+
+async function getWinningDefaultListenerEntries(
```
joelmccoy: I think since we block at the package admission this function isn't calculating "winners" any more. I think it should be called something more like this?:
```suggestion
async function getDefaultListenerEntries(
```

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
```diff
@@ -0,0 +1,280 @@
...
+  for (const entry of entries) {
+    const currentEntry = winners.get(entry.expose.port!);
+    if (currentEntry) {
+      throw new Error(
+        `UDP expose port ${entry.expose.port} on default Envoy Gateway is already used by ` +
+          `${currentEntry.namespace}/${currentEntry.pkg.metadata?.name}.`,
+      );
+    }
+    winners.set(entry.expose.port!, entry);
+  }
```
joelmccoy: I think since we block at package admission we can remove this winner logic from here if I am not mistaken?

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
```diff
@@ -0,0 +1,280 @@
...
+    );
+    await K8s(K8sUDPRoute).Apply(udpRoute, { force: true });
+  }
+
+  await purgeOrphans(generation, namespace, pkgName, K8sUDPRoute, log);
+
+  await reconcileDefaultGatewayListeners();
+}
+
+export async function reconcileDefaultGatewayListeners(packages?: UDSPackage[]): Promise<void> {
```
joelmccoy: do we need the `packages?: UDSPackage[]` argument here? We list all packages in the reconciliation workflow anyways and I don't think there is a case where we only want to reconcile based on a subset of packages in the cluster??

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
```diff
@@ -0,0 +1,280 @@
...
+    if (e?.status !== 404) {
+      throw e;
+    }
+    throw new Error(
+      "GatewayClass 'envoy-gateway' was not found. Ensure Envoy Gateway is deployed before using default UDP expose entries.",
+    );
+  }
+}
+
+async function performDefaultGatewayReconciliation(packages?: UDSPackage[]): Promise<void> {
```
joelmccoy: do we need the `packages?: UDSPackage[]` argument here? We list all packages in the reconciliation workflow anyways and I don't think there is a case where we only want to reconcile based on a subset of packages in the cluster??

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
```diff
@@ -0,0 +1,280 @@
...
+      },
+    },
+    { force: true },
+  );
+}
+
+async function getWinningDefaultListenerEntries(
+  packages?: UDSPackage[],
+): Promise<{ pkg: UDSPackage; expose: Expose; namespace: string }[]> {
+  const packageItems = packages ?? (await K8s(UDSPackage).Get()).items;
```
joelmccoy: do we need the packages?: UDSPackage[] argument here? We list all packages in the reconciliation workflow anyways and I don't think there is a case where we only want to reconcile based on a subset of packages in the cluster??

I think every time we generate the listeners for the gateway it is based off of the state of the packages in the cluster?

joelmccoy: The cluster-wide `K8s(UDSPackage).Get()` method for rebuilding the Gateway on every reconcile differs from our approach to other shared resources. 

Shared egress and blackbox configurations follow a pattern where each package manages its own portion of the shared state, either through an in-memory cache or a scoped read/modify/write. Could we apply the same approach here? In this case I think we should use an in-memory map of `port → namespace/expose`, allowing each package to add or remove its own entries and then rebuild the Gateway from this map. Since admission already ensures port uniqueness, no conflict resolution is necessary. This method avoids the need for a cluster-wide list, prevents packages from interfering with each other's state.  There is a case where two packages (A - B) are applied at the near same time and this current logic will apply B's listener in A's reconciliation.  Instead we should probably have tighter ownership here of the listeners on a per package basis.

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -156,35 +156,44 @@ const allow = {
...
       {
-        required: ["service", "selector", "port"],
+        required: ["host", "service", "selector", "port"],
       },
       {
-        required: ["advancedHTTP"],
+        required: ["host", "advancedHTTP"],
+      },
+      {
+        required: ["protocol", "service", "selector", "port"],
```
joelmccoy, line 174: nit: This branch matches when the protocol is present but does not restrict it to "UDP." Therefore, `protocol: HTTP` without a host would technically pass the schema. While this isn't a significant risk since the webhook validator catches invalid combinations, you could make it more precise by using `properties: { protocol: { const: "UDP" } }` if you want the schema to be exact.

joelmccoy (commented):
Mostly looking good here, but have a couple more comments/suggestions.  Once these are addressed I would like to play around with this a bit more locally to test.

src/pepr/operator/controllers/network/policies.ts
chance-coleman: moved the ingress-from-anywhere/egress-to-backends UDP rules onto the envoy-gateway package's own Allow rules (matching your suggested snippet)

src/pepr/operator/reconcilers/package-reconciler.ts
chance-coleman: finalizer no longer touches netpols at all now that nothing gets created outside the package's own namespace

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
chance-coleman: You're right, purgeOrphans handles it regardless.

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
chance-coleman: Replaced with Mutex from controllers/utils.ts. Added a test that exercises two concurrent reconcileDefaultGatewayListeners() calls with a map mutation in between to confirm the second (queued) call reflects current state

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
chance-coleman: it now just flattens the in-memory map, no conflict handling needed since admission blocks duplicates

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
chance-coleman: Removed from all three call sites

src/pepr/operator/controllers/envoy-gateway/udp-route-resources.ts
chance-coleman: Swapped the cluster-wide Get() for an in-memory map keyed by package, same pattern egress.ts already uses.

src/pepr/operator/crd/sources/package/v1alpha1.ts
chance-coleman: Tried this one but it breaks the codegen, tightening that property collapses the shared `ExposeProtocol` enum down to just UDP and drops HTTP entirely, which would've broken the validator. Leaving it as is since the webhook already catches this case like you said.

joelmccoy:
@greptileai review

joelmccoy:
On local testing, I found a couple of things that need to be addressed:

1. The actual `envoy-gateway` pod is deployed in `envoy-gateway-system`. Currently, the Gateway CR is in `envoy-default-gateway`, but its proxy Deployment and Service are created in `envoy-gateway-system`. Moving the data plane into the Gateway namespace provides cleaner isolation and aligns with our Istio patterns. We’ll also need to update the related NetworkPolicies, as the ones deployed by the package CR may not be useful after the move. I believe this requires an override to our `envoy-gateway` package. We need to ensure that this change does not interfere with people's custom gateway deployments.

2. We need to test the end-to-end workflow for the expose path. As I understand it, we are currently only testing the network path from within the cluster, which does not effectively test this feature. It never actually reaches the MetalLB LoadBalancer from outside the cluster. This means we're testing the Envoy proxy routing but not the complete ingress path that a real consumer would use. We should pass `K3D_EXTRA_ARGS` through `uds-core-e2e` so we can include `-p "5000:5000/udp@server:*"` when `validate_udp_expose` is enabled. The plumbing already exists; `uds-core` accepts `K3D_EXTRA_ARGS` and passes it to `deploy:k3d-standard-bundle`, but `uds-core-e2e` does not currently forward it. Once the port is mapped, this test could send UDP to `localhost:5000` from the `vitest` process itself (no `execInPod` needed) and verify the response end-to-end through MetalLB → Envoy → backend.  This path will also need validated e2e for IAC workflows where an actual load balancer is provisioned.  We need to update our testing path so we test the full expose path.

src/envoy-gateway/chart/templates/uds-package.yaml
```diff
@@ -26,12 +26,16 @@ spec:
...
-        description: "Intra-namespace communication for managed proxies and cert generation"
+        description: "Intra-namespace communication for cert generation"
       - direction: Ingress
         selector:
           control-plane: envoy-gateway
-        remoteGenerated: IntraNamespace
-        description: "Intra-namespace communication from managed proxies"
+        # Managed proxies now live in their own Gateway's namespace (default or
+        # user-managed), not alongside the controller, so this can't be scoped to
+        # IntraNamespace. Restricted to the xDS port to limit exposure.
```
joelmccoy, line 35: nit: weird AI comment that can probably be removed. the fact that is says NOW is definitely a comment on the current session and working tree rather than documenting behavior (and a bit confusing)

chance-coleman:
@greptileai review

chance-coleman:
> On local testing, I found a couple of things that need to be addressed:...

@joelmccoy Moved proxies into their own Gateway's namespace (matches how Istio does it). For custom gateways, that namespace just needs to be Zarf-managed + mesh-enrolled. same other workload, not new complexity.

K3D_EXTRA_ARGS is wired through, and the new e2e test sends UDP through the real external path now (not just pod-to-pod). Had to route it through the node's NodePort since MetalLB's LB IP isn't reachable outside docker on k3d.

.github/workflows/test.yaml
```diff
@@ -70,7 +70,16 @@ jobs:
 
       - name: Test UDS Core Install
         if: ${{ inputs.package == 'all' && inputs.test_type == 'install' }}
-        run: uds run test:uds-core-e2e --set FLAVOR=${{ inputs.flavor }} --set EXCLUDED_PACKAGES=${{ inputs.flavor == 'registry1' && 'portal' || '' }} --no-progress
+        # MetalLB's LoadBalancer IP isn't reachable from outside the docker network on
+        # k3d, but kube-proxy's NodePort forwarding works directly on the k3d node's own
+        # address, which is reachable from this runner (same docker host). udp_expose_lb_
+        # use_nodeport routes the UDP expose e2e test through that real external path
+        # instead of just pod-to-pod traffic in-cluster.
```
joelmccoy: nit: weird over explained AI comment. you can call this out but maybe put it next to the `udp_expose_lb_use_nodeport=true` flag as this comment is specific to that

.github/workflows/test.yaml
```diff
@@ -70,7 +70,16 @@ jobs:
...
+        run: |
+          uds run test:uds-core-e2e --set FLAVOR=${{ inputs.flavor }} --set EXCLUDED_PACKAGES=${{ inputs.flavor == 'registry1' && 'portal' || '' }} \
+            --with validate_udp_expose_lb=true \
+            --with udp_expose_lb_use_nodeport=true \
```
joelmccoy: I'm wondering if we should just default these to true like we do for egress tests? Is there any reason we can't?

.github/workflows/test.yaml
```diff
@@ -70,7 +70,16 @@ jobs:
 
       - name: Test UDS Core Install
         if: ${{ inputs.package == 'all' && inputs.test_type == 'install' }}
-        run: uds run test:uds-core-e2e --set FLAVOR=${{ inputs.flavor }} --set EXCLUDED_PACKAGES=${{ inputs.flavor == 'registry1' && 'portal' || '' }} --no-progress
+        # MetalLB's LoadBalancer IP isn't reachable from outside the docker network on
+        # k3d, but kube-proxy's NodePort forwarding works directly on the k3d node's own
```
joelmccoy: The nodeport feels like a lot of extra plumbing and doesn't test the actual LB -> EnvoyGateway flow which is super important for this feature.  

Can't we just use the metallb ip of the envoy-gateway on the docker network?  I tested this locally and it seems to work fine.  Docker networking ips from the metallb service for the envoy gateway should be reachable. I tested this locally and you can see what i did to make this work: https://github.com/defenseunicorns/uds-core/compare/chance/core-506...joel-udp-testing?expand=1

src/envoy-gateway/default-gateway/templates/uds-package.yaml
joelmccoy, line 1: this should be in its own zarf component and helm chart in the zarf package as it deploys to a different namespace than the envoy-gateway controller.  We shouldn't have to create namespaces in our template files. Zarf will handle this is we specify things correctly.

This needs to be its own zarf component with namespace: `envoy-default-gateway`, not part of the `uds-envoy-gateway-config` chart that deploys into `envoy-gateway-system`. This is the same pattern as the istio-gateways.

test/vitest/network.spec.ts
```diff
@@ -113,6 +273,44 @@ function expectUdpPingLog(log: string, message: string) {
 
 // Check if egress tests should run
 const runEgressTests = process.env.EGRESS_TESTS === "true";
+const runUDPExposeTests = process.env.UDP_EXPOSE_TESTS === "true";
+// Only true when the environment can actually reach the default Gateway through a real
+// external path: on k3d that's the node's own address plus the Service's NodePort (see
+// UDP_EXPOSE_LB_USE_NODEPORT, since MetalLB's LoadBalancer IP isn't reachable outside
+// the docker network); on a real cloud cluster (EKS/AKS) the provisioned LoadBalancer
+// address is reachable directly. This lets the test exercise the real external path
+// instead of just pod-to-pod traffic in-cluster.
```
joelmccoy: this comment is wrong.  with the nodeport setup we are bypassing metallb.  suggest moving away from nodeport however.

test/vitest/network.spec.ts
```diff
@@ -72,7 +228,11 @@ async function clearUdpLog(serverPodName: string): Promise<void> {
   await execInPod("curl-ns-udp-server", serverPodName, "udp-echo-server", [
     "sh",
     "-c",
-    "> /tmp/udp.log",
+    // busybox nc's UDP listener locks onto whichever peer sends the first datagram for
+    // its entire process lifetime, ignoring any other peer thereafter. Killing it forces
+    // the server's respawn loop to bind a fresh listener so each test gets its own peer,
+    // instead of silently losing to whichever earlier test (or client pod) contacted it first.
+    "> /tmp/udp.log; pkill -f 'nc -u -l -p 5000' || true; sleep 0.2",
```
joelmccoy: Looking at these e2e tests again and I am thinking we should take another approach for verification. This feels fragile for a couple reasons (and may produce false positives).

The verification process is one-directional: send UDP and then check a shared log file on the server to see if the message arrived. This is a bit fragile especially since you check the same file other tests are running in and can produce a false positive. You also need to stop the `nc` listener between tests, truncate the log, wait, and hope there’s no race with stale data.

Can we make the server echo back what it receives and check the response on the sender side?
In the end, we should check the response from the UDP packet for end-to-end verification. I tested something locally, and it worked. Can we implement the following change:

Modify the server command from `nc -u -l -p 5000 >> /tmp/udp.log` to `mkfifo /tmp/fifo; while true; do nc -u -l -p 5000 < /tmp/fifo | tee -a /tmp/udp.log > /tmp/fifo; done`. The FIFO loops the output of `nc` back to its stdin, allowing it to echo the received data back to the sender. I tested this on a live cluster and confirmed that the round-trip works through Envoy's UDP proxy.

With this approach, the tests can verify the response on the sender side instead of polling a server-side log file. This eliminates the need for the `clearUdpLog`, `waitForUdpLog`, and `pkill` hacks.

You can look here to see how I got this working: https://github.com/defenseunicorns/uds-core/compare/chance/core-506...joel-udp-testing?expand=1

test/vitest/envoy-gateway.spec.ts
```diff
@@ -48,13 +49,36 @@ async function createNamespace(): Promise<void> {
...
+      body: {
+        metadata: { name: "private-registry" },
+        type: source.type,
+        data: source.data,
+      },
+    });
+  } catch (error) {
+    if (!isConflict(error)) throw error;
+  }
 }
```
joelmccoy, line 82: this is a code smell and should go away once we have a separate zarf component for `envoy-default-gateway` (see previous comment)

tasks/test.yaml
```diff
@@ -171,6 +237,9 @@ tasks:
       - task: e2e-tests
         with:
           architecture: "amd64"
+          validate_egress: "false"
```
joelmccoy: why are we changing this to not validate egress here? we definitely want to test egress in our IAC envs

tasks/test.yaml
```diff
@@ -157,12 +187,48 @@ tasks:
...
+          external LoadBalancer address, rather than just pod-to-pod traffic in-cluster.
+          On k3d, pair with udp_expose_lb_use_nodeport="true" to reach it via the k3d
+          node's own address and the Service's NodePort instead (see e2e-tests).
+        default: "false"
+      udp_expose_lb_host:
+        description: "Overrides the target host for validate_udp_expose_lb (see e2e-tests)"
+        default: ""
+      udp_expose_lb_use_nodeport:
+        description: "For validate_udp_expose_lb on k3d (see e2e-tests)"
+        default: "false"
```
joelmccoy: I think we can remove all of these when moving away from the nodeport

tasks/test.yaml
```diff
@@ -95,6 +95,34 @@ tasks:
...
+          if the test runner needs a different address to reach it.
+        default: ""
+      udp_expose_lb_use_nodeport:
+        description: >-
+          For validate_udp_expose_lb on k3d: MetalLB's LoadBalancer IP isn't reachable
+          from outside the docker network, but kube-proxy's NodePort forwarding works
+          directly on the k3d node's own address (reachable from the test runner on the
+          same docker host). Setting this targets the node address at the Service's
+          dynamically-assigned NodePort instead of the LoadBalancer address.
+        default: "false"
```
joelmccoy: I'm pretty sure we can get rid of all these variables here when switching just to use the metallb ip on the docker network

tasks/test.yaml
```diff
@@ -95,6 +95,34 @@ tasks:
         description: "System architecture that the test-apps package should be built for."
         required: true
         default: ${UDS_ARCH}
+      validate_egress:
+        description: Whether to validate egress controls with the egress gateway
+        default: "true"
+      validate_udp_expose:
+        description: Whether to validate UDP expose routes through the default Envoy Gateway
+        default: "true"
```
joelmccoy, line 103: i wonder if we can just leave these off the top level and set them in the individual tasks... I'm a bit lost in the sauce here of tracing what we have set in each of the test workflows

test/vitest/network.spec.ts
```diff
@@ -614,14 +815,70 @@ test("UDP NetworkPolicy - custom allow and deny", { retry: 2, timeout: 60000 },
...
+    const receivedLog = await waitForUdpLog(udpServerPodName, "ping", 30000, 1000);
+
+    expectUdpPingLog(
+      receivedLog,
+      `UDP Envoy Gateway expose: log="${receivedLog}" nc="${send.stdout}" target="${serviceTarget}"`,
+    );
+  },
+);
+
+(runUDPExposeLBTest ? test : test.skip)(
```
joelmccoy: This should run all the time for `runUDPExposeTests` and should have the exact same behavior if in k3d or cloud. don't need this extra variable IMO runUDPExposeLBTest (called out in the task files). 

Essentially the full e2e path through the load balancers should always run even in cloud environments (super important).

test/vitest/network.spec.ts
```diff
@@ -614,14 +815,70 @@ test("UDP NetworkPolicy - custom allow and deny", { retry: 2, timeout: 60000 },
...
+  "UDP expose reaches the backend through the real external LoadBalancer path",
+  { retry: 2, timeout: 300000 },
+  async () => {
+    await assertDefaultEnvoyUDPResources(5000);
+    await clearUdpLog(udpServerPodName);
+
+    let host: string;
+    let port: number;
+    if (process.env.UDP_EXPOSE_LB_USE_NODEPORT === "true") {
```
joelmccoy: like commented earlier. we should get rid of this variable and not use nodeport

src/pepr/operator/controllers/envoy-gateway/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+ * Copyright 2026 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
+
+export const envoyDefaultGatewayName = "envoy-default-gateway";
+export const envoyDefaultGatewayNamespace = "envoy-default-gateway";
+export const envoyGatewaySystemNamespace = "envoy-gateway-system";
+
+export function getUDPGatewayPortKey(gateway: string | undefined, port: number): string {
+  return `${gateway ?? envoyDefaultGatewayName}:${port}`;
```
joelmccoy: The constants helpers use `??` (nullish), while `resolveGatewayTarget` uses truthiness (if `expose.gateway`). If the gateway is `""`, routing defaults to the standard gateway, but the port key resolves to `":port"`, which bypasses conflict detection. The same fix is needed on `resolveGatewayTarget`.lines 15 and 19
```suggestion
  return `${gateway || envoyDefaultGatewayName}:${port}`;
```

src/pepr/operator/controllers/envoy-gateway/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export function getUDPGatewayName(gateway: string | undefined): string {
+  return gateway ?? envoyDefaultGatewayName;
```
joelmccoy:
```suggestion
  return gateway || envoyDefaultGatewayName;
```

src/pepr/operator/controllers/envoy-gateway/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export function getUDPGatewayNamespace(gateway: string | undefined): string {
+  return gateway ?? envoyDefaultGatewayNamespace;
```
joelmccoy:
```suggestion
  return gateway || envoyDefaultGatewayNamespace;
```

src/pepr/operator/controllers/packages/package-store.ts
joelmccoy: nit: stale comment. does this for two other maps

src/pepr/operator/reconcilers/package-reconciler.ts
```diff
@@ -306,6 +317,28 @@ export async function packageFinalizer(pkg: UDSPackage) {
     return false;
   }
 
+  // Recompute the shared default UDP Gateway after package removal. Owner references clean up
+  // generated UDPRoutes and NetworkPolicies, but the shared Gateway has no package owner.
+  try {
+    if (hasDefaultModeUDPExpose(pkg)) {
+      await retryWithDelay(async function recomputeDefaultGateway() {
+        removeDefaultListenerMapEntry(pkg);
```
joelmccoy: nit: this can probably be removed from the retryWithDelay block. it's an in memory action

joelmccoy (commented):
Looking pretty good but a couple more overarching things I discovered in local testing we probably want to fix:
1. The `envoy-default-gateway` should be a separate zarf component and helm chart with the namespace declared in zarf (not in raw k8s manfests)
2. We should avoid using nodeport for the tests. This skips the loadbalancer and defeats the path we actually want to test.
3. The e2e tests should probably swap to a call/response verification so that they are less likely to produce false positives the way they are written.

test/vitest/envoy-gateway.spec.ts
chance-coleman: I've done that restructuring now, but I don't see how it connects to this test specifically: the test creates a completely different namespace (envoy-gateway-e2e, simulating a custom user Gateway), not envoy-default-gateway. Restructuring the default Gateway's namespace doesn't change what this test needs to bootstrap its own separate one.

joelmccoy: ahhh i see this now. gotcha. I think this is fine then (or we have this namespace deployed as a part of the test zarf package so we don't have to do this weird secret copy).

joelmccoy (commented):
Looking pretty good. Playing around with this a bit more locally. But we should also make a couple docs updates:

1. https://github.com/defenseunicorns/uds-core/blob/main/docs/how-to-guides/networking/configure-non-http-ingress.mdx#L18
   States _"UDP ingress is not currently supported by Istio,"_ which currently reads as _"UDS Core can't do UDP."_ This should include a callout directing users to the native UDP expose path through Envoy Gateway and clarify that this guide only covers **TCP over Istio**.

2. https://github.com/defenseunicorns/uds-core/blob/main/docs/how-to-guides/networking/expose-apps-on-gateways.mdx

   There is no mention of the `protocol` field or UDP exposes. At a minimum, this guide should include a section describing protocol selection or link to documentation covering UDP exposure via Envoy Gateway.

3. New how-to guide for UDP expose: enable the components, create a Package CR with protocol: UDP, verify. Cover default-Gateway mode vs user-managed mode and the security implications (no Istio mTLS/AuthorizationPolicy).  

We also need to add the new optional component in our IAC bundles like in the latest commit.  We want to definitely verify the e2e tests on all our IAC before merging this in

joelmccoy (commented):
Ok - happy with my testing locally. validated with multiple UDP servers and packages and the reconciliation and expose network paths work great.  However the e2e tests still need a small modification to make our UDP server resiliient. We should not rely on killing the process on the pod to reset.  This is very simple with the `-w 1` on nc.  I made these changes locally to validate a more resilient udp server.  I would recommend incorporating these: https://github.com/defenseunicorns/uds-core/compare/chance/core-506...joel-udp-2?expand=1

After that is done and:
1. Docs updated
2. IAC bundles updated with new optional component and tests run

We should be good to get this in.

docs/how-to-guides/networking/expose-udp-apps.mdx
```diff
@@ -0,0 +1,190 @@
...
+   spec:
+     network:
+       allow:
+         - direction: Ingress
+           selector:
+             app.kubernetes.io/name: my-udp-app
+           remoteGenerated: Anywhere
+           remoteProtocol: UDP
+           description: "UDP ingress from the managed Envoy Gateway proxy"
+   ```
```
joelmccoy: do we need this?? doesn't expose generate all the netpols for this?

chance-coleman: good catch, you're right. removed this

joelmccoy (approved):
LGTM. Awesome work on this one

PR #2773, opened 2026-06-25
chore(CORE-606): bump keycloak limits

slaskawi:
## Summary

This PR increases the Keycloak Limits which makes the bootstrapping process faster. Cuts roughly ~1m from the CI job. 

## Related tickets

CORE-606

### DIFF
--- a/src/keycloak/chart/values.yaml
+++ b/src/keycloak/chart/values.yaml
@@ -220,8 +220,8 @@ resources:
     cpu: "100m"
     memory: "700Mi"
   limits:
-    cpu: "1"
-    memory: "1Gi"
+    cpu: "2"
+    memory: "2Gi"
 
 # Persistence settings for PVC management (when enabled=false, an emptyDir will be used)
 # `devMode` overrides the enabled flag for `data` and `themes` to allow hot-reloads of the theme/plugin in k3d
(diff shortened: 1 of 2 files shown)


chance-coleman (approved):

joelmccoy (commented):
~don't we want to bump this for the standard bundle too? https://github.com/defenseunicorns/uds-core/blob/main/bundles/k3d-standard/uds-bundle.yaml~

nvm I see that we don't set the variables there.

src/keycloak/chart/values.yaml
```diff
@@ -220,8 +220,8 @@ resources:
     cpu: "100m"
     memory: "700Mi"
   limits:
-    cpu: "1"
-    memory: "1Gi"
+    cpu: "2"
+    memory: "2Gi"
```
joelmccoy, line 224: hmmmm. setting this could have an impact on people that don't override the limits for their environment (causing a sudden spike in usage?)... I think realistically we shouldn't have set these limits in default helm values in the first place.  

But since they are already there I think this change is fine and technically wouldn't be a breaking change? People just might notice a memory spike on their keycloak pods (but that ultimately may be a good thing).

All that to say I think this is fine, but wanted to spell out my concerns.

joelmccoy (approved):
LGTM. Left a comment on some concerns here, but I think they are minimal and acceptable.

src/keycloak/chart/values.yaml
slaskawi: I don't think this matters to be honest as these are the limits - the upper bound. These get allocated only if the Node has sufficient capacity. The requests stayed the same, so highly compacted nodes should not notice the difference (as they are very likely to operate on the lower bound). 

To sum it up - yes, I agree - customers may see increased memory and cpu usage but that's intentional and will happen only if there's some spare capacity in the cluster.

PR #2774, opened 2026-06-26
feat(portal): mount trust bundle via config chart

cross-du:
## Description

Mounts the UDS trust bundle into the Portal pod and moves the Portal `Package` CR into a Core-owned `uds-portal-config` chart.

This follows the config-chart pattern used by other Core apps so Core owns UDS-specific package configuration separately from the upstream Portal chart.

### Changes

- Adds `uds-portal-config` chart for the Portal `Package` CR.
- Disables the upstream Portal chart `Package` CR.
- Mounts `uds-trust-bundle` into Portal with `extraVolumes` and `extraVolumeMounts`.
- Bumps Portal chart/image refs to `v0.4.0` / `0.4.0`.

## Related Issue

Fixes [CORE-620](https://linear.app/defense-unicorns/issue/CORE-620/add-uds-trust-bundle-mount-to-portal-helm-chart)

## Type of change

## Steps to Validate
- `helm lint src/portal/chart`
- Rendered config chart and confirmed it creates the `Package` CR.
- Rendered Portal app chart and confirmed upstream `Package` is disabled.
- Rendered Portal app chart and confirmed trust bundle volume/mount exists.
- Private PKI deploy confirmed Portal pod has `uds-trust-bundle` mounted.
- `uds run -f src/portal/tasks.yaml validate`

## Checklist before merging


chance-coleman:
@greptileai review

src/portal/chart/templates/uds-package.yaml
joelmccoy, line 1: why are we adding the package CR here? the chart ships with this. not sure if I am missing some context here?

chance-coleman: yeah...the Package CR is what drives trust bundle creation, but the portal chart also deploys the pod. that lets the pod start before the trust bundle exists, and the mount is optional so it won’t block startup. moving the Package into a Core config chart follows the Grafana/Loki/etc pattern and lets Core create Package/trust-bundle config before the app chart deploys Portal.

chance-coleman: if we don’t want to move Package ownership in this PR, the fallback is to keep the Package in the portal chart and add an explicit rollout restart after deploy. that should fix runtime cert pickup, but it doesn’t solve the underlying ordering issue as cleanly as the config-chart pattern.

joelmccoy: I see the issue, but I don't think overriding the package `CR` is the right choice. The package `CR` is part of the upstream repository, and we would encounter drift if things change.

Here are some better options I’m considering:

1. Instruct the downstream consumers to override the mounts themselves in the chart, making it mandatory so the pod waits for the config map to be ready. This seems reasonable. We could add comments in the values to illustrate what this mandatory mount would look like.
2. Perform a post-deploy rollout restart, ensuring the deployment comes back up. However, this approach may be a bit disruptive.
3. Have `pepr` always create config maps, even when no CA certificates are provided, and make the mount mandatory in all instances.

I’ve also been wondering why the portal needs the trust bundle mount. @cross-du, is something breaking in your environment right now? Typically, trust bundle mounts are necessary when the application needs to connect to an endpoint using private PKI.

cross-du: Hi @joelmccoy

AI2C originally found the problem (and solution). They use private PKI and mounting the trust bundle gave them access to the Portal.

chance-coleman:
closing this since we’re not moving Portal Package ownership into Core.

the Portal-side follow-up merged in defenseunicorns/uds-portal#59. that keeps the Portal chart responsible for rendering its Package, removes the dead package.enabled API from the abandoned Core-side approach, and documents the trust bundle override using extraVolumes/extraVolumeMounts.

PR #2777, opened 2026-06-29
fix(falco): fix flaky Write below etc test and improve poll reliability

chance-coleman:
## Description
Falco's "Write below etc" integration test was intermittently failing because test pods are mutated by Pepr's `RequireNonRootUser` policy to run as`UID 1000`. Writing to `/etc/passwd` (root:root 644) silently fails as non-root, so `fd.num` is -1 and Falco's `open_write` macro never fires. The test only passed when an ambient "Write below etc" event from another cluster pod happened to be in falcosidekick logs.

Fix mounts an emptyDir volume at `/etc/testdir` with `fsGroup: 1000`, giving `UID 1000` write access. The test now touches a uniquely-named file there, guaranteeing it matches its own event rather than an ambient one.

Moves `execAndWait` calls inside the `pollUntilSuccess` loop in both spec files so a missed eBPF capture triggers a retry instead of a permanent miss. Increases timeouts accordingly.

## Type of change

## Checklist before merging

### DIFF
--- a/test/vitest/falco-integration.spec.ts
+++ b/test/vitest/falco-integration.spec.ts
@@ -95,7 +95,7 @@ describe("Falco Integration e2e Tests", () => {
             await execAndWait(
               "kube-system",
               podName,
-              ["touch", `${etcTestDir}/${testFile}`],
+              ["sh", "-c", `echo test >> ${etcTestDir}/${testFile}`],
               "main",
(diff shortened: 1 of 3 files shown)


briantwatson (approved):
Nice find and fix! 🚀

Issue #2784, opened 2026-07-02
Allow non-expanded hostnames in Package and Istio Gateways


doubleplush:
### Is your feature request related to a problem? Please describe.

Recently we needed to add an alias for Keycloak which was not part of the main domain for the cluster; for example, if the existing VirtualService for Keycloak was sso.cluster.domain.com, we needed to support an additional hostname sso.domain.com for Keycloak.  The Helm chart for uds-istio-config and the UDS Package don't currently support this functionality, since all hostnames are expanded to include the cluster domain.

### Describe the solution you'd like

- In order to support this use case, we would like to be able to provide a fully qualified domain name as a host, denoted by a trailing period at the end of the name.
- If a hostname contains a trailing period, then
- don't expand the hostname with the cluster domain.

PR #2786, opened 2026-07-02
feat: add support for fqdn as host in uds-istio-config

doubleplush:
## Description

This change allows a fully qualified domain name to be passed as a server host value for the uds-istio-config chart, which prevents the hostname from being expanded with the cluster domain.

For example, for the cluster domain `cluster.example.com`:

- Host `sso` should be expanded to `sso.cluster.example.com`.
- Host `.` should be expanded to `cluster.example.com`.
- Host `sso.example.com` should be expanded to `sso.example.com.cluster.example.com`.
- Host `sso.example.com.` should be expanded to `sso.example.com.`.

## Related Issue

Relates to #2784

## Type of change

## Steps to Validate

## Checklist before merging

### DIFF
--- a/src/istio/charts/uds-istio-config/templates/_helpers.tpl
+++ b/src/istio/charts/uds-istio-config/templates/_helpers.tpl
@@ -60,3 +60,17 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Conditionally expand a hostname with the domain name.
+Must be passed a dictionary containing the "host" and "domain" keys as a scope.
+*/}}
+{{- define "uds-istio-config.expandHost" -}}
+{{- if regexMatch "^\\.$" $.host }}
+{{- $.domain }}
+{{- else if regexMatch "\\.$" $.host }}
+{{- $.host }}
+{{- else }}
+{{- $.host }}.{{ $.domain }}
+{{- end }}
+{{- end }}
--- a/src/istio/charts/uds-istio-config/templates/gateway.yaml
+++ b/src/istio/charts/uds-istio-config/templates/gateway.yaml
@@ -16,10 +16,10 @@ spec:
   selector:
     app: {{ .Values.name }}-ingressgateway
   servers:
-    {{ range $name,$server := .Values.tls.servers }}
+    {{- range $name,$server := .Values.tls.servers }}
     - hosts:
         {{- range $server.hosts | default (list "*") }}
-        - "{{ . }}.{{ $domain }}"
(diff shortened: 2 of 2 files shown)


chance-coleman:
@greptileai review

chance-coleman (commented):
If i'm following the intention, this only fixes the Gateway chart side. [`getFqdn()`](https://github.com/defenseunicorns/uds-core/blob/0874c126772da2a6e6fbaa258dba892e135cb128/src/pepr/operator/controllers/domain-utils.ts#L15-L28) still doesn't check for a trailing dot on `expose.host`, so a Package with `host: "sso.example.com."` gets double-expanded into `sso.example.com..cluster.example.com` in [`spec.hosts`](https://github.com/defenseunicorns/uds-core/blob/0874c126772da2a6e6fbaa258dba892e135cb128/src/pepr/operator/controllers/istio/virtual-service.ts#L75).

The Keycloak alias case in #2784 goes through this Package path, not the gateway chart. Needs the same trailing-dot check here or the issue isn't actually closed.

doubleplush:
> If i'm following the intention, this only fixes the Gateway chart side. [`getFqdn()`](https://github.com/defenseunicorns/uds-core/blob/0874c126772da2a6e6fbaa258dba892e135cb128/src/pepr/operator/controllers/domain-utils.ts#L15-L28) still doesn't check for a trailing dot on `expose.host`, so a Package with `host: "sso.example.com."` gets double-expanded into `sso.example.com..cluster.example.com` in [`spec.hosts`](https://github.com/defenseunicorns/uds-core/blob/0874c126772da2a6e6fbaa258dba892e135cb128/src/pepr/operator/controllers/istio/virtual-service.ts#L75).
> 
> The Keycloak alias case in #2784 goes through this Package path, not the gateway chart. Needs the same trailing-dot check here or the issue isn't actually closed.

@chance-coleman The Package path currently has support for specifying a custom domain in an expose block, but currently the Package only supports a custom domain for non-standard Gateways.

https://github.com/defenseunicorns/uds-core/blob/8d6e3226136f2d3eefac39b1dad27951409885fd/src/pepr/operator/crd/validators/package-validator.ts#L68

Do you have context around why that functionality is limited to only custom gateways?

chance-coleman (commented):
> Do you have context around why that functionality is limited to only custom gateways?

package-validator.ts#L68 was added in PR #1912, closing issue #1170. That PR followed [this design doc](https://app.notion.com/p/defense-unicorns/Additional-Gateways-Support-for-Package-Expose-26ae512f24fc80c19166fc3dabcea86e).

domain was added because custom gateways have no assumable domain, so it "will only be valid when the gateway is non-standard." Standard gateways already resolve to a known domain via UDSConfig, so it was never meant as a general override. The case from #2784 just wasn't part of what #1170 was solving.

PR #2799, opened 2026-07-10
fix(operator): extract SAML IdP cert with a scoped, non-greedy regex

### LINKED ISSUE #2798: `samlIdpCertificate` in SSO secret is corrupted when realm has multiple active signing keys (greedy regex in `extractSamlCertificateFromXML`)
### Environment
Device and OS: macOS (client); cluster on AWS EKS
App version: UDS Core v1.7.0
Kubernetes distro being used: EKS, v1.35.6-eks
Other: Keycloak realm `uds` has two active realm signing keys (e.g., after a signing key rotation where the old key remains enabled)

### Steps to reproduce
1. In a Keycloak realm managed by UDS Core, have **two active RSA signing keys** (e.g., rotate the realm signing key without disabling/removing the old provider). The SAML IdP descriptor at `/realms/<realm>/protocol/saml/descriptor` will now contain two `<md:KeyDescriptor use="signing">` entries.
2. Deploy a UDS Package with a SAML SSO client (`protocol: saml`).
3. Inspect the provisioned SSO secret and look at the `samlIdpCertificate` key.

### Expected result
`samlIdpCertificate` contains a single valid base64 X.509 certificate (or, ideally, all advertised signing certificates in a form the SP can consume as multiple certs are legitimate during key rotation windows).

### Actual Result
`samlIdpCertificate` contains the **first certificate, followed by raw XML from the IdP descriptor, followed by the second certificate**, e.g.:

```
MII...pxbfEw=</ds:X509Certificate></ds:X509Data></ds:KeyInfo></md:KeyDescriptor><md:KeyDescriptor use="signing">ds:KeyInfods:KeyNameq8SbsBY9Lf0_...</ds:KeyName>ds:X509Datads:X509CertificateMII...zjM=
```

This is not a valid certificate and breaks SAML signature validation in the consuming application.

**Root cause:** `extractSamlCertificateFromXML` in the UDS Operator extracts the cert with a greedy regex:

```ts
export function extractSamlCertificateFromXML(xmlString: string) {
  const extractedIDPSSODescriptor = xmlString.match(idpSSODescriptorRegex)?.[1] || "";
  return extractedIDPSSODescriptor.match(x509CertRegex)?.[1] || "";
}
```

When the `IDPSSODescriptor` contains more than one `KeyDescriptor use="signing"`, the greedy `x509CertRegex` capture group matches from the first `<ds:X509Certificate>` opening tag to the *last* `</ds:X509Certificate>` closing tag, swallowing all intervening XML.

Suggested fixes: make the regex non-greedy and/or use `matchAll` to collect all certs and select the one matching the realm's active signing `KeyName` (or emit all of them), or parse the descriptor with a real XML parser instead of regex.

### Severity/Priority
High — silently produces an invalid `samlIdpCertificate`, breaking SAML SSO for any package whenever the realm has more than one active signing key (a normal state during key rotation). No error is surfaced; failure only appears downstream as SAML validation errors in the app.

### Additional Context
Workaround: manually patch the SAML IdP Certificate in the application if possible (using an admin web UI or equivalent) 

Note that "old key stays active during rotation" is the recommended Keycloak rotation practice (so outstanding tokens/assertions remain verifiable), so this state is expected to occur in normal operations, not just misconfiguration. This also occurs when Keycloak is configured for brokered IdP functions, and thereby inherits an additional signing key.

willswire:
## Description

The operator read the SAML IdP signing certificate from Keycloak's descriptor endpoint (`/realms/uds/protocol/saml/descriptor`) using a greedy regex:

```
/<[^>]*:X509Certificate[^>]*>((.|[\n\r])*)<\/[^>]*:X509Certificate>/
```

When a realm has more than one active signing key — the normal, recommended state during signing-key rotation — the `IDPSSODescriptor` advertises multiple `X509Certificate` elements. The greedy capture group matched from the first opening tag all the way to the last closing tag, so `samlIdpCertificate` in the generated SSO secret became a corrupt blob: `cert1 + intervening descriptor XML + cert2`. That is not a valid certificate and breaks SAML signature validation in the consuming application. The operator surfaced no error; the failure only appeared downstream. (Reproduction: [https://regex101.com/r/NjGZF5/2](<https://regex101.com/r/NjGZF5/2>))

### Fix

Scope the match to a single `use="signing"` `KeyDescriptor` block and make the capture non-greedy, so multiple certificates can no longer run together. Keycloak sorts signing `KeyDescriptor`s active-key first (verified against the Keycloak signing path, including priority ties), so the **first** signing certificate is the realm's active key — no key-name or descriptor-signature matching is required.

An earlier iteration matched the descriptor's `dsig:Signature` `KeyName` to pick the active key, but that block is only emitted when realm metadata signing is enabled, so it is not reliable; take-first is both simpler and correct. No XML-parser dependency is added.

### Behavior change

Failures now surface on the `Package` CR instead of silently provisioning a broken secret:

* A failed descriptor fetch (which previously returned `undefined` and was written to the secret as the literal base64 string `"undefined"`) now throws.
* A descriptor with no signing certificate now throws.

Both fail the `Package` reconcile with a message naming the client and package. The descriptor fetch also retries once on a transient error, using a small shared `retryOnce` helper that the existing `credentialsCreateOrUpdate` call now shares (previously a bespoke recursive retry).

### Workaround for affected versions

In Keycloak, remove the stale realm signing key so the descriptor advertises a single signing certificate, then trigger the `Package` CR to reconcile so the secret regenerates.

## Related Issue

Fixes defenseunicorns/uds-core#2798

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Steps to Validate

1. Deploy UDS Core with Identity & Authorization enabled.
2. In the `uds` realm, rotate the SAML signing key so two RSA signing keys are active at once. Confirm the descriptor advertises two `<md:KeyDescriptor use="signing">` entries:

   ```bash
   uds zarf tools kubectl run curl --rm -it --image=curlimages/curl --restart=Never -- \
     curl -s http://keycloak-http.keycloak.svc.cluster.local:8080/realms/uds/protocol/saml/descriptor
   ```
3. Deploy a `Package` with a `protocol: saml` SSO client.
4. Confirm `samlIdpCertificate` in the generated secret is a single valid base64 certificate (the first/active signing key) with no embedded XML:

   ```bash
   uds zarf tools kubectl get secret sso-client-<clientId> -n <namespace> \
     -o jsonpath='{.data.samlIdpCertificate}' | base64 -d | openssl x509 -noout -subject
   ```
5. Regression: with a single active signing key, confirm the same certificate is extracted as before.
6. Failure surfacing: point the operator at an unreachable/broken descriptor and confirm the `Package` enters `Failed` with an error naming the client and package, rather than writing a bad secret:

   ```bash
   uds zarf tools kubectl get package <name> -n <namespace> -o jsonpath='{.status}'
   ```

## Checklist before merging

- [ ] [Contributor Guide](<https://github.com/defenseunicorns/uds-core/blob/main/CONTRIBUTING.md>) followed

### DIFF
--- a/src/pepr/operator/controllers/keycloak/client-sync.spec.ts
+++ b/src/pepr/operator/controllers/keycloak/client-sync.spec.ts
@@ -1,12 +1,21 @@
 /**
- * Copyright 2024 Defense Unicorns
+ * Copyright 2024-2026 Defense Unicorns
  * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
  */
 
+import { X509Certificate } from "node:crypto";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
 import { beforeEach, describe, expect, it, MockedFunction, vi } from "vitest";
-import { Sso, UDSPackage } from "../../crd";
+import { Protocol, Sso, UDSPackage } from "../../crd";
 import { syncClient } from "./client-sync";
 import * as clientCredentials from "./clients/client-credentials";
+
+// Shared spy used inside the hoisted vi.mock factory for pepr's fetch. vi.hoisted
+// ensures it is initialized before the (also hoisted) mock factory runs.
+const { mockFetch } = vi.hoisted(() => ({
+  mockFetch: vi.fn(),
+}));
 
 // Mock the logger before importing the modules that use it
 vi.mock("../../../logger", () => ({
@@ -25,6 +34,7 @@ import {
   convertSsoToClient,
   extractSamlCertificateFromXML,
   generateSecretData,
+  getSamlCertificate,
 } from "./client-sync";
 import { Client } from "./types";
 
@@ -34,6 +44,7 @@ const mockApply = vi.fn().mockImplementation(resource => Promise.resolve(resourc
 // Mock the pepr module
 vi.mock("pepr", () => {
   return {
+    fetch: (...args: unknown[]) => mockFetch(...args),
     K8s: () => ({
       Apply: mockApply,
     }),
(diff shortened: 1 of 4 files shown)


src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -169,7 +196,14 @@ export async function syncClient(
...
   if (client.protocol === "saml") {
-    client.samlIdpCertificate = await getSamlCertificate();
+    try {
+      client.samlIdpCertificate = await getSamlCertificate();
+    } catch (err) {
+      throw new Error(
+        `Failed to get SAML IdP certificate for client '${client.clientId}', package ` +
+          `${pkg.metadata?.namespace}/${pkg.metadata?.name}. Error: ${err.message}`,
+      );
+    }
```
chance-coleman: There's no retry here, unlike `credentialsCreateOrUpdate` above it (line 164) which retries once on failure. A transient blip now fails the whole reconcile immediately. Suggest matching the retry-once pattern, ideally by extracting it into a shared helper since it'd exist twice.

src/pepr/operator/controllers/keycloak/client-sync.spec.ts
```diff
@@ -108,38 +117,173 @@ const mockClientStringified: Record<string, string> = {
...
-    </md:EntityDescriptor>
-    `;
-
-    expect(extractSamlCertificateFromXML(xmlString)).toEqual("FOUND THE CERT");
+// Builds a SAML IdP descriptor modeled on a real Keycloak descriptor. The
+// signature block uses the `dsig:` prefix while the KeyDescriptors use `ds:`, as
+// Keycloak does, and elements use the `md:` prefix. `signingKeys` describes the
+// signing KeyDescriptors; `signatureKeyName` is the KeyName Keycloak used to sign
+// the descriptor (i.e. the active signing key).
+function buildDescriptor({
```
chance-coleman, line 126: If im understanding your comment correctly, `buildDescriptor()` is hand-built. Can you pull a real descriptor (curl the in-cluster endpoint) and commit it as a fixture with its own test case? I'd like the testing to use real Keycloak output before we rely on it.

package.json
```diff
@@ -46,6 +46,7 @@
     "test:unit": "vitest run src/pepr/"
   },
   "dependencies": {
+    "fast-xml-parser": "5.9.3",
```
joelmccoy: haven't done a deep dive on the logic/problem here yet... but adding this as an npm dependency probably isn't going to fly (given a lot of the npm supply chain attacks).  We would have to go through a pretty rigorous analysis to want to introduce this dep into the operator.  Is there an approach we can take that doesn't involve bringing in this dep?

chance-coleman: yeah, i think we can avoid the dep. the actual bug is just that the current regex grabs everything from the first cert to the last cert instead of stopping at each one individually. if we match each key block one at a time instead of the whole thing at once, that goes away. same logic after that for picking the right key.

willswire: a no deps approach makes sense to me; reworking the PR

src/pepr/operator/controllers/keycloak/client-sync.ts
willswire: Good call. Extracted a shared `retryOnce(operation, failureContext)` helper and pointed both `credentialsCreateOrUpdate` and the SAML descriptor fetch at it. That also let me drop the old recursive `isRetry` param on `syncClient`, so the retry logic now lives in one place instead of being duplicated. Added tests covering the SAML fetch retrying once on a transient failure and failing with client/package context when it keeps failing.

src/pepr/operator/controllers/keycloak/client-sync.spec.ts
willswire: Added `src/pepr/operator/controllers/keycloak/testdata/saml-descriptor-rotation.xml` with its own test case that runs the extraction end-to-end and parses the result with `node:crypto` to confirm it's a single valid X509 cert (the first/active key), no XML fragments.

I modeled it on the real rotation snapshot from my environment with two signing keys, the enveloped RSA-SHA256 signature naming the first key, all four SLO/SSO bindings, the four NameIDFormats along with the `ID`/`Reference URI`. I swapped in self-signed test certs and a generic `sso.uds.dev` hostname rather than commit a real environment's certs/hostname. `buildDescriptor()` is still there for the focused edge cases (single key, encryption-key-ignored, whitespace stripping, empty).

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -20,15 +21,16 @@ const secretTemplateRegex = new RegExp(
...
+// which was the original bug. Keycloak sorts signing KeyDescriptors active-key
+// first, so the first match holds the realm's active signing key.
+const signingKeyDescriptorRegex =
+  /<[^>]*:KeyDescriptor[^>]*use="signing"[^>]*>([\s\S]*?)<\/[^>]*:KeyDescriptor>/;
 
-// Template regex to match the X509Certificate within the IDPSSODescriptor XML, see https://regex101.com/r/NjGZF5/1
-const x509CertRegex = new RegExp(
-  /<[^>]*:X509Certificate[^>]*>((.|[\n\r])*)<\/[^>]*:X509Certificate>/,
-);
+// Matches the X509Certificate within a KeyDescriptor block.
```
ericwyles:
```suggestion
// Matches the X509Certificate within a KeyDescriptor block. See https://regex101.com/r/NjGZF5/4
```

src/pepr/operator/controllers/keycloak/client-sync.ts
```diff
@@ -20,15 +21,16 @@ const secretTemplateRegex = new RegExp(
...
 
-// Template regex to match IDPSSODescriptor in the SAML IDP Descriptor XML, see https://regex101.com/r/DGvzjd/1
-const idpSSODescriptorRegex = new RegExp(
-  /<[^>]*:IDPSSODescriptor[^>]*>((.|[\n\r])*)<\/[^>]*:IDPSSODescriptor>/,
-);
+// Matches a signing KeyDescriptor block, non-greedy so that a descriptor
+// advertising multiple signing keys (the normal state during signing key
+// rotation) does not collapse into one greedy match spanning every certificate,
+// which was the original bug. Keycloak sorts signing KeyDescriptors active-key
+// first, so the first match holds the realm's active signing key.
```
ericwyles:
```suggestion
// advertising multiple signing keys (the normal state during signing key
// rotation) does not collapse into one greedy match spanning every certificate,
// which was the original bug. Keycloak sorts signing KeyDescriptors active-key
// first, so the first match holds the realm's active signing key. See https://regex101.com/r/YrDx8V/1
```

ericwyles (dismissed):

src/pepr/operator/controllers/keycloak/testdata/saml-descriptor-rotation.xml
```diff
@@ -0,0 +1,74 @@
+/**
+ * Copyright 2025-2026 Defense Unicorns
+ * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+ */
```
chance-coleman:
```suggestion
<!--
  Copyright 2025-2026 Defense Unicorns
  SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
-->
```

nit: I think this is currently invalid xml syntax. current license header checks don't include xml unfortunately so this wasn't picked up.

chance-coleman (approved):

ericwyles (approved):

joelmccoy:
@greptileai review

src/pepr/operator/controllers/keycloak/client-sync.spec.ts
```diff
@@ -108,38 +119,140 @@ const mockClientStringified: Record<string, string> = {
...
+      signingKeys: [{ keyName: "KEY-1" }],
+    });
+    expect(() => extractSamlCertificateFromXML(xml)).toThrow(/no signing X509Certificate/);
+  });
+
+  // Real-shape descriptor: exercises the extraction against a full Keycloak SAML
+  // IdP descriptor captured during a signing-key rotation (two signing keys, a
+  // signed descriptor, and the usual service endpoints).
+  it("extracts the active signing cert from a real rotation descriptor fixture", () => {
+    const xml = readFileSync(join(__dirname, "testdata", "saml-descriptor-rotation.xml"), "utf-8");
```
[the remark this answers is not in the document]
willswire: @joelmccoy you want me to address this finding? 🦎

slaskawi (approved):

joelmccoy (approved):

src/pepr/operator/controllers/keycloak/client-sync.spec.ts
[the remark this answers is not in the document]
joelmccoy: not necessary. minor nit and not affecting test results/workflows

PR #2811, opened 2026-07-17
chore: add retries to fix flakey fleet keycloak test

joelmccoy:
## Description

Fix flakey fleet admin tests with curl retries.  Example flake: https://github.com/defenseunicorns/uds-core/actions/runs/29604647464/job/87965197605?pr=2785

The flake is most likely coming from ambient mesh connectivity readiness when we spin up the pod that runs this test.  Simple retries should resolve.

## Type of change

## Checklist before merging

### DIFF
--- a/test/vitest/keycloak-fleet-admin.spec.ts
+++ b/test/vitest/keycloak-fleet-admin.spec.ts
@@ -135,7 +135,8 @@ set -eu
 token_response="$(mktemp)"
 trap 'rm -f "$token_response"' EXIT
 
-token_status="$(curl -sS -o "$token_response" -w "%{http_code}" -X POST "${keycloakPath(`/realms/${KEYCLOAK_REALM}/protocol/openid-connect/token`)}" \
+token_status="$(curl -sS --retry 5 --retry-all-errors --retry-delay 1 --retry-max-time 20 \
+  -o "$token_response" -w "%{http_code}" -X POST "${keycloakPath(`/realms/${KEYCLOAK_REALM}/protocol/openid-connect/token`)}" \
   -H "Content-Type: application/x-www-form-urlencoded" \
   --data-urlencode "grant_type=client_credentials" \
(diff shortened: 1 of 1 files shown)


test/vitest/keycloak-fleet-admin.spec.ts
```diff
@@ -135,7 +135,8 @@ set -eu
 token_response="$(mktemp)"
 trap 'rm -f "$token_response"' EXIT
 
-token_status="$(curl -sS -o "$token_response" -w "%{http_code}" -X POST "${keycloakPath(`/realms/${KEYCLOAK_REALM}/protocol/openid-connect/token`)}" \
+token_status="$(curl -sS --retry 5 --retry-all-errors --retry-delay 1 --retry-max-time 20 \
```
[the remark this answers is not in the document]
joelmccoy: We’re keeping `--retry-all-errors` because the observed failure is curl exit 56 (CURLE_RECV_ERROR). The `--retry-connrefused` option only addresses connection-refused failures and would not retry the actual issue. These commands do not use `--fail`, so regular HTTP 4xx responses are not considered curl errors.

joelmccoy:
@greptileai review

test/vitest/keycloak-fleet-admin.spec.ts
briantwatson, line 138: I believe if you omit the `--retry-delay` that will use built-in exponential backoff.  Would that be preferable or do we prefer the static interval?  

https://everything.curl.dev/usingcurl/downloads/retry.html#tweak-your-retries

chance-coleman (approved):

test/vitest/keycloak-fleet-admin.spec.ts
joelmccoy: its a good callout. exponential is usually preferred for backoff but i think the difference is so small in this case. having 5 second wait as the first jump seems reasonable.

briantwatson (approved):

PR #2838, opened 2026-07-24
fix: remove preview features from keycloak startup

chance-coleman:
## Description

Stop enabling Keycloak's generic `preview` feature group by default in UDS Core 1.10.

Keycloak 26.7 added `stateless:v1` to the preview group. UDS Core 1.9 implicitly enabled it through `--features=preview,fips`, increasing database utilization and causing a startup failure involving a null `NodeInfo.clusterName()` value in some PostgreSQL-backed deployments.

This change:

- Keeps FIPS enabled by default.
- Adds `keycloak.keycloak.features.additional` for explicit feature selection.
- Prevents newly introduced Keycloak preview features from activating implicitly.
- Updates release notes and upgrade guidance.

## Related Issue

Relates to #

## Type of change

## Steps to Validate

- Run the Helm unit tests and lint checks.
- Deploy UDS Core 1.8.0 with PostgreSQL.
- Upgrade to UDS Core 1.9.0 and verify the existing `--features=preview,fips` configuration reproduces the Keycloak 26.7 startup failure.
- Deploy the 1.10 package over the same database.
- Verify the Keycloak StatefulSet uses `--features=fips` and becomes Ready with zero restarts.
- Verify existing realms, clients, users, and database schema data are preserved.
- Run the documentation build and internal link validation.

## Checklist before merging


slaskawi (dismissed):

slaskawi:
As I highlited in the internal chat, turning the preview features off with UDS Core post 1.0 is probably the right thing to do.

docs/reference/configuration/identity-and-authorization.mdx
```diff
@@ -225,6 +225,26 @@ UDS Core configures Keycloak brute-force detection with the following defaults.
...
+| `features.additional` | `[]` | Keycloak feature names to enable in addition to `fips`. UDS Core passes these values to Keycloak's `--features` option. |
+
+For example, this values file explicitly enables the Keycloak `token-exchange:v1` feature:
+
+```yaml title="values.yaml"
+keycloak:
+  keycloak:
+    features:
+      additional:
+        - token-exchange:v1
```
slaskawi: The Token Exchange v1 is really badly broken. Let's pick some other example, like `spiffe`

slaskawi (dismissed):
LGTM but I would rather pick some other example than Token Exchange v1

slaskawi (dismissed):

packages/standard/zarf-values.schema.json
```diff
@@ -3934,6 +3934,14 @@
...
             },
+            "features": {
+              "properties": {
+                "additional": {
+                  "type": "array"
+                }
+              },
+              "type": "object"
+            },
             "httpRetry": {
```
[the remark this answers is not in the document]
chance-coleman: These schemas are generated, and the current Zarf schema generator drops nested array item types even though the Helm schema defines features.additional as string[]. We should not manually edit generated artifacts because regeneration would remove the change

packages/identity-authorization/zarf-values.schema.json
```diff
@@ -301,6 +301,14 @@
...
             },
+            "features": {
+              "properties": {
+                "additional": {
+                  "type": "array"
+                }
+              },
+              "type": "object"
+            },
             "httpRetry": {
```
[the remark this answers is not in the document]
chance-coleman: These schemas are generated, and the current Zarf schema generator drops nested array item types even though the Helm schema defines features.additional as string[]. We should not manually edit generated artifacts because regeneration would remove the change

slaskawi (dismissed):

docs/operations/release-notes/1-10.mdx
```diff
@@ -0,0 +1,57 @@
+---
+title: UDS Core 1.10
+description: UDS Core 1.10 keeps Keycloak FIPS mode enabled by default while requiring explicit configuration for additional Keycloak features.
+sidebar:
+  order: 3.985
```
joelmccoy, line 6: this should be a draft before the actual release notes are crafted probably? this is going to publish once we merge to main (or i guess when docs site rebuilds).

docs/operations/release-notes/1-10.mdx
```diff
@@ -0,0 +1,57 @@
+---
+title: UDS Core 1.10
+description: UDS Core 1.10 keeps Keycloak FIPS mode enabled by default while requiring explicit configuration for additional Keycloak features.
```
joelmccoy: very weird description here. why are we feeling the need to call out fips mode is still enabled in release notes??

docs/operations/release-notes/1-10.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+### Dependency updates
+
+This feature change does not update a runtime dependency.
+
+## Upgrade considerations
+
+### Keycloak database and session behavior
+
+The 1.10 upgrade does not downgrade the Keycloak database. Keycloak remains on version 26.7.0, and its schema migrations remain applied. Disabling stateless mode changes how volatile authentication data is stored; active login sessions and action tokens may not carry over and users may need to authenticate again. Durable data such as realms, clients, users, roles, and identity providers remains in the database.
```
joelmccoy: feels very weird to call this out.  almost makes me more concerned on this upgrade? does this provide anything useful here?

joelmccoy (commented):
A couple small docs things here, otherwise looks good to me

docs/operations/release-notes/1-10.mdx
chance-coleman: i can word this better. Primarily the database was what changed in the recent preview feature so wanted to specifically note that this removal of preview features wouldn't cause any issues with the database.

docs/operations/release-notes/1-10.mdx
chance-coleman: cleaned this up

docs/operations/release-notes/1-10.mdx
chance-coleman: added draft mode

chance-coleman:
@greptileai

docs/operations/release-notes/1-10.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+### Dependency updates
+
+This feature change does not update a runtime dependency.
+
+## Upgrade considerations
+
+### Keycloak session behavior
+
+Keycloak remains on version 26.7.0, and the upgrade leaves its schema and durable data unchanged. Realms, clients, users, roles, and identity providers remain intact. Because stateless mode is no longer enabled by default, active login sessions and action tokens may not carry over during the transition, so some users may need to authenticate again.
```
joelmccoy: nit: can we just keep this to the point? I am very confused by the wording here and as an external user I would be a bit confused what the takeaway is here

```suggestion
Because stateless mode is no longer enabled by default in this release, active Keycloak login sessions and action tokens may not carry over during the upgrade. Some users may need to sign in again.
```

docs/operations/release-notes/overview.mdx
```diff
@@ -27,8 +27,3 @@ If you support a deployment older than the release notes covered here, see [Lega
   description="A new Fleet Command Keycloak client, Portal tile customization, and a faster Keycloak startup."
   href="/operations/release-notes/1-8/"
 />
-<LinkCard
```
joelmccoy, line 30: do we need to remove this now? it will disappear from docs site

joelmccoy (dismissed):
Two comments, but not blocking

eoghanriley (approved):

joelmccoy (approved):

PR #2851, opened 2026-07-31
feat: support multiple classification banners

### LINKED ISSUE #2849: Support multiple classification banners with host and path matching

The Istio classification banner configuration supports only one banner definition selected by hostname. This forces every enabled host to use the same classification text and footer behavior, and it cannot assign different markings to URL path prefixes on the same host.

Operators need to support environments where applications or sections of one application have different classification markings without maintaining separate Envoy filters or modifying third-party applications.

- **Given** applications that require different classification markings
- **When** an operator configures multiple `classificationBanners` entries with independent `text`, `addFooter`, and `enabledHosts` values
- **Then** UDS Core injects the banner associated with the request host

- **Given** multiple classification levels served from one host
- **When** an operator adds optional `pathPrefixes` to banner entries
- **Then** UDS Core selects the banner with the longest matching URL path prefix and ignores the query string during matching

- **Given** a host-only banner and one or more path-specific banners for the same host
- **When** no configured path prefix matches the request
- **Then** UDS Core uses the host-only banner as the fallback

- **Given** an existing deployment that uses the singular `classificationBanner` value
- **When** the chart renders
- **Then** UDS Core prepends the legacy object to the effective banner list and emits a deprecation warning without breaking the deployment

The chart values schema should validate the plural configuration and require each path prefix to begin with `/`. The generated package schemas and classification banner documentation should remain synchronized.

- Deploy a separate `EnvoyFilter` for each classification. Multiple response filters introduce ordering and duplicate-injection risks.
- Implement banners directly in each application. This remains preferable for custom applications, but it is not always possible for third-party applications.
- Use separate hostnames for every classification level. This adds DNS and routing complexity and does not address applications that intentionally share a host.

### Additional context

Host-only behavior must remain unchanged when `pathPrefixes` is empty or omitted. When path prefixes overlap, longest-prefix matching provides deterministic precedence. The existing singular `classificationBanner` value should remain available as a deprecated compatibility path during migration.

willswire:
## Description

Adds support for configuring multiple Istio classification banners with independent classification text, footer behavior, hosts, and optional URL path prefixes.

The Envoy filter now:

* Selects banners by request host.
* Supports different banners on the same host using `pathPrefixes`.
* Uses the longest matching path prefix.
* Uses a host-only entry as the fallback when no path prefix matches.
* Excludes query strings from path matching.

The deprecated `classificationBanner` value remains supported. A Helm helper prepends it to the effective `classificationBanners` list, and the chart emits a deprecation note when it is enabled.

This change also updates the chart schema, generated Zarf schemas, tests, values examples, and classification banner documentation.

## Related Issue

Fixes defenseunicorns/uds-core#2849

## Type of change

## Steps to Validate

1. Run the classification banner values tests:

   ```bash
   npx vitest run test/values/base.spec.ts
   ```

   Verify all 52 tests pass.
2. Lint the Helm chart:

   ```bash
   helm lint src/istio/common/chart
   ```
3. Regenerate the package value schemas:

   ```bash
   uds run -f tasks/lint.yaml values-generate
   ```

   Verify only the expected generated schemas change.
4. Configure a host-only fallback and a path-specific banner on the same host:

   ```yaml
   classificationBanners:
     - text: "UNCLASSIFIED"
       addFooter: true
       enabledHosts:
         - app.uds.dev
     - text: "SECRET"
       addFooter: false
       enabledHosts:
         - app.uds.dev
       pathPrefixes:
         - /classified
   ```
5. Verify `https://app.uds.dev/` displays `UNCLASSIFIED` and `https://app.uds.dev/classified` displays `SECRET`.
6. Configure the deprecated `classificationBanner` value and verify Helm emits the migration warning and renders it before entries from `classificationBanners`.

## Checklist before merging

- [X] [Contributor Guide](<https://github.com/defenseunicorns/uds-core/blob/main/CONTRIBUTING.md>) followed


joelmccoy:
@willswire just a heads up this will likely require a refactor as we are changing some of the classification banner logic here: https://github.com/defenseunicorns/uds-core/pull/2710 (we are still working to test and validate that PR).

willswire:
@joelmccoy rebased to account for #2710

willswire:
@greptileai review

src/istio/common/chart/templates/classification-banner.yaml
```diff
@@ -162,11 +156,27 @@ spec:
...
+                  [{{ tpl . $ | quote }}] = true,
+                  {{- end }}
+                },
+                pathPrefixes = {
+                  {{- range ($banner.pathPrefixes | default (list)) }}
+                  [{{ . | quote }}] = true,
+                  {{- end }}
+                },
+                banner = {
+                  classification = {{ $banner.text | default "UNKNOWN" | quote }},
```
chance-coleman: Is `default "UNKNOWN"` intentional when the legacy `classificationBanner.text` is explicitly empty? `UNKNOWN` makes sense when the value is omitted, but the schema allows `""` and the old template kept it empty. If empty is meant to differ from unset, this changes the old behavior.

src/istio/common/chart/templates/classification-banner.yaml
```diff
@@ -1,7 +1,8 @@
 # Copyright 2024-2026 Defense Unicorns
 # SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
 
-{{- if .Values.classificationBanner.enabledHosts }}
+{{- $classificationBanners := include "uds-global-istio-config.classificationBanners" . | fromYamlArray }}
+{{- if $classificationBanners }}
```
chance-coleman, line 5: Should this check that at least one banner has an enabled host? With `classificationBanners: [{text: SECRET, enabledHosts: []}]`, the list is non-empty so the EnvoyFilter still gets installed even though it cannot match any request. The old chart skipped the filter when there were no enabled hosts.

chance-coleman (commented):
Two comment/questions, otherwise looking good.

src/istio/common/chart/templates/classification-banner.yaml
willswire: Good catch. That wasn’t intentional. I updated the template to distinguish an omitted `text` key from an explicitly empty value. Omitted text defaults to `UNKNOWN`, while `text: ""` preserves the legacy behavior. I also added a regression test for the legacy configuration.

src/istio/common/chart/templates/classification-banner.yaml
willswire: Agreed. I updated the helper to exclude banner entries without enabled hosts. If no configured banner has an enabled host, the chart no longer renders the EnvoyFilter. I added a regression test for this case as well.

willswire:
@greptileai please review my recent changes to address @chance-coleman's feedback

joelmccoy:
@willswire I think this is a reasonable contribution. I think deprecating the old value seems like a sane approach (it's just a matter to what level depth we want to support this feature in core. I'm ok with supporting this in core if we can more accurately test this in CI (we've had some regressions in the past that I would want to avoid).

I think this approach is reasonable but would want to see us:
1. Accurately test this in our e2e tests with multiple classification banners. Right now our demo bundles use the old classification banner field and put a sample banner on everything in Core. We then assert on that being present here: https://github.com/defenseunicorns/uds-core/blob/367d25f8a38386b6eb7bf4958991cd03c4bd43fc/test/playwright/classification-banner.test.ts#L10.  We should probably assert the logic e2e for this new feature instead.
2. callout this out in [DEPRECATIONS.md](https://github.com/defenseunicorns/uds-core/blob/main/DEPRECATIONS.md)

willswire:
Thanks @joelmccoy, agreed on both points. I updated the standard demo and CI configurations to use `classificationBanners` and expanded the Playwright coverage to verify multiple banner selections:

- A path-specific `SAMPLE BANNER` on SSO
- An `UNKNOWN` banner on Grafana with the footer disabled
- An `UNCLASSIFIED` banner on Portal

This exercises host matching, path-prefix matching, distinct banner text, and per-banner footer behavior through the deployed EnvoyFilter. I also added the legacy `classificationBanner` value to `DEPRECATIONS.md` with migration guidance and a `2.0.0` removal target.

I ran the Helm unit tests, values tests, formatting and lint checks, and the full E2E suite; everything seems to passing 👌🏼

bundles/k3d-standard/uds-bundle.yaml
```diff
@@ -110,16 +110,22 @@ packages:
...
-              value: "SAMPLE BANNER"
-            - path: classificationBanner.enabledHosts
+            # Demo classification banners use non-sensitive markings and provide E2E coverage for
+            # host, path, and footer selection across the standard dev bundle.
+            - path: classificationBanners
               value:
-                - "sso.uds.dev"
-                - "portal.uds.dev"
-                - "grafana.admin.uds.dev"
+                - text: "SAMPLE BANNER"
```
joelmccoy, line 117: the problem with this approach is that this shows up in the published uds demo bundles. it feels weird that we would set these for those.  The overrides worked before because we agreed to put SAMPLE Banner everywhere in the demo bundle.  I think for testing different combinations other than that we might need to override in our tasks with UDS variables/values but not bake into the demo bundles somehow?

willswire: good call. What if we change `classificationBanners` to a UDS variable with the published demo default remaining a single `SAMPLE BANNER` across the existing hosts? The multiple-banner matrix can live in a new `uds-test-config.yaml` and is applied by the test deployment task through `UDS_CONFIG`. I also added the matrix to the HA and upgrade test configs so those workflows retain the same Playwright coverage without baking test combinations into the published bundle.

DEPRECATIONS.md
```diff
@@ -21,6 +21,8 @@ When adding a deprecation, include:
...
 | `allow.podLabels`, `allow.remotePodLabels`, `expose.podLabels`, `expose.match`      | 0.12.0 ([#154](https://github.com/defenseunicorns/uds-core/pull/154))   | **Reason:** API naming improved.<br/>**Migration:** Use `allow.selector`, `allow.remoteSelector`, `expose.selector`, `expose.advancedHTTP.match` instead | Package `v1beta1` |
 | `sso.secretName`, `sso.secretLabels`, `sso.secretAnnotations`, `sso.secretTemplate` | 0.60.0 ([#2264](https://github.com/defenseunicorns/uds-core/pull/2264)) | **Reason:** Simplified field structure.<br/>**Migration:** Use `sso.secretConfig.name`, `.labels`, `.annotations`, `.template` instead                   | Package `v1beta1` |
+| `classificationBanner` Helm value                                                   | 1.10.0 ([#2851](https://github.com/defenseunicorns/uds-core/pull/2851)) | **Reason:** The value supports only one banner configuration.<br/>**Migration:** Move the configuration into the `classificationBanners` list.           | 2.0.0             |
```
joelmccoy: nit: can we just specify the helm chart here (istio)

joelmccoy (dismissed):
Awesome work on this one! Verified this against a bunch of sample apps locally.  Would want to get one more set of eyes on this before we merge in.  cc: @chance-coleman

test/playwright/classification-banner.test.ts
```diff
@@ -7,81 +7,85 @@ import { expect, test } from "@playwright/test";
...
+  await expectBanner(page, "SAMPLE BANNER");
 });
 
-test("grafana shows the classification banner", async ({ page }) => {
+test("grafana selects the host classification banner without a footer", async ({ page }) => {
   // Grafana's root returns a 302 redirect (the case that previously hung the gateway), so
   // reaching a rendered page here exercises both the redirect and the 200 HTML inject paths.
   await page.goto(`https://grafana.admin.${domain}/`);
-  await expectBanner(page);
+  await expectBanner(page, "UNKNOWN", false);
```
chance-coleman, line 81: Sorry, one more 🫣 i just realized this will fail our IAC CI jobs.  AKS, EKS, RKE2 CI jobs run this test will fail [because they expect "SAMPLE BANNER"](https://github.com/defenseunicorns/uds-core/blob/main/.github/bundles/eks/uds-bundle.yaml#L35) instead of the legacy "UNKNOWN" banner.

joelmccoy: ugh good catch. i think for our IAC bundles we can provide the correct overrides there since we don't publish those ones

chance-coleman (approved):
Thanks for sticking with this! LGTM

joelmccoy (approved):

PR #2871, opened 2026-08-06
fix(istio): gate gateway deployment on Pepr ambient readiness

chance-coleman:
## Description

Prevent intermittent Pepr admission webhook `EOF` failures during Istio Ambient gateway deployment.

- Istio Ambient can briefly make the Pepr admission webhook unavailable during startup.
- Waits for Pepr admission Pods to report Ready=True.
- Waits for each admission Pod to have `ambient.istio.io/redirection=enabled`.
- Uses bounded kubectl wait calls with 90-second condition timeouts and a 240-second action limit.
- Validates the admin gateway Service through a server-side apply dry-run.
- Emits Pod and dry-run diagnostics when validation fails.
- Supports both fresh installations and upgrades where the gateway Service already exists.

## Related Issue

Fixes CORE-660

## Type of change

## Steps to Validate
- `npm run test:unit`
- `uds zarf package create packages/standard --flavor upstream --confirm --skip-sbom`
- Verified server-side apply dry-run against an existing `admin-ingressgateway` Service.

## Checklist before merging


chance-coleman:
@greptileai

src/istio/common/zarf.yaml
```diff
@@ -140,6 +140,57 @@ components:
...
+              ./zarf tools kubectl wait \
+                -n "${namespace}" \
+                --for='jsonpath={.metadata.annotations.ambient\.istio\.io/redirection}'=enabled \
+                --timeout=90s \
+                pod -l "${pod_selector}"
+
+              ./zarf tools kubectl create namespace istio-admin-gateway --dry-run=client -o yaml \
+                | ./zarf tools kubectl apply --request-timeout=30s -f - >/dev/null
+
+              if dry_run_output=$(./zarf tools kubectl apply --request-timeout=30s --server-side --dry-run=server -f - -o yaml 2>&1 <<'EOF
```
joelmccoy: I'm not the biggest fan of this approach of targeting the istio admin gateway namespace and feel like this could be cleaner and more generic.. this create namespace dry-run doesn't do anything? and we are trying to create a resource in a namespace that possibly doesn't exist?

We basically just need to verify the webhook. Could we do something simpler in a namespace that we know is going to exist?

```yaml
          - description: "Wait for Pepr Ambient admission readiness"
            maxTotalSeconds: 90
            cmd: |
              set -e

              ./zarf tools kubectl wait \
                -n pepr-system \
                --for='jsonpath={.metadata.annotations.ambient\.istio\.io/redirection}'=enabled \
                --timeout=60s \
                pod -l 'app=pepr-uds-core,pepr.dev/controller=admission'

              until ./zarf tools kubectl create \
                --dry-run=server \
                -o name \
                -f - <<'EOF'
              apiVersion: v1
              kind: Service
              metadata:
                generateName: pepr-admission-probe-
                namespace: default
              spec:
                ports:
                  - port: 443
              EOF
              do
                sleep 2
              done

```

src/istio/common/zarf.yaml
```diff
@@ -140,6 +140,57 @@ components:
...
 
+          - description: "Verify Pepr ambient readiness before gateway deployment"
+            cmd: |
+              set -eu
+
+              namespace="pepr-system"
+              pod_selector="app=pepr-uds-core,pepr.dev/controller=admission"
+              ./zarf tools kubectl wait \
+                -n "${namespace}" \
+                --for=condition=Ready \
```
joelmccoy: curious is this even needed? ok with keeping it... but as far as i can tell pepr shouldn't restart here?

joelmccoy (commented):
Thanks for tracking this down. Small comments for consideration!

src/istio/common/zarf.yaml
chance-coleman: yeah good call, remnant from an earlier approach that i didnt clean up.

src/istio/common/zarf.yaml
chance-coleman: got this working with a simple approach,  now waits for Ambient enrollment, then uses a generated Service in the existing default namespace

bradclawsie (dismissed):

src/istio/common/zarf.yaml
```diff
@@ -140,6 +140,35 @@ components:
...
+              kind: Service
+              metadata:
+                generateName: pepr-admission-probe-
+                namespace: default
+              spec:
+                ports:
+                  - port: 443
+              EOF
+              do
+                sleep 2
```
Racer159: supernit: nowhere else (pretty sure) do we depend on the `sleep` binary - it is possible to do mostly what this does in a more native-Zarf way.  Not a huge deal since `busybox`'s toolset is generally the accepted requirements to deploy Zarf packages (and Zarf does need an `sh` POSIX shell for actions) but calling out this technically adds a dependency for Core.  Marking it as supernit because the below behavior is different (it will retry immediately without waiting necessarily) and `sleep` is still super common if you have a shell.

```
- description: "Wait for Pepr Ambient enrollment"
  maxTotalSeconds: 60
  wait:
    cluster:
      kind: pod
      name: app=pepr-uds-core,pepr.dev/controller=admission
      namespace: pepr-system
      condition: '{.metadata.annotations.ambient\.istio\.io/redirection}=enabled'

- description: "Verify Pepr admission readiness"
  maxTotalSeconds: 30
  maxRetries: 10
  cmd: |
    ./zarf tools kubectl create \
      --request-timeout=5s \
      --dry-run=server \
      -o name \
      -f - <<'EOF'
    apiVersion: v1
    kind: Service
    metadata:
      generateName: pepr-admission-probe-
      namespace: default
    spec:
      ports:
        - port: 443
    EOF
```

Racer159 (dismissed):

src/istio/common/zarf.yaml
chance-coleman: thats a good call, i've updated to use native Zarf wait and retry actions. Removes the sleep dependency while preserving Ambient enrollment and admission webhook readiness checks.

joelmccoy (approved):

PR #2873, opened 2026-08-10
feat: expose full scope allowed for SSO clients

slaskawi:
## Description

Exposes `spec.sso[].fullScopeAllowed` on Package SSO clients. Omitted values default to `true` to preserve existing Keycloak behavior.

For security hardening, set `fullScopeAllowed: false` whenever the application works with explicit role scope mappings and client scopes. Uptime probe clients set this to `false` explicitly.

## Related Issue

Relates to CORE-653

## Type of change

- [ ] Bug fix, non breaking change which fixes an issue
- [x] New feature, non breaking change which adds functionality
- [ ] Other, security configuration, docs update, etc

## Steps to Validate

1. Run `npm run test:unit`.
2. Regenerate CRD artifacts with `uds run -f src/pepr/tasks.yaml gen-crds`.

## Checklist before merging


src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -535,6 +535,12 @@ const sso = {
         type: "boolean",
         default: false,
       },
+      fullScopeAllowed: {
+        description:
+          "Controls which roles Keycloak includes in tokens for this client. When enabled, Keycloak includes all roles assigned to the authenticated subject. When disabled, Keycloak limits roles according to the client's role scope mappings and linked client scopes. UDS Core defaults this setting to true for backwards compatibility. Set this to false whenever possible to limit role scope and harden security.",
+        type: "boolean",
+        default: true,
```
slaskawi: This shouldn't be here. Otherwise, once we pass 2.0, we'd need to introduce a breaking change. There should be no default here (if possible).

docs/reference/configuration/identity-and-authorization.mdx
```diff
@@ -142,6 +142,8 @@ The plugin enforces a `UDS Client Profile` Keycloak client policy for all client
 
 To extend the allow-list, set `SECURITY_HARDENING_ADDITIONAL_PROTOCOL_MAPPERS` or `SECURITY_HARDENING_ADDITIONAL_CLIENT_SCOPES` in `realmInitEnv` (see [Realm initialization variables](#realm-initialization-variables)).
 
+UDS Core defaults `sso[].fullScopeAllowed` to `true` for backwards compatibility. Set it to `false` whenever your application works with explicit role scope mappings and client scopes, which limits token roles to only those the client requires. Reference the [`Package` CR SSO fields](/reference/operator-and-crds/packages-v1alpha1-cr/#sso) for the field definition.
```
chance-coleman: The docs say that leaving out `fullScopeAllowed` always sets it to true. The code leaves the field out, so Keycloak chooses the value. For example, a new client with `consentRequired: true` can get false. Probably should say that omission keeps Keycloak’s normal behavior.

chance-coleman (commented):
looks good, one small nit and then will approve

docs/reference/configuration/identity-and-authorization.mdx
slaskawi: Updated the documentation to clarify that omitting fullScopeAllowed preserves Keycloak's normal behavior. The generated CRD reference remains synchronized.

slaskawi:
@codex review

src/pepr/operator/crd/sources/package/v1alpha1.ts
```diff
@@ -535,6 +535,11 @@ const sso = {
         type: "boolean",
         default: false,
       },
+      fullScopeAllowed: {
+        description:
+          "Controls which roles Keycloak includes in tokens for this client. When enabled, Keycloak includes all roles assigned to the authenticated subject. When disabled, Keycloak limits roles according to the client's role scope mappings and linked client scopes. UDS Core defaults this setting to true for backwards compatibility. Set this to false whenever possible to limit role scope and harden security.",
+        type: "boolean",
```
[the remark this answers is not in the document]
slaskawi: Fixed in cbd39ffc. The field description now states that omission preserves Keycloak’s normal behavior. I regenerated the CRD YAML, TypeScript types, JSON schema, and CRD reference. The focused tests pass (22/22).

chance-coleman (dismissed):

src/pepr/operator/controllers/keycloak/client-sync.spec.ts
joelmccoy, line 1: should we add a test when fullscopeclient is set to false?

docs/reference/configuration/identity-and-authorization.mdx
joelmccoy, line 1: do we eventually want to default this to false? if so, maybe we should add a deprecation in `DEPRECATIONS.md`

joelmccoy (commented):
Generally lgtm - two small comments/questions

src/pepr/operator/controllers/keycloak/client-sync.spec.ts
slaskawi: Added a focused false-preservation test alongside the default and explicit-value coverage. It protects the `?? true` behavior in client conversion.

docs/reference/configuration/identity-and-authorization.mdx
slaskawi: Added the deprecation for omitting `sso[].fullScopeAllowed`, with an explicit `true` migration path to preserve current behavior before the `2.0.0` default change.

chance-coleman (approved):

joelmccoy (approved):

PR #1865, opened 2025-08-29
fix: use fixed Keycloak hostnames (strict variant)

slaskawi:
## Description

This Pull Request has been created mostly for discussion and it's a more strict version of https://github.com/defenseunicorns/uds-core/pull/1864 with the following Well Known Endpoint payload:

```yaml
{
    "issuer": "https://sso.uds.dev/realms/uds",
    "authorization_endpoint": "https://sso.uds.dev/realms/uds/protocol/openid-connect/auth",
    "token_endpoint": "https://sso.uds.dev/realms/uds/protocol/openid-connect/token",
    "introspection_endpoint": "https://sso.uds.dev/realms/uds/protocol/openid-connect/token/introspect",
    "userinfo_endpoint": "https://sso.uds.dev/realms/uds/protocol/openid-connect/userinfo",
    "end_session_endpoint": "https://sso.uds.dev/realms/uds/protocol/openid-connect/logout",
    "frontchannel_logout_session_supported": true,
    "frontchannel_logout_supported": true,
    "jwks_uri": "https://sso.uds.dev/realms/uds/protocol/openid-connect/certs",
    "check_session_iframe": "https://sso.uds.dev/realms/uds/protocol/openid-connect/login-status-iframe.html",
    "grant_types_supported": [
        "authorization_code",
        "client_credentials",
        "implicit",
        "password",
        "refresh_token",
        "urn:ietf:params:oauth:grant-type:device_code",
        "urn:ietf:params:oauth:grant-type:token-exchange",
        "urn:ietf:params:oauth:grant-type:uma-ticket",
        "urn:openid:params:grant-type:ciba"
    ],
    "acr_values_supported": [
        "0",
        "1"
    ],
    "response_types_supported": [
        "code",
        "none",
        "id_token",
        "token",
        "id_token token",
        "code id_token",
        "code token",
        "code id_token token"
    ],
    "subject_types_supported": [
        "public",
        "pairwise"
    ],
    "prompt_values_supported": [
        "none",
        "login",
        "consent",
        "create"
    ],
    "id_token_signing_alg_values_supported": [
        "PS384",
        "RS384",
        "EdDSA",
        "ES384",
        "HS256",
        "HS512",
        "ES256",
        "RS256",
        "HS384",
        "ES512",
        "PS256",
        "PS512",
        "RS512"
    ],
    "id_token_encryption_alg_values_supported": [
        "ECDH-ES+A256KW",
        "ECDH-ES+A192KW",
        "ECDH-ES+A128KW",
[...]

PR #1864, opened 2025-08-29
fix: use fixed Keycloak hostname

slaskawi:
## Description

This Pull Request introduces a fixed Keycloak hostname. When calling the Well Known OIDC endpoint from the inside of the cluster, the following content is produced:

```json
{
    "issuer": "https://sso.uds.dev/realms/uds",
    "authorization_endpoint": "https://sso.uds.dev/realms/uds/protocol/openid-connect/auth",
    "token_endpoint": "http://keycloak-http.keycloak/realms/uds/protocol/openid-connect/token",
    "introspection_endpoint": "http://keycloak-http.keycloak/realms/uds/protocol/openid-connect/token/introspect",
    "userinfo_endpoint": "http://keycloak-http.keycloak/realms/uds/protocol/openid-connect/userinfo",
    "end_session_endpoint": "https://sso.uds.dev/realms/uds/protocol/openid-connect/logout",
    "frontchannel_logout_session_supported": true,
    "frontchannel_logout_supported": true,
    "jwks_uri": "http://keycloak-http.keycloak/realms/uds/protocol/openid-connect/certs",
    "check_session_iframe": "https://sso.uds.dev/realms/uds/protocol/openid-connect/login-status-iframe.html",
    "grant_types_supported": [
        "authorization_code",
        "client_credentials",
        "implicit",
        "password",
        "refresh_token",
        "urn:ietf:params:oauth:grant-type:device_code",
        "urn:ietf:params:oauth:grant-type:token-exchange",
        "urn:ietf:params:oauth:grant-type:uma-ticket",
        "urn:openid:params:grant-type:ciba"
    ],
    "acr_values_supported": [
        "0",
        "1"
    ],
    "response_types_supported": [
        "code",
        "none",
        "id_token",
        "token",
        "id_token token",
        "code id_token",
        "code token",
        "code id_token token"
    ],
    "subject_types_supported": [
        "public",
        "pairwise"
    ],
    "prompt_values_supported": [
        "none",
        "login",
        "consent",
        "create"
    ],
    "id_token_signing_alg_values_supported": [
        "PS384",
        "RS384",
        "EdDSA",
        "ES384",
        "HS256",
        "HS512",
        "ES256",
        "RS256",
        "HS384",
        "ES512",
        "PS256",
        "PS512",
        "RS512"
    ],
    "id_token_encryption_alg_values_supported": [
        "ECDH-ES+A256KW",
        "ECDH-ES+A192KW",
        "ECDH-ES+A128KW",
[...]

PR #2893, opened 2026-08-19
fix: separate  keycloak public and admin hostnames

chance-coleman:
## Description

Configure Keycloak with separate public and admin hostnames.

- Use `sso.<domain>` for frontend and fixed backchannel URLs.
- Use `keycloak.<adminDomain>` for the administration console and API.
- Evaluate templated domain values before applying admin-domain fallbacks.
- Add Helm coverage for hostname and fallback behavior.
- Add runtime coverage for reset-password and OIDC discovery URLs.

## Related Issue

Relates to CORE-672

## Type of change

## Steps to Validate
- Run `helm unittest src/keycloak/chart --color`.
- Run `helm lint src/keycloak/chart`.
- Run `./node_modules/.bin/vitest run keycloak-reset-password.spec.ts` from `test/vitest`.
- Run `npm run format:check`.

## Checklist before merging


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -161,7 +161,18 @@ spec:
...
             - "--http-enabled=true"
+            {{- if .Values.devMode }}
+            # Keep forwarded hostnames dynamic for the local zarf connect workflow.
             - "--hostname-strict=false"
+            {{- else }}
+            {{- $domain := include "keycloak.domain" . }}
+            {{- $adminDomain := include "keycloak.adminDomain" . }}
+            # Keep browser-facing URLs on the public hostname while serving the
+            # administration console from the dedicated admin hostname.
+            - "--hostname=https://sso.{{ $domain }}"
```
slaskawi, line 172: Unfortunately this doesn't seem to work and it hangs exactly the same way as here: https://github.com/defenseunicorns/uds-core/pull/1864#pullrequestreview-3169106064

Just to add more context - I previously tried simialar approach in these Pull Requests:

- https://github.com/defenseunicorns/uds-core/pull/1864
- https://github.com/defenseunicorns/uds-core/pull/1865

None of them really worked and based on the conversation within https://github.com/keycloak/keycloak/issues/42264 - it's by design. 

Personally, I'm not convinced by any of the arguments from the above discussion and I think the API design is so misleading, that it's really easy to get the configuration wrong. At the same time, I think we can address this ourselves by writing our own `HostnameProvider` that extends `HostnameV2Provider`.

PR #2898, opened 2026-08-21
feat: configure separate public and admin keycloak hostnames

chance-coleman:
## Description

Configures Keycloak with separate public and admin frontend hostnames.

Public frontend URLs use `https://sso.<domain>`. Admin-console frontend URLs use `https://keycloak.<admin-domain>`. The admin domain defaults to `admin.<domain>` and can be configured explicitly.

Adds Helm tests and live regression tests for public discovery, admin discovery, public admin-path redirects, and admin-console frontend URLs.

## Related Issue

Relates to CORE-672

## Type of change

## Steps to Validate
- Deploy the slim-dev bundle.
- Verify public discovery returns `https://sso.uds.dev/realms/uds`.
- Verify admin discovery returns `https://keycloak.admin.uds.dev/realms/uds`.
- Verify the admin console uses `keycloak.admin.uds.dev` for its frontend URLs.
- Verify public user login continues using `sso.uds.dev`.

## Checklist before merging

### DIFF
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -218,6 +218,8 @@ spec:
               value: "{{ $domain }}"
             - name: UDS_ADMIN_DOMAIN
               value: "{{ $adminDomain }}"
+            - name: UDS_CLUSTER_DOMAIN
+              value: "{{ .Values.clusterDomain }}"
 
             # Enable health and metrics endpoints
             - name: KC_HEALTH_ENABLED
--- a/src/keycloak/chart/templates/statefulset.yaml
+++ b/src/keycloak/chart/templates/statefulset.yaml
@@ -164,12 +164,6 @@ spec:
             - "--spi-authentication-sessions--map--auth-sessions-limit={{ .Values.realmConfig.maxInFlightLoginsPerUser | default 300 }}"
             - "--proxy-headers=xforwarded"
             - "--http-enabled=true"
-            - "--hostname-strict=false"
-            # Keep public and admin frontend URLs separate while the UDS hostname provider
-            # selects the admin origin for requests entering through the admin gateway.
-            - "--hostname=https://sso.{{ .Values.domain }}"
-            - "--hostname-admin=https://keycloak.{{ $adminDomain }}"
-            - "--hostname-backchannel-dynamic=true"
             {{- if .Values.jsonLogFormat }}
             - "--log-console-output=json"
             {{- end }}
@@ -220,6 +214,16 @@ spec:
               value: "{{ $adminDomain }}"
             - name: UDS_CLUSTER_DOMAIN
               value: "{{ .Values.clusterDomain }}"
+
+            # Keep native Keycloak hostname settings together in the pod environment.
+            - name: KC_HOSTNAME
+              value: "https://sso.{{ $domain }}"
+            - name: KC_HOSTNAME_ADMIN
+              value: "https://keycloak.{{ $adminDomain }}"
+            - name: KC_HOSTNAME_BACKCHANNEL_DYNAMIC
+              value: "true"
+            - name: KC_HOSTNAME_STRICT
+              value: "false"
 
             # Enable health and metrics endpoints
             - name: KC_HEALTH_ENABLED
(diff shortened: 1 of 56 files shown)


src/keycloak/chart/templates/statefulset.yaml
```diff
@@ -206,12 +214,12 @@ spec:
               value: "-Djava.security.properties=/opt/keycloak/data/fips-sunjce.security"
             {{- end }}
             # Common configuration
-            {{- $domain := .Values.domain }}
-            {{- $adminDomain := .Values.adminDomain | default (printf "admin.%s" $domain) }}
             - name: UDS_DOMAIN
               value: "{{ $domain }}"
             - name: UDS_ADMIN_DOMAIN
               value: "{{ $adminDomain }}"
+            - name: UDS_CLUSTER_DOMAIN
```
slaskawi, line 215: I wonder if we could align the configuration to use one way of passing the values in? Either all the hostnames through the environment variables or bootstrap parameters.

slaskawi (dismissed):
LGTM with one comment

src/keycloak/chart/templates/statefulset.yaml
chance-coleman: I might be misunderstanding what you're suggesting here. The `--hostname*` values are native Keycloak bootstrap settings, while `UDS_DOMAIN` and `UDS_ADMIN_DOMAIN` are used by the realm configuration, and `UDS_CLUSTER_DOMAIN` is read directly by the UDS hostname plugin. Moving them all to one mechanism would require changing those contracts and could break custom cluster-domain support.

slaskawi: All the bootstrap settings passed into the `kc.sh` script can be replaced with env vars and vice-versa. 

For example, the `--hostname` can also be set via `KC_HOSTNAME` environment variable. The full manual for these settings can be found [here](https://www.keycloak.org/server/all-config#category-hostname_v2). 

I'm basically suggesting to use one convention (doesn't matter which one) for all Hostname-related settings.

slaskawi:
@codex review

src/keycloak/chart/templates/statefulset.yaml
chance-coleman: I moved the native Keycloak hostname settings in Core to the KC_* environment variable convention:
```
- KC_HOSTNAME
- KC_HOSTNAME_ADMIN
- KC_HOSTNAME_BACKCHANNEL_DYNAMIC
- KC_HOSTNAME_STRICT
```
I left `UDS_DOMAIN`, `UDS_ADMIN_DOMAIN`, and `UDS_CLUSTER_DOMAIN` unchanged because those are UDS/plugin inputs, not native Keycloak hostname settings. The Core Helm tests pass with this change.

src/keycloak/zarf.yaml
```diff
@@ -36,8 +36,7 @@ components:
           - "values/upstream-values.yaml"
     images:
       - quay.io/keycloak/keycloak:26.7.2
-      - ghcr.io/defenseunicorns/uds/identity-config:0.30.0
-
+      - ttl.sh/uds-identity-config-core-672-20260824-e0dc391:24h
```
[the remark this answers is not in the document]
chance-coleman: this ttl image will be removed before merge.

test/vitest/keycloak-hostname.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+    });
+  });
+
+  it("uses the admin origin for admin realm discovery", async () => {
+    const response = await fetch(`${adminOrigin}/realms/uds/.well-known/openid-configuration`);
+
+    expect(response.ok).toBe(true);
+    await expect(response.json()).resolves.toMatchObject({
+      issuer: `${adminOrigin}/realms/uds`,
+    });
```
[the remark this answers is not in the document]
chance-coleman: I believe this is a false positive. The full Core test matrix passes with the admin discovery issuer set to keycloak.admin.uds.dev, which is the intended behavior for requests through the admin gateway. Changing this to the public issuer would regress the admin hostname behavior this PR is adding.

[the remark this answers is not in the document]
chance-coleman: The stock HostnameV2Provider behavior you cited is correct, but Core is using the UDSHostnameProvider from Identity Config 0.31.0. That provider overrides getBaseUri for UrlType.FRONTEND requests and rewrites admin-gateway requests to hostname-admin. OIDCWellKnownProvider builds the issuer from UrlType.FRONTEND, so admin discovery is expected to return keycloak.admin.uds.dev.

The full Core matrix also passes this test with Identity Config 0.31.0, so I’m leaving the assertion unchanged.

slaskawi (approved):

joelmccoy (approved):

PR #2930, opened 2026-09-01
fix: order metrics server after gateway

slaskawi:
## Description

Private PKI deployment intermittently failed while Helm was validating a following chart. The saved k3d control-plane log shows that Metrics Server registered the aggregated `v1beta1.metrics.k8s.io` API before its Service had an endpoint. Kubernetes then returned HTTP 503 while its aggregation controller fetched that API's OpenAPI specification:

```
Error updating APIService "v1beta1.metrics.k8s.io" with err: failed to retrieve openAPI spec, http error: ResponseCode: 503, Body: service unavailable
error resolving metrics-server/metrics-server: no endpoints available for service "metrics-server"
```

Helm downloads the cluster OpenAPI schema before applying every chart. Once the incomplete Metrics Server API is included in discovery, that download fails and prevents the next chart (observed as Keycloak or Envoy Gateway) from installing.

This change deploys Metrics Server last in the standard package. It is the only component that registers an aggregated API, so no later Helm chart performs OpenAPI validation during its registration window. Before the deployment returns, it waits for recovery and then requires 60 consecutive seconds of successful server-side `kubectl apply` validation, protecting the immediately following package validation and test actions with the same client path that failed.

Current validation uncovered a separate blocker: the Unicorn/FIPS Metrics Server image publishes an OpenAPI document rejected by k3s 1.35 (`ArbitraryKeys`, `SchemaProps`, and `SwaggerSchemaProps` are invalid). The upstream and registry1 images pass the same check. A corrected Unicorn image is required before this PR can be green; the readiness check intentionally does not hide that incompatibility.

This is needed to make standard bundle and Private PKI deployments deterministic without bypassing Kubernetes schema validation or relying on retry-only CI behavior.

## Related Issue

Relates to failed workflow run [33461886447](https://github.com/defenseunicorns/uds-core/actions/runs/33461886447).

## Type of change

## Steps to Validate

- Run `yamllint packages/standard/zarf.yaml -c .yamllint --no-warnings`.
- Run `yamllint src/metrics-server/zarf.yaml -c .yamllint --no-warnings`.
- Run the Private PKI workflow for all three flavors.
- Verify server-side validation is successful after Metrics Server deploys.
- Confirm server-side `kubectl apply --dry-run=server` succeeds continuously for 60 seconds after Metrics Server deploys.

## Checklist before merging

### DIFF
--- a/.github/workflows/private-pki-test.yaml
+++ b/.github/workflows/private-pki-test.yaml
@@ -14,6 +14,7 @@ on:
     paths:
       - ".github/workflows/private-pki-test.yaml"
       - "bundles/k3d-standard/**"
+      - "packages/standard/**"
       - "test/playwright/private-pki/**"
 
 concurrency:
--- a/.github/workflows/private-pki-test.yaml
+++ b/.github/workflows/private-pki-test.yaml
@@ -15,7 +15,6 @@ on:
       - ".github/workflows/private-pki-test.yaml"
       - "bundles/k3d-standard/**"
       - "packages/standard/**"
-      - "src/metrics-server/**"
       - "test/playwright/private-pki/**"
 
 concurrency:
--- a/.github/workflows/private-pki-test.yaml
+++ b/.github/workflows/private-pki-test.yaml
@@ -15,6 +15,7 @@ on:
       - ".github/workflows/private-pki-test.yaml"
       - "bundles/k3d-standard/**"
       - "packages/standard/**"
+      - "src/metrics-server/**"
       - "test/playwright/private-pki/**"
 
 concurrency:
--- a/packages/standard/zarf.yaml
+++ b/packages/standard/zarf.yaml
@@ -163,16 +163,25 @@ components:
       onDeploy:
         after:
           - description: Wait for cluster OpenAPI validation to stabilize
-            maxTotalSeconds: 180
-            maxRetries: 3
+            maxTotalSeconds: 300
             cmd: |
-              for _ in $(seq 1 30); do
-                printf '%s\n' \
+              stable_checks=0
+              for attempt in $(seq 1 120); do
+                if printf '%s\n' \
                   'apiVersion: v1' \
                   'kind: ConfigMap' \
                   'metadata:' \
                   '  name: openapi-stability-probe' \
                   '  namespace: default' |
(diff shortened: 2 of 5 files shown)


slaskawi:
### Root-cause report: Unicorn metrics-server v0.9.0 OpenAPI failure

This is a Chainguard image build defect, not a UDS NetworkPolicy or deployment-order issue.

Evidence:

- The live Unicorn pod is Ready and `v1beta1.metrics.k8s.io` is Available, so connectivity and APIService health checks pass.
- Its aggregated OpenAPI document contains an invalid `io.k8s.apimachinery.pkg.api.resource.Quantity_v2` definition with top-level fields `ArbitraryKeys`, `SchemaProps`, and `SwaggerSchemaProps`. This exactly matches the kube-apiserver error from CI.
- The Unicorn `cgr.dev/defenseunicorns.com/metrics-server-fips:v0.9.0` binary is built with Go 1.27.0, reports module `v0.9.0+dirty`, and includes `encoding/json/v2`.
- The official `registry.k8s.io/metrics-server/metrics-server:v0.9.0` binary is built with Go 1.26.4. The v0.9.0 source also declares Go 1.26.4: https://github.com/kubernetes-sigs/metrics-server/blob/v0.9.0/go.mod#L3
- Its kube-openapi dependency serializes these named fields using `json:",inline"`: https://github.com/kubernetes/kube-openapi/blob/aa012df4f4af/pkg/validation/spec/schema.go#L490-L512
- Go 1.27 renamed the JSON v2 tag from `inline` to `embed`: https://go.dev/doc/go1.27#encodingjsonv2

Result: the Go 1.27 build treats `inline` as an unknown option and serializes the wrapper field names instead of embedding their contents. Kube-apiserver rejects that malformed aggregated schema and intermittently returns 503 for `/openapi/v2`, which makes normal manifest validation fail.

Required fix: rebuild the Chainguard v0.9.0 image with the upstream-supported Go 1.26.4 toolchain, or update its kube-openapi dependency/source for Go 1.27 `embed` semantics. UDS-side ordering, retries, `--validate=false`, or a version downgrade only mask the defective image and are not appropriate fixes.

slaskawi:
Closing as we're waiting for a fix from Chainguard.

PR #2940, opened 2026-09-03
chore: manage Gateway API resources with stable Helm chart

joelmccoy:
## Description

When using manifests in zarf packages the helm chart name is generated based on the content causing the helm chart name to change.  We should avoid using manifests directly as it can cause some ownership issues when using different resource managers.  This swaps the gateway api resources to a local stable helm chart.  This is needed for a clean migration to using cli next to deploy on top of cli legacy packages.

Instead of reference the manifests via a remote link, I updated our CI/tasks to pull down and sync them in the repo.

## Type of change

## Steps to Validate
- `uds run test-uds-core-upgrade --set FLAVOR=upstream --no-progress`

## Checklist before merging


src/istio/gateway-api-crds/chart/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright 2026 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+apiVersion: v2
+name: gateway-api-crds
+version: 0.1.0
+description: Gateway API experimental CRDs
```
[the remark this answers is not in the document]
joelmccoy: fixed

joelmccoy:
@greptileai @codex review

joelmccoy:
@greptileai review

joelmccoy:
@greptileai @codex review

src/istio/gateway-api-crds/chart/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright 2026 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+apiVersion: v2
+name: gateway-api-crds
```
chance-coleman, line 5: Curious what the lifecycle is here for the old zarf-* release? The upgrade leaves it installed alongside `gateway-api-crds`, just want to confirm the intended behavior when the legacy package is later removed

joelmccoy: hmmm. this is a good callout. for the velero swap it seems we left the helm release in. i don't think there is much harm in leaving in the helm release. I hesitate to do a helm uninstall as it might try to cleanup these resources we changed ownership for.

I think our best option is to just leave it around unless there is another clever suggestion here. we could maybe delete the helm secret specifically?

chance-coleman: yeah agreed, helm uninstall probably isnt best option. Could we delete the Helm release records for old release after adoption instead? That retires the old release while leaving the resources intact, and i think Zarf treats a missing release as already removed during a later package removal? Also fine just leaving it, i just like the idea of keeping things cleaned up.

joelmccoy: i went and added a command to cleanup the helm secret and added some e2e tests to verify the behavior of this whole process.

joelmccoy:
@codex @greptileai review

chance-coleman (dismissed):

chance-coleman (approved):

PR #2947, opened 2026-09-09
docs(release-notes): clarify reset password links

slaskawi:
## Description

Clarifies the UDS Core 1.12 release note with the CORE-672 outcome: locked-out Registry users can reset passwords through the public SSO URL without disrupting Realm administration through the admin URL.

## Related Issue

Relates to [CORE-672](https://linear.app/defense-unicorns/issue/CORE-672/fix-the-reset-password-link)

## Type of change

## Steps to Validate

- Run `git diff --check` and confirm `docs/operations/release-notes/1-12.mdx` describes the Reset Password link use case.

## Checklist before merging


docs/operations/release-notes/1-12.mdx
```diff
@@ -13,7 +13,7 @@ UDS Core 1.12 gives you more control over network access and Keycloak client con
 ### Notable features
 
 - **Envoy Gateway network access:** Set `additionalNetworkAllow` to add `network.allow` entries to the Envoy Gateway `Package`, enabling integrations such as Envoy AI Gateway. See [Configure Core network access](/how-to-guides/networking/configure-core-network-access/) ([#2891](https://github.com/defenseunicorns/uds-core/pull/2891)).
-- **Separate Keycloak public and admin hostnames:** Keycloak now serves public frontend URLs from `sso.<domain>` and administrative frontend URLs from `keycloak.<admin_domain>`. UDS Core defaults the admin domain to `admin.<domain>`, and you can configure it explicitly ([#2898](https://github.com/defenseunicorns/uds-core/pull/2898)).
+- **Reset Password links:** Registry users who are locked out can now reset their password through the public SSO URL without disrupting Realm administration through the admin URL ([#2898](https://github.com/defenseunicorns/uds-core/pull/2898)).
```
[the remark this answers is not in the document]
slaskawi: I think the edited text sounds better. Closing this one.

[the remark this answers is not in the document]
slaskawi: Yes, and that's the point! It was too low-level!

chance-coleman: I think `Registry` is a typo? assuming you meant `Registered users.. `

joelmccoy: I'm not sure this is specific to "registry users"... And wasn't this change more than just reset password links?  This was to route and template things better based on whether you were accessing keycloak from tenant vs admin path.  cc: @chance-coleman can you confirm?

joelmccoy (commented):
If we make this change on main, we will want to also make this change on the release branch (release-1.12)

docs/operations/release-notes/1-12.mdx
chance-coleman: Yeah. The reset password link was the symptom that led to this, but the actual change is broader request-aware hostname routing for the frontend URLs.  Keycloak now determines which gateway the request came through and uses the corresponding public or admin hostname when generating links, instead of applying one hostname to every request.

slaskawi: Changed `Registry users` to `Registered users` in the [release-note bullet](https://github.com/defenseunicorns/uds-core/blob/docs/release-notes-reset-password-links/docs/operations/release-notes/1-12.mdx#L16).

slaskawi: I'm assuming there's no action for this in this PR. Let me resolve.

docs/operations/release-notes/1-12.mdx
```diff
@@ -13,7 +13,7 @@ UDS Core 1.12 gives you more control over network access and Keycloak client con
 ### Notable features
 
 - **Envoy Gateway network access:** Set `additionalNetworkAllow` to add `network.allow` entries to the Envoy Gateway `Package`, enabling integrations such as Envoy AI Gateway. See [Configure Core network access](/how-to-guides/networking/configure-core-network-access/) ([#2891](https://github.com/defenseunicorns/uds-core/pull/2891)).
-- **Separate Keycloak public and admin hostnames:** Keycloak now serves public frontend URLs from `sso.<domain>` and administrative frontend URLs from `keycloak.<admin_domain>`. UDS Core defaults the admin domain to `admin.<domain>`, and you can configure it explicitly ([#2898](https://github.com/defenseunicorns/uds-core/pull/2898)).
+- **Reset Password links:** Registered users who are locked out can now reset their password through the public SSO URL without disrupting Realm administration through the admin URL ([#2898](https://github.com/defenseunicorns/uds-core/pull/2898)).
```
jasonwashburn, line 16: Apologize if I'm just behind on this or missing some greater part of the fix. But my understanding, based on previous experience from operating registry, was that users _could_ always reset their passwords (when they weren't already locked out) and the email they received worked correctly and had the correct `sso.<domain>` link. 

The problem, as I remember it, was _when_ a user was locked out completely, and an admin tried to trigger a password reset email for a user from the admin console at `keycloak.admin.<domain>` ...the email the user received contained a password reset link to `keycloak.admin.<domain>` instead of `sso.<domain>` which they weren't able to use.

I'm honestly not really clear on what we mean by `...can now reset their password through the public SSO URL without disrupting Realm administration through the admin URL` here?

Assuming I'm not missing some additional functionality that was restored... I would think the fixed functionality would be something more like 
"... user password reset emails triggered by administrators via the admin console will now contain a useable link to the correct, user-facing `sso.<domain>` url rather than `keycloak.admin.<domain>`."

slaskawi: Agreed, the [release note](https://github.com/defenseunicorns/uds-core/blob/docs/release-notes-reset-password-links/docs/operations/release-notes/1-12.mdx#L16) is unclear about the reset flow. The hostname checks don't establish that administrator-triggered emails use the public URL, so I'm holding the wording change until we confirm CORE-672's outcome.

slaskawi: You're right: self-service reset emails already used `sso.<domain>`. The fix covers administrator-triggered emails for locked-out users that previously pointed to the admin hostname; I [updated the release note](https://github.com/defenseunicorns/uds-core/blob/docs/release-notes-reset-password-links/docs/operations/release-notes/1-12.mdx#L16) to make that distinction and describe the public/admin hostname hardening.

jasonwashburn (approved):

chance-coleman (approved):

PR #675, opened 2024-08-15
chore(neuvector): update source for unicorn images

mjnagel:
## Description

Switches neuvector unicorn flavor to use cgr images for all images except the scanner (this image is not supported by cgr). This requires a cert for internal comms which is generated by helm and valid for the same period of time as the upstream generated cert.

## Related Issue

Fixes https://github.com/defenseunicorns/uds-core/issues/568

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Other (security config, docs update, etc)

## Checklist before merging

- [x] Test, docs, adr added or updated as needed
- [x] [Contributor Guide](https://github.com/defenseunicorns/uds-template-capability/blob/main/CONTRIBUTING.md) followed

PR #2948, opened 2026-09-09
feat(ci): test K3s 1.34-1.36 compatibility

slaskawi:
## Description

Adds a static Kubernetes compatibility matrix covering exact K3s releases v1.34.4+k3s1, v1.35.8+k3s1, and v1.36.4+k3s1 across the upstream, registry1, and unicorn flavors.

- Runs the full nine-cell matrix nightly, on demand, and when the compatibility workflow changes.
- Deploys the requested K3s version through uds-k3d and verifies the cluster version before testing.
- Updates the default uds-k3d package to 0.20.3.
- Keeps registry1 Portal exclusions in the calling workflows.
- Documents tested K3s versions and validates documentation against the matrix.
- Configures Renovate to update K3s and uds-k3d versions.

## Related Issue

Relates to #675

## Type of change

## Steps to Validate

Verified in CI

## Checklist before merging


slaskawi:
@codex review

slaskawi:
@greptileai review

.github/workflows/test-kubernetes-matrix.yaml
```diff
@@ -0,0 +1,163 @@
...
+        with:
+          ghToken: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Parse and verify matrix
+        id: matrix
+        env:
+          # Supported Kubernetes minors: https://kubernetes.io/releases/
+          # Initial set only: these are the only versions backed by required uds-k3d custom images.
+          # renovate: datasource=github-releases depName=k3s-io/k3s versioning=regex:^v(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)\+k3s(?<build>\d+)$
+          K3S_VERSION_1: "v1.34.4+k3s1"
```
slaskawi: This looks very cumbersome. We should define a matrix of the k3d versions and flavors somewhere at the top of the workflow file. This way, it will be much easier for further maintenance.

.github/workflows/test-kubernetes-matrix.yaml
```diff
@@ -0,0 +1,163 @@
...
+          # renovate: datasource=github-releases depName=k3s-io/k3s versioning=regex:^v(?<major>\d+)\.(?<minor>\d+)\.(?<patch>\d+)\+k3s(?<build>\d+)$
+          K3S_VERSION_2: "v1.35.8+k3s1"
+          K3S_VERSIONS_INPUT: ${{ inputs.k3s_versions }}
+          FLAVORS_INPUT: ${{ inputs.flavors }}
+        shell: bash
+        run: |
+          set -euo pipefail
```
slaskawi: We don't need this amount of validation. If something is wrong - it will fail as part of the job. This seems like an excessive handling of an error case.

.github/workflows/test.yaml
```diff
@@ -78,7 +83,7 @@ jobs:
 
       - name: Test UDS Core values deploy equivalency
         if: ${{ inputs.package == 'all' && inputs.test_type == 'values' }}
-        run: uds run test:uds-core-values-e2e --set FLAVOR=${{ inputs.flavor }} --no-progress
+        run: uds run test:uds-core-values-e2e --set FLAVOR=${{ inputs.flavor }} --set EXCLUDED_PACKAGES=${{ inputs.flavor == 'registry1' && 'portal' || '' }} --with k3s_version=${{ inputs.k3s_version }} --no-progress
```
slaskawi, line 93: The Portal exclusion seems to be shared across all the jobs. As a matter of fact, invoking `uds run test:uds-core-values-e2e` with the registry1 flavor doesn't make sense. 

We should modify the task itself to perform the exclusion and simplify the CI job invocations.

tasks/setup.yaml
```diff
@@ -3,19 +3,55 @@
...
         # renovate: datasource=docker depName=ghcr.io/defenseunicorns/packages/uds-k3d versioning=docker
+        if: ${{ eq .inputs.k3s_version "" }}
         cmd: "./uds zarf package deploy oci://defenseunicorns/uds-k3d:0.20.2-airgap --confirm"
 
+      - description: "Create the requested K3d cluster"
+        if: ${{ ne .inputs.k3s_version "" }}
+        shell:
+          darwin: bash
+          linux: bash
+        cmd: |
```
slaskawi, line 20: This command should be collapsed with the previous one. Right now it's very hard to read this.

.github/workflows/test-kubernetes-matrix.yaml
```diff
@@ -0,0 +1,163 @@
...
+      matrix:
+        k3s_version: ${{ fromJSON(needs.prepare.outputs.k3s_versions) }}
+        flavor: ${{ fromJSON(needs.prepare.outputs.flavors) }}
+    uses: ./.github/workflows/test.yaml
+    with:
+      package: all
+      flavor: ${{ matrix.flavor }}
+      test_type: values
+      k3s_version: ${{ matrix.k3s_version }}
+    secrets: inherit
```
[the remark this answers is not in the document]
slaskawi: Fixed by declaring the reusable workflow’s required secrets and passing only those from the matrix caller.

renovate.json
```diff
@@ -182,6 +182,17 @@
...
       "allowedVersions": "<10"
     },
+    {
+      "matchFileNames": [
+        ".github/workflows/test-kubernetes-matrix.yaml"
+      ],
+      "matchDepNames": [
+        "k3s-io/k3s"
+      ],
+      "allowedVersions": "/^v{{major}}\\.{{minor}}\\.\\d+\\+k3s\\d+$/",
```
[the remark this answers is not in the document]
slaskawi: No change: Renovate templates `allowedVersions` with each dependency’s parsed major and minor before filtering candidates, so these entries remain constrained to same-minor K3s updates.

tasks/lint.yaml
```diff
@@ -209,8 +209,42 @@ tasks:
         cmd: |
           set -euo pipefail
           DOC_FILE="docs/concepts/platform/supported-distributions.mdx"
+          MATRIX_WORKFLOW=".github/workflows/test-kubernetes-matrix.yaml"
           FAILED=0
 
+          MATRIX_VERSIONS=$(sed -n 's/^[[:space:]]*K3S_VERSION_[0-9][0-9]*: "\([^"]*\)"$/\1/p' "$MATRIX_WORKFLOW")
+          MATRIX_FLAVORS="upstream
+          registry1
+          unicorn"
```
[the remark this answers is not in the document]
slaskawi: No change: YAML strips the block-scalar indentation before Bash executes it, and `uds run lint:k8s-version-check --no-progress` passes.

docs/concepts/platform/supported-distributions.mdx
```diff
@@ -8,14 +8,16 @@ sidebar:
...
+> UDS Core tests EKS, AKS, and RKE2 against **Kubernetes 1.35**. The target is typically **n-1** (one minor version behind the latest release, latest patch). This version may lag slightly behind new Kubernetes releases.
 
 | Distribution | K8s Version | Status | Testing Schedule |
 |-------------|-------------|--------|-----------------|
-| [K3s](https://k3s.io/) / [k3d](https://k3d.io/) | **1.35** | [![K3d HA Test](https://github.com/defenseunicorns/uds-core/actions/workflows/test-k3d-ha.yaml/badge.svg?branch=main&event=schedule)](https://github.com/defenseunicorns/uds-core/actions/workflows/test-k3d-ha.yaml?query=event%3Aschedule+branch%3Amain) | Nightly and before each release |
+| [K3s](https://k3s.io/) / [k3d](https://k3d.io/) | **1.34**, **1.35** | [![Kubernetes Compatibility](https://github.com/defenseunicorns/uds-core/actions/workflows/test-kubernetes-matrix.yaml/badge.svg?branch=main&event=schedule)](https://github.com/defenseunicorns/uds-core/actions/workflows/test-kubernetes-matrix.yaml?query=event%3Aschedule+branch%3Amain) | Nightly and on release-please pull requests |
```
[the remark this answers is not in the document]
slaskawi: No change: the table summarizes coverage cadence, while the path-filtered PR triggers are workflow implementation details.

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -185,7 +185,10 @@ jobs:
       package: ${{ matrix.package }}
       flavor: ${{ matrix.flavor }}
       test_type: ${{ matrix.test_type }}
-    secrets: inherit # Inherits all secrets from the parent workflow.
+    secrets:
```
slaskawi: This change seems out of the scope. Can we remove it?

tasks/test.yaml
```diff
@@ -180,11 +180,21 @@ tasks:
 
   - name: uds-core-values-e2e
     description: "Build and test UDS Core deployed via Zarf values (equivalency to bundle deploy)"
+    inputs:
+      k3s_version:
+        default: ""
+        description: "Exact K3s version for the test cluster"
     actions:
+      - description: "Exclude Portal from registry1 values tests"
```
slaskawi: This should be moved to tasks.yaml rather than here in the pipeline.

tasks/lint.yaml
```diff
@@ -209,8 +209,41 @@ tasks:
         cmd: |
           set -euo pipefail
           DOC_FILE="docs/concepts/platform/supported-distributions.mdx"
+          MATRIX_WORKFLOW=".github/workflows/test-kubernetes-matrix.yaml"
           FAILED=0
 
+          MATRIX_JSON=$(./uds zarf tools yq -r '.env.K3S_MATRIX' "$MATRIX_WORKFLOW")
+          MATRIX_VERSIONS=$(printf '%s' "$MATRIX_JSON" | jq -r '.versions[]')
```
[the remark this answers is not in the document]
slaskawi: No change: `set -euo pipefail` already fails immediately if `jq` is unavailable, so an explicit preflight would only duplicate that behavior.

.github/workflows/test-kubernetes-matrix.yaml
```diff
@@ -0,0 +1,135 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+name: Kubernetes Compatibility
+
+env:
+  K3S_MATRIX: |-
+    {
+      "versions": [
+        "v1.34.4+k3s1",
+        "v1.35.8+k3s1"
```
joelmccoy: we should be adding 1.35 and 1.37 to this instead... 1.37 is current. we should test normally on 1.36 (need to bump uds-k3d). This matrix should test `n-2` and `n`

.github/workflows/test-kubernetes-matrix.yaml
```diff
@@ -0,0 +1,135 @@
...
+    types: [milestoned, opened, reopened, synchronize]
+    branches:
+      - main
+      - "release/**"
+    paths:
+      - ".release-please-manifest.json"
+      - ".github/workflows/test-kubernetes-matrix.yaml"
+      - ".github/workflows/test.yaml"
+      - "tasks/setup.yaml"
+      - "tasks/test.yaml"
```
joelmccoy: I think it is overkill to include these. These will trigger on way more prs than we intend.  And we are generally more conservative. I am ok trigger in the `.github/workflows/test-kubernetes-matrix.yaml` but i think everything else is not necessary.

.github/workflows/test-kubernetes-matrix.yaml
```diff
@@ -0,0 +1,135 @@
...
+          fi
+          {
+            echo "k3s_versions<<EOF"
+            echo "$k3s_versions"
+            echo "EOF"
+            echo "flavors<<EOF"
+            echo "$flavors"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
```
joelmccoy: can't we get ride of this whole prepare step?? i feel like it is overkill if we can just use a native github matrix like so:

```yaml
   jobs:
     test:
       name: Kubernetes compatibility (${{ matrix.k3s_version }}, ${{ matrix.flavor }})
       strategy:
         fail-fast: false
         matrix:
           k3s_version:
             - v1.34.4+k3s1
             - v1.35.8+k3s1
           flavor:
             - upstream
             - registry1
             - unicorn
       uses: ./.github/workflows/test.yaml
```

.github/workflows/test-kubernetes-matrix.yaml
joelmccoy: going to bump this here: https://github.com/defenseunicorns/uds-k3d/pull/382

joelmccoy (commented):
A couple initial comments that might effect the shape of this pr

.github/workflows/pull-request-conditionals.yaml
```diff
@@ -185,7 +185,7 @@ jobs:
       package: ${{ matrix.package }}
       flavor: ${{ matrix.flavor }}
       test_type: ${{ matrix.test_type }}
-    secrets: inherit # Inherits all secrets from the parent workflow.
```
slaskawi, line 188: This should be reverted to the origin/main

.github/workflows/test.yaml
```diff
@@ -34,6 +34,18 @@ on:
         type: string
         description: "The type of test to perform"
         required: true
+      k3s_version:
+        type: string
+        description: "Exact K3s version for values tests"
+        required: false
+        default: ""
```
slaskawi, line 41: It's very hard to track what is the default Kubernetes version we're testing with. We need to use one default (for the non-matrix jobs) explicitly somewhere. It must be easy to spot.

.github/workflows/test-kubernetes-matrix.yaml
slaskawi: Updated normal coverage to uds-k3d 0.20.3 / K3s 1.36.4 and expanded the matrix across 1.34, 1.35, and 1.36 using the latest available custom images. K3s 1.37 is not released yet.

.github/workflows/test-kubernetes-matrix.yaml
slaskawi: Limited pull request runs to compatibility workflow changes and updated the documented schedule to match.

.github/workflows/test-kubernetes-matrix.yaml
slaskawi: Replaced the prepare job with the visible static matrix at the top of the workflow. Manual dispatch remains available and now always runs all nine combinations.

renovate.json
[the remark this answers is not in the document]
slaskawi: No change: Renovate documents `allowedVersions` as templatable and compiles it per dependency with the current version's `major`, `minor`, and `patch`. Each matrix entry is extracted independently, so this rule keeps 1.34, 1.35, and 1.36 on their respective minor lines.

.github/workflows/test-kubernetes-matrix.yaml
```diff
@@ -0,0 +1,73 @@
...
+jobs:
+  test:
+    name: Kubernetes compatibility (${{ matrix.k3s_version }}, ${{ matrix.flavor }})
+    strategy:
+      fail-fast: false
+      matrix:
+        k3s_version:
+          - v1.34.4+k3s1
+          - v1.35.8+k3s1
+          - v1.36.4+k3s1
```
joelmccoy: I think we can remove this one since we should test this on all prs anyways? we are bumping to 1.36 in this pr here: https://github.com/defenseunicorns/uds-core/pull/2953. Basically we should just test on two versions here? it should only cover the gap

.github/workflows/test-kubernetes-matrix.yaml
```diff
@@ -0,0 +1,73 @@
...
+  packages: read
+
+# Keep the complete compatibility matrix at the top of the workflow.
+jobs:
+  test:
+    name: Kubernetes compatibility (${{ matrix.k3s_version }}, ${{ matrix.flavor }})
+    strategy:
+      fail-fast: false
+      matrix:
+        k3s_version:
```
joelmccoy, line 33: Can we add a small job here to run `uds run lint:k8s-version-check`? This check may not run for matrix-only changes today, which could allow the supported distributions documentation to drift out of sync.

joelmccoy (commented):
Two small comments here. But mostly LGTM

.github/workflows/test-kubernetes-matrix.yaml
slaskawi: Added a dedicated [Kubernetes version validation job](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-675/.github/workflows/test-kubernetes-matrix.yaml#L52) and included it in the aggregate result.

.github/workflows/test-kubernetes-matrix.yaml
slaskawi: Removed 1.36 from the dedicated compatibility matrix because the [normal and nightly suites](https://github.com/defenseunicorns/uds-core/blob/sebastianlaskawiec/core-675/.github/workflows/test-k3d-ha.yaml#L29) already test that default across all flavors. Documentation validation now combines the gap matrix with the configured default.

chance-coleman (dismissed):

chance-coleman (approved):

PR #2951, opened 2026-09-10
chore: upgrade eks test to k8s 1.36

ntwkninja:


slaskawi:
I believe we'll address this in a broader change here: https://github.com/defenseunicorns/uds-core/pull/2953

Thanks for the PR though!

PR #2962, opened 2026-09-14
feat: support disabling 421 envoyfilter

mjnagel:
## Description

The `misdirected-request` EnvoyFilter in Core returns HTTP `421` when an HTTP/2 request’s `:authority` does not match the TLS SNI that established the connection. It prevents clients from reusing a connection opened for one hostname to send requests for another hostname. Without the check, the gateway that accepted the original TLS connection may process routing for the different authority. This matters in the context of Core when tenant and passthrough gateways have overlapping certificates (or other custom gateways that may overlap). In these cases the existing TLS connection cannot be reused as the connection needs to go through a separate TLS handshake (for passthrough handling) and likely a different loadbalancer endpoint.

This PR modifies this EnvoyFilter:
- Moves the filter into its own Helm template and simplifies its Lua logic without changing the default global behavior
- Adds `misdirectedRequest.enabled` to disable the filter 

This EnvoyFilter is often not applicable to environments and can cause issues with Safari based clients in particular since they do not retry the connection when receiving a 421. By adding a conditional to disable it, environments can opt out of this behavior without changing the default.

## Type of change


src/istio/common/chart/templates/misdirected-request-envoyfilter.yaml
```diff
@@ -0,0 +1,45 @@
...
+{{- $misdirectedRequest := .Values.misdirectedRequest }}
+{{- if $misdirectedRequest.enabled }}
+{{- $gateways := $misdirectedRequest.gateways | default (list "") }}
+{{- range $gateway := $gateways }}
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: misdirected-request
+  namespace: {{ if $gateway }}istio-{{ $gateway }}-gateway{{ else }}istio-system{{ end }}
```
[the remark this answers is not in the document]
mjnagel: Added the gateway template in here. I had been thinking Zarf did this for us, but definitely was wrong.

packages/base/zarf-values.schema.json
```diff
@@ -927,6 +927,17 @@
...
               "type": "string"
+            },
+            "misdirectedRequest": {
+              "properties": {
+                "enabled": {
+                  "type": "boolean"
+                },
+                "gateways": {
+                  "type": "array"
+                }
```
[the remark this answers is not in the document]
mjnagel: Added basic non-empty validation. I don't think we want enum here or anything too specific since it would prevent using with custom gateways?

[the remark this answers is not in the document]
mjnagel: Yep that seems legit. added.

src/istio/common/chart/templates/misdirected-request-envoyfilter.yaml
```diff
@@ -0,0 +1,52 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+{{- $misdirectedRequest := .Values.misdirectedRequest }}
+{{- if $misdirectedRequest.enabled }}
+{{- $gateways := $misdirectedRequest.gateways | default (list "") }}
+{{- range $gateway := $gateways }}
+{{- if $gateway }}
+---
+apiVersion: v1
+kind: Namespace
```
chance-coleman: Could this common chart's ownership of `istio-<gateway>-gateway` conflict with the separate gateway components that use those namespaces, causing upgrades to fail or disabling this setting to delete the gateway namespace and its resources?

src/istio/common/chart/values.schema.json
```diff
@@ -43,6 +43,22 @@
...
+    "misdirectedRequest": {
+      "type": "object",
+      "properties": {
+        "enabled": {
+          "type": "boolean"
+        },
+        "gateways": {
+          "type": "array",
+          "uniqueItems": true,
+          "items": {
```
chance-coleman: Could values such as `Tenant`, `tenant/foo`, or an overly long string pass this schema and then make the generated Namespace or `app` label invalid, since the schema currently checks only `minLength` and `uniqueItems`?

chance-coleman (commented):
couple questions as im wrapping my head around this

src/istio/common/chart/templates/misdirected-request-envoyfilter.yaml
mjnagel: Short answer: no I don't believe so.

Longer answer:

Ultimately these namespace templates are part of the same zarf component as the gateways (for standard core gateways) and the "prereqs" to run gateways (Istio). Ordering wise the ns template here is created first and would be removed last. So (assuming namespaces got handled normally) I think the timing here would be fine for deletion.

More broadly in terms of ownership - namespaces aren't really owned by helm releases, at least in Zarf land. [Here's](https://github.com/zarf-dev/zarf/blob/v0.84.0/src/internal/packager/helm/post-render.go#L217-L230) a bit of the code to back that up - they get filtered and handled specially, and removed "from helm's brain" 😄. If there are further concerns here I'm happy to test a specific scenario although I'm unsure on a good way to hit a timing issue (per the other paragraph).

src/istio/common/chart/values.schema.json
mjnagel: Added a restriction that I think covers this relatively well. I know greptile commented on further restrictions earlier and I think there's probably a balance here in trying to bake too much into the schema vs just trusting the user (and this would fail at apply time quickly too). I do like some good basic schema adds like this one tho.

src/istio/common/chart/values.yaml
```diff
@@ -29,3 +29,13 @@ adminDomain: '{{default (default (printf "admin.%s" (default .Variables.DOMAIN .
...
+
+# Reject HTTP/2 requests whose :authority differs from the TLS SNI that selects
+# a gateway. This prevents a connection established for one gateway from being
+# reused for a hostname owned by another gateway.
+misdirectedRequest:
+  enabled: true
+  # An empty list applies the filter to all gateway workloads from istio-system.
+  # Each name selects an ingress gateway using istio-<name>-gateway and
+  # <name>-ingressgateway. Egress gateways are not selected by this setting.
+  gateways: []
```
joelmccoy: I think in order to simplify things we can remove the gateways field here and just have this as a general on/off switch.  I'm not seeing the value in allowing this per gateway. Generally we really shouldn't need this feature on for out of the box UDS core deployments.  And for cases where there is no overlap I think it is ok to disable this envoyfilter across the board

src/istio/common/chart/values.yaml
```diff
@@ -29,3 +29,13 @@ adminDomain: '{{default (default (printf "admin.%s" (default .Variables.DOMAIN .
 # Optional annotations for grouping dashboards in Grafana folders.
 # see docs for more info: https://docs.defenseunicorns.com/core/how-to-guides/monitoring--observability/add-custom-dashboards/
 dashboardAnnotations: {}
+
+# Reject HTTP/2 requests whose :authority differs from the TLS SNI that selects
+# a gateway. This prevents a connection established for one gateway from being
+# reused for a hostname owned by another gateway.
+misdirectedRequest:
+  enabled: true
```
joelmccoy, line 38: as discussed in slack thread. I think the preferred default behavior for this would be to have this off by default?  Out of the box uds core deployments don't need this on.  Could we just add a deprecation in our `DEPRECATIONS.md` and track this for flipping this to off by default in UDS Core 2.0?

joelmccoy (commented):
Two small comments that will likely change the shape of this PR. Definitely happy to support having this as a toggle of some sort

src/istio/common/chart/values.schema.json
mjnagel: OBE.

src/istio/common/chart/templates/misdirected-request-envoyfilter.yaml
mjnagel: OBE.

src/istio/common/chart/values.yaml
mjnagel: Simplified to just the enabled true/false toggle.

src/istio/common/chart/values.yaml
mjnagel: Added.

joelmccoy:
@greptileai review

src/istio/common/chart/templates/misdirected-request-envoyfilter.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright 2026 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+{{- $misdirectedRequest := .Values.misdirectedRequest }}
+{{- if $misdirectedRequest.enabled }}
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: misdirected-request
```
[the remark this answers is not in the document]
joelmccoy: this is intentional to keep things simple

joelmccoy (approved):

chance-coleman (approved):

jeff-mccoy:
Thanks for working this so quickly gents

PR #2964, opened 2026-09-15
fix(ci): improving reliability and flaky test cases

chance-coleman:
## Description

This PR improves CI reliability by addressing Kubernetes reconciliation races, test-environment timing issues, and failure diagnostics.

Key changes:

- Make Save Logs non-gating so artifact-upload failures do not mask the actual test result.
- Ensure AKS provisioning waits for the NAT gateway public IP association before creating the cluster.
- Coalesce concurrent shared-egress package events and reconcile from a complete live package snapshot.
- Retry transient Kubernetes API reads and only replace in-memory egress state after a valid snapshot is available.
- Re-read orphaned resources before deletion and verify labels and generations to reduce cleanup race conditions.
- Improve network-test diagnostics by exposing `curl` connection and TLS errors.
- Prepare the active gateway serving CA for Node-based Vitest tests while retaining normal TLS verification.
- Run Vitest network tests separately from the remaining suites while preserving global setup and cleanup behavior.
- Ensure the temporary `policy-tests` namespace is deleted and wait for namespace termination during Vitest teardown.
- Preserve a 30-second Vitest teardown timeout for namespace cleanup.
- Add bounded HTTP requests with method and URL context for Vitest helper calls.
- Use a healthy in-cluster target for probe admission-policy fixtures so they are not coupled to unrelated public endpoint alert behavior.
- Add diagnostics for slow or failed Zarf manifest renders, including package, values, variables, duration, and process information. Render concurrency remains unchanged.
- Keep IaC pull-request triggering enabled; the temporary investigation gate is not included in this PR.

## Related Issue

Relates to [CORE-679](https://linear.app/defenseunicorns/issue/CORE-679/investigate-core-ci-flakiness)

## Type of change

## Steps to Validate

- `npx vitest run --reporter=dot` — root unit suite: 67 files and 1,234 tests passed locally.
- `npx eslint test/values/helpers.ts` passed.
- `git diff --check` passed.
- Cluster-backed Vitest, IaC, and flavor-matrix workflows should be validated by GitHub Actions.
- No local cluster was created or modified during validation.

## Checklist before merging


chance-coleman:
@codex @greptile review these changes

chance-coleman:
@codex @greptile review the changes and resolve your previous comments if addressed.

chance-coleman:
@greptile @codex review latest changes

test/vitest/vitest.setup.js
```diff
@@ -1,11 +1,15 @@
...
  * SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
  */
 
-import { K8s, kind } from 'pepr';
+import { K8s, kind } from "pepr";
+
+export async function setup(project) {
+  // The root project owns the shared namespace. Child projects inherit this
+  // setup, but must not create or delete the namespace independently.
+  if (!project.isRootProject()) return;
```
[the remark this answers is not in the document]
chance-coleman: I don’t believe this finding applies. Vitest documents that root-level globalSetup runs once per test run and is not inherited by child projects. isRootProject() is the supported API for identifying the root project. This guard ensures only the root setup creates and tears down the shared policy-tests namespace, preventing duplicate cleanup. [Vitest global setup](https://main.vitest.dev/config/globalsetup) [Vitest projects](https://main.vitest.dev/guide/projects)

joelmccoy:
Haven't done a deep dive into this yet, but I also noticed that the EKS tests took 5 reruns :/ with a timeout on the admin-gateway-deployment. Have we looked into this at all as a part of this discovery?  Is this something introduced by this PR or already existing?

Run: https://github.com/defenseunicorns/uds-core/actions/runs/35249139427/job/105381939968?pr=2964#step:14:300

chance-coleman:
existing unfortunately. I've hit this a number of times when we have lots of PRs that are triggering IAC tests because of the limit on number of load balancers in AWS. Happens when a load balancer becomes orphaned because of a failed cleanup but still counts towards our AWS quota. Waiting till after nightly nuke runs or manually cleaning them up has worked in the past for me.

joelmccoy:
Also noticed this flake: https://github.com/defenseunicorns/uds-core/actions/runs/35249139176/job/105304055311?pr=2964

And this flake: https://github.com/defenseunicorns/uds-core/actions/runs/35249139176/job/105298678897?pr=2964#step:5:198

Not sure these can be actioned on in this PR, but doesn't feel great that our PR to improve flakes is still having many flakes.  I wonder if we can add retries to individual steps like zarf package creates? or even deploys?  Doesn't need to be done this PR, but something to consider in the future.

chance-coleman:
Yeah, i considered adding generic retries for those since its an endpoint thats not accessible at the time of pull. I decided not too cause i was trying to avoid just adding retries in this PR, for this specific flake that is potentially the only fix though.

scripts/retry-zarf-package-create.sh
```diff
@@ -0,0 +1,53 @@
...
+readonly MAX_ATTEMPTS=2
+readonly RETRY_DELAY_SECONDS=15
+readonly TRANSIENT_REGISTRY_ERROR_PATTERN='https?://[^" ]+/v2/[^" ]*".*(i/o timeout|TLS handshake timeout|connection reset by peer|connection refused|unexpected EOF|no such host|network is unreachable)'
+
+attempt=1
+while [ "${attempt}" -le "${MAX_ATTEMPTS}" ]; do
+  log_file=$(mktemp)
+  echo "Creating ${PACKAGE_PATH} (attempt ${attempt}/${MAX_ATTEMPTS})"
+
+  if ZARF_CONFIG="${CONFIG}" uds zarf package create "${PACKAGE_PATH}" \
```
slaskawi: This script seems a bit complicated. Could we implement a really simple stop-the-gap solution, such as https://github.com/minfrin/retry, and contribute an upstream Zarf patch to introduce registry pull retries? 

This way, we'd end up with smaller patch for the meantime and a clear path forward for the future.

slaskawi (commented):
Generally LGTM but I'd vote for simplifying the retries script.

scripts/retry-zarf-package-create.sh
chance-coleman: I think this could make sense. The caveat is that we’d be agreeing to retry every failed package create at least once. If we only want to retry likely registry or network flakes, the current implementation is a better fit because it lets us be more selective. I see pros and cons to both approaches. What do you think? 

@joelmccoy, what’s your opinion?

joelmccoy: side note: Would getting this PR help our case here? https://github.com/zarf-dev/zarf/pull/5303

For this PR though I would get rid of this script. Feels yucky and I don't think we need that much complexity here. I would vote for a simple 1 time retry on a single failure of a zarf package create in our tasks (not a separate script).

Basically try to create the package once, and if it fails for any reason retry once.  This would keep it simple and address most flakes I imagine... And then maybe when that pr is merged upstream we can remove this single retry altogether?

chance-coleman: Yeah, I think [Zarf #5303](https://github.com/zarf-dev/zarf/pull/5303) is the right long-term direction since it retries individual OCI requests instead of rerunning the entire package create.

For this PR, I agree this script is way too complex. i'll add a simple one-time retry directly in the package-create task. That gives us a straightforward temporary fix, and we can remove it once the upstream Zarf change is released.

PR #2976, opened 2026-09-16
chore: modernize dev workflow w/ mise and hk

eoghanriley:
## Description

This PR moves to using mise to manage the lifecycle of dependency tools. This includes having users and CI run `mise install` to download and use pinned version of dev tools. Additionally we swap out husky for hk to bring core in line with our other repos. 

## Related Issue

Fixes CORE-690

## Type of change

## Steps to Validate

## Checklist before merging


eoghanriley:
Seems like the mise install for helm unit test installs it as a binary `untt` but not as a helm plugin. I have the tasks invoke it as untt, but will have the helm plugin install process in the CONTRIBUTING.md as it is meant to be used a helm plugin.

tasks/lint.yaml
```diff
@@ -152,7 +151,7 @@ tasks:
           for chart in $(find . -name Chart.yaml -type f -exec dirname {} \;); do
             if [ -d "$chart/tests" ]; then
               echo "Testing $chart"
-              if ! helm unittest "$chart" --color; then
+              if ! untt "$chart" --color; then
```
bradclawsie, line 139: for anyone else reading and wondering what `untt` is (I thought it was a typo!), it is installed via Helm unittest (https://github.com/helm-unittest/helm-unittest)

CONTRIBUTING.md
```diff
@@ -60,13 +55,19 @@ git checkout -b my-feature-branch
...
 We use pre-commit hooks to ensure code quality and consistency. These checks run automatically when you commit code and help catch issues early.
 
-**Setup Pre-Commit with Husky:**
+**Set up pre-commit with hk:**
 
 ```console
-# Initialize husky hooks (only needed once after cloning)
-npx husky
+# Install pinned development tools
+mise install
```
bradclawsie: Is there a need for the `mise trust` step here?

tasks/lint.yaml
eoghanriley: Yes I added a note here for some context on why this is and why I left in the direction in the contributing.md to still install it as a helm plugin. https://github.com/defenseunicorns/uds-core/pull/2976#issuecomment-5702088336

CONTRIBUTING.md
eoghanriley: Good callout, fixed in both spots where we direct users to `mise install`

bradclawsie (dismissed):

.github/actions/setup-tools/action.yaml
```diff
@@ -3,67 +3,14 @@
...
 
 runs:
   using: "composite"
   steps:
-    # We always need UDS CLI for running tasks
-    - name: Install UDS CLI
-      uses: defenseunicorns/setup-uds@d6add59d4c2ca789ed7c92741e91bd340b4e950b # v1.0.2
+    - uses: jdx/mise-action@c2a87611a18de5b3828c5652fe268e992400cb5c # v4.3.0
       with:
-        # renovate: datasource=github-tags depName=defenseunicorns/uds-cli versioning=semver
```
joelmccoy, line 32: red diff here feels soo good :)

CONTRIBUTING.md
```diff
@@ -80,6 +81,9 @@ Our pre-commit and linting checks validate:
 The easiest way to install all required dependencies is:
 
 ```console
+# Install pinned development tools
+mise trust && mise install
+
 # Install the helm-unittest plugin
 helm plugin install https://github.com/helm-unittest/helm-unittest.git
```
joelmccoy: do we still need this??

mise.toml
```diff
@@ -0,0 +1,29 @@
...
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+[tools]
+helm = "4.2.4"
+hk = "2.0.0"
+jq = "1.8.2"
+k3d = "5.9.0"
+node = "24.21.0"
+opentofu = "1.12.6"
+uv = "0.12.10"
```
joelmccoy: do we need uv if we are installing deps directly in mise?

mise.toml
```diff
@@ -0,0 +1,29 @@
+# Copyright 2026 Defense Unicorns
+# SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+[tools]
+helm = "4.2.4"
```
joelmccoy: why do we have helm here? helm should be vedored in by uds cli -> zarf right? we should defer to that if possible unless there is a gap  somehwhere that I am not understanding

hk.pkl
```diff
@@ -0,0 +1,28 @@
+// Copyright 2026 Defense Unicorns
+// SPDX-License-Identifier: AGPL-3.0-or-later OR LicenseRef-Defense-Unicorns-Commercial
+
+amends "package://github.com/jdx/hk/releases/download/v2.0.0/hk@2.0.0#/Config.pkl"
+
+steps {
+    ["yamllint"] {
+        check = "uvx yamllint . -c .yamllint --no-warnings"
```
joelmccoy: do we need uvx if we are installing yamllint with mise?

hk.pkl
```diff
@@ -0,0 +1,28 @@
...
+        glob = List("**/*.yaml", "**/*.yml", ".yamllint")
+    }
+    ["codespell"] {
+        check = "uvx codespell"
```
joelmccoy: same comment about uvx

hk.pkl
```diff
@@ -0,0 +1,28 @@
...
+        fix = "uvx codespell -w"
+        glob = List("**/*")
+        exclude = List("build/**", "vendor/**", "uds-docs/**", "package-lock.json")
+    }
+    ["typescript-format"] {
+        check = "npx eslint {{files}} && npx prettier --check {{files}}"
```
joelmccoy, line 18: why npx here? should eslint be in mise and we use that version? maybe we can have mise install our package.json dev deps??

mise.toml
joelmccoy, line 1: we need to update our renovate to include all mise deps in support deps renovate grouping.

the only mise dep that affects other things may be opentofu (which shoudl be a part of IAC deps i believe)

mise.toml
```diff
@@ -0,0 +1,29 @@
...
+hk = "2.0.0"
+jq = "1.8.2"
+k3d = "5.9.0"
+node = "24.21.0"
+opentofu = "1.12.6"
+uv = "0.12.10"
+
+# renovate: datasource=pypi depName=codespell versioning=pep440
+"pypi:codespell" = "2.4.3"
```
joelmccoy, line 12: getting an error on pypi things. any additional steps i need to do here?

```console
mise install
mise ERROR pypi is not a valid plugin name
mise ERROR Version: 2026.8.2 macos-arm64 (2026-08-05)
mise ERROR Run with --verbose or MISE_VERBOSE=1 for more information
mise WARN  Failed to resolve tool version list for pypi:codespell: [~/github/uds-core.review-1/mise.toml] pypi:codespell@2.4.3: pypi is not a valid plugin name
mise WARN  Failed to resolve tool version list for pypi:yamllint: [~/github/uds-core.review-1/mise.toml] pypi:yamllint@1.38.0: pypi is not a valid plugin name
```

joelmccoy (commented):
A couple initial comments on first pass.

CONTRIBUTING.md
eoghanriley: I think you are calling out the helm plugin install (not sure since GH grabbed a couple lines there). See this comment for more context, but mise can't run the plugin install so it is there for people to invoke the unit tests through the helm cli. [https://github.com/defenseunicorns/uds-core/pull/2976#issuecomment-5702088336](https://github.com/defenseunicorns/uds-core/pull/2976#issuecomment-5702088336)

We can remove it, but the binary name for it that mise installs is untt which isn't really human and ai friendly so I figured its best to still include the step, but have mise manage it for CI and the official supported version.

hk.pkl
eoghanriley: Previously with husky we used `npx lint-staged`, which I swapped out in favor of eslint and prettier since hk gives us access to the staged files via `{{files}}`. Joonas pointed me towards [hk built in linters](https://hk.jdx.dev/builtins.html) which may be a better fit here. We still get version management through the package.json & node_modules so I don't think we neccessarily gain anything by moving it to mise strictly for that.

hk.pkl
bradclawsie: it still has to call out to a python module installer, true? So you either use `uv` version-managed by `mise`, or you need `pipx`?

mise.toml
eoghanriley: Thats interesting it worked out of the box for me, checking the docs doesn't seem like the pypi backend needs to be installed/enabled https://mise.jdx.dev/dev-tools/backends/pipx.html. Perhaps its a matter of mise version? I am currently running 2026.9.7

eoghanriley: In relation to your comments about removing uv from the config since I already have codespell and yamllint in there, if updating your mise doesn't fix it or we don't want to require that then I think we should just remove those 2 then. Otherwise I would prefer directly using codespell and yamlllint.

hk.pkl
eoghanriley: Seems like the hk built in linters only modify how you execute and configure them, but still uses mise or npm to install them. Personally I am in favor of still using npm to manage eslint and prettier but I don't feel too strongly about it, and would be willing to go the mise way. 

I think moving to the hk built in linter gives us slightly more speed as it separates eslint and prettier into parallel steps but thats about it. We already have the staged files functionality with the current command

CONTRIBUTING.md
joelmccoy: I'd vote for removing the helm dep and plugin if we can just have it as a binary. removes an extra dep. any linting or test tasks should be wrapped with a uds command anyways and we can just call the untt binary in the tasks

hk.pkl
joelmccoy: i was hoping mise install added yamllint as a binary. uvx downloads and reinstalls the package with uv to run which feels redundant

mise.toml
joelmccoy: seems updating mise resolved this. pypi must be a new plugin. in favor of getting rid of uv if we can just install natively with mise

hk.pkl
joelmccoy: ack. seems reasonable. happy with his now

hk.pkl
eoghanriley: Removed uv and updated tests

hk.pkl
eoghanriley: Removed uv

mise.toml
eoghanriley: Removed

mise.toml
eoghanriley: Removed uv

mise.toml
eoghanriley: Removed uv

mise.toml
eoghanriley: Added

CONTRIBUTING.md
eoghanriley: Removed helm and the instruction to install the plugin

chance-coleman (dismissed):
No new comments from me, looks good!

renovate.json
```diff
@@ -236,6 +236,11 @@
       ],
       "versioning": "regex:^distroless-v(?<major>\\d+)\\.(?<minor>\\d+)\\.(?<patch>\\d+)$",
       "description": "Envoy distroless images use a 'distroless-v<semver>' tag prefix that the default docker versioning rejects as invalid-value, blocking updates"
+    },
+    {
+      "matchManagers": ["mise"],
+      "groupName": "support-deps",
+      "commitMessageTopic": "support dependencies"
```
joelmccoy, line 244: I think we also need to put opentofu in iac-support-deps too. maybe add this after:

```json
 {
   "matchManagers": ["mise"],
   "matchPackageNames": ["opentofu"],
   "groupName": "iac-support-deps",
   "commitMessageTopic": "iac support dependencies"
 }
```

joelmccoy (commented):
Mostly LGTM - one small comment on the renovate

renovate.json
eoghanriley: Fixed, I also excluded it from the support-deps since that is how uds-cli puts go in another group, not sure if its required.

joelmccoy (approved):

PR #2992, opened 2026-09-20
fix(ci): preserve node log writer exemption

slaskawi:
## Description

Fix the Vector node-log test DaemonSet's interaction with `RequireNonRootUser`.

### Why the failure started

[#2971](https://github.com/defenseunicorns/uds-core/pull/2971) added the root node-log writer DaemonSet and its `Exemption` on 18 September. The first scheduled Kubernetes compatibility run after it merged ([run 35420816507](https://github.com/defenseunicorns/uds-core/actions/runs/35420816507)) failed the writer in every v1.34 flavor. The test was absent from the preceding successful scheduled run.

The writer intentionally runs as UID 0 to write controlled files under the host's `/var/log` for the Vector host-log test. Its manifest creates the `Exemption` and DaemonSet together.

### Root cause

`RequireNonRootUser` reads exemptions from the informer-backed `ExemptionStore` during mutation. If the cache has not observed this just-created `Exemption`, it adds `runAsNonRoot: true` to the pod security context. The writer already declares `runAsUser: 0`. Validation can observe the exemption shortly afterward and admit the pod, but kubelet then rejects that contradictory context with `container's runAsUser breaks non-root policy`.

The patch adds the same exemption marker normally written by the policy. Mutation preserves the writer's intentional root context. Validation still independently requires the matching `Exemption`; this does not bypass the policy.

Production Vector was healthy in the failing runs. It is a separate workload whose chart explicitly sets `podSecurityContext.runAsNonRoot: false` and deploys its own `vector` exemption separately. The failed pod is only the test writer; the upstream flavor failed too.

### Why #2971 CI missed it

The introducing PR passed the full install matrix for upstream, registry1, and unicorn. Those jobs run `uds-core-e2e` on the default v1.36.4 bundle-deploy path. They did create the test package and run the new Vitest test.

The compatibility workflow exercises the separate values-deploy path on v1.34 and v1.35. It runs on schedule, manual dispatch, or changes to `.github/workflows/test-kubernetes-matrix.yaml`; #2971 changed none of those paths. No PR-time v1.34 values-deploy job exercised the cache timing that causes this failure.

This is not a Kubernetes v1.34 semantic regression: v1.35 values-deploy jobs passed, and the v1.36 PR jobs used a different deploy path. An asynchronous cache has no ordering guarantee; the v1.34 values jobs consistently exposed the timing window in these runs.

### Evidence

Unchanged commit `724e1397`:

- [First scheduled run](https://github.com/defenseunicorns/uds-core/actions/runs/35420816507): all three v1.34 flavors failed with the same kubelet error; all v1.35 flavors passed.
- [Reported run](https://github.com/defenseunicorns/uds-core/actions/runs/35488725974): upstream and registry1 v1.34 failed; rerun attempts 2 and 3 failed the same two jobs. Attempt 3 logs show `CreateContainerConfigError` and `container's runAsUser breaks non-root policy` for `vector-node-log-writer`.

Patched commit `22c9a3f`:

- [Run 1](https://github.com/defenseunicorns/uds-core/actions/runs/35518153691), [run 2](https://github.com/defenseunicorns/uds-core/actions/runs/35519466093), and [run 3](https://github.com/defenseunicorns/uds-core/actions/runs/35521412261) each passed all six Kubernetes-version/flavor jobs: 18 successful matrix jobs.

No documentation change is needed: this only configures an internal test fixture, and the public `Exemption` CRD documentation makes no readiness-ordering guarantee.

## Related Issue

Relates to https://github.com/defenseunicorns/uds-core/actions/runs/35488725974/job/106085179592

## Type of change

## Steps to Validate

- `yq eval 'select(.kind == "DaemonSet").spec.template.metadata.annotations."uds-core.pepr.dev/uds-core-policies.RequireNonRootUser"' src/test/vector-node-log-test.yaml`
- `yamllint src/test/vector-node-log-test.yaml`
- Run the Kubernetes compatibility workflow three times against this branch.

## Checklist before merging


chance-coleman (dismissed):

src/test/vector-node-log-test.yaml
```diff
@@ -38,6 +38,8 @@ spec:
       app.kubernetes.io/name: vector-node-log-writer
   template:
     metadata:
+      annotations:
+        uds-core.pepr.dev/uds-core-policies.RequireNonRootUser: exempted
```
joelmccoy: this is not how this annotation is expected to be used. end users shouldn't be setting this on their templates (this is a operator controlled annotation).  This is a hack to bypass some mutation logic our operator code. It just buys some extra time (exemptions are still checked after the mutation)

The correct fix for this would be to make sure the exemption is deployed/processed by admission pods before the daemonset is admitted.  This could be done maybe by:

1. Split `src/test/vector-node-log-test.yaml` into:
   - `vector-node-log-test-exemption.yaml`
     - Namespace
     - Exemption
   - `vector-node-log-test-daemonset.yaml`
     - DaemonSet

2. In `src/test/zarf.yaml`, move the exemption to an earlier component and maybe have a sleep command  onDeploy after as there is no way we can check an exemption is truly ready by the admission pods

slaskawi: Agreed. I’ll deploy the matching exemption before the writer and remove the pre-seeded operator annotation.

joelmccoy (approved):

